PR #2, opened 2014-12-26
Add lots of docs, mostly best practices.

brian-brazil:
Start out on the visualisation docs.

Expand overview, remove federation as a feature we don't have that yet.

Switch example variable to follow naming scheme.

### DIFF
--- a/content/docs/practices/alerting.md
+++ b/content/docs/practices/alerting.md
@@ -1,0 +1,74 @@
+---
+title: Alerting
+sort_rank: 4
+---
+
+# Alerting
+
+We recommend that you read [My Philosophy on Alerting](https://docs.google.com/a/boxever.com/document/d/199PqyG3UsyXlwieHaqbGiWVa8eMWi8zzAn0YfcApr8Q/edit) based on Rob Ewaschuk's observations at Google.
+
+To summarize, keep alerting simple, alert on symptoms, have good consoles
+to allow pinpointing causes and avoid having pages where there is nothing to
+do.
+
+## What to alert on
+
+Aim to have as few alerts as possible, by alerting on symptoms that are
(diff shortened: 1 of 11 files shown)


content/docs/practices/alerting.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Metamonitoring
+
+It's important to have confidence that your monitoring is working, accordingly
+you should have alerts that there's sufficient Prometheuses, Alertmanagers,
+PushGateways and other monitoring infrastructure running to alert on any
+problems.
+
+As always, if you can alert on symptoms rather than causes this helps reduce
+noise. For example a blackbox test that alerts are getting from PushGateway to
```
brian-brazil: We're referring to our particular Alertmanager and PushGateway implementations, thus I think they're proper nouns and get capitals.

brian-brazil:
I think that's all of your comments covered.

Not using contractions feels very unnatural to me.

juliusv:
@brian-brazil Thanks! Re-reading and then reading the second half now.

I agree in some cases about the contractions. @opensourcegrrrl (our doc writer) has always told me to avoid them, but I'm not sure if we have to be so strict about it.

juliusv:
As discussed on IRC, I'm going to merge this as-is now and send a followup PR with some changes later. Easier for everyone :)

opensourcegrrrl:
"Expand abbreviations, such as e.g., because they add complexity for non-native readers, are used inconsistently, and are unnecessary." http://infobits.eu/tech-doc/style-guide/

opensourcegrrrl:
In an instance such as, "When doesn't it fit?", the emphasis should be on _not_, yet that part is contracted. "When does it not fit?" might be the clearer option.

PR #14, opened 2015-01-23
Document deriv, update/correct docs for rate and delta.

brian-brazil:


content/docs/querying/functions.md
```diff
@@ -37,28 +37,33 @@ allows grouping by labels via a `by` clause.
...
-Example which returns the difference in CPU temperature between now and 2 hours
-ago:
+`delta` should only be used with gauges.
+
+## `deriv()`
+
+`deriv(v range-vector)` calculates the derivative of the time series in a range
+vector `v`, using [simple linear regression](http://en.wikipedia.org/wiki/Simple_linear_regression).
+
+Example which predicts the CPU temperature in 5 minutes time based on the previous hour of data:
```
juliusv: "...which predicts the CPU temperature 5 minutes into the future..."?

content/docs/querying/functions.md
```diff
@@ -37,28 +37,33 @@ allows grouping by labels via a `by` clause.
...
 
 ```
-delta(cpu_temp_celsius{host="zeus"}[2h], 0)
+cpu_temp_celsius{host="zeus"} + deriv(cpu_temp_celsius{host="zeus"}[1h]) * 5 * 60
 ```
 
+`deriv` should only be used with gauges.
```
juliusv: "..., since it does not correct for counter resets."?

content/docs/querying/functions.md

content/docs/querying/functions.md
brian-brazil: I'm not sure we should get too deep into the theory in the reference documentation, as it could get pretty long across all of the functions (the *_over_time functions are complicated for example).

content/docs/querying/functions.md
juliusv: From our tech writer :)

"The following example call returns..." (to make it a sentence)

Both here and in the other example.

juliusv: (in this case: "The following example call predicts")

content/docs/querying/functions.md
juliusv: Feels weird to say "should only be used with gauges" without giving a reasoning though?

content/docs/querying/functions.md
brian-brazil: Done, I've worded the predict slightly differently.

content/docs/querying/functions.md
```diff
@@ -37,28 +37,34 @@ allows grouping by labels via a `by` clause.
...
-ago:
+`delta` should only be used with gauges.
+
+## `deriv()`
+
+`deriv(v range-vector)` calculates the derivative of the time series in a range
+vector `v`, using [simple linear regression](http://en.wikipedia.org/wiki/Simple_linear_regression).
+
+The following example expression returns the predicted CPU temperature in 5
+minutes into the future based on the previous hour of data:
```
juliusv: "in 5 minutes into the future" -> then I'd just say "in 5 minutes" now.

content/docs/querying/functions.md
brian-brazil: I think this is one where I'll need another blog post.

PR #15, opened 2015-01-23
Expand faq.

brian-brazil:
@beorn7 Is that the right plural?

PR #16, opened 2015-01-24
Picture

brian-brazil:
Depends on #15


juliusv:
So while I think the picture is cute, I feel kind of awkward having a picture of me/us in the FAQ...

Maybe there is a better place to put this kind of picture, like a less timeless place than the FAQ (in one of the blog posts maybe)? What do others think? @beorn7 @discordianfish

brian-brazil:
Maybe the community page?

beorn7:
I'm a public person. I don't mind to have a pic of me up there, but I don't insist to have it, either.

juliusv:
Ok, I feel kind of weird about having it on any "permanent" (that is, timeless, non-timestamped) page. If it happened to appear in a blog post (not sure there's a good place for it in ours now though) or other page that "ages", it'd feel much more natural to me...

discordianfish:
Agreeing with @juliusv, I think it's a bit awkward there. And hope it's outdated by end of the week.
I'm not against having something like this in there in general, but not at that page. If we had some "history" page about the "story behind prometheus" we could put that picture there, but -1 on putting it on such 'general' page.

brian-brazil:
Okay, lets not have it then.

PR #17, opened 2015-01-25
Make it clear what the full config is.

brian-brazil:


content/docs/introduction/getting_started.md
```diff
@@ -266,7 +266,32 @@ global: {
   rule_file: "prometheus.rules"
 }
 
-[...]
+job: {
```
juliusv: :+1: except that the sentence "The global section should now look like this" should now read "The config should now look like this"

brian-brazil: Updated

PR #21, opened 2015-01-26
Use a protocol-relative URL to load Google Fonts

aeriff:
When viewing the documentation site over HTTPS, the CSS is not loaded from Google Fonts due to explicitly requesting it via HTTP. Switch it to a protocol-relative URL so the fonts are loaded correctly.

screen shot 2015-01-26 at 5 16 42 pm


juliusv:
Thanks!

PR #23, opened 2015-01-27
Update 'random-example' section.

mjkelly:
1) Use https to clone 'client_golang' repository, like the 'prometheus'
   repository.
2) Update build instructions to specify 'example_random' target instead
   of 'random_example'. I updated the sample configs, too.


juliusv:
Oh, thanks for catching and fixing that! :+1:

brian-brazil:
:+1: Thank!

PR #24, opened 2015-01-28
Add roadmap documentation.

juliusv:
@beorn7 @brian-brazil

### DIFF
--- a/content/docs/introduction/roadmap.md
+++ b/content/docs/introduction/roadmap.md
@@ -1,0 +1,82 @@
+---
+title: Roadmap
+sort_rank: 6
+---
+
+# Roadmap
+
+The following is only a selection of some of the major features we plan to
+implement in the near future. To get a more complete overview of planned
+features and current work, see the issue trackers for the various repositories,
+for example, the [Prometheus
+server](https://github.com/prometheus/prometheus/issues).
+
+**Hierarchical federation**
+
+Hierarchical federation will allow higher-level Prometheus servers to collect
+aggregated time series data from subordinated servers. This will enable more
+scalable monitoring topologies. For example, a setup might consist of
+per-datacenter Prometheus servers that collect data in high detail, and a set
+of global Prometheus servers which collect and store only aggregated data from
+those local servers. This allows you to have an aggregate global view and
+detailed local views.
+
+GitHub issue: [#480](https://github.com/prometheus/prometheus/issues/480)
+
+**Aggregatable histograms**
+
+The current client-side [summary
+types](http://localhost:3000/docs/concepts/metric_types/#summaries) do not
+support aggregation of quantiles. For example, it is [statistically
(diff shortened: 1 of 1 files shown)


content/docs/introduction/roadmap.md
```diff
@@ -0,0 +1,72 @@
...
+GitHub issue: [#9](https://github.com/prometheus/prometheus/issues/9)
+
+**More flexible label matching in binary operations**
+
+[Binary operations](/docs/querying/operators/) between time series vectors
+currently require exact matches of label sets on both sides of the operation
+in order for paired vector elements to propagate into the result. We plan to
+add more flexible label-matching features to support advanced query use cases.
+
+GitHub issue: [#393](https://github.com/prometheus/prometheus/issues/393)
```
brian-brazil: That's just one of the features we'll be adding.

content/docs/introduction/roadmap.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Roadmap
+sort_rank: 6
+---
+
+# Roadmap
```
brian-brazil: More clients, more exporters?

beorn7:
Perhaps mention our vision of actually making use of the type of a metric to find out about semantically wrong expressions and warn the user.

juliusv:
@beorn7 Hmm, I would have considered that more of a nice-to-have minor feature that is not firmly planned yet. Whereas the rest are quite vital stuff we really need for Prometheus to be successful in the future. But open to debate...

content/docs/introduction/roadmap.md
juliusv: It is, but I found it a quite vital one. What do you propose? Renaming the section to "Query language improvements" and listing several examples?

content/docs/introduction/roadmap.md
juliusv: Besides the Python client lib, do we have more concrete plans here?

content/docs/introduction/roadmap.md
brian-brazil: Maybe clarify that the github issue is just one of the features planned for binary operations?

beorn7:
@juliusv  The problem I see without that note is that smart people will ask us why we have types in the exposition format at all (and not just 'gauges'), Cf. discussion with our principal engineer. Without the vision behind it, it seems like unneeded complexity.

content/docs/introduction/roadmap.md
juliusv: Ah, what else do we have planned for binops? If we do have other things, we should create issues for them.

brian-brazil: Matching on a subset of labels without the logic in that issue, and various modifiers as to what happens to the labels you don't match on

juliusv:
Added second binop issue and section about metric types. @brian-brazil feel free to add more sections about exporters, client libraries, staleness.

PR #29, opened 2015-02-02
Add staleness and clients/exporters

brian-brazil:
@juliusv


content/docs/introduction/roadmap.md
```diff
@@ -72,6 +72,15 @@ other long-term storage backends.
 
 GitHub issue: [#10](https://github.com/prometheus/prometheus/issues/10)
 
+**Improved staleness handling**
+
+Currently Prometheus considers a time series stale when it has no data points
+within the past 5 minutes. This prevents things like propogating timestamps
```
juliusv: s/propogating/propogating/

I think the concept of staleness itself needs further explanation here. People might fear that they can't see time series at all anymore if they didn't get an update for 5 minutes. Rather, they only disappear from queries at times that are >5m away from the closest sample.

How about writing it like this (still not clear enough for my taste, but it's a complicated subject):

"Currently Prometheus omits time series from query results if the timestamp for which the query is executed is more than 5 minutes away from the nearest sample. This means that time series will grow "stale" in instant queries after not receiving samples for 5 minutes. This currently prevents usage of client-side timestamps from the Pushgateway or CloudWatch Exporter, which might indicate a time more than 5 minutes in the past. We plan on only considering a time series stale if it was not present in the most recent scrape."

===> Actually, this whole topic still raises a lot of implementation and design questions. With the current design, I don't know how I'd do it (scrapers are completely separated from the stored time series - how would we communicate that a scrape happened for a time series without actually updating that time series?), and what happens to staleness that is not at the *end* of a time series, but somewhere in between? Given that we don't have a good idea how to do this yet, I'm hesitant to promise it this concretely on the official roadmap until we have a clearer vision of how this can actually work. At least the last sentence should be more of a "we will see how we can improve this situation, etc.".

juliusv: For the issue of staleness in between queries, the whole issue of interpolation also comes up. If a time series has a gap of an hour, we probably want to drop results for it in the middle, but the behavior of still interpolating 5 minutes worth of stale data into this interval from both sides is probably also not ideal. Not exactly sure how it should behave in the ideal case.

PR #31, opened 2015-02-05
Document rounding functions.

juliusv:
@brian-brazil @mmikulicic

### DIFF
--- a/content/docs/querying/functions.md
+++ b/content/docs/querying/functions.md
@@ -28,6 +28,11 @@ the 1-element output vector from the input vector:
 This is useful for alerting on when no time series
 exist for a given metric name and label combination.
 
+## `ceil()`
+
+`ceil(v instant-vector)` rounds the sample values of all elements in `v` up to
+the nearest integer.
+
 ## `count_scalar()`
 
 `count_scalar(v instant-vector)` returns the number of elements in a time series
@@ -70,6 +75,11 @@ cpu_temp_celsius{host="zeus"} + deriv(cpu_temp_celsius{host="zeus"}[1h]) * 5 * 6
 `drop_common_labels(instant-vector)` drops all labels that have the same name
 and value across all series in the input vector.
 
+## `floor()`
+
+`floor(v instant-vector)` rounds the sample values of all elements in `v` down
+to the nearest integer.
+
 ## `rate()`
 
 `rate(v range-vector)` calculate the per-second average rate of increase of the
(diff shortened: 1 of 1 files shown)


content/docs/querying/functions.md
```diff
@@ -85,6 +95,15 @@ rate(http_requests_total{job="api-server"}[5m])
 
 `rate` should only be used with counters.
 
+## `round()`
+
+`round(v instant-vector, places=0 scalar)` rounds the sample values of all
```
brian-brazil: I'm not sure that 'places' is the best API to offer, as sometimes you want to round to a non-decimal base and doing round(x, log(2)/log(10)) isn't exactly the most user friendly vs round(x, 1e-2) for when you want more decimal places.

I'd suggest round(x, 5) would round to the nearest 5 etc., and have similar for floor/ceil.

juliusv: Good point, and admittedly this was the interface I was initially naively expecting as well. Would you be up for a PR around that (since you seem to be into the math)?

brian-brazil: It's been a while since I've touched done it, but let me have a go.

juliusv:
@brian-brazil Updated to match https://github.com/prometheus/prometheus/pull/508.

content/docs/querying/functions.md
```diff
@@ -85,6 +95,13 @@ rate(http_requests_total{job="api-server"}[5m])
 
 `rate` should only be used with counters.
 
+## `round()`
+
+`round(v instant-vector, to_nearest=0 scalar)` rounds the sample values of all
```
brian-brazil: toNearest=1

juliusv: Fixed.

brian-brazil:
:+1:

content/docs/querying/functions.md
```diff
@@ -85,6 +95,13 @@ rate(http_requests_total{job="api-server"}[5m])
 
 `rate` should only be used with counters.
 
+## `round()`
+
+`round(v instant-vector, to_nearest=1 scalar)` rounds the sample values of all
+elements in `v` to the nearest integer. Ties are resolved by rounding up. The
+optional `to_nearest` argument allows specifying the nearest multiple to which
+the sample values should be rounded.
```
mmikulicic: Should we make it clearer that the multiple can be a fraction ?

Given that the text starts by stating that it the value gets rounded to the nearest _integer_, the reader might not get it works also with e.g. `0.1`.

juliusv: Created new PR: https://github.com/prometheus/docs/pull/32

PR #33, opened 2015-02-05
Add some FAQs.

beorn7:
Those were actually frequently asked...

Also, fix issue swap.

NOTE: Only merge once I have reproduced the benchmark numbers. Midway,
my server fell apart. (Prometheus is also blowing up old harddisks...)
I should get a new hard disk tonight...

@juliusv  @brian-brazil Just to get your opinion early. Will run that benchmark again once my RAID is back together. (I could run it on one disk, but first of all, if that blows up, too, I'm in trouble, and second, I want the RAID to prove that even on a slow writer like RAID1, everything is fine... :)

### DIFF
--- a/content/docs/introduction/faq.md
+++ b/content/docs/introduction/faq.md
@@ -89,6 +89,16 @@ Yes, with the experimental [Alertmanager](https://github.com/prometheus/alertman
 Yes, with [PromDash](/docs/visualization/promdash/) and [Console
 templates](/docs/visualization/consoles/).
 
+### Can I change the timezone? Why is everything in UTC?
+
+To avoid any kind of timezone confusion, especially when the so-called
+daylight saving time is involved, we decided to exclusively use Unix
+time internally and UTC for display purposes in all components of
+Prometheus. A carefully done timezone selection could be introduced
+into the UI. Contributions are welcome. See
+[issue #500](https://github.com/prometheus/prometheus/issues/500)
+for the current state of this effort.
+
 ## Instrumentation
 
 ### Which languages have instrumentation libraries?
@@ -153,3 +163,53 @@ should take less than a minute under normal circumstances. See [crash recovery](
 
 You have run into a bug of ZFS on Linux. See [issue #484](https://github.com/prometheus/prometheus/issues/484)
 for details. Upgrading to ZFS on Linux v0.6.4 should fix the issue.
+
+## Implementation
+
+### Why are all sample values 64-bit floats? I want integers.
+
+We restrained ourselves to 64-bit floats to simplify the design. The
+[IEEE 754 double-precision binary floating-point
+format](http://en.wikipedia.org/wiki/Double-precision_floating-point_format)
+supports integer precision for values up to 2<sup>53</sup>. Supporting
+native 64 bit integers would (only) help if you need integer precision
+above 2<sup>53</sup> but below 2<sup>63</sup>. In principle, support
+for different sample value types (including some kind of big integer,
+supporting even more than 64 bit) could be implemented, but it is not
+a priority right now. Note that a counter, even if incremented
+one million times per second, will only run into precision issues
+after over 285 years.
+
+### Why does Prometheus use a custom storage backend rather than [some other storage method]? Isn't the "one file per time series" approach killing performance?
+
+Initially, Prometheus ran completely on LevelDB, but to achieve better
+performance, we had to change the storage for bulk sample data. We
+evaluated many storage backends that were available at the time,
+without getting satisfactory results. So we implemented exactly the
+parts we needed, while keeping LevelDB for indexes and making heavy
+use of file system capabilities. Obviously, we could not evaluate
+every single storage backend out there, and storage backends have
+evolved meanwhile. However, the performance of the solution
+implemented now is satisfactory for most use-cases. Our most important
+requirements are an acceptable query speed for common queries and a
+sustainable ingestion rate of many thousands of samples per
+second. The latter depends on the compressibility of the sample data
+and on the number of time series the samples belong to, but to give
+you an idea, here are some results from benchmarks:
+
+* On an older 8-core machine with Intel Core i7 CPUs and two spinning
+  disks (Samsung HD753LJ) in a RAID-1 setup, Prometheus sustained an
+  ingestion rate of 20k samples per second, belonging to 450k time
+  series, scraped from 1500 targets.
+* On a modern server with SSD, Prometheus sustained an ingestion rate
+  of more than 100k samples per second, belonging to millions of time
+  series, scraped from thousands of targets.
+
+In both cases, the bottleneck was identified as insufficiently
+parallelized hash calculation, which happens before samples even hit
+the storage backend.
+
+Running out of inodes is highly unlikely in a usual set-up. There is a
+possible downside: If you want to delete Prometheus's storage
+directory, you will notice that some file systems are very slow when
+deleting files.
--- a/content/docs/introduction/roadmap.md
+++ b/content/docs/introduction/roadmap.md
@@ -21,7 +21,7 @@ of global Prometheus servers which collect and store only aggregated data from
 those local servers. This allows you to have an aggregate global view and
 detailed local views.
 
-GitHub issue: [#480](https://github.com/prometheus/prometheus/issues/480)
+GitHub issue: [#9](https://github.com/prometheus/prometheus/issues/9)
 
 **Aggregatable histograms**
 
@@ -32,7 +32,7 @@ incorrect](http://latencytipoftheday.blogspot.de/2014/06/latencytipoftheday-you-
 to average over the 90th percentile latency of multiple monitored instances.
 We plan to implement server-side histograms which will allow for this use case.
 
-GitHub issue: [#9](https://github.com/prometheus/prometheus/issues/9)
+GitHub issue: [#480](https://github.com/prometheus/prometheus/issues/480)
 
 **More flexible label matching in binary operations**
 
@@ -86,12 +86,15 @@ GitHub issue: [#398](https://github.com/prometheus/prometheus/issues/398)
 
 **Server-side metric metadata support**
 
-At this time, metric types and other metadata are only used in the client
-libaries and in the exposition format, but not persisted or utilized in the
-Prometheus server. We plan on making use of this metadata in the future. For
-example, we could suggest automatic rates over counters, warn users if they
-take the rate of a gauge, or display metric documentation strings. The details
-of this are still to be determined.
+At this time, metric types and other metadata are only used in the
+client libaries and in the exposition format, but not persisted or
+utilized in the Prometheus server. We plan on making use of this
+metadata in the future. For example, we could suggest automatic rates
+over counters, warn users if they take the rate of a gauge, or display
+metric documentation strings. Some metric types, like the upcoming
+[server-side histograms](https://github.com/prometheus/prometheus/issues/480),
+could also be stored and processed in a more efficient way.  The
+details of this are still to be determined.
 
 **More client libraries and exporters**
 
--- a/content/docs/operating/storage.md
+++ b/content/docs/operating/storage.md
@@ -86,3 +86,16 @@ storage directory:
    1. Stop Prometheus.
    1. `rm -r <storage path>/*`
    1. Start Prometheus.
+
+## Hash collisions
+
+Prometheus currently uses 64-bit fingerprints to identify time
+series. On a large server with several million time series, the chance
+of a hash collision is about one to one million (assuming the FNV-1a
+hash function works well). While that might appear safe enough, the
+problem is that a hash collision will effectively lead to undetected
+data corruption. Also, with more powerful hardware and future
+improvements of the Prometheus code, much higher numbers of time
+series might be handled by a single server, increasing the chance of a
+collision. See [Prometheus issue #509] for efforts to deal with the
+problem.
--- a/content/docs/concepts/jobs_instances.md
+++ b/content/docs/concepts/jobs_instances.md
@@ -35,5 +35,3 @@ For each instance scrape, Prometheus stores a sample of the form
 `up{job="<job-name>", instance="<instance-url>"}` with a value of `1` if the
 instance was scraped successfully or a value of `0` if the scrape failed. This
 time series is useful for instance availability monitoring.
-
-TODO: move alerts stuff somewhere else (Prometheus also stores * `ALERTS`: for pending and firing alerts, a time series of the form `ALERTS{alertname="...", alertstate="pending|firing",...alertlabels...}` is written out. The sample value is `1` as long as the alert is in the indicated active (pending/firing) state, but a single `0` value gets written out when an alert transitions from active to inactive state.)
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -12,10 +12,11 @@ metrics directly (for example, HAProxy or Linux system stats). The
 following is a list of existing third-party exporters:
 
    * [Node/system metrics exporter](https://github.com/prometheus/node_exporter)
-   * [HAProxy exporter](https://github.com/prometheus/haproxy_exporter)
-   * [AWS CloudWatch exporter](https://github.com/prometheus/cloudwatch_exporter)
-   * [StatsD bridge](https://github.com/prometheus/statsd_bridge)
+   * [Collectd exporter](https://github.com/prometheus/collectd_exporter)
    * [JMX exporter](https://github.com/prometheus/jmx_exporter)
+   * [HAProxy exporter](https://github.com/prometheus/haproxy_exporter)
+   * [StatsD bridge](https://github.com/prometheus/statsd_bridge)
+   * [AWS CloudWatch exporter](https://github.com/prometheus/cloudwatch_exporter)
    * [Hystrix metrics publisher](https://github.com/prometheus/hystrix)
 
 The [JMX exporter](https://github.com/prometheus/jmx_exporter) can export from a
--- a/content/docs/introduction/getting_started.md
+++ b/content/docs/introduction/getting_started.md
@@ -93,6 +93,20 @@ about itself from its own HTTP metrics endpoint.
 You can also verify that Prometheus is serving metrics about itself by
 navigating to its metrics exposure endpoint: http://localhost:9090/metrics
 
+By default, Prometheus will only execute at most one OS thread at a
+time. In production scenarios on multi-CPU machines, you will most
+likely achieve better performance by setting the `GOMAXPROCS`
+environment variable to a value similar to the number of available CPU
+cores:
+
+```language-bash
+GOMAXPROCS=8 ./prometheus -config.file=prometheus.conf
+```
+
+Blindly setting `GOMAXPROCS` to a high value can be
+counterproductive. See the relevant [Go
+FAQs](http://golang.org/doc/faq#Why_no_multi_CPU).
+
 ## Using the expression browser
 
 Let's try looking at some data that Prometheus has collected about itself. To
--- a/content/docs/querying/rules.md
+++ b/content/docs/querying/rules.md
@@ -118,6 +118,13 @@ To manually inspect which alerts are active (pending or firing), navigate to
 the "Alerts" tab of your Prometheus instance. This will show you the exact
 label sets for which each defined alert is currently active.
 
+For pending and firing alerts, Prometheus also stores synthetic time series of
+the form `ALERTS{alertname="<alert name>", alertstate="pending|firing", <additional alert labels>}`.
+The sample value is set to `1` as long as the alert is in the indicated active
+(pending or firing) state, and a single `0` value gets written out when an alert
+transitions from active to inactive state. Once inactive, the time series does
+not get further updates.
+
 ### Sending alert notifications
 Prometheus's alerting rules are good at figuring what is broken *right now*,
 but they are not a fully-fledged notification solution. Another layer is needed


content/docs/introduction/faq.md
```diff
@@ -89,6 +89,16 @@ Yes, with the experimental [Alertmanager](https://github.com/prometheus/alertman
 Yes, with [PromDash](/docs/visualization/promdash/) and [Console
 templates](/docs/visualization/consoles/).
 
+### Can I change the timezone? Why is everything in UTC?
+
+For monitoring purposes, it is absurd to use any timezone observing
+so-called daylight saving time. To avoid any kind of timezone
+confusion, we decided to exclusively use UTC for Prometheus. For pure
```
brian-brazil: I'd say unixtime rather than UTC

content/docs/introduction/faq.md
```diff
@@ -89,6 +89,16 @@ Yes, with the experimental [Alertmanager](https://github.com/prometheus/alertman
 Yes, with [PromDash](/docs/visualization/promdash/) and [Console
 templates](/docs/visualization/consoles/).
 
+### Can I change the timezone? Why is everything in UTC?
+
+For monitoring purposes, it is absurd to use any timezone observing
```
brian-brazil: Not just monitoring, anything that's not a display for human consumption. Maybe avoid the word 'absurd'? It might come off a bit strong.

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+
+## Implementation
+
+### Why are all sample values 64-bit floats? I want integers.
+
+We restrained ourselves to 64-bit floats to simplify the design. The
+[IEEE 754 double-precision binary floating-point format](http://en.wikipedia.org/wiki/Double-precision_floating-point_format)
+supports integer precision for values up to 2<sup>53</sup>. Supporting
+native 64 bit integers would (only) help if you need integer precision
+above 2<sup>53</sup> but below 2<sup>63</sup>. In principle, support
```
brian-brazil: I'd add mention that with a counter in a process being incremented at a very high rate of 1M events per second, it'd take over 285 years to run into precision issues.

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+
+### Why does Prometheus use a custom storage backend rather than [some other storage system]? Isn't the "one file per time series" approach killing performance?
+
+Initially, Prometheus ran completely on LevelDB, but to achieve better
+performance, we had to change the storage for bulk sample data. We
+evaluated many storage systems that were available at the time, to no
+avail. So we implemented exactly the parts we needed, while keeping
+LevelDB for indexes and making heavy use of file system
+capabilities. Obviously, we could not evaluate every single storage
+system out there, and storage systems have evolved meanwhile. However,
```
brian-brazil: In absolute terms, storage systems haven't evolved that much in the past few years. SSD is the only thing really pushing anything fundamentally new.

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+the performance of the solution implemented now is satisfying for most
```
brian-brazil: satisfactory

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+We restrained ourselves to 64-bit floats to simplify the design. The
+[IEEE 754 double-precision binary floating-point format](http://en.wikipedia.org/wiki/Double-precision_floating-point_format)
+supports integer precision for values up to 2<sup>53</sup>. Supporting
+native 64 bit integers would (only) help if you need integer precision
+above 2<sup>53</sup> but below 2<sup>63</sup>. In principle, support
+for different sample value types (including some kind of big integer,
+supporting even more than 64 bit) could be implemented, but it is not
+really a priority right now.
+
+### Why does Prometheus use a custom storage backend rather than [some other storage system]? Isn't the "one file per time series" approach killing performance?
```
brian-brazil: There is one downside to one file per timeseries: Linux is really slow deleting files when you want to blow the directory away.

Probably make mention that you'd not likely to run out of inodes.

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+capabilities. Obviously, we could not evaluate every single storage
+system out there, and storage systems have evolved meanwhile. However,
+the performance of the solution implemented now is satisfying for most
+use-cases. The most important metric is the sustainable ingestion
+rate. It depends on the compressability of the sample data and on the
+number of time series the samples belong to, but to give you an idea,
+here are some results from benchmarks:
+
+* On an older 8-core machine with Intel Core i7 CPUs and two spinning
+  disks (Samsung HD753LJ) in a Raid-1 setup, Prometheus could ingest
```
brian-brazil: RAID

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+  50k samples per seconds, belonging to 300k time series. (TODO:
```
brian-brazil: Mention the bottleneck?

content/docs/introduction/roadmap.md
```diff
@@ -86,12 +86,16 @@ GitHub issue: [#398](https://github.com/prometheus/prometheus/issues/398)
...
-of this are still to be determined.
+At this time, metric types and other metadata are only used in the
+client libaries and in the exposition format, but not persisted or
+utilized in the Prometheus server. We plan on making use of this
+metadata in the future. For example, we could suggest automatic rates
+over counters, warn users if they take the rate of a gauge, or display
+metric documentation strings. Some metric types, like the upcoming
+[server-side histograms](https://github.com/prometheus/prometheus/issues/480),
+could also be stored and processed in a more efficient way.
+characteristics are known. The details of this are still to be
```
brian-brazil: > characteristics are known.
I think you meant to remove this

content/docs/introduction/faq.md
juliusv: Yes please. @beorn7 I know how we all feel about daylight savings time, but this is not the best place for these feelings :) I would word the whole paragraph less prescriptively (as in, "we know what's best for you"). I can understand why some people would want to look at stuff in local time in their situations, and I think "this is potentially very tedious work with little benefit" is a bit exaggerated / prescriptivist as well. I'd rather say that so far, we haven't had the need for it (and maybe explain in somewhat more neutral terms why), but we welcome UI-side contributions which would make this possible.

content/docs/introduction/faq.md
juliusv: I think UTC is less confusing than "Unix time". What the user sees is simply UTC, not Unix time, even if some (but not all) of the interfaces below speak in Unix timestamps.

content/docs/introduction/faq.md
juliusv: That's a good one.

content/docs/introduction/faq.md
beorn7: Yeah, it's unixtime internally and UTC for display purposes. The FAQ is probably asked because somebody saw UTC.

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+### Why are all sample values 64-bit floats? I want integers.
+
+We restrained ourselves to 64-bit floats to simplify the design. The
+[IEEE 754 double-precision binary floating-point format](http://en.wikipedia.org/wiki/Double-precision_floating-point_format)
+supports integer precision for values up to 2<sup>53</sup>. Supporting
+native 64 bit integers would (only) help if you need integer precision
+above 2<sup>53</sup> but below 2<sup>63</sup>. In principle, support
+for different sample value types (including some kind of big integer,
+supporting even more than 64 bit) could be implemented, but it is not
+really a priority right now.
```
juliusv: s/really//

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+
+Initially, Prometheus ran completely on LevelDB, but to achieve better
+performance, we had to change the storage for bulk sample data. We
+evaluated many storage systems that were available at the time, to no
+avail. So we implemented exactly the parts we needed, while keeping
+LevelDB for indexes and making heavy use of file system
+capabilities. Obviously, we could not evaluate every single storage
+system out there, and storage systems have evolved meanwhile. However,
+the performance of the solution implemented now is satisfying for most
+use-cases. The most important metric is the sustainable ingestion
```
juliusv: I'd say query speed is just as important, if not more important.

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+rate. It depends on the compressability of the sample data and on the
```
juliusv: s/compressability/compressibility/

content/docs/introduction/faq.md
beorn7: I already wrote that a million times more benign than I feel about this scourge of humanity. I will dilute it down another trillion times...

content/docs/introduction/faq.md
juliusv: s/could/was able to/

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+number of time series the samples belong to, but to give you an idea,
+here are some results from benchmarks:
+
+* On an older 8-core machine with Intel Core i7 CPUs and two spinning
+  disks (Samsung HD753LJ) in a Raid-1 setup, Prometheus could ingest
+  50k samples per seconds, belonging to 300k time series. (TODO:
+  Reproduce values.)
+* On a modern server with SSD, Prometheus could ingest more than 100k
```
juliusv: dito

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+  samples per second, belonging to millions of time series. The
```
juliusv: s/belonging to/spread across/

content/docs/introduction/roadmap.md
```diff
@@ -32,7 +32,7 @@ incorrect](http://latencytipoftheday.blogspot.de/2014/06/latencytipoftheday-you-
 to average over the 90th percentile latency of multiple monitored instances.
 We plan to implement server-side histograms which will allow for this use case.
 
-GitHub issue: [#9](https://github.com/prometheus/prometheus/issues/9)
+GitHub issue: [#480](https://github.com/prometheus/prometheus/issues/480)
```
juliusv: Oopsie. Thanks for fixing these.

content/docs/introduction/faq.md
grobie: Add a link at the end: "For the author's real opinion on this topic, click
here".

content/docs/introduction/roadmap.md
```diff
@@ -86,12 +86,16 @@ GitHub issue: [#398](https://github.com/prometheus/prometheus/issues/398)
 
 **Server-side metric metadata support**
 
-At this time, metric types and other metadata are only used in the client
-libaries and in the exposition format, but not persisted or utilized in the
-Prometheus server. We plan on making use of this metadata in the future. For
-example, we could suggest automatic rates over counters, warn users if they
-take the rate of a gauge, or display metric documentation strings. The details
-of this are still to be determined.
+At this time, metric types and other metadata are only used in the
```
juliusv: We might want to standardize on a line length for wrapping to cause less diff noise if the only change to a paragraph is an addition at the end. (e.g. I use the vim default of 79 chars plus newline)

content/docs/introduction/faq.md
beorn7: Good idea.

content/docs/introduction/faq.md
beorn7: Oops... By "storage system", I don't mean hardware but software that organizes the data before it is saved to a device. What would be a better word for that?

content/docs/introduction/faq.md
beorn7: Done.

content/docs/introduction/faq.md
juliusv: Homeopathic dillution please :D

content/docs/introduction/faq.md
beorn7: Funny enough, the deletion speed wasn't really an issue in practice. (Originally, we had one _directory_ per time series, with files in it, but ever since we got rid of that, we had no problems when blowing away a massive amount of files.) Perhaps XFS (which we are using) is faster in deleting files?

juliusv: Yeah, with initial versions of the new storage, deletes of large structures took hours. But then we switched to a different dir/file layout and it took only seconds. Or are we completely sure it wasn't the switch to SSD?

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,40 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+above 2<sup>53</sup> but below 2<sup>63</sup>. In principle, support
+for different sample value types (including some kind of big integer,
+supporting even more than 64 bit) could be implemented, but it is not
+really a priority right now.
+
+### Why does Prometheus use a custom storage backend rather than [some other storage system]? Isn't the "one file per time series" approach killing performance?
+
+Initially, Prometheus ran completely on LevelDB, but to achieve better
+performance, we had to change the storage for bulk sample data. We
+evaluated many storage systems that were available at the time, to no
```
juliusv: s/to no avail/without getting satisfying results/

content/docs/introduction/faq.md
beorn7: fixed

content/docs/introduction/faq.md
beorn7: Most likely disk... but at the moment, I can't investigate. Will add when done.

content/docs/introduction/roadmap.md
beorn7: Yup.

content/docs/introduction/faq.md
beorn7: Done.

content/docs/introduction/faq.md
beorn7: reworded

content/docs/introduction/faq.md
beorn7: done

content/docs/introduction/faq.md
```diff
@@ -89,6 +89,16 @@ Yes, with the experimental [Alertmanager](https://github.com/prometheus/alertman
 Yes, with [PromDash](/docs/visualization/promdash/) and [Console
 templates](/docs/visualization/consoles/).
 
+### Can I change the timezone? Why is everything in UTC?
+
+To avoid any kind of timezone confusion, especially when the so-called
+daylight saving time is involved, we decided to exclusively use Unix
```
juliusv: s/the so-called//

Too snarky... I know it is hard to control your feelings ;)

content/docs/introduction/roadmap.md
beorn7: Obviously, we have to use the emacs default of 70. :)

content/docs/introduction/faq.md
beorn7: changed

content/docs/introduction/faq.md
beorn7: I'll cannot call something by the wrongest name ever. It is neither saving daylight nor a time.

You can change that on your own, if you want. My name will not be spoiled with this excrement of Satan.

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,49 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+every single storage system out there, and storage systems have
+evolved meanwhile. However, the performance of the solution
+implemented now is satisfactory for most use-cases. After an
+acceptable query speed for common queries, an important metric is the
+sustainable ingestion rate. It depends on the compressibility of the
+sample data and on the number of time series the samples belong to,
+but to give you an idea, here are some results from benchmarks:
+
+* On an older 8-core machine with Intel Core i7 CPUs and two spinning
+  disks (Samsung HD753LJ) in a RAID-1 setup, Prometheus could ingest
```
juliusv: Hmm, still "could" here and below. Note that could means "könnte" (https://github.com/prometheus/docs/pull/33/files#r24201643). You'll want "can" or "is able to" or "was able to".

content/docs/introduction/faq.md
juliusv: Oooook.

content/docs/introduction/faq.md
brian-brazil: The grammar seems fine to me, I think "Prometheus ingested" might read better

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,49 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+  50k samples per seconds, belonging to 300k time series. (TODO:
+  Reproduce values. Identify bottleneck.)
+* On a modern server with SSD, Prometheus could ingest more than 100k
+  samples per second, belonging to millions of time series. The
+  bottleneck was identified as the hash calculation, which happens
```
brian-brazil: clarify that this is a cpu bottleneck?

content/docs/introduction/faq.md
beorn7: That's simple past. I'm reporting findings from the past.

content/docs/introduction/faq.md
beorn7: I'm not sure if that strictly applies. The hash calculation is happening sequentially in a pipeline. It could be the overhead of the channel receive and send, and the fact that it is single-threaded at that point...

content/docs/introduction/faq.md
beorn7: how about 'sustained an ingestion rate' - to emphasize that it was not a peak rate.

juliusv: Sounds good.

Interesting about "could" (http://simple.wiktionary.org/wiki/can). You're right. Normally I'm only used to it as "könnte" (in German). (and I accidentally posted the wrong link anyways)

content/docs/introduction/faq.md
juliusv: How about then: "was identified as insufficiently parallelized metric hash calculation"...

beorn7: "insufficiently parallelized" is a nic euphemism for "serial, wich communication overhead" :)
But I like it. at least, it is not triggering a misunderstanding.

beorn7:
Added note about hash collisions. Please check my math and my reasoning... Perhaps we shouldn't worry too much.

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,50 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+[IEEE 754 double-precision binary floating-point
+format](http://en.wikipedia.org/wiki/Double-precision_floating-point_format)
+supports integer precision for values up to 2<sup>53</sup>. Supporting
+native 64 bit integers would (only) help if you need integer precision
+above 2<sup>53</sup> but below 2<sup>63</sup>. In principle, support
+for different sample value types (including some kind of big integer,
+supporting even more than 64 bit) could be implemented, but it is not
+a priority right now. Note that a counter, even if incremented
+one million times per second, will only run into precision issues
+after more than 285 years.
```
brian-brazil: "after over 285 years"

content/docs/operating/storage.md
```diff
@@ -86,3 +86,12 @@ storage directory:
    1. Stop Prometheus.
    1. `rm -r <storage path>/*`
    1. Start Prometheus.
+
+## Hash collisions
+
+Prometheus uses 64-bit fingerprints to identify time series. On a
```
brian-brazil: The rule of thumb for the twin paradox is that the probability of collision goes over .5 when you've got sqrt(max size) elements.

So for a space 2^64, we'd be looking at series problems at 2^32. Thus 10^6 ~= 2^20 should be okay from rough math - but the chances of hardware/software failure is much smaller than 2^12.

content/docs/operating/storage.md
```diff
@@ -86,3 +86,12 @@ storage directory:
...
+large server with serveral million time series, the chance of a hash
```
juliusv: s/serveral/several/

content/docs/operating/storage.md
juliusv: Yeah, but then the details depend a lot on the hash quality (FNV-1a). Someone did some collision and randomness comparisons here: http://programmers.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed

But it's only for the 32-bit versions... but at least FNV-1a seems to have reasonably good (but *really* good) randomness.

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,50 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+evolved meanwhile. However, the performance of the solution
+implemented now is satisfactory for most use-cases. After an
+acceptable query speed for common queries, an important metric is the
+sustainable ingestion rate. It depends on the compressibility of the
+sample data and on the number of time series the samples belong to,
+but to give you an idea, here are some results from benchmarks:
+
+* On an older 8-core machine with Intel Core i7 CPUs and two spinning
+  disks (Samsung HD753LJ) in a RAID-1 setup, Prometheus sustained an
+  ingestion rate of 50k samples per seconds, belonging to 300k time
```
juliusv: s/per seconds/per second/

content/docs/introduction/faq.md
```diff
@@ -153,3 +163,50 @@ should take less than a minute under normal circumstances. See [crash recovery](
...
+  series. (TODO: Reproduce values. Identify bottleneck.)
+* On a modern server with SSD, Prometheus sustained an ingestion rate
+  of more than 100k samples per second, belonging to millions of time
+  series. The bottleneck was identified as insufficiently parallelized
+  hash calculation, which happens before samples even hit the storage
+  backend.
+
+Running out of inodes is highly unlikely in a usual set-up. There is a
+possible downside: If you want to delete Prometheus's storage
+directory, you will notice, that some file systems are very slow when
```
juliusv: s/notice,/notice/ (no comma)

content/docs/operating/storage.md
```diff
@@ -86,3 +86,12 @@ storage directory:
...
+collision is about one to one million (assuming the FNV-1a hash
+function works well). We expect that the probability of a data
```
juliusv: "that the probability ... to be higher than that" -> either lose the "that" or the "to"

juliusv:
Typo in first commit message: s/same/some/

content/docs/introduction/faq.md
beorn7: Done.

content/docs/introduction/faq.md
beorn7: done.

content/docs/operating/storage.md
beorn7: Done.

content/docs/operating/storage.md
beorn7: Assuming optimal hashing, the 1e-6 probability of a collision is reached at 6.1e6 entries.
See http://en.wikipedia.org/wiki/Birthday_attack#Mathematics
Which coincides nicely with the largest usable Prometheus server.

Where is the 2^12 number for hardware or software failure from? Over which runtime?
The nice thing about the hash collision is that it doesn't increase over time if you handle the same 6.1M time series all the time.

In my previous jobs, we had data corruptions only detected by some later checksumming, quite regularly. That was a lot of data handled there, but the probability was high enough so it happened every other month or so... not one in a million...

content/docs/introduction/faq.md
beorn7: Done.

beorn7:
I'll squash all those commits later with a nice message.

I'll need another check of yours about the hash collision section and the storage FAQ.

juliusv:
:+1:

juliusv:
:+1:

PR #39, opened 2015-02-12
Add note about memory usage during build.

beorn7:
@juliusv @brian-brazil Do you remember the exact threshold we have seen?


brian-brazil:
:+1: 

Would splitting the files into 1kB chunks help?

PR #41, opened 2015-02-19
Make the benchmark results more precise.

beorn7:
I have run a number of benchmarks by now. Results depend on a lot of
boundary conditions, sometimes in surprising ways. But I guess that's
true for any benchmark... Anyway, the text now gives at least one data
point to get an impression of what is doable.

The reduction of performance for my spinning-disk server is mainly because the filesystem tricked me as it could handle many small files nicely but degraded in performance when those files got bigger after hours of ingestion...

The improvement for the moderns server is due to recent tweaks. (They also increased the spinning-disk old-server performance. With the bigger time series files, the performance with the previous code was pretty abysmal.)

@juliusv @brian-brazil

### DIFF
--- a/content/docs/introduction/faq.md
+++ b/content/docs/introduction/faq.md
@@ -199,15 +199,15 @@ you an idea, here are some results from benchmarks:
 
 * On an older 8-core machine with Intel Core i7 CPUs and two spinning
   disks (Samsung HD753LJ) in a RAID-1 setup, Prometheus sustained an
-  ingestion rate of 20k samples per second, belonging to 450k time
-  series, scraped from 1500 targets.
+  ingestion rate of 18k samples per second, belonging to 56k time
+  series, scraped from 200 targets.
 * On a modern server with SSD, Prometheus sustained an ingestion rate
-  of more than 100k samples per second, belonging to millions of time
-  series, scraped from thousands of targets.
+  of 320k samples per second, belonging to 1.7M time
+  series, scraped from 1100 targets.
 
-In both cases, the bottleneck was identified as insufficiently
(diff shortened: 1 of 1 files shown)


juliusv:
:+1:

content/docs/introduction/faq.md
```diff
@@ -199,15 +199,15 @@ you an idea, here are some results from benchmarks:
 
 * On an older 8-core machine with Intel Core i7 CPUs and two spinning
   disks (Samsung HD753LJ) in a RAID-1 setup, Prometheus sustained an
-  ingestion rate of 20k samples per second, belonging to 450k time
-  series, scraped from 1500 targets.
+  ingestion rate of 18k samples per second, belonging to 56k time
```
brian-brazil: Are you missing a 0 on the 56k?

beorn7: Unfortunately no. With more time series (and thus more time series files), you hit a cliff where all of a sudden, you cannot write more than 100 chunks a sec anymore, but more something like 50 (or, if there is other load, only 30). Perhaps the file system knows how to optimize smaller amount of files or data, or, more likely, the write cache spans the writes to 56k files nicely, but not 560k. With my latest tweaks, everything still works with 560k, because Prometheus is doing the queuing itself (if you have enough memory chunks configured, say 3x number of time series), but then you have millions of chunks in the persist queue. If you want to shutdown the server, those millions of chunks have to be persisted, which will take a long time (O(hours)). So it's not a scenario you would recommend to users right now. See https://github.com/prometheus/prometheus/issues/533 for the details how to remedy the situation.

PR #43, opened 2015-02-24
Document histograms and histogram_quantile.

beorn7:
@juliusv @brian-brazil

### DIFF
--- a/content/docs/concepts/metric_types.md
+++ b/content/docs/concepts/metric_types.md
@@ -63,13 +63,14 @@ during a scrape:
   * the **count** of events that have been observed, exposed as `<basename>_count` (identical to `<basename>_bucket{le="+Inf"}` above)
 
 Use the [`histogram_quantile()`
-function](/docs/querying/functions/#histogram_quantile()) to calculate quantiles
-from histograms or even aggregations of histograms. A histogram is also suitable
-to calculate an [Apdex score](http://en.wikipedia.org/wiki/Apdex). See [summary
-vs. histogram](/docs/practices/instrumentation/#summary-vs.-histogram) for
-details of histogram usage and differences to [summaries](#summary).
+function](/docs/querying/functions/#histogram_quantile()) to calculate
+quantiles from histograms or even aggregations of histograms. A
+histogram is also suitable to calculate an [Apdex
+score](http://en.wikipedia.org/wiki/Apdex). See [histograms and
+summaries](/docs/practices/histograms) for details of histogram usage
+and differences to [summaries](#summary).
 
-Client library usage documentation for summaries:
+Client library usage documentation for histograms:
 
    * [Go](http://godoc.org/github.com/prometheus/client_golang/prometheus#Histogram)
    * [Java](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Histogram.java) (histograms are only supported by the simple client but not by the legacy client)
@@ -79,7 +80,7 @@ Client library usage documentation for summaries:
 
 Similar to a _histogram_, a _summary_ samples observations (usually things like
 request durations and response sizes). While it also provides a total count of
-observation and a sum of all observed values, it calculates configurable
+observations and a sum of all observed values, it calculates configurable
 quantiles over a sliding time window.
 
 A summary with a base metric name of `<basename>` exposes multiple time series
@@ -89,9 +90,9 @@ during a scrape:
   * the **total sum** of all observed values, exposed as `<basename>_sum`
   * the **count** of events that have been observed, exposed as `<basename>_count`
 
-See [summary
-vs. histogram](/docs/practices/instrumentation/#summary-vs.-histogram) for
-details of summary usage and differences to [histograms](#histogram).
+See [histograms and summaries](/docs/practices/histograms) for
+detailed explanations of φ-quantiles, summary usage, and differences
+to [histograms](#histogram).
 
 Client library usage documentation for summaries:
 
--- a/content/docs/concepts/metric_types.md
+++ b/content/docs/concepts/metric_types.md
@@ -24,7 +24,7 @@ tasks completed, errors occurred, etc. Counters should not be used to expose
 current counts of items whose number can also go down, e.g. the number of
 currently running goroutines. Use gauges for this use case.
 
-See the client library usage documentation for counters:
+Client library usage documentation for counters:
 
    * [Go](http://godoc.org/github.com/prometheus/client_golang/prometheus#Counter)
    * [Java](https://github.com/prometheus/client_java/blob/master/client/src/main/java/io/prometheus/client/metrics/Counter.java)
@@ -41,7 +41,7 @@ Gauges are typically used for measured values like temperatures or current
 memory usage, but also "counts" that can go up and down, like the number of
 running goroutines.
 
-See the client library usage documentation for gauges:
+Client library usage documentation for gauges:
 
    * [Go](http://godoc.org/github.com/prometheus/client_golang/prometheus#Gauge)
    * [Java](https://github.com/prometheus/client_java/blob/master/client/src/main/java/io/prometheus/client/metrics/Gauge.java)
@@ -49,26 +49,51 @@ See the client library usage documentation for gauges:
    * [Ruby](https://github.com/prometheus/client_ruby#gauge)
    * [Python](https://github.com/prometheus/client_python#gauge)
 
-## Summaries
+## Histogram
 
-A _summary_ samples observations (usually things like request durations) over
-sliding windows of time and provides instantaneous insight into their
-distributions, frequencies, and sums.
+A _histogram_ samples observations (usually things like request durations or
+response sizes) and counts them in configurable buckets. It also provides a sum
+of all observed values.
+
+A histogram with a base metric name of `<basename>` exposes multiple time series
+during a scrape:
+
+  * cumulative counters for the observation buckets, exposed as `<basename>_bucket{le="<upper inclusive bound>"}`
+  * the **total sum** of all observed values, exposed as `<basename>_sum`
+  * the **count** of events that have been observed, exposed as `<basename>_count` (identical to `<basename>_bucket{le="+Inf"}` above)
+
+Use the [`histogram_quantile()`
+function](/docs/querying/functions/#histogram_quantile()) to calculate quantiles
+from histograms or even aggregations of histograms. A histogram is also suitable
+to calculate an [Apdex score](http://en.wikipedia.org/wiki/Apdex). See [summary
+vs. histogram](/docs/practices/instrumentation/#summary-vs.-histogram) for
+details of histogram usage and differences to [summaries](#summary).
+
+Client library usage documentation for summaries:
+
+   * [Go](http://godoc.org/github.com/prometheus/client_golang/prometheus#Histogram)
+   * [Java](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Histogram.java) (histograms are only supported by the simple client but not by the legacy client)
+   * [Python](https://github.com/prometheus/client_python#histogram)
+
+## Summary
+
+Similar to a _histogram_, a _summary_ samples observations (usually things like
+request durations and response sizes). While it also provides a total count of
+observation and a sum of all observed values, it calculates configurable
+quantiles over a sliding time window.
 
 A summary with a base metric name of `<basename>` exposes multiple time series
 during a scrape:
 
-  * streaming **quantile values** of observed events, exposed as `<basename>{quantile="<quantile label>"}`
+  * streaming **φ-quantiles** (0 ≤ φ ≤ 1) of observed events, exposed as `<basename>{quantile="<φ>"}`
   * the **total sum** of all observed values, exposed as `<basename>_sum`
   * the **count** of events that have been observed, exposed as `<basename>_count`
 
-This is quite convenient, for if you are interested in tracking latencies of an
-operation in real time, you get three types of information reported for free
-with one metric.
+See [summary
+vs. histogram](/docs/practices/instrumentation/#summary-vs.-histogram) for
+details of summary usage and differences to [histograms](#histogram).
 
-A typical use-case is the observation of request latencies or response sizes.
-
-See the client library usage documentation for summaries:
+Client library usage documentation for summaries:
 
    * [Go](http://godoc.org/github.com/prometheus/client_golang/prometheus#Summary)
    * [Java](https://github.com/prometheus/client_java/blob/master/client/src/main/java/io/prometheus/client/metrics/Summary.java)
--- a/content/docs/practices/histograms.md
+++ b/content/docs/practices/histograms.md
@@ -1,0 +1,220 @@
+---
+title: Histograms and summaries
+sort_rank: 4
+---
+
+# Histograms and summaries
+
+Histograms and summaries are more complex metric types. Not only
+creates a single histogram or summary a multitude of time series, it
+is also more difficult to use them correctly. This section helps you
+to pick and configure the appropriate metric type for your use case.
+
+## Library support
+
+First of all, check the library support for
+[histograms](/docs/concepts/metric_types/#histogram) and
+[summaries](/docs/concepts/metric_types/#summary). Full support for
+both currently only exists in the Go client library. Many libraries
+support only one of the two types, or they support summaries only in a
+limited fashion (lacking [quantile
+calculation](#quantiles)). [Contributions are welcome](/community/),
+of course. In general, we expect histograms to be more urgently needed
+than summaries. Histograms are also easier to implement in a client
+library, so we recommend to implement histograms first, if in
+doubt. The reason why some libraries offer summaries but not
+histograms (Ruby, the legacy Java client) is that histograms are a
+more recent feature of Prometheus.
+
+## Count and sum of observations
+
+Histograms and summaries both sample observations, typically request
+durations or response sizes. They track the number of observations
+*and* the sum of the observed values, allowing you to calculate the
+*average* of the observed values. Note that the number of observations
+(showing up in Prometheus as a time series with a `_count` suffix) is
+inherently a counter (as described above, it only goes up). The sum of
+observations (showing up as a time series with a `_sum` suffix)
+behaves like a counter, too, as long as all observations are
+positive. Obviously, request durations or response sizes are always
+positive. In principle, however, you can use summaries and histograms
+to observe negative values (e.g. temperatures in centigrade). In that
+case, the sum of observations can go down, so you cannot apply
+`rate()` to it anymore.
+
+To calculate the average request duration during the last 5 minutes
+from a histogram or summary called `http_request_duration_second`, use
+the following expression:
+
+    rate(http_request_duration_seconds_sum[5m])
+      /
+    rate(http_request_duration_seconds_count[5m])
+
+## Apdex score
+
+A straight-forward use of histograms (but not summaries) is to count
+observations falling into particular buckets of observation
+values.
+
+You might have an SLA to serve 95% of requests within 300ms. In that
+case, configure a histogram to have a bucket with an upper limit of
+0.3 seconds. You can then directly express the relative amount of
+requests served within 300ms and easily alert if the value drops below
+0.95. The following expression calculates it by job for the requests
+served in the last 5 minutes. The request durations were collected with
+a histogram called `http_request_duration_seconds`.
+
+    sum(rate(http_request_duration_seconds_bucket{le="0.3"}[5m])) by (job)
+      /
+    sum(rate(http_request_duration_seconds_count[5m])) by (job)
+
+
+You can calculate the well-known [Apdex
+score](http://en.wikipedia.org/wiki/Apdex) in a similar way. Configure
+a bucket with the target request duration as upper bound and another
+bucket with the tolerated request duration (usually 4 times the target
+request duration) as upper bound. Example: The target request duration
+is 300ms. The tolerable request duration is 1.2s. The following
+expression yields the Apdex score over the last 5 minutes:
+
+    (
+      rate(http_request_duration_seconds_bucket{le="0.3"}[5m])
+        +
+      rate(http_request_duration_seconds_bucket{le="1.2"}[5m])
+    ) / 2 / rate(http_request_duration_seconds_count[5m])
+
+## Quantiles
+
+You can use both summaries and histograms to calculate so-called φ-quantiles,
+where 0 ≤ φ ≤ 1. The φ-quantile is the observation value that ranks at number
+φ*N among the N observations. Examples for φ-quantiles: The 0.5-quantile is
+known as the median. The 0.95-quantile is the 95th percentile.
+
+The essential difference between summaries and histograms is that summaries
+calculate streaming φ-quantiles on the client side and expose them directly,
+while histograms expose bucketed observations counts and the calculation of
+quantiles from the buckets of a histogram happens on the server side using the
+[`histogram_quantile()`
+function](/docs/querying/functions/#histogram_quantile()).
+
+The two approaches have a number of different implications:
+
+|   | Histogram | Summary
+|---|-----------|---------
+| Required configuration | Pick buckets suitable for the expected range of observed values. | Pick desired φ-quantiles and sliding window. Other φ-quantiles and sliding windows cannot be calculated later.
+| Client performance | Observations are very cheap as they only need to increment counters. | Observations are expensive due to the streaming quantile calculation.
+| Server performance | The server has to calculate quantiles. You can use [recording rules](/docs/querying/rules/#recording-rules) should the ad-hoc calculation take too long (e.g. in a large dashboard). | Low server-side cost.
+| Number of time series (in addition to the `_sum` and `_count` series) | One time series per configured bucket. | One time series per configured quantile.
+| Quantile error (see below for details) | Error is limited in the dimension of observed values by the width of the relevant bucket. | Error is limited in the dimension of φ by a configurable value.
+| Specification of φ-quantile and sliding time-window | Ad-hoc with [Prometheus expressions](/docs/querying/functions/#histogram_quantile()). | Preconfigured by the client.
+| Aggregation | Ad-hoc with [Prometheus expressions](/docs/querying/functions/#histogram_quantile()). | In general [not aggregatable](http://latencytipoftheday.blogspot.de/2014/06/latencytipoftheday-you-cant-average.html).
+
+Note the importance of the last item in the table. Let us return to
+the SLA of serving 95% of requests within 300ms. This time, you do not
+want to display the percentage of requests served within 300ms, but
+instead the 95th percentile, i.e. the request duration within which
+you have served 95% of requests. To do that, you can either configure
+a summary with a 0.95-quantile and (for example) a 5-minute decay
+time-window, or you configure a histogram with a few buckets around
+the 300ms mark, e.g. `{le="0.1"}`, `{le="0.2"}`, `{le="0.3"}`, and
+`{le="0.45"}`. If your service runs replicated with a number of
+instances, you will collect request durations from every single one of
+them, and then you want to aggregate everything into an overall 95th
+percentile. However, aggregating the precomputed quantiles from a
+summary rarely makes sense. In this particular case, averaging the
+quantiles yields statistically nonsensical values.
+
+    avg(http_request_duration_seconds{quantile="0.95"}) // BAD!
+
+Using histograms, the aggregation is perfectly possible with the
+[`histogram_quantile()`
+function](/docs/querying/functions/#histogram_quantile()).
+
+    histogram_quantile(0.95, sum(rate(http_request_duration_seconds_bucket[5m])) by (le)) // GOOD.
+
+Furthermore, should your SLA change and you now want to plot the 90th
+percentile, or you want to take into account the last 10 minutes
+instead of the last 5 minutes, you only have to adjust the expression
+above and you do not need to reconfigure the clients.
+
+## Errors of quantile estimation
+
+Quantiles, whether calculated client-side or server-side, are
+estimated. It is important to understand the errors of that
+estimation.
+
+Continuing the histogram example from above, imagine your usual
+request durations are almost all very close to 220ms, or in other
+words, if you could plot the "true" histogram, you would see a very
+sharp spike at 220ms. In the Prometheus histogram metric as configured
+above, almost all observations, and therefore also the 95th percentile,
+will fall into the bucket labeled `{le="0.3"}`, i.e. the bucket from
+200ms to 300ms. The histogram implementation guarantees that the true
+95th percentile is somewhere between 100ms and 200ms. To return a
+single value (rather than an interval), it applies linear
+interpolation, which yields 295ms in this case. The calculated
+quantile gives you the impression that you are close to breaking the
+SLA, but in reality, the 95th percentile is a tiny bit above 220ms,
+a quite comfortable distance to your SLA.
+
+Next step in our *Gedenkenexperiment*: A change in backend routing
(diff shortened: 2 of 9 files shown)


content/docs/practices/instrumentation.md
```diff
@@ -206,11 +206,55 @@ Gauges can be set, go up, and go down. They are useful for snapshots of state,
...
-aggregatable](http://latencytipoftheday.blogspot.de/2014/06/latencytipoftheday-you-cant-average.html).
+### Summary vs. histogram
+
+Summaries and histograms are more complex metric types. They both sample
+observations. They track the number of observations *and* the sum of the
+observed values, allowing you to calculate the average observed value (useful
+for latency, for example). Note that the number of observations (showing up in
+Prometheus as a time series with a `_count` suffix) is inherently a counter (as
+described above, it only goes up), while the sum of observations (showing up as
+a time series with a `_sum` suffix) is inherently a gauge (if a negative value
```
brian-brazil: This is a bit messy, the cases where it won't be a counter are very small - and in that case the _sum is of questionable use.

content/docs/practices/instrumentation.md
```diff
@@ -206,11 +206,55 @@ Gauges can be set, go up, and go down. They are useful for snapshots of state,
...
+server side using the [`histogram_quantile()`
+function](/docs/querying/functions/#histogram_quantile()).
+
+Both approaches have specific advantages and disadvantages:
+
+|   | Histogram | Summary
+|---|-----------|---------
+| Configuration | Need to configure buckets suitable for the expected range of observed values. | Need to configure φ-quantiles and sliding window, other φ-quantiles and sliding windows cannot be calculated later.
+| Client performance | Observations are very cheap as they only need to increment counters. | Observations are expensive due to the streaming quantile calculation.
+| Server performance | Calculating quantiles is expensive, consider [recording rules](/docs/querying/rules/#recording-rules) as a remedy. | Very low resource needs.
```
brian-brazil: Do we know how expensive calculating quantiles is?

content/docs/practices/instrumentation.md
```diff
@@ -206,11 +206,55 @@ Gauges can be set, go up, and go down. They are useful for snapshots of state,
...
+| Number of time series | Low for Apdex score (see below), very high for accurate quantiles. Each bucket creates a time series. | Low, one time series per configured quantile.
+| Accuracy | Depends on number and layout of buckets. Higher accuracy requires more time series. | Configurable. Higher accuracy requires more client resources but is relatively cheap.
+| Specification of φ-quantile and sliding time window | Ad-hoc in Prometheus expressions. | Preconfigured by the client.
+| Aggregation | Ad-hoc aggregation with [Prometheus expressions](/docs/querying/functions/#histogram_quantile()). | In general [not aggregatable](http://latencytipoftheday.blogspot.de/2014/06/latencytipoftheday-you-cant-average.html).
+
```
brian-brazil: I think we should make mention that in practice having exact quantile values isn't necessary for most use cases, having an indication of which direction it's moving in is the more useful aspect and that can be at a much lower precision.

content/docs/practices/instrumentation.md
```diff
@@ -206,11 +206,55 @@ Gauges can be set, go up, and go down. They are useful for snapshots of state,
...
+score](http://en.wikipedia.org/wiki/Apdex). Configure a bucket with the target
+request duration as upper bound and another bucket with 4 times the request
+duration as upper bound. Example: The target request duration is 250ms. The
+tolerable request duration is 1s. The request duration are collected with a
+histogram called `http_request_duration_seconds`. The following expression
+yields the Apdex score:
+
+```
+(
+  http_request_duration_seconds_bucket{le="0.25"} + http_request_duration_seconds_bucket{le="1"}
```
brian-brazil: It could be le="1.0" with some clients, not sure we should mention that.

We should be taking rates of these and aggregating up

content/docs/querying/functions.md
```diff
@@ -28,6 +28,12 @@ the 1-element output vector from the input vector:
 This is useful for alerting on when no time series
 exist for a given metric name and label combination.
 
+## `bottomk()`
+
+`bottomk(k integer, v instant-vector` returns the `k` smallest elements of `v`
```
brian-brazil: missing a )

content/docs/querying/functions.md
```diff
@@ -80,6 +86,55 @@ and value across all series in the input vector.
 `floor(v instant-vector)` rounds the sample values of all elements in `v` down
 to the nearest integer.
 
+## `histogram_quantile()`
+
+`histogram_quantile(φ float, b instant-vector)` calculates the φ-quantile (0 ≤ φ
```
brian-brazil: I think using phi might scare people off

content/docs/querying/functions.md
```diff
@@ -80,6 +86,55 @@ and value across all series in the input vector.
...
+calculate the 90th percentile of request durations over the last 10m, use the
+following expression:
+
+```
+histogram_quantile(0.9, rate(http_request_duration_seconds_bucket[10m]))
+```
+
+The quantile is calculated for each label combination in
+`http_request_duration_seconds`. To aggregate, use the `sum()` aggregator
+outside of the `rate()` function. Since the `le` label is required by
```
brian-brazil: s/outside/around/ - an example maybe? Probably belongs in the rules best practices.

content/docs/practices/instrumentation.md
beorn7: In the next commit, I'll try to make the usual case clearer without denying the possibility of negative observations.

content/docs/practices/instrumentation.md
beorn7: It's two orders of magnitude slower than rendering a time series straight... but the latter is very fast... I didn't want to make the table even more verbose so I didn't want to start a scientific discussion there about the exact costs...

content/docs/practices/instrumentation.md
beorn7: I think we should not speculate on what the users want or need. Let's wait how the histograms work in practice, and with a bit of experience, we might make bolder statements here.

content/docs/practices/instrumentation.md
beorn7: *gulp* (about the 1.0 vs. 1)

And fixed the rates.

content/docs/querying/functions.md
beorn7: Done.

content/docs/querying/functions.md
beorn7: It's the usual symbol for a quantile value. The problem is that usually, quantiles are named the "4-quantile" (for quartiles) and such. specifying the fraction, as we do, is not so common, but if it is done, it's oftes connected to φ. So I think it creates the least amount of confusion.

content/docs/querying/functions.md
beorn7: Done.

(I'll leave the best practice part to a different PR.)

content/docs/practices/instrumentation.md
brian-brazil: The current wording makes it seem that calculating quantiles is very expensive, can we make it less so?

content/docs/practices/instrumentation.md
brian-brazil: I'm going off my own experience working similar things in the past. You'd be surprised how much you can do with just a handful of buckets.

content/docs/querying/functions.md
brian-brazil: My concern is that most people are not good at math and are scared by anything that's more complex than addition and multiplication.

Using Greek letters is going to make many users discard this function (and possibly all of Prometheus) out of hand.

This documentation needs to not do that, we need to aim for the lowest-common denominator for at least the first paragraph.

juliusv: I agree. Greek letters are "cool", but scary for the uninitiated, and I think very few Prometheus users will know the meaning of φ.

beorn7: Happy to insert your suggestion here.

content/docs/practices/instrumentation.md
beorn7: Sure. But I have to argue with 100+ Soundcloud engineers, some of them very opinionated. So I'll prefer not to open a discussion front here. Stating that you can't aggregate pre-calculated quantiles in a meaningful way is a much more convincing argument (because of... math!) then opening that can of worms in which situation you want which level of accuracy...

content/docs/practices/instrumentation.md
beorn7: But it is pretty expensive. At least if you plot 2000 quantiles in one graph. You wouldn't believe what otherwise reasonable people do over here once they have the power of Prometheus in their mortal hands... :)

brian-brazil: Doing 2000 of anything is slow, that doesn't say much about histogram_quantile (I'd expect the timeseries lookups to dominate that particular query for example).
This language needs softening, or at least for this row to use the same language as the previous row (cheap vs expensive).

content/docs/practices/instrumentation.md
```diff
@@ -206,11 +206,61 @@ Gauges can be set, go up, and go down. They are useful for snapshots of state,
...
+function](/docs/querying/functions/#histogram_quantile()).
+
+Both approaches have specific advantages and disadvantages:
+
+|   | Histogram | Summary
+|---|-----------|---------
+| Configuration | Need to configure buckets suitable for the expected range of observed values. | Need to configure φ-quantiles and sliding window, other φ-quantiles and sliding windows cannot be calculated later.
+| Client performance | Observations are very cheap as they only need to increment counters. | Observations are expensive due to the streaming quantile calculation.
+| Server performance | Calculating quantiles is expensive, consider [recording rules](/docs/querying/rules/#recording-rules) as a remedy. | Very low resource needs.
+| Number of time series | Low for Apdex score (see below), very high for accurate quantiles. Each bucket creates a time series. | Low, one time series per configured quantile.
```
brian-brazil: > very high for accurate quantiles

Citation needed.

"High if you have many buckets", accuracy is a separate question from precision.

content/docs/practices/instrumentation.md
brian-brazil: I think we should generally aim our docs at new users, and encourage them down the right path where possible. 

What we want to avoid is users summarily choosing summaries as they think "perfect accuracy is critical", without understanding the broader implications or indeed that it isn't critical in most use cases (and if it is critical, you probably shouldn't be using Prometheus for it).

beorn7:
I'll comment and rework the text ASAP. Just very busy with other stuff right now. Will get back to this tomorrow or so.

content/docs/concepts/metric_types.md
```diff
@@ -49,26 +49,51 @@ See the client library usage documentation for gauges:
...
+  * the **count** of events that have been observed, exposed as `<basename>_count` (identical to `<basename>_bucket{le="+Inf"}` above)
+
+Use the [`histogram_quantile()`
+function](/docs/querying/functions/#histogram_quantile()) to calculate quantiles
+from histograms or even aggregations of histograms. A histogram is also suitable
+to calculate an [Apdex score](http://en.wikipedia.org/wiki/Apdex). See [summary
+vs. histogram](/docs/practices/instrumentation/#summary-vs.-histogram) for
+details of histogram usage and differences to [summaries](#summary).
+
+Client library usage documentation for summaries:
```
juliusv: s/for summaries/for histograms/

content/docs/concepts/metric_types.md
```diff
@@ -49,26 +49,51 @@ See the client library usage documentation for gauges:
...
+
+   * [Go](http://godoc.org/github.com/prometheus/client_golang/prometheus#Histogram)
+   * [Java](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Histogram.java) (histograms are only supported by the simple client but not by the legacy client)
+   * [Python](https://github.com/prometheus/client_python#histogram)
+
+## Summary
+
+Similar to a _histogram_, a _summary_ samples observations (usually things like
+request durations and response sizes). While it also provides a total count of
+observation and a sum of all observed values, it calculates configurable
```
juliusv: s/observation/observations/

content/docs/practices/instrumentation.md
juliusv: I agree with @brian-brazil here. Users will be looking for some kind of "guidance" to know what actually matters, and we might either scare them or send them on the wrong path otherwise.

content/docs/practices/instrumentation.md
```diff
@@ -206,11 +206,61 @@ Gauges can be set, go up, and go down. They are useful for snapshots of state,
...
+probably collect request durations from every single instance in your fleet, and
+then you want to aggregate everything into an overall 95th percentile. You can
+only do that with histograms, but not with summaries. Aggregating the
+precomputed quantiles from a summary rarely makes sense.
+
+A histogram is suitable to calculate the [Apdex
+score](http://en.wikipedia.org/wiki/Apdex). Configure a bucket with the target
+request duration as upper bound and another bucket with 4 times the request
+duration as upper bound. Example: The target request duration is 250ms. The
+tolerable request duration is 1s. The request duration are collected with a
```
juliusv: s/duration/durations/

content/docs/querying/basics.md
```diff
@@ -107,6 +107,30 @@ a `job` label set to `prometheus`:
...
 
+### Offset modifier
+
+The `offset` modifier allows changing the time offset for individual
+instant and range vectors in a query.
+
+For example, the following expression returns the value of `foo` 5
+minutes in the past relative to the current query evaluation time:
+
+    foo offset 5m
```
juliusv: For the official docs, we might want to use a name other than `foo` that is more instantly recognizable as a metric name.

Like `api_http_requests_total`.

It would also be good to include a `{...}` label selector in at least one of these examples so that people see that that syntactical combination is possible: `api_http_requests_total{method="GET"} offset 5m`

juliusv: (actually, s/api_http_requests_total/http_requests_total/, since that is the example metric used multiple times above.

content/docs/querying/functions.md
juliusv: Just `q` or `quantile`?

content/docs/querying/functions.md
```diff
@@ -80,6 +86,55 @@ and value across all series in the input vector.
...
+histogram_quantile(0.9, sum(rate(http_request_duration_seconds_bucket[10m])) by (job, le))
+```
+
+To aggregate everything, specify only the `le` label:
+
+```
+histogram_quantile(0.9, sum(rate(http_request_duration_seconds_bucket[10m])) by (le))
+```
+
+The `histogram_quantile()` interpolates quantile values by assuming a linear
```
juliusv: "The `histogram_quantile()` function"?

content/docs/practices/instrumentation.md
beorn7: I'd expect that it will be dominated by quantile calculation. But that's not the point.

The point is that users will have done something heavy with summaries, and if they do the same with histograms, things will be slower. They need to be aware of that. The remedy is already in the table.

I have reworded the whole table anyway, though (wait for next commit). I aim to keep it factual but short (and neither soft nor harsh).

content/docs/practices/instrumentation.md
beorn7: I would like to remind you that we already had bucketed histograms in a very early version of the client library. The whole shebang with precalculated quantiles stemmed from the bad experience with the bucketed approach. Histograms are hard to use correctly. We can't change that. By giving newbies some handwavy "guidance", we'll make things worse. Claiming that accuracy wrt buckets will not matter so much is like telling them, they could just do weighted average of their precalculated quantiles. The latter will also be "almost" correct in normal cases, and it will also give them an "indication of which direction it's moving". The big problem with all of that is that the "normal" cases are not the one where you have to rely on your monitoring. You need your monitoring when exploring exceptional cases, and that's exactly when the pointlessness of averaging quantiles strikes. Or, in terms of buckets, when all your relevant percentiles fall into the same bucket, and the distribution within the bucket is so uneven, that the linear interpolation we apply here is misleading at best.
Scientific ethics requires that we don't gloss over those relevant facts. (I feel a bit like Matt now... :)

In the next commit, you'll find a reworded version that goes into a bit more depth about those error cases.

content/docs/practices/instrumentation.md
beorn7: I went down the loser's way and changed the values to avoid 1.0. We should keep this thing in mind and think about unified number formatting with client libraries.

content/docs/querying/functions.md
beorn7: As said, that will lure people down the more common way of a '4-quantile' (for 25th percentile). They will get weird results once they try to run `histogram_quantile(4, series)`.

Again, I don't think we can gain anything by pretending a hard thing is simple.

In the next commit, I add a section that explains φ-quantiles, hopefully in a way everybody understands it (or let's say only people will not understand it that should better not touch histograms at all).

content/docs/practices/instrumentation.md
beorn7: This will be completely reworded in the next commit.

content/docs/concepts/metric_types.md
beorn7: done.

content/docs/practices/instrumentation.md
beorn7: Done.

content/docs/querying/functions.md
juliusv: Oooook.

content/docs/querying/basics.md
beorn7: Done.

content/docs/querying/functions.md
beorn7: Done.

beorn7:
Still polishing.... commit might only happen tonight. Stay tuned. :)

beorn7:
Sorry for the longer-then-expected delay... Here is the heavily reworked version.

Most importantly, I have created a completely new section for the delicate subject of histograms and
summaries and updated all the references.

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,220 @@
+---
+title: Histograms and summaries
+sort_rank: 4
+---
+
+# Histograms and summaries
+
+Histograms and summaries are more complex metric types. Not only
+creates a single histogram or summary a multitude of time series, it
```
brian-brazil: Grammar is off here

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,220 @@
...
+
+Histograms and summaries both sample observations, typically request
+durations or response sizes. They track the number of observations
+*and* the sum of the observed values, allowing you to calculate the
+*average* of the observed values. Note that the number of observations
+(showing up in Prometheus as a time series with a `_count` suffix) is
+inherently a counter (as described above, it only goes up). The sum of
+observations (showing up as a time series with a `_sum` suffix)
+behaves like a counter, too, as long as all observations are
+positive. Obviously, request durations or response sizes are always
```
brian-brazil: s/positive/not negative/

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,220 @@
...
+to observe negative values (e.g. temperatures in centigrade). In that
+case, the sum of observations can go down, so you cannot apply
+`rate()` to it anymore.
+
+To calculate the average request duration during the last 5 minutes
+from a histogram or summary called `http_request_duration_second`, use
+the following expression:
+
+    rate(http_request_duration_seconds_sum[5m])
+      /
```
brian-brazil: This formatting will be hard to read when nested, use:
  a
/
  b

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,220 @@
...
+    sum(rate(http_request_duration_seconds_count[5m])) by (job)
+
+
+You can calculate the well-known [Apdex
+score](http://en.wikipedia.org/wiki/Apdex) in a similar way. Configure
+a bucket with the target request duration as upper bound and another
+bucket with the tolerated request duration (usually 4 times the target
+request duration) as upper bound. Example: The target request duration
+is 300ms. The tolerable request duration is 1.2s. The following
+expression yields the Apdex score over the last 5 minutes:
```
brian-brazil: The above example is across all instances, this one is per instance. We should probably have both across all.

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,220 @@
...
+will fall into the bucket labeled `{le="0.3"}`, i.e. the bucket from
+200ms to 300ms. The histogram implementation guarantees that the true
+95th percentile is somewhere between 100ms and 200ms. To return a
+single value (rather than an interval), it applies linear
+interpolation, which yields 295ms in this case. The calculated
+quantile gives you the impression that you are close to breaking the
+SLA, but in reality, the 95th percentile is a tiny bit above 220ms,
+a quite comfortable distance to your SLA.
+
+Next step in our *Gedenkenexperiment*: A change in backend routing
```
brian-brazil: You seem to have accidentally slipped into German here :)

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,220 @@
...
+dimension of the observed value (via choosing the appropriate bucket
+layout). With a broad distribution, small changes in φ result in
+large deviations in the observed value. With a sharp distribution, a
+small interval of observed values covers a large interval of φ.
+
+Two rules of thumb:
+
+  1. If you need to aggregate, choose histograms.
+
+  2. Otherwise, choose a histogram if you need accuracy in the
```
brian-brazil: This is a bit wordy for a rule of thumb.

Maybe something along the lines of:
choose a histogram if you have an idea of the range/distribution of values that will be seen and have a good idea of the buckets to use. Choose a summary if you need an accurate quartile, no matter what the range/distribution of the values is.

content/docs/querying/functions.md
```diff
@@ -80,6 +86,54 @@ and value across all series in the input vector.
 `floor(v instant-vector)` rounds the sample values of all elements in `v` down
 to the nearest integer.
 
+## `histogram_quantile()`
+
+`histogram_quantile(φ float, b instant-vector)` calculates the
```
brian-brazil: Can we drop the greek letter from here please? 'q' or 'quantile' should do.

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Library support
+
+First of all, check the library support for
+[histograms](/docs/concepts/metric_types/#histogram) and
+[summaries](/docs/concepts/metric_types/#summary). Full support for
+both currently only exists in the Go client library. Many libraries
+support only one of the two types, or they support summaries only in a
+limited fashion (lacking [quantile
+calculation](#quantiles)). [Contributions are welcome](/community/),
```
brian-brazil: The historical note and call for contributions feels off at the start of the doc, move to the end maybe?

content/docs/practices/histograms.md
beorn7: In which way? Need language assistance...

content/docs/practices/histograms.md
beorn7: Done.

content/docs/practices/histograms.md
beorn7: Thanks. That's exactly the advice I was fishing for.

content/docs/practices/histograms.md
beorn7: Done.

content/docs/practices/histograms.md
beorn7: Not accidentally.

http://www.merriam-webster.com/dictionary/gedankenexperiment

It's very physics... if that feels off too much, I'm happy tho change it.

content/docs/practices/histograms.md
beorn7: Gave it another try.

content/docs/querying/functions.md
beorn7: Surely we can do that, but then we have to write a line further down that `d` stands for the φ in "φ-quantile". It's an additional layer of indirection. I can't see how that will help to reduce confusion.

content/docs/practices/histograms.md
beorn7: Done.

content/docs/practices/histograms.md
juliusv: It's "Gedankenexperiment" though, not "Gedenkenexperiment" :)

But yeah - please strive for maximum clarity in docs, not maximum witfulness. "Thought experiment" or something like that will be easier to understand for most people.

content/docs/practices/histograms.md
juliusv: "Not only creates a single histogram" is missing a subject, or at least it seems that way, because I think you mean "histogram or summary" to be the subject, but the location they appear in makes them seem like the object.

Probably you want to say "Not only do they create a multitude of time series, it is also ..."

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,224 @@
...
+observations (showing up as a time series with a `_sum` suffix)
+behaves like a counter, too, as long as there are no negative
+observations. Obviously, request durations or response sizes are
+never negative. In principle, however, you can use summaries and
+histograms to observe negative values (e.g. temperatures in
+centigrade). In that case, the sum of observations can go down, so you
+cannot apply `rate()` to it anymore.
+
+To calculate the average request duration during the last 5 minutes
+from a histogram or summary called `http_request_duration_second`, use
```
juliusv: s/http_request_duration_second/http_request_duration_seconds/

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,224 @@
...
+a histogram called `http_request_duration_seconds`.
+
+      sum(rate(http_request_duration_seconds_bucket{le="0.3"}[5m])) by (job)
+    /
+      sum(rate(http_request_duration_seconds_count[5m])) by (job)
+
+
+You can calculate the well-known [Apdex
+score](http://en.wikipedia.org/wiki/Apdex) in a similar way. Configure
+a bucket with the target request duration as upper bound and another
```
juliusv: "as upper bound" -> "as the upper bound"

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,224 @@
...
+bucket with the tolerated request duration (usually 4 times the target
+request duration) as upper bound. Example: The target request duration
```
juliusv: Dito.

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,224 @@
...
+## Quantiles
+
+You can use both summaries and histograms to calculate so-called φ-quantiles,
+where 0 ≤ φ ≤ 1. The φ-quantile is the observation value that ranks at number
+φ*N among the N observations. Examples for φ-quantiles: The 0.5-quantile is
+known as the median. The 0.95-quantile is the 95th percentile.
+
+The essential difference between summaries and histograms is that summaries
+calculate streaming φ-quantiles on the client side and expose them directly,
+while histograms expose bucketed observations counts and the calculation of
```
juliusv: "observations counts" -> "observation counts"

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,224 @@
...
+| Specification of φ-quantile and sliding time-window | Ad-hoc with [Prometheus expressions](/docs/querying/functions/#histogram_quantile()). | Preconfigured by the client.
+| Aggregation | Ad-hoc with [Prometheus expressions](/docs/querying/functions/#histogram_quantile()). | In general [not aggregatable](http://latencytipoftheday.blogspot.de/2014/06/latencytipoftheday-you-cant-average.html).
+
+Note the importance of the last item in the table. Let us return to
+the SLA of serving 95% of requests within 300ms. This time, you do not
+want to display the percentage of requests served within 300ms, but
+instead the 95th percentile, i.e. the request duration within which
+you have served 95% of requests. To do that, you can either configure
+a summary with a 0.95-quantile and (for example) a 5-minute decay
+time-window, or you configure a histogram with a few buckets around
```
juliusv: "time-window" -> "time window"

content/docs/practices/histograms.md
beorn7: I see, the "do" is missing. http://dictionary.cambridge.org/grammar/british-grammar/not-only-but-also

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,224 @@
...
+200ms to 300ms. The histogram implementation guarantees that the true
+95th percentile is somewhere between 100ms and 200ms. To return a
+single value (rather than an interval), it applies linear
+interpolation, which yields 295ms in this case. The calculated
+quantile gives you the impression that you are close to breaking the
+SLA, but in reality, the 95th percentile is a tiny bit above 220ms,
+a quite comfortable distance to your SLA.
+
+Next step in our *Gedenkenexperiment*: A change in backend routing
+adds a fixed amount of 100ms to all requent durations. Now the request
```
juliusv: "requent" -> "request"

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,224 @@
...
+duration has its sharp spike at 320ms and almost all observations will
+fall into the bucket from 300ms to 450ms. The 95th percentile is
+calculated to be 442.5ms, although the correct values is close to
```
juliusv: "values" -> "value"

content/docs/practices/histograms.md
```diff
@@ -0,0 +1,224 @@
...
+
+
+## What can I do if my client library does not support the metric type I need?
+
+Implement it! [Code contributions are welcome](/community/). In
+general, we expect histograms to be more urgently needed than
+summaries. Histograms are also easier to implement in a client
+library, so we recommend to implement histograms first, if in
+doubt. The reason why some libraries offer summaries but not
+histograms (Ruby, the legacy Java client) is that histograms are a
```
juliusv: "," -> " and"

content/docs/querying/functions.md
juliusv: Given that we had this discussion last week (and no clear outcome), I think it's ok to keep it as it is for now.

content/docs/practices/histograms.md
juliusv: Well, also the subject is in the wrong position, and the verb is not conjugated correctly? I would just copy my suggestion :)

beorn7: "a single histogram or summary" is clearly singular. A "does" was missing, and then "creates" becomes infinite and goes to a different place. The subject was correctly placed from the beginning.

Should be good now.

content/docs/practices/histograms.md
beorn7: Removed the nerdiness.

content/docs/practices/histograms.md
beorn7: Done.

content/docs/practices/histograms.md
beorn7: Done.

content/docs/practices/histograms.md
beorn7: Done.

content/docs/practices/histograms.md
beorn7: DOne.

content/docs/practices/histograms.md
beorn7: I thought it's a time window for the decay rather than a window of decay time.

I guess it's clearer anyway to call it a '5-minute decay time' without the window.

content/docs/practices/histograms.md
beorn7: Done.

content/docs/practices/histograms.md
beorn7: Doen.

content/docs/practices/histograms.md
beorn7: DOne

beorn7:
PHAL

juliusv:
:+1:

PR #45, opened 2015-02-26
Move by clauses on aggregators to all be in front.

brian-brazil:
With by clauses on binary operators imminent,
mixing the two forms would be confusing. It's
also easier to understand when expressions are deep.

@jrv @fabx

### DIFF
--- a/content/docs/introduction/getting_started.md
+++ b/content/docs/introduction/getting_started.md
@@ -249,7 +249,7 @@ preserving the `job` and `service` dimensions) as measured over a window of 5
 minutes. We could write this as:
 
 ```
-avg(rate(rpc_durations_microseconds_count[5m])) by (job, service)
+avg by (job, service)(rate(rpc_durations_microseconds_count[5m]))
 ```
 
 Try graphing this expression.
@@ -259,7 +259,7 @@ called `job_service:rpc_durations_microseconds_count:avg_rate5m`, create a file
 with the following recording rule and save it as `prometheus.rules`:
 
 ```
-job_service:rpc_durations_microseconds_count:avg_rate5m = avg(rate(rpc_durations_microseconds_count[5m])) by (job, service)
+job_service:rpc_durations_microseconds_count:avg_rate5m = avg by (job, service)(rate(rpc_durations_microseconds_count[5m]))
 ```
 
 To make Prometheus pick up this new rule, add a `rule_files` statement to the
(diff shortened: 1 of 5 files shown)


juliusv:
So in my opinion, that is less readable for the simple cases and will look really weird to newbies ("sum X by Y" vs. "sum by X the Y" just doesn't read like English anymore).

I think it's good that we support both orders, but not sure about changing the docs. Would like to have another opinion, like @beorn7's.

If we do go this route, do we really want no space between the `)(` closing brace of the by-clause and the opening brace of the aggregated expression?

beorn7:
Now we have swapped roles, it seems... :)

I believe things should be accessible for newbies where possible. I also believe, the current by clause syntax in the docs is more accessible for newbies.

I might be convinced otherwise when I see by clauses on binary operators.

juliusv:
I actually wonder if `by` clauses for binary operators should be named `on` instead, because they are more for matching between both left/right sides, than output grouping, like in:

```
node_cpu / group_left(cpu,mode) by(instance) node_boot_time
```

instead:

```
node_cpu / group_left(cpu,mode) on(instance) node_boot_time
```

brian-brazil:
> So in my opinion, that is less readable for the simple cases and will look really weird to newbies ("sum X by Y" vs. "sum by X the Y" just doesn't read like English anymore).

They both read fine to me.

> If we do go this route, do we really want no space between the )( closing brace of the by-clause and the opening brace of the aggregated expression?

I'm unsure what the best spacing is.


> I might be convinced otherwise when I see by clauses on binary operators.

The real issue is when you start nesting expressions, it's not at all obvious which by clause goes with which operator.

grobie:
So I started to get used to @brian-brazil's way to write expressions and putting the by clause in the front. Though, I used a different spacing to make the relation between the `by` keyword and its labels more obvious and, subjectively, more readable.

```
avg by(job, service) (rate(rpc_durations_microseconds_count[5m]))
```

And now I wonder whether we could make the parentheses around the metric clause optional and allow the following:

```
avg by(job, service) rate(rpc_durations_microseconds_count[5m])
```

/me eagerly waits for a blessed formatting style and `promtool fmt`.

fabxc:
@grobie Regarding fmt tool - I'd be happy to have that. There are implementation concerns but it's not worth investing time in those as long as we don't have a clear specification of "the one and only" valid formatting. If you look into `go fmt` you'll see that it can get fairly complex.

So for the most part it needs someone leading the efforts of agreeing on formatting rules.

PR #46, opened 2015-02-26
add mesos_exporter to list of exporters

antonlindstrom:
This commit adds the [mesos_exporter](https://github.com/antonlindstrom/mesos_exporter) for tasks in mesos in the documentation.

Thanks!

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -18,6 +18,7 @@ following is a list of existing third-party exporters:
    * [StatsD bridge](https://github.com/prometheus/statsd_bridge)
    * [AWS CloudWatch exporter](https://github.com/prometheus/cloudwatch_exporter)
    * [Hystrix metrics publisher](https://github.com/prometheus/hystrix)
+   * [Mesos task exporter](https://github.com/antonlindstrom/mesos_exporter)
 
 The [JMX exporter](https://github.com/prometheus/jmx_exporter) can export from a
 wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.org/) and


juliusv:
@antonlindstrom Thanks! So far, the list only has exporters in the `prometheus` organization because those are vetted to a certain degree to adhere to code and metrics conventions. But I think it's time to mention the various external exporters as well. Just wondering if we should mark them as external in the doc (but I can add that later). Or perhaps you're interested in moving this to the `prometheus` org and maintaining it there anyways? Either way is fine...

juliusv:
Btw., it looks like your code is already 95% following our conventions anyways, so we would probably only want to change minor things when including it in the `prometheus` org.

grobie:
What about a github.com/prometheus/exporters repo? Should be easier to handle than tens of repos.

juliusv:
@grobie Then we'd have to be ok with every exporter maintainer to have push access to the entire repo. We might be ok with that, but we should at least give it some thought - especially in the scenarios where such an approach would be most helpful (like 100 exporters), we'd also potentially have a huge amount of people with commit access to everything. At the moment, I don't see a big problem with adding more exporters to the main org yet.

juliusv:
I'm merging this for now because I don't want to hold things up regardless of outcome - I might add some things later to point out `prometheus`-org vs. external repos.

antonlindstrom:
@juliusv Thanks! I don't mind moving it to the prometheus org, just let me know what do to to make it happen.

Cheers!

juliusv:
@antonlindstrom Cool! I'll try to have a closer look soon and then it'd be cool to make it happen :)

juliusv:
@antonlindstrom Hey, if you're still interested in moving this to the `prometheus` GitHub org, we could do https://help.github.com/articles/transferring-a-repository/#transferring-from-a-user-to-an-organization. I think I'll need to give you temporary admin access to the org to make that work. Alternatively, I could create a new repo and give you push access to that. Let me know what you prefer :)

antonlindstrom:
@juliusv Sorry for the late reply. Would be awesome, I'm fine with transferring the repository to the Prometheus Github org. Let me know when I can do it and I'll transfer it right away.

Thanks!

juliusv:
@antonlindstrom Sorry, I'm a bit slow/overloaded these days as well. I made you a temporary admin user on github.com/prometheus, so you should be able to move the repo now! One thing though to clarify, we enforce the same LICENSE/AUTHORS/etc. boilerplate (and copyright "The Prometheus authors") for all projects, in order to stay consistent and not become insane. Just want to check if that's ok with you first. We might also want to change some of the metric names in the 
Mesos exporter a tiny bit to match our conventions (i.e. secs->seconds), but probably nothing major.

antonlindstrom:
@juliusv No worries. Just tried to move and it didn't seem to work. Maybe it takes some time to propagate?

No worries about updating license and authors, you are free to update it as you see fit. Just let me know what to change and I'll do it.

Thanks!

juliusv:
@antonlindstrom Ah, when trying to add you to the Admin group again, GitHub says, "Invitation already sent". So I think you have to accept the invitation first before you're in the group...

antonlindstrom:
@juliusv Ah, thanks! I thought that would show up in the web ui but it didn't. Transferring the repo now.

juliusv:
@antonlindstrom Awesome, thanks! Will send followup PRs about the above-mentioned topics soon!

antonlindstrom:
@juliusv Perfect, thanks!

PR #48, opened 2015-02-27
Documentation for vector matching added.

fabxc:


content/docs/querying/operators.md
```diff
@@ -25,13 +31,13 @@ value of every data sample in the vector. E.g. if a time series instant vector
...
 the original vector is multiplied by 2.
 
-**Between two instant vectors**, a binary arithmetic operator only applies to
-vector elements that have identical sets of labels between the two vectors.
-Vector elements that don't find an exact label match on the other side get
-dropped from the result. The metric name of the result vector is carried over
-from the left hand side of the expression.
+**Between two instant vectors**, a binary arithmetic operator is applied to
+each entry in the left-hand-side vector and its [matching element](#vector-matching)
+in the right hand side. The result is carried over to the element to the left
```
brian-brazil: carried over doesn't make sense here, as the result doesn't have a left side

content/docs/querying/operators.md
```diff
@@ -50,19 +56,18 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
 comparison result is `false` get dropped from the result vector.
 
-**Between two instant vectors**, these operators behave as a filter: They apply
-to vector elements that have identical sets of labels between the two vectors.
-Vector elements for which the expression evaluates to `false` or which don't
-find an exact label match on the other side of the expression get dropped from
-the result, while the others get carried over into a result vector with their
-original (left-hand-side) metric names and data values.
+**Between two instant vectors**, these operators behave as a filter, applied to
+matching entries. Vector elements for which the expression evaluates to false or
```
brian-brazil: "isn't true" rather than "evaluates to false", as otherwise we'd need to explain how the operators work internally

content/docs/querying/operators.md
```diff
@@ -25,13 +31,13 @@ value of every data sample in the vector. E.g. if a time series instant vector
...
-dropped from the result. The metric name of the result vector is carried over
-from the left hand side of the expression.
+**Between two instant vectors**, a binary arithmetic operator is applied to
+each entry in the left-hand-side vector and its [matching element](#vector-matching)
+in the right hand side. The result is carried over to the element to the left
+and the metric name is dropped. Entries for which no matching entry in the
+right vector can be found are dropped from the result.
 
-## Comparison / filter binary operators
+### Comparison/filter binary operators
```
brian-brazil: == too?

content/docs/querying/operators.md
```diff
@@ -71,10 +76,63 @@ over from the left-hand-side vector.
...
+By means of the `on` keyword one can reduce the set of considered labels to a provided list:
+
+  <vector expr> <bin-op> on(<label list>) <vector expr>
+
+_Modified matching methods can only be used for [comparison/filterting](#comparison-/-filter-binary-operators)
+and [airthmetic](#arithmetic-binary-operators) operations._
+
+Example:
+
+  http_errors_total / on(method) http_requests_total
```
brian-brazil: Any examples should follow best practice, so this example needs a sumb by(method)(rate()) on both sides

content/docs/querying/operators.md
```diff
@@ -71,10 +76,63 @@ over from the left-hand-side vector.
...
+
+This returns a result vector of average errors per request for the different
+HTTP methods. Without `on(method)` there would have been no match as the metrics
+do not share the same set of labels, e.g. `http_errors_total` has a label for the
```
brian-brazil: That's not at all clear from the example, you'd have to show the input vectors

content/docs/querying/operators.md
```diff
@@ -71,10 +76,63 @@ over from the left-hand-side vector.
...
+error type.
+
+**Many-to-one** and **one-to-many** matching refer to the case that each vector entry on
+the "one"-side can match with numerous elements on the "many"-side. This has to
+be explicitly enforced using the `group_left` or `group_right` modifier, where
```
brian-brazil: requested rather than enforced

content/docs/querying/operators.md
```diff
@@ -71,10 +76,63 @@ over from the left-hand-side vector.
...
+left/right determines which vector has the higher cardinality.
+
+  <vector expr> <bin-op> group_left(<label list>) on(<label list>) <vector expr>
```
brian-brazil: Per comment elsewhere, `on` should be first. This makes it easier for a human to parse as they don't need to know the various special modifiers to the `on` (group_left/group_right are merely the first of them) to determine if label matching is going on

content/docs/querying/operators.md
```diff
@@ -71,10 +76,63 @@ over from the left-hand-side vector.
...
+the result metrics. The intersection of both lists must be empty. Labels appearing on
+either side of the operation may be used but they must not occur on both. Every time
+series of the result vector must be uniquely identifiable by the labels from both lists
+combined.
+
+Example:
+
+  cpu_rate / group_left(mode) on(instance) cpu_count
+
+In this case the `cpu_rate` vector contains metrics for different CPU modes and instances.
```
brian-brazil: This would benefit from showing the input and output vectors

content/docs/querying/operators.md
fabxc: Would `method:http_errors_total:rate5m` be good as well? Just to reduce some noise.

content/docs/querying/operators.md
```diff
@@ -5,7 +5,13 @@ sort_rank: 2
 
 # Operators
 
-## Arithmetic binary operators
+## Binary operators
+
+Prometheus' query language supports the basic logical and arithmetic operators.
```
juliusv: Though it's not totally clear cut, we say "Prometheus's" for possession (http://grammar.ccc.commnet.edu/grammar/possessives.htm).

I don't know if I'd include a "the" in "the basic logical ...". It might confuse some readers ("the? which ones?"). Maybe just "supports basic logical..."?

content/docs/querying/operators.md
juliusv: "propagated into the result vector"?

juliusv: s/in the right hand side/on the right hand side/

(if it said "in the right hand side vector", "in" would be ok, but if it's just "on the right hand side", it's on)

content/docs/querying/operators.md
juliusv: Oh yeah, this whole page is missing `==`. Would be good to add it :)

content/docs/querying/operators.md
```diff
@@ -50,19 +56,18 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
 
-**Between two instant vectors**, these operators behave as a filter: They apply
-to vector elements that have identical sets of labels between the two vectors.
-Vector elements for which the expression evaluates to `false` or which don't
-find an exact label match on the other side of the expression get dropped from
-the result, while the others get carried over into a result vector with their
-original (left-hand-side) metric names and data values.
+**Between two instant vectors**, these operators behave as a filter, applied to
+matching entries. Vector elements for which the expression evaluates to false or
+which don't find a match on the other side of the expression get dropped from the
```
juliusv: We have a tech-writer rule: no contractions in docs :) don't->do not

content/docs/querying/operators.md
```diff
@@ -71,10 +76,63 @@ over from the left-hand-side vector.
...
+## Vector matching
+
+Operations between vectors attempt to find a matching element in the right-hand-side
+vector for each entry in the left-hand-side. There are two basic types of
+matching behaviour:
+
+**One-to-one** finds a unique pair of entries from each side of the operation.
+In the default case, that is an operation following the format `vector1 <operator> vector2`.
+Two entries match if they have the exact same set of labels and corresponding values.
+By means of the `on` keyword one can reduce the set of considered labels to a provided list:
```
juliusv: Simpler formulation: "The `on` keyword allows reducing the set..."

content/docs/querying/operators.md
```diff
@@ -71,10 +76,63 @@ over from the left-hand-side vector.
...
+
+  <vector expr> <bin-op> on(<label list>) <vector expr>
```
juliusv: This isn't indented sufficiently, thus not showing up as a code block and thus also things in pointy brackets become invisible. Please use [the development server](https://github.com/prometheus/docs#development-server) to visually test the docs.

content/docs/querying/operators.md
```diff
@@ -71,10 +76,63 @@ over from the left-hand-side vector.
...
+
+_Modified matching methods can only be used for [comparison/filterting](#comparison-/-filter-binary-operators)
```
juliusv: s/filterting/filtering/

content/docs/querying/operators.md
```diff
@@ -71,10 +76,63 @@ over from the left-hand-side vector.
...
+and [airthmetic](#arithmetic-binary-operators) operations._
```
juliusv: airthmetic -> arithmetic

content/docs/querying/operators.md
juliusv: Yes, please start out by specifying the input vectors, in a format like in https://github.com/prometheus/prometheus/issues/488#issuecomment-75257886

content/docs/querying/operators.md
fabxc: != too

content/docs/querying/operators.md
fabxc: Just adapted what was given :P

content/docs/querying/operators.md
```diff
@@ -71,10 +76,63 @@ over from the left-hand-side vector.
...
+Example:
+
+  http_errors_total / on(method) http_requests_total
+
+This returns a result vector of average errors per request for the different
+HTTP methods. Without `on(method)` there would have been no match as the metrics
+do not share the same set of labels, e.g. `http_errors_total` has a label for the
+error type.
+
+**Many-to-one** and **one-to-many** matching refer to the case that each vector entry on
```
juliusv: s/matching/matchings/?

Alternatively, s/refer/refers/ to make the verb fit the singularity of the noun.

content/docs/querying/operators.md
juliusv: Oh right! Burn the contractions anyways :P

content/docs/querying/operators.md
juliusv: Indeed! Someone must have been drunk when initially writing these docs. Ehmmm. :)

content/docs/querying/operators.md
juliusv: If we do `by(method)(rate())`, can we do nicer whitespacing, like `by (method) (rate())`. E.g. in my brain's visual pattern matching, a word followed without space by an opening brace is always a function call.

brian-brazil: > Would method:http_errors_total:rate5m be good as well?

It'd be method:http_errors:rate5m, as you remove the _total when taking the rate.

I think getting people into good habits on rule naming would be good here.

fabxc:
I've switched to a single input and extended the example for many-to-one. I think it's easier to digest but in case in causes unclarity in any way, I'll come up with something else.

content/docs/querying/operators.md
```diff
@@ -25,16 +31,18 @@ value of every data sample in the vector. E.g. if a time series instant vector
...
 
-**Between two instant vectors**, a binary arithmetic operator only applies to
-vector elements that have identical sets of labels between the two vectors.
-Vector elements that don't find an exact label match on the other side get
-dropped from the result. The metric name of the result vector is carried over
-from the left hand side of the expression.
+**Between two instant vectors**, a binary arithmetic operator is applied to
+each entry in the left-hand-side vector and its [matching element](#vector-matching)
+on the right hand side. The result is propagated into the result vector and the metric
+name is dropped. Entries for which no matching entry in the right vector can be
```
brian-brazil: right hand vector

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
 
 * `and` (intersection)
-* `or` (union) (**not yet implemented!**)
+* `or` (union)
 
 `vector1 and vector2` results in a vector consisting of the elements of
 `vector1` for which there are elements in `vector2` with exactly matching
-labelsets. Other elements are dropped. The metric name and values are carried
+label sets. Other elements are dropped. The metric name and values are carried
 over from the left-hand-side vector.
```
brian-brazil: this isn't always true for `or`

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+Example query:
+
+    method:http_errors:rate5m{code="500"} / on(method) method:http_requests:rate5m
+
+This returns a result vector that tells us what fraction of requests had an HTTP
+status code of 500 for each method over the last 5 minutes. Without `on(method)` there
+would have been no match as the metrics do not share the same set of labels.
+The entries with methods `put` and `del` have no match and will not show up in the result:
+
+	{method="get"}  0.025 			//  1.64 / 65.6
```
brian-brazil: Maybe keep the examples to integers?

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+
+**Many-to-one** and **one-to-many** matchings refer to the case that each vector entry on
+the "one"-side can match with numerous elements on the "many"-side. This has to
+be explicitly requested using the `group_left` or `group_right` modifier, where
+left/right determines which vector has the higher cardinality.
+
+    <vector expr> <bin-op> on(<label list>) group_left(<label list>) <vector expr>
+    <vector expr> <bin-op> on(<label list>) group_right(<label list>) <vector expr>
+
+The label list provided with the modifier contains additional labels to be included in
```
brian-brazil: "with group modifier" would be clearer I think

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+the result metrics. The intersection of both lists must be empty. Labels appearing on
```
brian-brazil: "A label name can only appear in on list" maybe?

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+either side of the operation may be used but they must not occur on both. Every time
```
brian-brazil: > Labels appearing on either side of the operation may be used but they must not occur on both.

I'm not sure about this sentence, I think it should be dropped.

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+series of the result vector must be uniquely identifiable by the labels from both lists
+combined.
+
+Example query:
+
+    method:http_errors:rate5m / on(method) group_left(code) method:http_requests:rate5m
```
brian-brazil: I'm not sure this example is meaningful.

I'd suggest subtracting a timestamp on a master to the timestamps on the slaves, and leading with the example.

brian-brazil:
Generally looks fine, just a few wording matters.

content/docs/querying/operators.md
fabxc: Why? It's part of the implementation behaviour. I think especially as long as we don't provide error messages everything that may cause failure should be documented somehow.

brian-brazil: I found that statement hard to figure out, and it doesn't quite capture the semantics.

content/docs/querying/operators.md
```diff
@@ -5,7 +5,13 @@ sort_rank: 2
 
 # Operators
 
-## Arithmetic binary operators
+## Binary operators
+
+Prometheus's query language supports basic logical and arithmetic operators.
+For operations between two instant vectors the [matching behaviour](#vector-matching)
```
juliusv: comma after "vectors" (https://owl.english.purdue.edu/owl/resource/607/03/)

content/docs/querying/operators.md
```diff
@@ -17,7 +23,7 @@ The following binary arithmetic operators exist in Prometheus:
 Binary arithmetic operators are defined between scalar/scalar, vector/scalar,
 and vector/vector value pairs.
 
-**Between two scalars**, the behavior is obvious: they evaluate to another
+**Between two scalars**, the behaviour is obvious: they evaluate to another
```
juliusv: Let's stick to US English and keep it as is... otherwise we'll also have to say "colour" and "centre".

content/docs/querying/operators.md
```diff
@@ -25,16 +31,18 @@ value of every data sample in the vector. E.g. if a time series instant vector
...
 the original vector is multiplied by 2.
 
-**Between two instant vectors**, a binary arithmetic operator only applies to
-vector elements that have identical sets of labels between the two vectors.
-Vector elements that don't find an exact label match on the other side get
-dropped from the result. The metric name of the result vector is carried over
-from the left hand side of the expression.
+**Between two instant vectors**, a binary arithmetic operator is applied to
+each entry in the left-hand-side vector and its [matching element](#vector-matching)
+on the right hand vector. The result is propagated into the result vector and the metric
```
juliusv: I think I had a comment here: it's either "in the right hand vector" or "on the right hand [side]", but not "on the right hand vector".

content/docs/querying/operators.md
fabxc: So should we go with double comma in ", i.e.,"/", e.g.," then as well? Otherwise it's becoming a mixture of both.

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
 `vector1` for which there are elements in `vector2` with exactly matching
-labelsets. Other elements are dropped. The metric name and values are carried
+label sets. Other elements are dropped. The metric name and values are carried
 over from the left-hand-side vector.
 
 `vector1 or vector2` results in a vector that contains all original elements
-(labelsets + values) of `vector1` and additionally all elements of `vector2`,
-which don't have matching labelsets in `vector1`. The metric name is carried
-over from the left-hand-side vector in both cases.
+(label sets + values) of `vector1` and additionally all elements of `vector2`,
```
juliusv: remove the comma at the end of this line (it actually changes the meaning of the sentence, saying that elements of vector2 in general don't have matching label sets in vector1)

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+which do not have matching label sets in `vector1`.
+
+## Vector matching
+
+Operations between vectors attempt to find a matching element in the right-hand-side
+vector for each entry in the left-hand-side. There are two basic types of
```
juliusv: s/left-hand-side/left-hand side/ in this case because "side" is not part of the compound adjective anymore here, but the described noun itself (http://www.gingersoftware.com/content/grammar-rules/adjectives/compound-adjectives/).

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+matching behaviour:
```
juliusv: US spelling :)

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+
+    method:http_requests:rate5m{method="get"}  600
+    method:http_requests:rate5m{method="del"}  34
+    method:http_requests:rate5m{method="post"} 120
+
+Example query:
+
+    method:http_errors:rate5m{code="500"} / on(method) method:http_requests:rate5m
+
+This returns a result vector that tells us what fraction of requests had an HTTP
```
juliusv: "tells us" -> "describes" (avoid informal language in reference docs)

Or: "This returns a result vector containing the fraction of HTTP requests with status code of 500 for each method, as measured over the last 5 minutes."

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+status code of 500 for each method over the last 5 minutes. Without `on(method)` there
+would have been no match as the metrics do not share the same set of labels.
+The entries with methods `put` and `del` have no match and will not show up in the result:
+
+	{method="get"}  0.04 			//  24 / 600
+	{method="post"} 0.1				//  12 / 120
+
+**Many-to-one** and **one-to-many** matchings refer to the case that each vector entry on
```
juliusv: s/case that/case where/?

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+Example query:
+
+    method:http_errors:rate5m{code="500"} / on(method) method:http_requests:rate5m
+
+This returns a result vector that tells us what fraction of requests had an HTTP
+status code of 500 for each method over the last 5 minutes. Without `on(method)` there
+would have been no match as the metrics do not share the same set of labels.
+The entries with methods `put` and `del` have no match and will not show up in the result:
+
+	{method="get"}  0.04 			//  24 / 600
```
juliusv: Please use spaces instead of tabs for indent, both at the beginning of the line, as well as before the comment (e.g. in the GitHub diff, this doesn't look aligned anymore because of different tab width).

content/docs/querying/operators.md
juliusv: "entry" -> "element" (consistent terminology)

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+	{method="post"} 0.1				//  12 / 120
+
+**Many-to-one** and **one-to-many** matchings refer to the case that each vector entry on
+the "one"-side can match with numerous elements on the "many"-side. This has to
```
juliusv: numerous -> multiple (numerous has the connotation of "a lot")

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+be explicitly requested using the `group_left` or `group_right` modifier, where
+left/right determines which vector has the higher cardinality.
+
+    <vector expr> <bin-op> on(<label list>) group_left(<label list>) <vector expr>
+    <vector expr> <bin-op> on(<label list>) group_right(<label list>) <vector expr>
+
+The label list provided with the group modifier contains additional labels to be included in
```
juliusv: It might be a good idea to include examples that have multiple labels in `on(...)` and/or `group_x(...)` just so that users see that it's syntactically possible. Don't spend too much energy on that if it's hard to fit in though.

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+the result metrics. A label can only appear in one of the lists. Every time
+series of the result vector must be uniquely identifiable by the labels from both lists
+combined.
+
+_Grouping modifieres can only be used for [comparison/filtering](#comparison-/-filter-binary-operators)
```
juliusv: modifieres -> modifiers

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+Example query:
+
+    method:http_errors:rate5m / on(method) group_left(code) method:http_requests:rate5m
+
+In this case the left vector contains more than one entry per `method` label value. Thus,
+we indicate this using `group_left` and provide `code` as the label that ensures that
+the result vector entries are unique. The elements from the right side are now matched with
+multiple elements with the same `method` label on the left:
+
+    {method="get", code="500"}  0.04			//  24 / 600
```
juliusv: Same thing about tabs here.

content/docs/querying/operators.md
```diff
@@ -50,31 +58,101 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+    {method="get", code="404"}  0.05			//  30 / 600
+    {method="post", code="500"} 0.1				//  12 / 120
+    {method="post", code="404"} 0.175			//  21 / 120
+
+_Many-to-one and one-to-many matching are advanced use cases that should be carefully considered.
+Often times a proper use of `on(<labels>)` provides the desired outcome._
```
juliusv: "Often times" -> "Often" ("Oftentimes" (sp) is colloquial)

juliusv:
Looks great otherwise!

content/docs/querying/operators.md
fabxc: For `group_x` it's no problem. A proper example of `on` with multiple labels makes the example more complex, I think. So I would leave that out to keep the example easy to understand. In the end people have to experiment a bit to get a full understanding anyway.

juliusv: That sounds good.

content/docs/querying/operators.md
juliusv: Oh, I didn't know that rule. Actually, I'm not used to seeing commas after "i.e."/"e.g." - do you perhaps have a link pointing at rules? If that's the way, I'm fine with it.

Though with the British/US thing I'm mostly after the spelling of words. I didn't even know there were comma differences between US/British, except for the Oxford comma, which I think really adds clarity, but can also be used in US English.

juliusv: Oh, and btw., our tech writer recommends also not using abbreviations at all, but using "for example", "that is" (or equivalent), etc. Not super urgent though.

fabxc: http://grammarist.com/usage/ie-eg/
Doesn't seem super strict, though. But I also like the Oxford comma - so I guess going British with commas and US with words is fine. There are also some differences regarding width of dashes and whether or not they are surrounded by spaces.

Usage of full phrases noted.

juliusv: Thanks! Yeah, I think for dashes we'll want to stick with a normal "-" (minus, but surrounded by spaces if it's not a dash joining two words) just for simplicity. It's on everyone's keyboard, and is the most common in docs websites, I think.

content/docs/querying/operators.md
```diff
@@ -25,16 +31,18 @@ value of every data sample in the vector. E.g. if a time series instant vector
...
 
-**Between two instant vectors**, a binary arithmetic operator only applies to
-vector elements that have identical sets of labels between the two vectors.
-Vector elements that don't find an exact label match on the other side get
-dropped from the result. The metric name of the result vector is carried over
-from the left hand side of the expression.
+**Between two instant vectors**, a binary arithmetic operator is applied to
+each entry in the left-hand-side vector and its [matching element](#vector-matching)
+in the right hand vector. The result is propagated into the result vector and the metric
+name is dropped. Entries for which no matching entry in the right vector can be
```
brian-brazil: right hand vector

Without the 'hand', 'right' way be interpreted as 'correct'

content/docs/querying/operators.md
```diff
@@ -50,31 +58,102 @@ either `0` (`false`) or `1` (`true`), depending on the comparison result.
...
+(label sets + values) of `vector1` and additionally all elements of `vector2`
+which do not have matching label sets in `vector1`.
+
+## Vector matching
+
+Operations between vectors attempt to find a matching element in the right-hand-side
+vector for each entry in the left-hand side. There are two basic types of
+matching behavior:
+
+**One-to-one** finds a unique pair of entries from each side of the operation.
```
brian-brazil: Can we add a note or something that you almost always want one-to-one?

content/docs/querying/operators.md
brian-brazil: I'd add job labels to the existing examples

brian-brazil:
:+1:

content/docs/querying/operators.md
juliusv: But then "right-hand vector", not "right hand vector" please :)

content/docs/querying/operators.md
fabxc: At the section's end: "Many-to-one and one-to-many matching are advanced use cases that should be carefully considered. Often a proper use of `on(<labels>)` provides the desired outcome."

I can move it higher up, though.

juliusv:
(and this is a good candidate for squashing before merging)

juliusv:
Merging, thanks a lot! :+1:

PR #51, opened 2015-03-04
Fix usage of include labels in group modifiers.

fabxc:
The final change to the group modifiers requires included labels
to be on the "many"-side of the operation. This adjusts the example
and description to match this behavior.


juliusv:
Thanks, totally forgot about this one.

Issue #35, opened 2015-02-09
Document promdash label filters


grobie:
I couldn't find the documentation about legend format filters. Maybe there are more promdash features which should be documented?

@stuartnelson3 @juliusv

juliusv:
Yeah, from the bottom of http://prometheus.io/docs/visualization/promdash/:

----------
TODOs
===

Add documention about:

* annotation tags
* fullscreen / fullscreenTitle URL parameters
* document the dashboard JSON format

juliusv:
@stuartnelson3 stu stu... @grobie really wants this :)

sdurrheimer:
@grobie @juliusv I think this can be closed now :)

PR #57, opened 2015-03-27
Add format filter documentation.

stuartnelson3:
@juliusv feel free to move or change this as you see fit

Solves #35

### DIFF
--- a/content/docs/visualization/promdash.md
+++ b/content/docs/visualization/promdash.md
@@ -129,6 +129,9 @@ interpolated version of the given format string. To reference specific label
 values in the format string, use double curly braces: `{{label-name}}`. For
 example: `{{host}} - cluster {{cluster}}`.
 
+Format strings support filters. See the Filters section below for a list of
+currently available filters, expected inputs, and outputs.
+
 ### Link to graph
 The "Link to this graph" menu tab allows you to generate a link to a specific
 graph. This link will show the graph in a single-widget fullscreen view as it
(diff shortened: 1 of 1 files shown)


content/docs/visualization/promdash.md
```diff
@@ -129,6 +129,19 @@ interpolated version of the given format string. To reference specific label
 values in the format string, use double curly braces: `{{label-name}}`. For
 example: `{{host}} - cluster {{cluster}}`.
 
+### Legend format filters
+The aforementioned `Legend format` builder also supports filters for the
+labels. The format is `{{label | filter}}`. The following filters are available:
+
+- `toPercent`
```
juliusv: Could you mention for each of these what input they expect, and what the output is? For some, it's not really clear just from the name.

content/docs/visualization/promdash.md
```diff
@@ -129,6 +129,19 @@ interpolated version of the given format string. To reference specific label
 values in the format string, use double curly braces: `{{label-name}}`. For
 example: `{{host}} - cluster {{cluster}}`.
 
+### Legend format filters
```
juliusv: These filters work also outside of legend format strings (for example, in expressions). So I think it would make more sense to put this into the "Using template variables" section?

juliusv: Hmm. Though labels in legend strings are not dashboard template variables. It's more like both cases use this feature, so maybe it should be separate somehow?

juliusv:
Ping.

content/docs/visualization/promdash.md
stuartnelson3: gosh no idea. your choice :)

stuartnelson3:
made some updates

grobie:
Let's merge @juliusv?!

juliusv:
I still think my old comment applies and the format filters should be documented not as a legend-specific feature, but separately, as a feature that applies wherever labels or variables can be used.

grobie:
@juliusv @stuartnelson3 I extracted the section and made it more generic. What else is missing to finally unblock this PR.

juliusv:
:+1: wonderful, thanks!

PR #67, opened 2015-04-21
Add docs for hidden labels.

fabxc:


content/docs/concepts/data_model.md
```diff
@@ -28,8 +28,10 @@ allows filtering and aggregation based on these dimensions. Changing any label
 value, including adding or removing a label, will create a new time series.
 
 Label names may contain ASCII letters, numbers, as well as underscores. They
-must match the regex `[a-zA-Z_][a-zA-Z0-9_]*`. Label names beginning with `__`
-are reserved for internal use.
+must match the regex `[a-zA-Z_][a-zA-Z0-9_]*`.
+Label names beginning with `_` are hidden and will, in genereal, not propagate
```
juliusv: s/genereal/general/

juliusv:
Just talked with @fabxc. I think introducing `_` as another type of internal prefix gets a bit messy:

- things with `_` are hidden
- things with `__` are a kind of a subset, but e.g. `__name__` is actually attached verbatim (although that happens after the place where labels with a `_` are interpreted as hidden (in the SD relabeler), it's still confusing)

How about just keeping `__` as the single internal prefix and making the "hidden" metadata labels a clean subset of those, prefixed with `__meta`, like `__meta_etcd_path__`.

juliusv:
@brian-brazil see comment above

brian-brazil:
I see _ being special only in the context of service discovery, otherwise it's a label like any other

fabxc:
But we'll want to drop all those _* labels given by the SD if we didn't
extract them during relabeling.

brian-brazil:
I can see dropping _ in any other circumstance getting very messy, and would suggest we just stick with __ as the general reserved prefix. __meta_* would also work.

juliusv:
Cool, let's do that then...

fabxc:
:+1:

PR #72, opened 2015-04-29
Document Grafana integration.

juliusv:


juliusv:
@fabxc @brian-brazil @beorn7

grobie:
Thanks, looks great.

As we haven't made any decision about the future of promdash yet, I think we should give users some orientation. Something like "Comparison with PromDash" listing and linking to advantages and limitations. The top 3 for each should be more than enough for an overview.

juliusv:
Uhm... What would one list there? Takes way more clicks to make graphs with
Grafana than with PromDash? Grafana generally more mature, but less mature
for Prometheus?

beorn7:
:+1: A list for orientation sounds good, but could be added later, once we have a bit more experience gathered...

juliusv:
Merging for now, happy to accept an orientation list later.

PR #77, opened 2015-05-08
Fix rule_checker example

stapelberg:
-ruleFile doesn’t exist, and the existing -rule-file flag is deprecated in https://github.com/prometheus/prometheus/commit/c4e762adbf6ee7f8fd24af8dc584a64d14974a7e


matthiasr:
:+1:

fabxc:
Thanks :+1:

PR #78, opened 2015-05-08
Fix examples: comments start with #, not //

stapelberg:
See https://github.com/prometheus/prometheus/issues/683


fabxc:
Thanks :+1:

juliusv:
@stapelberg Thanks and sorry for the breakage!

PR #81, opened 2015-05-16
Remove comment about hash collisions

fabxc:
This can certainly be removed.

@beorn7: In case you wanted to replace it with something about the way the storage deals with collisions now, this PR can be discarded.


beorn7:
\o/
:+1:

PR #92, opened 2015-05-20
document caveat about combining sum and rate

stapelberg:
As per https://groups.google.com/d/msg/prometheus-developers/wfGhGkDcZ3Q/SXv4aDGgKY8J

### DIFF
--- a/content/docs/querying/functions.md
+++ b/content/docs/querying/functions.md
@@ -184,6 +184,11 @@ rate(http_requests_total{job="api-server"}[5m])
 
 `rate` should only be used with counters.
 
+Note that when combining `rate()` with an aggregation operator (e.g. `sum()`)
+or a function aggregating over time (any function ending in `_over_time`),
+always take a `rate()` first, then aggregate. Otherwise `rate()` cannot detect
+counter resets when your target restarts.
+
 ## `resets()`
 
 For each input time series, `resets(v range-vector)` returns the number of


matthiasr:
This isn't really limited to `sum` though, it would apply to combining rate with any other function AFAICT. I would suggest only mentioning this in the documentation of `rate` and replacing the mention of sum with "any other function or operator". Maybe also add that this is to make counter reset detection reliable?

stapelberg:
> This isn't really limited to sum though, it would apply to combining rate with any other function AFAICT. I would suggest only mentioning this in the documentation of rate and replacing the mention of sum with "any other function or operator".

Done.

> Maybe also add that this is to make counter reset detection reliable?

But I already mention that? Did you overread that part or do you want me to change the wording?

matthiasr:
Thank you! Indeed, missed that.

:+1:

content/docs/querying/functions.md
```diff
@@ -178,6 +178,10 @@ rate(http_requests_total{job="api-server"}[5m])
 
 `rate` should only be used with counters.
 
+Note that when combining functions (e.g. `sum()`) with `rate()`, you must
```
beorn7: `sum` is actually not a function, but an aggregation operator.

Suggestion:

When combining `rate()` with an aggregation operator (e.g. `sum()`) or a function aggregating over time (any function that ends with `_over_time`), always take the `rate()` first and then aggregate, otherwise ...

stapelberg:
@beorn7 Thanks, changed the pull request.

content/docs/querying/functions.md
```diff
@@ -184,6 +184,11 @@ rate(http_requests_total{job="api-server"}[5m])
 
 `rate` should only be used with counters.
 
+Note that when combining `rate()` with an aggregation operator (e.g. `sum()`)
+or a function aggregating over time (any function ending in `_over_time`),
+always take a `rate()` first, then aggregate, otherwise `rate()` cannot detect
```
juliusv: Run-on sentence. Either add an "as" after the "aggregate," or end the sentence there and start a new one.

:+1: otherwise.

stapelberg:
@juliusv Done.

juliusv:
Thanks a bunch!

PR #95, opened 2015-05-22
Two minor docs improvements.

beorn7:
@juliusv @fabxc

### DIFF
--- a/content/docs/practices/naming.md
+++ b/content/docs/practices/naming.md
@@ -6,24 +6,24 @@ sort_rank: 1
 # Metric and label naming
 
 The metric and label conventions presented in this document are not required
-for using Prometheus, but can serve as both a style-guide and collection of
+for using Prometheus, but can serve as both a style-guide and a collection of
 best practices. Individual organizations may want to approach e.g. naming
 conventions differently.
 
 ## Metric names
 
-A metric name:
+A metric name...
 
-* should have a (single-word) application prefix relevant to the containing Prometheus domain
- * <code><b>prometheus</b>\_notifications\_total</code>
- * <code><b>indexer</b>\_requests\_latencies\_milliseconds</code>
- * <code><b>processor</b>\_requests\_total</code>
-* must have a single unit (i.e. do not mix seconds with milliseconds)
-* should have a units suffix
+* ...should have a (single-word) application prefix relevant to the domain the metric belongs to. The prefix is sometimes referred to as `namespace` by client libraries. For metrics specific to an application, the prefix is usually the application name itself. Sometimes, however, metrics are more generic, like standardized metrics exported by client libraries. Examples:
+ * <code><b>prometheus</b>\_notifications\_total</code> (specific to the Prometheus server)
+ * <code><b>process</b>\_cpu\_seconds\_total</code> (exported by many client libraries)
+ * <code><b>http</b>\_request\_duration\_microseconds</code> (for all HTTP requests)
+* ...must have a single unit (i.e. do not mix seconds with milliseconds).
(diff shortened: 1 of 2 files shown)


content/docs/practices/naming.md
```diff
@@ -6,24 +6,24 @@ sort_rank: 1
...
-A metric name:
+A metric name...
 
-* should have a (single-word) application prefix relevant to the containing Prometheus domain
- * <code><b>prometheus</b>\_notifications\_total</code>
- * <code><b>indexer</b>\_requests\_latencies\_milliseconds</code>
- * <code><b>processor</b>\_requests\_total</code>
-* must have a single unit (i.e. do not mix seconds with milliseconds)
-* should have a units suffix
+* ...should have a (single-word) application prefix relevant to the domain the metric belongs to. The prefix is sometimes referred to as `namespace` by client libraries. For metrics specific to an application, the prefix is usually the application name itself. Sometimes, however, metrics are more generic, like standardized metrics exported by client libraries. Examples:
```
juliusv: I think you should be able to wrap this text, even if it's in a list (indent everything by two spaces?)?

content/docs/practices/naming.md
```diff
@@ -6,24 +6,24 @@ sort_rank: 1
...
+ * <code><b>prometheus</b>\_notifications\_total</code> (specific to the Prometheus server)
+ * <code><b>process</b>\_cpu\_seconds\_total</code> (exported by many client libraries)
+ * <code><b>http</b>\_request\_duration\_microseconds</code> (for all HTTP requests)
+* ...must have a single unit (i.e. do not mix seconds with milliseconds).
+* ...should have a suffix describing the unit.
```
juliusv: `_total` is not a really a unit. To make it even more confusing, we have metrics with just `_total` and others with `<unit>_total` (like `foobar_seconds_total`). Maybe add an extra comment that counters should have an additional `_total` suffix *after* the unit?

content/docs/practices/naming.md
```diff
@@ -44,7 +44,7 @@ Use labels to differentiate the characteristics of the thing that is being measu
 Do not put the label names in the metric name, as this introduces redundancy
 and will cause confusion if the respective labels are aggregated away.
 
-CAUTION: <b>CAUTION:</b> Remember that every unique key-value label pair
+**CAUTION:** Remember that every unique key-value label pair
```
juliusv: The initial `CAUTION` here triggers a special "caution" markup block and doesn't actually show up in the text. See http://prometheus.io/docs/practices/naming/. It'll need to be preserved.

content/docs/practices/naming.md
beorn7: This section is not about units. The _total thing is explained in the units section below.

content/docs/practices/naming.md
beorn7: Trying now...

content/docs/practices/naming.md
beorn7: Oops...

content/docs/practices/naming.md
beorn7: Yeah, worked.

content/docs/practices/naming.md
juliusv: This section says, "should have a suffix describing the unit", and then lists as one of the examples a metric without any unit, but a `_total` suffix. That's the confusing bit. It was there before, but we could make it clearer what suffix is expected when.

beorn7: I see, I missed what line you are commenting on.

Will improve wording.

PR #98, opened 2015-05-26
Document changes() and resets() functions.

juliusv:


juliusv:
@fabxc @brian-brazil

content/docs/querying/functions.md
```diff
@@ -40,6 +40,12 @@ by sample value.
 `ceil(v instant-vector)` rounds the sample values of all elements in `v` up to
 the nearest integer.
 
+## `changes()`
+
+For each input time series, `changes(v range-vector)` returns the number of
+times its value has changed within the provided time range as an instant
+vector.
```
fabxc: Should we mention that it's fitted towards integers? For floats with small changes the result is fairly unpredictable.

fabxc:
:+1: otherwise

content/docs/querying/functions.md
juliusv: Hmm, not sure. If two samples are ingested with the same value on the same machine, they will always end up having the same bit representation and be equal, no? The time where you have to be careful is when you do math on values before processing the result further, and slight orders in operations etc. could change the output value ever so slightly each time.

I can imagine this function being used with e.g. timestamps that are stored as seconds, but have sub-second precision. And those should compare as equal as long as they don't change at all, AFAIU...

fabxc: If protobuf preserves the representation, then yes - don't know the internals.
We can certainly add a notice later if there are people asking questions about it.

PR #99, opened 2015-05-28
Sanitize and uniquify in-page anchors properly.

juliusv:
Things like question marks or parentheses don't play nicely in URLs in
many contexts. Replace all special characters with dashes, and also
ensure that anchors are unique within a page if there are headers with
identical texts.


juliusv:
@beorn7

juliusv:
@fabxc

fabxc:
:+1:

PR #102, opened 2015-05-30
Document increase() function.

brian-brazil:


content/docs/querying/functions.md
```diff
@@ -149,6 +149,23 @@ for quantiles located in the lowest bucket.
 If `b` contains fewer than two buckets, `NaN` is returned. For φ < 0, `-Inf` is
 returned. For φ > 1, `+Inf` is returned.
 
+## `increase()`
+
+`increase(v range-vector)` calculate the increase in the
```
fabxc: calculate**s**

"of the time series in the range vector"?

content/docs/querying/functions.md
```diff
@@ -149,6 +149,23 @@ for quantiles located in the lowest bucket.
...
+
+The following example expression returns the number of HTTP requests as measured
+over the last 5 minutes, per time series in the range vector:
+
+```
+increase(http_requests_total{job="api-server"}[5m])
+```
+
+`increase` should only be used with counters. `increase` should be used
+primarily for human readability, use `rate` for rules so that everything is
```
fabxc: s/, use/. Use

fabxc:
:+1:

content/docs/querying/functions.md

PR #110, opened 2015-06-03
Set up continuous deployment

grobie:
Use travis CI to continuously deploy prometheus.io after every push to master. A test run can be found here: https://travis-ci.org/prometheus/docs

For authentication a [deploy key](https://developer.github.com/guides/managing-deploy-keys/) is used, as that allows for the most granular access rights (access to only a single repo: prometheus/prometheus.github.io). More details can be found here: http://docs.travis-ci.com/user/private-dependencies/. The deploy key is a 4096 bit RSA key, stored encrypted inside of the repo as well as our shared LastPass folder.

@juliusv @fabxc


juliusv:
:+1: Nice, thanks!

PR #113, opened 2015-06-13
Change default storage.local.path mentions

sdurrheimer:
Or maybe:

```
By default, Prometheus stores its database in a `data` folder in the current path. (flag -storage.local.path).
```

```
The default path is a `data` folder in the current path, which is good to try something out quickly but most likely not what you want for actual operations.
```

@juliusv


juliusv:
:+1: Thanks!

PR #114, opened 2015-06-13
Add logo and icon to Atom Feed

sdurrheimer:
@juliusv


fabxc:
Are there constraints on the file format for feed logos? If not, an SVG version would be preferable.

:+1: otherwise.

sdurrheimer:
Atom specs are not so clear. No real size limitation, just a hint about a 1:1 ratio for the logo. Same for the image format so I really don't know if a SVG file would work. Already talk about that with @juliusv on IRC.

It's kinda hard to test it to be honest, seems like there is too few aggregators that are taking it into account.

juliusv:
:+1: Thanks! Yeah, I tried deploying an SVG for you earlier and it didn't work, right?

sdurrheimer:
Hum don't think we tested out with the SVG logo. It was a png file.

But again, the fact I didn't find a aggregator taking it into account, I followed the few informations I found in the Atom specs and Nanoc documentation.

juliusv:
Ok, let's try it...

PR #116, opened 2015-06-15
Document metric relabeling.

brian-brazil:
@fabxc @juliusv

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -118,8 +118,12 @@ file_sd_configs:
 target_groups:
   [ - <target_group> ... ]
 
-# List of relabel configurations.
+# List of target relabel configurations.
 relabel_configs:
+  [ - <relabel_config> ... ]
+
+# List of metric relabel configurations.
+metric_relabel_configs:
   [ - <relabel_config> ... ]
 ```
 
@@ -239,7 +243,7 @@ Where `<filename_pattern>` may be a path ending in `.json`, `.yml` or `.yaml`. T
 may contain a single `*` that matches any character sequence, e.g. `my/path/tg_*.json`.
 
 
-### Relabeling `<relabel_config>`
+### Target relabeling `<relabel_config>`
 
 Relabeling is a powerful tool to dynamically rewrite the label set of a target before
 it gets scraped. Multiple relabeling steps can be configured per scrape configuration.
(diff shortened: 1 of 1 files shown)


content/docs/operating/configuration.md
```diff
@@ -239,7 +243,7 @@ Where `<filename_pattern>` may be a path ending in `.json`, `.yml` or `.yaml`. T
 may contain a single `*` that matches any character sequence, e.g. `my/path/tg_*.json`.
 
 
-### Relabeling `<relabel_config>`
+### Target Relabeling `<relabel_config>`
```
fabxc: We don't do title case (unfortunately). This also apply to the uppercase Rs below.

We could make it uppercase globally though as it is more of a custom concept name and not self-explanatory. This would require changes to other parts of the docs as well.

content/docs/operating/configuration.md
```diff
@@ -290,3 +294,11 @@ regex: <regex>
   (`${1}`, `${2}`, ...) in `replacement` substituted by their value.
 * `keep`: Drop targets for which `regex` does not match the concatenated `source_labels`.
 * `drop`: Drop targets for which `regex` matches the concatenated `source_labels`.
+
+### Metric Relabeling `<metric_relabel_configs>`
+
+Metric Relabeling is applied to samples as the last step before ingestion . It
```
fabxc: Space before ` .`

content/docs/operating/configuration.md
```diff
@@ -290,3 +294,11 @@ regex: <regex>
...
+has the same configuration format and actions as Target Relabeling. Metric
+Relabeling does not apply to automatically generated timeseries such as `up`.
```
fabxc: Should we just list them all (it's just two right?). We haven't defined what automatically generated time series are, IIRC.
Also space between timeseries.

content/docs/operating/configuration.md
brian-brazil: Fixed.

content/docs/operating/configuration.md
brian-brazil: Fixed.

content/docs/operating/configuration.md
brian-brazil: They're described over in jobs and instances, I'd rather not have copies lying around to keep updated as we add more. `up` is the one the vast majority of users will encounter.

Fixed.

juliusv:
:+1:

PR #122, opened 2015-06-22
Update label matching documentation.

fabxc:
This commit adds documentation about the limitations of empty
label matching and documents the `__name__` label which is useful
for federation.

@beorn7 @juliusv

### DIFF
--- a/content/docs/querying/basics.md
+++ b/content/docs/querying/basics.md
@@ -84,6 +84,28 @@ For example, this selects all `http_requests_total` time series for `staging`,
 
     http_requests_total{environment=~"staging|testing|development",method!="GET"}
 
+Label matchers that match empty label values also select all time series that do
+not have the specific label set at all.
+
+Vector selectors must either specify a name or at least one label matcher
+that does not match the empty string. The following expression is illegal:
+
+    {job=~".*"} # Bad!
+
+In contrast, these expressions are valid as they both have a selector that does not
+match empty label values.
+
+    {job=~".+"}              # Good!
+    {job=~".*",method="get"} # Good!
+
+Label matchers can also be applied to metric names by matching against the internal
+`__name__` label. For example, the expression `http_requests_total` is equivalent to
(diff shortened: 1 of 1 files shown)


content/docs/querying/basics.md
```diff
@@ -84,6 +84,27 @@ For example, this selects all `http_requests_total` time series for `staging`,
 
     http_requests_total{environment=~"staging|testing|development",method!="GET"}
 
+Label matchers that match empty label values also select all time series that do
+not have the specific label set at all.
+
+Vector selectors must either specify a name or at least one label matcher
+that does not match the empty string. The following expression is illegal:
+
+    {job=".*"} # Bad!
```
juliusv: `=` -> `=~`

fabxc: Done.

content/docs/querying/basics.md
```diff
@@ -84,6 +84,27 @@ For example, this selects all `http_requests_total` time series for `staging`,
...
+
+In contrast, these expressions are valid as they both have a selector that does not
+match empty label values.
+
+    {job=".+"}              # Good!
+    {job=".*",method="get"} # Good!
+
+Metrics may also be selected by their names using the non-equal matchers (`!=`, `=~`, `!~`)
```
juliusv: `non-equal` sounds too close to `not-equal` (i.e. `!=). How about:

"Label matchers can also be applied to metric names by matching against the internal `__name__` label. For example, the expression `http_requests_total` is equivalent to `{__name__="http_requests_total"}`. Matchers other than `=` (`!=`, `=~`, `!~`) may also be used. The following expression [...]:"

juliusv:
Not sure if you noticed - I added another comment before you force-pushed: https://github.com/prometheus/docs/pull/122#discussion-diff-32937572

content/docs/querying/basics.md
fabxc: Done.

beorn7:
:+1:

PR #124, opened 2015-06-23
Document binary operator precedence.

juliusv:


juliusv:
@grobie @fabxc

grobie:
:+1: Thanks!

juliusv:
I added a sentence about left-associativity on the same level. @fabxc is that ok to expose in the specs, or better to leave undefined?

fabxc:
Yes, that should be documented. I'd drop the "to change the order" part, though. In these cases it's usually easier to always use parentheses for explicitness.

:+1:

juliusv:
Ok, I dropped the last sentence.

PR #127, opened 2015-06-24
Document hashmod

brian-brazil:


fabxc:
Can we note somewhere that `regex` is still mandatory for other actions?

brian-brazil:
@fabxc Done

PR #135, opened 2015-07-01
Document HTTP API v1.

juliusv:


juliusv:
@fabxc 

The one thing I want to change before merging is the location of this document in the menu structure. I haven't found a good place for it yet. Maybe rename one of the sections to make it fit in better? Like the Visualization section?

content/docs/querying/api.md
```diff
@@ -0,0 +1,341 @@
+---
+title: HTTP API
+sort_rank: 7
+---
+
+# HTTP API
+
+The current stable HTTP API is reachable under `/api/v1` on a Prometheus
+server. Any non-breaking additions will be added under that endpoint. There are
+no minor versions of the API.
```
fabxc: This was more for the design doc. I don't think we have to document what we not have.

content/docs/querying/api.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Format overview
+
+The API response format is JSON. Every successful API request returns a `2xx`
+status code.
+
+Invalid requests that reach the API handlers return a JSON error object
+and the HTTP response code `422 Unprocessable Entity`
+([RFC4918](http://tools.ietf.org/html/rfc4918#page-78)). Other non-`2xx`
+codes may be returned from outer HTTP handling layers.
```
fabxc: Implementation detail.

Other non-`2xx` codes may be returned for errors occurring before reaching the application.

Mostly we want to return `400` for invalid JSON (not invalid JSON data) - right now we simply don't have any.

content/docs/querying/api.md
```diff
@@ -0,0 +1,341 @@
...
+
+The JSON response envelope format is as follows:
+
+```
+{
+  "status": "success" | "error",
+  "data": <data>,
+
+  // Only set if status == "error". The data field may still hold
```
fabxc: `==` to `equals`

content/docs/querying/api.md
```diff
@@ -0,0 +1,341 @@
...
+
+Names of query parameters that may be repeated end with `[]`.
+
+`<series_selector>` placeholders refer to Prometheus [time series
+selectors](/docs/querying/basics/#time-series-selectors) like
+`http_requests_total` or `http_requests_total{method=~"^GET|POST$"}` and need
+to be URL-encoded.
+
+`<duration>` placeholders refer to Prometheus duration strings of the form
+`[0-9]+[smhdwy]`. For example, `5m` refers to a duration of 5 minutes.
```
fabxc: What's the status on this. I removed `w` and `y` in `model/common`. We should get rid of those units.

content/docs/querying/api.md
```diff
@@ -0,0 +1,341 @@
...
+
+### Range vectors
+Range vectors are returned as result type `matrix`. The corresponding
+`result` property has the following format:
+
+```
+[
+  {
+    "metric": { "<label_name>": "<label_value>", … },
+    "values": [ [ <unix_time>, "<sample_value>" ], ... ]
```
fabxc: Inconsistent `...`s.

content/docs/querying/api.md
```diff
@@ -0,0 +1,341 @@
...
+### Instant vectors
+Instant vectors are returned as result type `vector`. The corresponding
+`result` property has the following format:
+
+```
+[
+  {
+    "metric": { "<label_name>": "<label_value>", … },
+    "timestamp": <unix_time>,
+    "value": "<sample_value>"
```
fabxc: I think we agreed to consistently make it `[<ts>, <val>]` as this is just confusing and the user has to handle the array representation anyway.
If I missed to implement this, I would like to do this before releasing the official docs.

content/docs/querying/api.md
juliusv: Hmm, that sounds like unnecessary breakage. Is there a strong motivation to remove them?

Btw., why only remove `w` and `y` not also `d` (days also don't have a well-defined length)?

content/docs/querying/api.md
juliusv: Yeah, I was wondering about that and thinking the same thing. I'm happy to change it in the docs, but it needs to be changed in the code as well.

content/docs/querying/api.md
juliusv: Removed the last sentence.

content/docs/querying/api.md
juliusv: Changed to your suggested text.

content/docs/querying/api.md
fabxc: Because they are hardly used and everyone has a very clear perception of what `30d`, `90d`, or `360d` are.
Days have leap seconds - years have whole days...

Meh - we can keep them - still think days are superior in any scenario.

content/docs/querying/api.md
juliusv: Changed to `is` instead of `equals` instead.

content/docs/querying/api.md
juliusv: Fixed.

content/docs/querying/api.md
juliusv: Yeah, I probably wouldn't add `w` and `y` nowadays, but there's so little cost to keep supporting them that I think it's good to keep them to not create unnecessary breaking changes.

fabxc: Yes, sure - more of a question whether we still want to document it.

juliusv:
I've updated the "Query language" menu section to just "Querying" now, so that the "HTTP API" section fits in better. Also, linked to it from the beginning of the "Basics" section. It's still not perfect, as the API has one endpoint so far that isn't strictly "querying" (the series deletion), but probably the best we can do for without major restructuring of the menus.

fabxc:
Yes, that will work. :+1:

juliusv:
Before merging this, we need to fix the vector output format to have the same timestamp/value representation as the matrix one.

I already changed this PR to reflect the anticipated change now.

content/docs/querying/api.md
```diff
@@ -0,0 +1,338 @@
...
+
+### Instant vectors
+Instant vectors are returned as result type `vector`. The corresponding
+`result` property has the following format:
+
+```
+[
+  {
+    "metric": { "<label_name>": "<label_value>", ... },
+    "timestamp": <unix_time>,
```
fabxc: To be removed.

juliusv: Ah right, done.

juliusv:
@fabxc Hmm, now it would be most consistent to also bundle up the timestamp with the value for `scalar` and `string` return types.

fabxc:
Yes, I've thought about that - would be :+1: on this. Even though it requires more back porting for the legacy API.

fabxc:
`[<timestamp>, <value>]` is in Prometheus master for all types.

juliusv:
Updated to match https://github.com/prometheus/prometheus/pull/872. Should be good to merge now.

fabxc:
Yes :+1:

PR #138, opened 2015-07-07
Add a Help page to help direct users in need of assistance.

brian-brazil:
@juliusv

### DIFF
--- a/content/help.html
+++ b/content/help.html
@@ -1,0 +1,28 @@
+---
+title: Help
+---
+<div class="col-md-8 col-md-offset-2 doc-content">
+  <h1>Help</h1>
+  <p>
+  If you are having a problem using Prometheus, there are several ways to get
+  help from the community.
+  </p>
+  <h2>Documentation</h2>
+  <p>The <a href="/docs/introduction/overview/">official documentation</a> for
+  Prometheus is extensive, and covers the core system in depth.</p>
+
+  <h2>IRC</h2>
+  <p>IRC is good for getting interactive help. As our users are spread around the world, please be patient and hang around
+  for a while if you do not get an immediate answer.</p>
+ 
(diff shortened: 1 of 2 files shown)


grobie:
Any specific reasons not to use markdown? Should help to decouple content from presentation (like all the css classes a content author shouldn't need to know about.

content/help.html
```diff
@@ -0,0 +1,28 @@
+---
+title: Help
+---
+<div class="col-md-8 col-md-offset-2 doc-content">
+  <h1>Help</h1>
+  <p>
+  If you are having a problem using Prometheusm, there are several ways to get
```
juliusv: Prometheusm -> Prometheus

juliusv:
Hmm, this is duplicating so much of the "COMMUNITY" page. Also, we have to use the top menu sparingly, and adding another distinct nav item for "HELP" makes it look noticeably more blown up (harder for the eye to scan it).

Could some of this information be put into the COMMUNITY page somehow? Right now HELP is basically pointing people to (information in) COMMUNITY and DOCUMENTATION, which each have titles which are sufficiently descriptive to be found on their own.

content/help.html
brian-brazil: Fixed.

brian-brazil:
@grobie I'm following what the community page was doing.

@jrv I see your concern, this is more task oriented as a new user would have to guess which of docs/community would be of assistance - I personally always hate having to reverse engineer a how a website is laid out when looking for basic information.

juliusv:
So personally I still feel that the extra visual overhead isn't worth the gains. An approach that would work without adding an extra menu item for mainly duplicated/pointer content would be much preferable IMO.

If others do think it'd be worth adding another menu point just for this, I'd suggest moving it next to "BLOG" since that is also four letters and looks less awkward than if one small word is squeezed between two larger ones.

PR #141, opened 2015-07-20
Fix broken link for alerting details.

dato:


beorn7:
Thank you.

Once the whole alerting section is more comprehensive, we should probably point to the "Overview" page of the "Alerting" section. (But currently, the overview doesn't lead to to the more important other Alerting gaps.)

PR #162, opened 2015-08-13
Document new dns_sd_config options

fabxc:
The meta label `__meta_dns_srv_name` only partially makes sense. We should consider just renaming it `__meta_dns_name`.

@brian-brazil

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -174,7 +174,7 @@ number.
 
 ### DNS-SD configurations `<dns_sd_config>`
 
-A DNS-SD configuration allows specifying a set of DNS SRV record names which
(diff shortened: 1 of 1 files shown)


content/docs/operating/configuration.md
```diff
@@ -187,11 +187,18 @@ record name that produced the discovered target.
 names:
   [ - <record_name> ]
 
+# The type of DNS query to perform.
+[ type: <query_type> | default = 'SRV' ]
+
+# The used port number if the query type is not SRV.
```
brian-brazil: The port number used, if the query type is not SRV

Reads more naturally

brian-brazil:
:+1: That sounds like a better name to me

fabxc:
Done.

PR #166, opened 2015-08-13
Fix Apdex score example in Histograms page.

mateusbraga:
I noticed the mistake comparing it to the formula in Wikipedia: https://en.wikipedia.org/wiki/Apdex

### DIFF
--- a/content/docs/practices/histograms.md
+++ b/content/docs/practices/histograms.md
@@ -76,8 +76,8 @@ following expression yields the Apdex score for each job over the last
     (
       sum(rate(http_request_duration_seconds_bucket{le="0.3"}[5m])) by (job)
     +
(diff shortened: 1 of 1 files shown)


fabxc:
Good catch, thanks!

matthiasr:
Noooo! This is frequently confused so it should be explained, but the original formula is correct. The reason is that the histogram buckets are cumulative; the `le="0.3"` bucket is also contained in the `le="1.2"` bucket; dividing by 2 corrects for that.

fabxc:
@matthiasr  Oh boy, sorry.
Mind reverting and adding a brief explanation? – I probably should not stack up my incompetence here :)

matthiasr:
@fabxc yep, was already on it

PR #172, opened 2015-08-18
Revert #166 and add explanation why the formula is in fact correct.

matthiasr:
Reverts prometheus/docs#166

This is frequently confused so it should be explained, but the original formula is correct. The reason is that the histogram buckets are cumulative; the le="0.3" bucket is also contained in the le="1.2" bucket; dividing by 2 corrects for that.

Add an explanation.


fabxc:
:+1: thanks for walking through the old stuff and catching this.

PR #174, opened 2015-08-18
Reduce confusion on file SD format

brian-brazil:
The missing [] confused users, so make the example a list so it's the base for a complete file.


brian-brazil:
Let's use #175 instead

PR #184, opened 2015-08-26
Remove duplicate line in the getting started sample config

ChaoticMind:
`scrape_interval` is again repeated two lines below (in both cases).


fabxc:
Thanks :+1:

PR #195, opened 2015-08-28
Add `go get` to the build instructions for the example client.

beorn7:
@fabxc

### DIFF
--- a/content/docs/introduction/getting_started.md
+++ b/content/docs/introduction/getting_started.md
@@ -160,6 +160,7 @@ processes:
 # Fetch the client library code and compile example.
 git clone https://github.com/prometheus/client_golang.git
 cd client_golang/examples/random
+go get -d
 go build
 
 # Start 3 example targets in separate terminals:


content/docs/introduction/getting_started.md
```diff
@@ -160,6 +160,7 @@ processes:
 # Fetch the client library code and compile example.
 git clone https://github.com/prometheus/client_golang.git
 cd client_golang/examples/random
+go get
```
juliusv: `go get` actually installs the random example, so it wouldn't even have to be built anymore. But for that to work, the user would have to have the right output dir in their `$PATH`. So maybe just do `go get -d` here?

fabxc: That seems sane.

beorn7: Hmm, shouldn't we assume the user has a proper Go setup? Then it would really just be `go get`.

fabxc: As a user with a proper go setup, I don't want example binaries in my $PATH.

beorn7: Fair enough... so `go get -d` it is.

juliusv:
:+1:

PR #198, opened 2015-08-30
Remove obsolete reference to closed issue

ChaoticMind:
As far as I understand it, the interpolation handling (at least in a way relevant to the linked issue) did not and will not change as a result of that issue.

I could be wrong.

### DIFF
--- a/content/docs/querying/basics.md
+++ b/content/docs/querying/basics.md
@@ -170,23 +170,20 @@ in detail in the [expression language functions](/docs/querying/functions/) page
 ### Interpolation and staleness
 
 When queries are run, timestamps at which to sample data are selected
-independent of the actual present time series data. This is mainly to support
+independently of the actual present time series data. This is mainly to support
 cases like aggregation (`sum`, `avg`, and so on), where multiple aggregated
-time series do not exactly align in time. At every one of these predetermined
-sampling times, Prometheus searches for the closest surrounding samples and
-linearly interpolates a timestamp-value pair between the actual stored samples.
-That means that the generated sample has timestamp and sample values which are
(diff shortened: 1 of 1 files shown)


content/docs/querying/basics.md
```diff
@@ -184,9 +184,8 @@ sampling timestamp, no interpolated sample is generated for this time series at
 this point in time. This effectively means that time series "disappear" from
 graphs at times where their latest collected sample is 5 older than 5 minutes.
 
-NOTE: <b>NOTE:</b> Staleness and interpolation handling might change. See
```
brian-brazil: Around line 177 is no longer true, could you update that too?

brian-brazil:
https://github.com/prometheus/prometheus/issues/581 also relates to interpolation

ChaoticMind:
Well, I opted to simply link that issue instead, since that's possibly relevant. Thanks for the link! (first attempt was `a5e0722`)

> Around line 177 is no longer true, could you update that too?

It's not actually obvious to me how/why line 177 is no longer true. Could you clarify?

brian-brazil:
> Prometheus searches for the closest surrounding samples and
 linearly interpolates a timestamp-value pair between the actual stored samples.

This is no longer the behaviour, we now take the newest sample before the timestamp.

ChaoticMind:
Aha, ok. I updated the text, could you review it?

I speculated on the "If no stored sample is found (by default) [...]" paragraph. It could be that this paragraph is no longer true and should be removed entirely.

Out of curiosity, do you have a link to a discussion/issue about the revised "interpolation"?

brian-brazil:
Thanks!

PR #208, opened 2015-09-08
Require that metrics be grouped together for text format.

brian-brazil:
While it'd be nice for TYPE to be first,
that's not what the examples do so we leave the ordering
of HELP/TYPE unspecified.

@beorn7 @fabxc


beorn7:
We could change the examples... how important is the order of HELP/TYPE really?

brian-brazil:
It's not too important in practise, the grouping is what matters. Requiring the other ordering would likely be a breaking change for all clients which I'd like to avoid.

juliusv:
:+1:

PR #175, opened 2015-08-18
explain how to make file_sd_config work.

robey:

PR #210, opened 2015-09-10
Better explain file_sd's expected format

evanphx:
Indicate in the docs that the files need to contain an array of target groups, not a standalone one. This was only discoverable by looking at the output of prometheus saying the file was bad.

I know the docs indicate "zero or more" which implies it is an array, but that detail is extremely easy to miss.

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -428,6 +428,20 @@ The JSON version of a target group has the following format:
 }
 ```
 
+Also, the files must contain an array of target groups, not a standalone one. Here is
+an example of a simple file:
+
+```
+[
+  {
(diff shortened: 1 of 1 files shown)


matthiasr:
I would leave out "also" and "not a standalone one", just say " The file must contain an array of target groups. Here …"

matthiasr:
Otherwise, thank you for spotting this :+1:

content/docs/operating/configuration.md
```diff
@@ -428,6 +428,20 @@ The JSON version of a target group has the following format:
 }
 ```
 
+Also, the files must contain an array of target groups, not a standalone one. Here is
```
brian-brazil: One example with a list would probably be clearer, see #175

brian-brazil:
Superseded by #257

PR #205, opened 2015-09-03
Document bool modifier

brian-brazil:
@juliusv 

Fixes #204 

At least this will all be simpler to explain when `0 == 0` is changed.

PR #211, opened 2015-09-11
Document regex anchoring.

brian-brazil:
@juliusv


juliusv:
Thanks! We should get into the habit of using the `next-release` branch for docs changes that are only relevant for the next release and misleading before that. Could you reopen the PR against that branch?

brian-brazil:
I think we've already lost that battle for this release (see #205), do you still want me to reopen?

juliusv:
Yup, please do. No need to pile more stuff into the docs which doesn't "exist" yet and will confuse users.

PR #214, opened 2015-09-12
Document new vector() function.

brian-brazil:
For https://github.com/prometheus/prometheus/pull/1073

@juliusv

### DIFF
--- a/content/docs/querying/functions.md
+++ b/content/docs/querying/functions.md
@@ -269,6 +269,10 @@ expression is to be evaluated.
 `topk(k integer, v instant-vector)` returns the `k` largest elements of `v` by
 sample value.
 
+## `vector()`
+
+`vector(s scalar)` returns the scalar `s` as a vector with no labels.
+
 ## `<aggregation>_over_time()`: Aggregating values over time:
 
 The following functions allow aggregating each series of a given range vector


grobie:
Maybe worth to explain that non-equal matchers are going to be ignored, so
that the instance example becomes more understandable. Assuming that is the
intention of the second example.

I'd actually expect that using a regexp matcher, etc. would not be be
allowed here. I guess it would require to use something like a label setter
type maybe?

juliusv:
Yeah, it is all a bit weird. But also tricky:

Introducing such a "label-setter-only" type for `vector()` would make it inconsistent with `absent()`. `absent()` can take any vector value (not only series selectors, but any expression which evaluates to vector type), but derives labels from the passed in vector in the special case when it's a series selector and it evaluates to an empty result. So doing that same label-deriving fundamentally differently in `vector()` would be inconsistent. On the other hand, it *is* a bit weird to pass a selector to `absent()` that will never even be evaluated, and where only parts of its options (the `=` selectors) are even considered. Still seems like the least bad option to me right now though.

Alternatively, we could choose to get rid of the second argument of `vector()` completely and require people to use `label_replace()` to add labels. Adding arguments later is always easier - removing them less so.

grobie:
I don't see the inconstency. Functions work on different operands. While
some need "getters" aka matchers other are setters. I see it as type of
typesafety to not allow meaningless arguments.

I'm fine with postponing this and to use label_replace. Not to
straightforward though and a bit tedious. We can also leave it for now and
restrict the interface to only use meaningful types later.

fabxc:
As this is obviously confusing, we should just do it properly and introduce
a label set AST node. The parser already implements this, so it is fairly
simple.

juliusv:
The inconsistency would be that `absent()` also derives output labels from an input series selector in the same way as `vector()` currently does here (by looking at the `=` matchers and ignoring the others). If `vector()` would do the same thing completely differently, that would be inconsistent.

juliusv:
Thinking about this more, the use of a series selector here is indeed not unproblematic: even if we never fully evaluate the selector, all series that it would resolve to are looked up from the indexes during the query analysis stage. For `absent()` this is totally fine because we *want* to look if there are any series matching it. But for `vector()`, it could potentially lead to a huge number of index lookups that are not useful for anything.

Unless we change the analyser to understand that index lookups are not required here and because I feel that having a tottally different type here than in `absent()` would be weirdly inconsistent, I vote for removing the `vector()` argument completely for now.

brian-brazil:
Docs updated, PTAL.

juliusv:
:+1:

PR #215, opened 2015-09-14
Link the Clojure client library (wrapper).

beorn7:
@juliusv

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -68,8 +68,18 @@ separate exporters are needed:
 
    * [cAdvisor](https://github.com/google/cadvisor)
    * [Etcd](https://github.com/coreos/etcd)
-   * [go-metrics instrumentation library](https://github.com/armon/go-metrics)
-   * [gokit](https://github.com/peterbourgon/gokit)
    * [Kubernetes-Mesos](https://github.com/mesosphere/kubernetes-mesos)
    * [Kubernetes](https://github.com/GoogleCloudPlatform/kubernetes)
    * [RobustIRC](http://robustirc.net/)
+
+## Other third-party utilities
+
+This section lists libraries and other utilities that help you instrument code
+in a certain language. They are not Prometheus client libraries themselves but
+make use of one of the normal Prometheus client libraries under the hood. As
+for all independently maintained software, we cannot vet all of them for best
+practices.
+
+   * Clojure: [prometheus-clj](https://github.com/soundcloud/prometheus-clj)
+   * Go: [go-metrics instrumentation library](https://github.com/armon/go-metrics)
+   * Go: [gokit](https://github.com/peterbourgon/gokit)


content/docs/instrumenting/clientlibs.md
```diff
@@ -21,6 +21,7 @@ HTTP endpoint on your application’s instance:
 Unofficial third-party client libraries:
 
 * [Bash](https://github.com/aecolley/client_bash)
+* [Clojure](https://github.com/soundcloud/prometheus-clj) (wraps the Java client library)
```
brian-brazil: This repo isn't public

brian-brazil: Also I'm strongly against wrappers of client libraries, I'd want to give this a thorough review to ensure it follows the client guidelines and doesn't distort the java client before even mentioning it in our docs

beorn7: Loads for me in an anonymous tab.

juliusv: Hmm, I can see it in an incognito window though?

brian-brazil: Seems to be working for me now

beorn7: @brian-brazil Feel free to review. I put it into the unofficial section because I wanted to spare us the work. We were pretty relaxed with including links as long as we marked them as unofficial, i.e. not necessarily following approved best practices etc.

brian-brazil: This appears to be a wrapper of the original client, as that's due for deprecation soon (once we settle what to do about the Hystrix collector) it'd be best not to link it.

beorn7: Which is (ironically) exactly the reason why I want to link it here. I talked to @tomcz about the necessary migration work for Hystrix, and then he told me he needs to convert the Clojure wrapper, too. It was that lack of visibility that made us not see it.

If you think a wrapper shouldn't be linked here, perhaps it would be better located where we also put gokit and such?

brian-brazil: @beorn7 This is the first one that's claimed to be a wrapper and has additional text calling that out in this PR.

It's actually a pile of utility functions which I'm mostly okay with, but some are clearly bad practice (app name shouldn't go in to the metric name, that's what the job label is for; and having an outcome label).

I don't think we should be directing Clojure users to this rather than the Java client.

brian-brazil: I think we're getting crossed communication :)

> If you think a wrapper shouldn't be linked here, perhaps it would be better located where we also put gokit and such?

It's not a wrapper, nor is it a client library, nor directly instrumented code so we don't really have anywhere to put it currently. I guess a new category under 3rd party integrations for utilities?

beorn7: @brian-brazil The various go metric abstractions are essentially wrappers, whether they claim so or not.

In any case, since @tomcz created this, being familiar with both Clojure and Prometheus, I'd assume that many other Clojure programmers will feel the same urge when they use the naked Java client. Rather than letting them all create utility functions (with likely the same or even more errors), wouldn't it be better if we point them to a repo where they can join forces and even enjoy our consulting on best practices?

Speaking of which, perhaps you could file issues in https://github.com/soundcloud/prometheus-clj to help @tomcz getting naming and labeling in line with best practices?

beorn7: The "directly instrumented software" section is perhaps a bit misnamed, as gokit and go-metrics don't really fit. So perhaps we should move them too to the new section, which we could name in a very broad way to not get into discussions what's a wrapper and what's not.

brian-brazil: > The various go metric abstractions are essentially wrappers, whether they claim so or not.

Those are utility functions, you're providing something extra rather than replacing core functionality of the client library.

> wouldn't it be better if we point them to a repo where they can join forces and even enjoy our consulting on best practices?

Some of this appears to be a little organisation specific.

brian-brazil: > The "directly instrumented software" section is perhaps a bit misnamed, as gokit and go-metrics don't really fit. 

It's a little odd, but I'd say go-kit is directly instrumented. go-metrics is more a collector.

beorn7:
Now the whole discussion is in the outdated diff... :-/

Anyway, I have created the section suggested.

gokit might have parts that are directly instrumented software, but as it is not something you just run, but you use in another program, I still find it a bit confusing.

So to avoid all of that discussion, the new section does not claim anything, just talks about "libraries and other utilities that help instrument code".

brian-brazil:
:+1:

beorn7:
@fabxc or @juliusv does this look good to you, too?

fabxc:
:+1:

PR #217, opened 2015-09-24
Move hystrix down to other section, point to SC

brian-brazil:
We're going to drop the hystrix repo from the prometheus
repositry as it's a stale fork. Rather refer
to the original, and put it in the appropriate section.

@beorn7 @tomcz

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -22,7 +22,6 @@ These exporters are maintained as part of the official
    * [Consul exporter](https://github.com/prometheus/consul_exporter)
    * [Graphite exporter](https://github.com/prometheus/graphite_exporter)
    * [HAProxy exporter](https://github.com/prometheus/haproxy_exporter)
-   * [Hystrix metrics publisher](https://github.com/prometheus/hystrix)
    * [JMX exporter](https://github.com/prometheus/jmx_exporter)
    * [Mesos task exporter](https://github.com/prometheus/mesos_exporter)
    * [MySQL server exporter](https://github.com/prometheus/mysqld_exporter)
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/exporters.md
```diff
@@ -83,3 +82,4 @@ practices.
    * Clojure: [prometheus-clj](https://github.com/soundcloud/prometheus-clj)
    * Go: [go-metrics instrumentation library](https://github.com/armon/go-metrics)
    * Go: [gokit](https://github.com/peterbourgon/gokit)
+   * [Hystrix metrics publisher](https://github.com/soundcloud/prometheus-hystrix)
```
beorn7: Should we prefix "Java:" as Hystrix is a Java library?

beorn7:
In general, much cleaner this way.

@tomcz IIRC you weren't involved in the fork anyway. We can't even remember who did it. Since we have a large number of contributed exporters and other tools, and we don't fork any of them in the Prometheus org as a pointer, it's more consistent to not do it with Hystrix, either.

content/docs/instrumenting/exporters.md
brian-brazil: I was unsure on that, it's more JVM than Java per-se due to how the JVM works.

Let's do Java/JVM.

content/docs/instrumenting/exporters.md
```diff
@@ -83,3 +82,4 @@ practices.
    * Clojure: [prometheus-clj](https://github.com/soundcloud/prometheus-clj)
    * Go: [go-metrics instrumentation library](https://github.com/armon/go-metrics)
    * Go: [gokit](https://github.com/peterbourgon/gokit)
+   * Java/JXM: [Hystrix metrics publisher](https://github.com/soundcloud/prometheus-hystrix)
```
juliusv: s/JXM/JMX/?

tomcz:
@beorn7 I was not involved in the fork. Didn't really make sense to me but I thought that the Prometheus team wanted to take ownership. I wasn't asked nor did I intend to maintain the fork.

PR #218, opened 2015-09-24
Fix typo, JXM -> JVM

brian-brazil:
@juliusv


juliusv:
Ah, that's what it meant :)

PR #224, opened 2015-10-05
Correct improper use of 'labels' setting.

varunpatro:
The 'labels' setting in the global configuration section has been renamed to 'external_labels' and now has changed semantics. Hence, the fix.


varunpatro:
@fabxc This should fix all our problems.

fabxc:
Thanks a lot :+1:

PR #245, opened 2015-10-17
Rename statsd_bridge to statsd_exporter

grobie:


juliusv:
:+1: Maybe announce on the list?

PR #254, opened 2015-11-04
Remove wrong anchoring statement for regex matchers

grobie:
@fabxc @juliusv

### DIFF
--- a/content/docs/querying/basics.md
+++ b/content/docs/querying/basics.md
@@ -85,7 +85,7 @@ For example, this selects all `http_requests_total` time series for `staging`,
     http_requests_total{environment=~"staging|testing|development",method!="GET"}
 
 Label matchers that match empty label values also select all time series that do
-not have the specific label set at all. Regex-matches are fully anchored.
+not have the specific label set at all. Regex-matches are not anchored.
 
 Vector selectors must either specify a name or at least one label matcher
 that does not match the empty string. The following expression is illegal:
@@ -154,7 +154,7 @@ The same works for range vectors. This returns the 5-minutes rate that
 `http_requests_total` had a week ago:
 
     rate(http_requests_total[5m] offset 1w)
-    
+
 ## Operators
 
 Prometheus supports many binary and aggregation operators. These are described


brian-brazil:
:-1: Regex documentation should always specify what the anchoring behaviour is, rather than leaving the user to read the code.

grobie:
The current behavior is to be unanchored. The statement is just wrong.

fabxc:
Yes, let's change it to state that it does partial matching then.

brian-brazil:
Yes, let's make it clear what the behaviour is.

grobie:
updated.

content/docs/querying/basics.md
```diff
@@ -85,7 +85,8 @@ For example, this selects all `http_requests_total` time series for `staging`,
     http_requests_total{environment=~"staging|testing|development",method!="GET"}
 
 Label matchers that match empty label values also select all time series that do
-not have the specific label set at all. Regex-matches are fully anchored.
+not have the specific label set at all. Regex-matchers match partially and are not
```
brian-brazil: "match partially" isn't clear, does that mean that only some of my regex is applied? Stick to discussing anchoring.

grobie:
done

content/docs/querying/basics.md
```diff
@@ -85,7 +85,7 @@ For example, this selects all `http_requests_total` time series for `staging`,
     http_requests_total{environment=~"staging|testing|development",method!="GET"}
 
 Label matchers that match empty label values also select all time series that do
-not have the specific label set at all. Regex-matches are fully anchored.
+not have the specific label set at all. Regex-matches are not anchored.
```
beorn7: I think the usual term is "partial match".

beorn7: Sorry, did not see all the comments first...

I don't feel strongly, but "partial match" _is_ pretty much the correct technical term in my experience...

brian-brazil: I'd take partial match to mean that only some of the regex matched, not that only some of the text matched.

beorn7: I'd say 99% of the use of "partial match" in the context of regular expressions means "the whole regular expression matches a part of the target string". But when doing my little research about what the standard anchoring behavior is, I found a few occurrences (namely in Boost) that use "partial match" as in "parts of the regexp matches". So I agree we should just talk about anchoring to avoid misunderstandings.

PR #257, opened 2015-11-06
Clarify the format for file SD files.

brian-brazil:
This example has confused many potential users who
didn't realise it needed to be put in a list.

#175 and #210 don't seem to be going anywhere, so let's not let this languish any longer.

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -472,15 +472,18 @@ all defined files are detected via disk watches and applied immediately. Files m
 provided in YAML or JSON format. Only changes resulting in well-formed target groups
 are applied.
 
-The JSON version of a target group has the following format:
+The JSON file must contain a list of target groups, using this format:
 
 ```
-{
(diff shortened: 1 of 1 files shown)


content/docs/operating/configuration.md
```diff
@@ -472,15 +472,16 @@ all defined files are detected via disk watches and applied immediately. Files m
...
-  "targets": [ "<host>", ... ],
-  "labels": {
-    [ "<labelname>": "<labelvalue>", ... ]
+[
+  {
+    "targets": [ "<host>", ... ],
+    "labels": {
+      [ "<labelname>": "<labelvalue>", ... ]
+    }
   }
```
grobie: Maybe worth making it extra clear that more than on object can be specified here, something like this could be enough:

```json
  },
  ...
]
```

grobie:
:+1:

content/docs/operating/configuration.md
brian-brazil: Good idea, done.

PR #262, opened 2015-11-13
Grafana supports prom by default now.

alicebob:
As per their release notes: http://grafana.org/blog/2015/10/28/Grafana-2-5-Released.html Grafana has the prometheus datasource by default, so the installation instructions can be simpeler now.

(The text could also sell Grafana a bit more, since Grafana and Prometheus work really rather well together)

### DIFF
--- a/content/docs/visualization/grafana.md
+++ b/content/docs/visualization/grafana.md
@@ -4,36 +4,26 @@ sort_rank: 7
 ---
 
 # Grafana support for Prometheus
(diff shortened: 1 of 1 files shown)


matthiasr:
:+1:

juliusv:
:+1: Thanks! I agree with your proposal to sell Grafana more, especially since we're all tending to go the Grafana way as well. Contributions welcome :)

PR #265, opened 2015-11-18
Adding a query filter example for better understanding in the API documentation

alexissavin:
Following my question in "prometheus/prometheus#1216", may I suggest adding the following example in the documentation ?

Regards.

### DIFF
--- a/content/docs/querying/api.md
+++ b/content/docs/querying/api.md
@@ -57,7 +57,7 @@ to be URL-encoded.
 
 Query language expressions may be evaluated at a single instant or over a range
 of time. The sections below describe the API endpoints for each type of
-expression query.
+expression query and how to filter their result.
 
 ### Instant queries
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR, I don't think this belong in the reference documentation as the docs already state that `expr` can be any Prometheus expression which includes filtering, aggregation, prediction etc. Filtering isn't special in this regard.

alexissavin:
I agree, filtering is not special in this regard, but an example would be appreciated I think, considering that people reading the documentation are probably not yet familiar with the `expr` potential.
In my case, such an example would have saved me a few time understanding how to interact with prometheus.

beorn7:
http://prometheus.io/docs/querying/examples/ is for expression examples. Providing more complex examples for _expressions_ in the API documentation would be weird, especially given that the API documentation is at the end of the chapter about the query language. So the whole expression shebang was explained before. If that's not obvious enough, we could add "... an expression as described in previous sections of the QUERYING chapter".

davidmankin:
Here's what would have helped me with the query api:

    GET /api/v1/label/__name__/values

This is the way to get access to all the different metrics.  If you remember that `__name__` is a magic label then you can extrapolate from the information in "Querying label values", but it would be so great to have this specific example there.

Issue #200, opened 2015-09-02
Table of contents on pages


brian-brazil:
For some pages like the FAQ and SD it'd be good to have a ToC

juliusv:
:+1: I can give this a go.

grobie:
:+1: as discussed a few times.

RichiH:
@ddfreyne can you help here? This would be quite useful.

PR #266, opened 2015-11-26
Upgrade to Nanoc 4

ddfreyne:
This PR upgrades the Prometheus site to Nanoc 4. It also makes use of the new identifiers with extensions, as well as globs.

**Note:** Ruby 2.2+ or JRuby 9000+ is required to use Nanoc 4.

Also see the [Nanoc 4 upgrade guide](http://nanoc.ws/doc/nanoc-4-upgrade-guide/).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'nanoc'
+gem 'nanoc', '~> 4.0'
 gem 'adsf'
 gem 'kramdown'
 gem 'guard-nanoc'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,7 @@ GEM
       timers (~> 4.0.0)
     coderay (1.1.0)
     colored (1.2)
-    cri (2.6.1)
+    cri (2.7.0)
       colored (~> 1.2)
     em-websocket (0.5.1)
       eventmachine (>= 0.12.9)
@@ -40,7 +40,7 @@ GEM
     method_source (0.8.2)
     mini_portile (0.6.0)
     multi_json (1.11.2)
-    nanoc (3.7.3)
+    nanoc (4.0.0)
       cri (~> 2.3)
     nokogiri (1.6.3.1)
       mini_portile (= 0.6.0)
@@ -74,10 +74,13 @@ DEPENDENCIES
   guard-livereload
   guard-nanoc
   kramdown
-  nanoc
+  nanoc (~> 4.0)
   nokogiri
   pygments.rb
   rb-fchange
   rb-fsevent
   rb-inotify
   redcarpet
+
+BUNDLED WITH
+   1.10.6
--- a/Rules
+++ b/Rules
@@ -1,45 +1,21 @@
 #!/usr/bin/env ruby
 
-# A few helpful tips about the Rules file:
-#
-# * The string given to #compile and #route are matching patterns for
-#   identifiers--not for paths. Therefore, you can’t match on extension.
-#
-# * The order of rules is important: for each item, only the first matching
-#   rule is applied.
-#
-# * Item identifiers start and end with a slash (e.g. “/about/” for the file
-#   “content/about.html”). To select all children, grandchildren, … of an
-#   item, use the pattern “/about/*/”; “/about/*” will also select the parent,
-#   because “*” matches zero or more characters.
-
-compile '/assets/*' do
-end
-
-route '/assets/*' do
-  # /assets/foo.html/ → /foo.html
-  item.identifier[0..-2]
-end
-
-route '/README/' do
-    '/README.md'
-end
+passthrough '/static/**/*'
+passthrough '/README.md'
 
 # RSS Feed
-compile '/blog/feed/' do
+compile '/blog/feed.*' do
     filter :erb
 end
 
-route '/blog/feed/' do
-    '/blog/feed.xml'
+route '/blog/feed.*' do
+  '/blog/feed.xml'
 end
 
-compile '*' do
+compile '/**/*' do
   filter :erb
 
-  if item[:title] == 'README.md'
-    # Don't filter; this should propagate verbatim to the output GitHub repository.
-  elsif item[:extension] == 'md'
+  if item[:extension] == 'md'
     #filter :kramdown
     filter :redcarpet, options: {filter_html: true, autolink: true, no_intraemphasis: true, fenced_code_blocks: true, gh_blockcode: true, tables: true}, renderer_options: {with_toc_data: true}
     filter :add_anchors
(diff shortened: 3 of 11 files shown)


ddfreyne:
CC @beorn7 @fabxc

brian-brazil:
Ruby 2.2 is not available on Ubuntu 12.04, we need to stick with 1.9 at at most.

matthiasr:
Ruby 2.2 is readily available through RVM everywhere. Ruby 1.9 is out of support. By the same argument, Prometheus would need to be Go 1.0 compatible.

brian-brazil:
I'm running a modern OS that's supported by my vendor, however:

```
$ rvm 2.2.0
ERROR: Unknown ruby interpreter version: '2.2.0'
```

> By the same argument, Prometheus would need to be Go 1.0 compatible.

This is why I argued that we should pull down the version of Go we need, Go 1.5 is not available in the majority of OSes. Similarly there is no LTS release of Ubuntu with Ruby 2.2 for example, as it was released less than a year ago.

We can't assume that all of our users and developers are running at the bleeding edge for every library and language we choose to use, or that they're even familiar with how to get these environments working.

denis-soundcloud:
FYI, Nanoc 4’s requirement will drop from Ruby 2.2 to Ruby 2.1 with the next patch release (4.0.1).

Ruby 2.0 will be EOL in February, and Ruby 2.1 appears to be provided by all major distros (it’s in Debian Jessie for instance).

brian-brazil:
Even if a given version is in some release, it'll in the best case take an enterprise 6-12 months to upgrade to it. Jessie is only out 7 months, so I wouldn't presume that most Debian users support it especially given LTS for Wheezy extends until 2018.
The latest Ubuntu LTS only has 1.9, and it'll be the end of next year before it'd be okay to even begin presuming a good number of users have 16.04. Thus Ubuntu is at least one major distro without support  for 2.0.

In general I've found that 2-3 years after release is a good rule of thumb as to when you can assume a given piece of software is reasonably available. 11 months is not a reasonable amount of time.

denis-soundcloud:
@brian-brazil Ruby 2.1 was released late 2013, so we’re pretty close to the 2-3 years that you mentioned.

brian-brazil:
That's just a rule of thumb, it can take 5-6 years sometimes if the timing is bad or due to other circumstances. For example it would be unwise to switch to only supporting systemd in a generic .deb until 2019 or so.

There is no Ubuntu LTS with Ruby 2.1, and there won't be a newer version until next April at the earliest (I'm not sure why 14.04 is missing it if it was released in 2013, but them's the breaks).

fabxc:
We are talking about the docs here. It's markdown, no need for hosting to
read them. Hosted via our website too.
So is this seriously a discussion worth spending time on?

Is there a tl;dr on the improvements we gain from the new version?

denis-soundcloud:
The main improvements in Nanoc 4 are identifiers that include the file extension and glob patterns ([details here](http://nanoc.ws/doc/identifiers-and-patterns/))—the old identifiers/patterns are a common source of confusion, and so Nanoc 4 solves this problem once and for all.

Nanoc 4 is spiritually the same as Nanoc 3 and only has small changes, but it breaks backwards compatibility in order to be able to focus on future improvements.

brian-brazil:
> So is this seriously a discussion worth spending time on?

We're telling our users that they can't easily contribute documentation unless they've reinstalled their desktops in the past 6-12 months. That's quite a barrier to entry.

If this goes in I'll no longer be able to test the docs locally. My desktop is only 2 years old. I need my desktop to be reliable, so it purposely doesn't run at the bleeding edge.

matthiasr:
I'm all for upgrading. We won't need a newer Ruby server-side for this.

fabxc:
FWIW, even though I've all sorts of Ruby versions installed I could never test it locally anyway. Of course caused by me not being familiar with Ruby and not spending too much time on the issue. But that's the average doc-contributor – if we argue for users being easily able to test, that ship has sailed way before.

I'm really glad that someone with the required knowledge takes time to look into the code behind the docs and where it can be improved.

Lastly, interpreter/compiler version requirements are perfectly normal for any project. It saves time, keeps things lean, up to date, and allows moving forward on the important aspects.

Downloading and installing a compiler/interpreter at build time is still madness to me. I've never seen it and there are many extremely successful projects out there that have not died because of it.

denis-soundcloud:
As an alternative to requiring the user to have Ruby 2.1+ installed, we could download/vendor JRuby 9000 and use that—no installation of Ruby required.

(It does mean we depend on JVM 8 or later though. Is that an acceptable requirement?)

brian-brazil:
> FWIW, even though I've all sorts of Ruby versions installed I could never test it locally anyway. Of course caused by me not being familiar with Ruby and not spending too much time on the issue. But that's the average doc-contributor – if we argue for users being easily able to test, that ship has sailed way before.

I've already had to spend time figuring out how to get 1.9 working on my desktop and figuring out what 'bundle' is and how to run it with the non-default version of Ruby (1.8 is the default, but 1.9 is provided by my distro). Once that's figured out, the instructions in the readme work. I've caught many syntactic mistakes in my PRs by testing my changes, and I would have thought that everyone else was doing the same.

> Lastly, interpreter/compiler version requirements are perfectly normal for any project. It saves time, keeps things lean, up to date, and allows moving forward on the important aspects.

The discussion here isn't about version requirements, it's about relatively recent versions being required. That's not free. It might save some people time but it can cost others large amounts of time, or indeed make it impractical for them to contribute. For example the discussion we're having here threatens to eat a chunk of my time for little apparent benefit to me or the project.

> (It does mean we depend on JVM 8 or later though. Is that an acceptable requirement?)

It's not acceptable for similar reasons. Java 6 is about the safest you can assume when writing code (per our user survey, about 2/3 of our users are using Java 6 in some way still), Java 7 is pushing it a bit, but is probably okay in terms of the JVM being available.

Which is fun, considering Java 7 was released in 2011 and is publicly EOLed. Java 8 has only been out a year and a half.

fabxc:
There are simpler ways to validate Markdown syntax than running the whole thing.

Can we provide a Dockerfile that allows us to upgrade while allowing anyone (with Docker) to test?

brian-brazil:
Github markdown is not quite the same as markdown elsewhere (presuming you're using the preview functions there), I've had to fix some problems there for other projects. It's always safest to run the generator you're using.

A dockerfile would help.

Someday I'll have time for round 3 of trying to get Docker working on my desktop...

beorn7:
This kind of problem seems to be an ideal usecase for Docker. Would also be nice to have automated checks in Github PRs via some CI so that contributors don't even have to think about the testing.

In general, if the Nanoc upgrade has tangible benefits, we should not keep it back because of obsolete Java or Ruby versions in distributions. Only very few people have to run the site generator. It's a (small) burden for a few. I never had problems generating the site. I run Ubuntu 14.04 LTS and use rvm if I require a specific Ruby version.

Obviously, if they are no real benefits with the upgrade, we don't need to bother right now.

ddfreyne:
FYI, the last Nanoc release lowers the minimum required Ruby version from 2.2 to 2.1.

>> (It does mean we depend on JVM 8 or later though. Is that an acceptable requirement?)
> 
It's not acceptable for similar reasons. Java 6 is about the safest you can assume when writing code (per our user survey, about 2/3 of our users are using Java 6 in some way still), Java 7 is pushing it a bit, but is probably okay in terms of the JVM being available.

Oracle is no longer releasing updates to JDK 7 and encourages moving to JDK 8.

beorn7:
I'd be totally fine with Ruby 2.1 (speaking as Ubuntu LTS user).

@ddfreyne How would you sell us the benefits of the Nanoc upgrade for our docs site? As said, if there is nothing to gain for us right now, it's difficult to justify the effort it would cause for some.

ddfreyne:
The main reason for upgrading to Nanoc 4 is that it’s easier to understand, because of the new identifiers with extensions (identifiers no longer end with a slash, but behave like filenames) and globs (wildcards now behave like standard *nix globs). This enhancement also removes the need for the workaround with the “static” data source, which has been a thorn in the flesh for a long time.

Nanoc 3 is in maintenance mode and will not get new features or speed improvements. Nanoc 4, on the other hand, is actively developed and will get features and enhancements which should be generally beneficial to anyone using Nanoc.

beorn7:
@ddfreyne Thanks.

I'm a pretty light Nanoc user so far, so while I think it's good to upgrade in general (for the stated general reasons) and I have no trouble running Ruby 2.1, I can't really claim I have concrete need for any of the immediately available new features.

Could more Nanoc-involved contributors to the Prometheus docs speak up for or against an upgrade?

brian-brazil:
If this could get us #200 that'd be good, otherwise what we have seems to work okay currently.

ddfreyne:
#200 can be (pretty easily) implemented without going to Nanoc 4.

I’m OK with leaving the upgrade for a later time (even though I’d be a bit sad!), but I hope the Prometheus documentation will switch to use software that’s maintained.

RichiH:
As a data point, not to bikeshed:

Personally, I don't expect to write anything complex enough so that I can't do it in my head. While having  docker as an option sounds nice, a CI system which publishes to some random subdomain with a `robots.txt` to keep the search engines at bay sounds useful.

ddfreyne:
Closing due to lack of activity (1+ year) and merge conflicts.

PR #274, opened 2015-12-21
Update clientlibs.md

discordianfish:
https://github.com/siimon/prom-client seems like the more active and more feature complete project


brian-brazil:
:+1: 

@siimon

siimon:
Cool! I'll try to be around at irc as much as I can!

PR #275, opened 2015-12-23
[WIP] new alerting/AM docs

fabxc:
Some of the new concepts are certainly not trivial to explain.
So comments and ideas are welcome at this early stage.

We have to think about restructuring since it's currently one large 'Prometheus server documentation' that doesn't really accommodate Alertmanager docs.


content/docs/alerting/alertmanager.md
```diff
@@ -6,196 +6,101 @@ nav_icon: sliders
...
+Silences are a straightforward way to simply mute alerts for a given time.
+A silence is configured based on matchers, just as the routing tree. Incoming
+alerts are checked whether they match all the equality or regular expression
+matchers of an active silence.
+If they do, no notifications will be send out for that alert.
+
+Silences are configured in the web interface of the Alertmanager.
+
+
+## Sending alerts
```
fabxc: A section for pushing clients will do for the first iteration.
The other endpoints are not stable enough to document an official API yet.

brian-brazil: I think we should keep this separate from the main alertmanager docs, as it's something that extremely few users will need and it'll cause confusion for the rest.

fabxc: That's why the initial disclaimer below, which we can make even more visible.
I think the very bottom of an existing page is more subtle than a separate section in the navigation bar, which users will click by just looking around.

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Route `<route>`
+
+A route block defines a node in a routing tree and its children. Its optional
+configuration parameters are inherited from its parent node if not set.
+
+Every alert enters the routing tree at the configured top-level route, which
+must match all alerts.
+It then traverses the child nodes. If `continue` is set to false, it stops
+after the first matching child. Otherwise it potentially matches all child
```
brian-brazil: "it attempts to match the following child nodes"?

Continue only says to keep going, not to try everything.

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,207 @@
...
+```
+[ receiver: <string> ]
+[ group_by: '[' <labelname>, ... ']' ]
+
+# Zero or more child routes.
+routes:
+  [ - <route> ... ]
+
+# Whether an alert should continue matching further child nodes
+# after the first match with a child node.
```
brian-brazil: s/with a child node//

content/docs/alerting/rules.md
```diff
@@ -15,53 +15,63 @@ Alerting rules are configured in Prometheus in the same way as [recording
...
+information such as alert descriptions or runbook links. The annotation values
+can be templated.
+
+#### Templating
 
-The `SUMMARY` should be a short, human-readable summary of the alert (suitable
-for e.g. an email subject line), while the `DESCRIPTION` clause should provide
-a longer description. Both string fields allow the inclusion of template
-variables derived from the firing vector elements of the alert:
+Label and annotation values can be templated using Go's template language.
```
brian-brazil: This should have a pointer over to the template reference, as it's not purely Go's template language.

content/docs/alerting/alertmanager.md
brian-brazil: I'd suggest just dropping these docs for now then rather than causing confusion, maybe put them in the AM repo somewhere as implementation reference docs. The warning currently reads as though it generally applies.

content/docs/alerting/configuration.md
fabxc: We had a bit of discussion around that offline. Eventually we found that a flag for match-first or match-all children is easier to grasp and also newly inserted nodes to starve their following siblings by consuming everything and having `continue=false`.

Do you see any use cases not covered by this?

content/docs/alerting/rules.md
fabxc: Yes, need reference. But what do you mean?
It's purely Go's template language + functions and pre-defined templates, which I consider still part of that.

content/docs/alerting/alertmanager.md
fabxc: That strongly depends on how independently we want to advertise AM as an alert handling solution that works without Prometheus.

content/docs/alerting/configuration.md
brian-brazil: I expect this to work similar to procmail, where the `c` flag is on per-rule basis.

What you're describing isn't a 'continue' it's a 'apply to all children'. I've seen the latter behaviour cause endless confusion and complexity with similar systems in the past.

I think you can do everything in both systems, but having it per child will be much more understandable. Usually what someone wants to CC alerts to somewhere, not to have to write all their following rules to exclude the once that are already "matched".

content/docs/alerting/rules.md
brian-brazil: I think mentioning Go here may cause users to think it's yet another templating language inside Prometheus, rather us just having the one.

fabxc: The thing is, Go's templating is documented. Ours is not. Which is luckily not necessary, because Go's is, and our's is Go's.

content/docs/alerting/configuration.md
fabxc: Changing the semantics and keeping the word wasn't ideal.

You said your's is more understandable but didn't give reasons. I don't see it that way. You didn't address the starvation aspect, which is very real, especially at the top level.
For a level with 20 siblings, you would have to go through all 19 to know whether or not you receive the alerts you think you do – and that's only valid until the next matcher changes.

I couldn't find too many use cases yet for stop-first semantics actually. Most of the time the matchers are mutually exclusive to begin with.

content/docs/alerting/rules.md
brian-brazil: The reference here should be to our general templates (which mention Go), not to Go's. To do otherwise risks confusion.

content/docs/alerting/configuration.md
brian-brazil: > You said your's is more understandable but didn't give reasons.

It comes from experience, it breaks down as soon as you try to do anything even mildly complicated. As routing rules are business logic that changes with the systems and teams, they tend to end up very complicated with lots of nesting and special cases.

> I couldn't find too many use cases yet for stop-first semantics actually. Most of the time the matchers are mutually exclusive to begin with.

The usual way alert routing ends up is something like:

  * Send copy of all alerts to email
  * specialisation for team X
  * specialisation for subsystem B
  * specialisation for team Y
 *  ...
  * general rule for paging alerts
  * catchall

These rules are not mutually exclusive as you usually want a given alert to only go to one team, and writing them with the all-match is very tricky to do correctly as for every specialisation you add you'd need to change every other rule that could also apply. First match wins is easier to understand and maintain, as you can read it top to bottom and know what happens.

> You didn't address the starvation aspect, which is very real, especially at the top level.

That's what unit tests of your routing rules are for.

content/docs/alerting/rules.md
fabxc: Yes, of course there should be a section on our own templating first with a transitive reference to Go. We don't have that for AM yet though.

content/docs/alerting/alertmanager.md
```diff
@@ -6,196 +6,101 @@ nav_icon: sliders
...
+
+The following describes the core concepts the Alertmanager implements. Consult
+the [configuration documentation](../configuration) to learn how to use them
+in more detail.
+
+## Grouping
+
+Grouping categorizes alerts of similar nature into a single notification. This
+is especially useful during larger outages when many systems fail at once and
+hundreds the thousands of alerts may be firing simultaniously.
```
juliusv: s/the/of/

juliusv: s/simultaniously/simultaneously/

content/docs/alerting/alertmanager.md
```diff
@@ -6,196 +6,101 @@ nav_icon: sliders
...
+
+**Example:** Dozens or hundreds of instances of a service are running in your
+cluster when a network partition occurs. Half our your service instances
```
juliusv: s/our/of/

content/docs/alerting/alertmanager.md
```diff
@@ -6,196 +6,101 @@ nav_icon: sliders
...
+Alertmanager to group alerts by their cluster and alertname so it sends a
+single compact notification.
+
+Grouping of alerts, timing for the grouped notifications, and the receivers
+of those notificiations are configured by a routing tree in the configuration
+file.
+ 
+## Inhibition
+
+Inhibition is a concept of surpressing notifications for certain alerts if
```
juliusv: s/surpressing/suppressing/

content/docs/alerting/alertmanager.md
```diff
@@ -6,196 +6,101 @@ nav_icon: sliders
...
+certain other alerts are already firing.
+
+**Example:** An alert is firing that informs that an entire cluster is not
+reachable. Alertmanager can be configured to mute all other alerts concerning
+this cluster if that particular alert is firing.
+This prevents hundreds to thousands of alerts firing unrelated to the actual
```
juliusv: The alerts are still firing, but notifications are inhibited. Maybe, "This prevents notifications for hundreds or thousands of firing alerts that are unrelated to the actual issue."?

content/docs/alerting/alertmanager.md
```diff
@@ -6,196 +6,101 @@ nav_icon: sliders
...
+
+
+## Sending alerts
+
+__Prometheus automatically takes care of sending alerts generated by its
+configured [alerting rules](../rules). The following is a general documentation for clients.__
+
+The Alertmanager listens for alerts on an API endpoint at `/api/v1/alerts`.
+Clients are expected to continously re-send alerts as long as they are still
+active (usually at the order of 30 seconds to 3 minutes).
```
juliusv: "on the order of" (or "in" or "of" seem to be also allowed, but not "at" :) )

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,207 @@
...
+* `<filename>`: a valid path in the current working directory
+* `<boolean>`: a boolean that can take the values `true` or `false`
+* `<string>`: a regular string
+* `<tmpl_string>`: a string which is template-expanded before usage
+
+The other placeholders are specified separately.
+
+A valid example file can be found [here](https://github.com/prometheus/alertmanager/blob/master/doc/examples/simple.yml).
+
+The global configuration specifies parameters that are valid in all other configuration contexts. They also serve as defaults for other configuration
```
juliusv: Long line.

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,207 @@
...
+
+A route block defines a node in a routing tree and its children. Its optional
+configuration parameters are inherited from its parent node if not set.
+
+Every alert enters the routing tree at the configured top-level route, which
+must match all alerts.
+It then traverses the child nodes. If `continue` is set to false, it stops
+after the first matching child. Otherwise it potentially matches all child
+nodes. The procedure repeats until an alert does not match any child nodes of
+the current node or there are none. Then, the routing and grouping paramteres
```
juliusv: s/paramteres/parameters/

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,207 @@
...
+
+# A set of equality matchers an alert has to fulfill to match the node.
+match:
+  [ <labelname>: <labelvalue>, ... ]
+
+# A set of regex-matchers an alert has to fulfill to match the node.
+match_re:
+  [ <labelname>: <regex>, ... ]
+
+# How long to wait initially wait to send a notification for a group
```
juliusv: one "wait" too much

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,207 @@
...
+# of alerts. Allows to wait for an inhibiting alert to arrive or collect
+# more initial alerts for the same group. (Usually ~0s to few minutes.)
+[ group_wait: <duration> ]
+
+# How long to send a notification about new alerts and existing group
```
juliusv: "to send" -> "before sending"

"and existing group received"? This phrase does not make sense to me.

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,207 @@
...
+# received. (Usually ~5min or more.)
+[ group_interval: <duration> ]
+
+# How long to wait to send a notification again if it has already
```
juliusv: "to send" -> "before sending"

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,207 @@
...
+# been sent successfully for an alert. (Usually ~3h or more).
+[ repeat_interval: <duration> ]
+```
+
+### Example
+
+```
+# The root route with all paramteres, which are inherited by the child
```
juliusv: s/paramteres/parameters/

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,207 @@
...
+Both alerts must have a set of equal labels.
+
+```
+# Matchers that have to be fulfilled in the alerts to be muted.
+target_match:
+  [ <labelname>: <labelvalue>, ... ]
+target_match_re:
+  [ <labelname>: <regex>, ... ]
+
+# Matchers for which one or more alerts have to exists for the
```
juliusv: s/exists/exist/

content/docs/alerting/rules.md
```diff
@@ -15,53 +15,63 @@ Alerting rules are configured in Prometheus in the same way as [recording
...
     # To insert the numeric expression value of the firing element:
-    {{$value}}
+    {{ $value }}
 
 Examples:
 
     # Alert for any instance that is unreachable for >5 minutes.
     ALERT InstanceDown
-      IF up == 0
+      IF up == 0
```
juliusv: Is this introducing a trailing space?

content/docs/alerting/rules.md
juliusv: I think this is fine. How about something like:

"Label and annotation values can be templated using Go's template language. The same [template functions](/docs/visualization/template_reference/) as in Prometheus's console templates are available here."

brian-brazil: That's inaccurate, it is the console templates in use here and that's how the console template docs approach it too. This should only reference console templates, and not confuse users with mentions of Go.

content/docs/alerting/alertmanager.md
brian-brazil: I think we should treat it similarly to the exposition formats. It an important technical detail that needs to be documented, but kept well away from beginners.

content/docs/alerting/configuration.md
brian-brazil: Following offline discussion with @fabxc it turns out it is already implemented the way I proposed, so we just need to update the docs in this PR.

content/docs/alerting/rules.md
fabxc: How about leaving this for the first iteration until we have a templating documentation for AM templates. AM templates are not the same as console templates. The available functions are different, so a reference to Go is more helpful than to console templates.

By the time the docs for the new AM hit master, we will have something in place. But having a branch with an initial version of the docs is necessary for a beta release. For beta testers that should be understandable enough.

fabxc:
Addressed many comments.

We should improve our structure anyway, so this doesn't have to be final and perfect. But if we can merge this as an initial version, we can cut a beta version for AM and have docs to point testers to.

content/docs/alerting/rules.md
brian-brazil: This section is about alert templates though on the Promethus side (which are console templates), not AM notification templates.

fabxc: My bad, all correct then.

fabxc:
Fixed templating reference.
Added documentation for receiver configurations.

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,347 @@
...
+
+# The root node of the routing tree.
+route: <route>
+
+# A list of inhibition rules.
+inhibit_rules:
+  [ - <inhibit_rule> ... ]
+
+# A list of notification receivers.
+receivers:
```
brian-brazil: I'd say put the receivers above the inhibits, as inhibits are optional but receivers aren't.

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,347 @@
...
+none exist), the alert is handled based on the configuration paramters of the
+current node.
+
+
+```
+[ receiver: <string> ]
+[ group_by: '[' <labelname>, ... ']' ]
+
+# Zero or more child routes.
+routes:
```
brian-brazil: routes would be better after match I think, as it doesn't apply until they do

content/docs/alerting/alertmanager.md
brian-brazil: This section still needs to be removed from here. Implementation details not relevant to 99% of users have no place on an overview document that's otherwise looking at concepts. It makes it seem as though this is a essential thing for users to be aware of.

fabxc:
@brian-brazil can you give this another pass?

content/docs/alerting/alertmanager.md
```diff
@@ -6,196 +6,64 @@ nav_icon: sliders
...
+alerts are checked whether they match all the equality or regular expression
+matchers of an active silence.
+If they do, no notifications will be send out for that alert.
+
+Silences are configured in the web interface of the Alertmanager.
+
+
+## Client behavior
+
+The Alertmanager has [special requirements](../clients) for behavior of its
```
brian-brazil: Did you forget to add this file to the PR?

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,348 @@
...
+defined by the scheme described below.
+Brackets indicate that a parameter is optional. For non-list parameters the
+value is set to the specified default.
+
+Generic placeholders are defined as follows:
+
+* `<duration>`: a duration matching the regular expression `[0-9]+[smhdwy]`
+* `<labelname>`: a string matching the regular expression `[a-zA-Z_][a-zA-Z0-9_]*`
+* `<labelvalue>`: a string of unicode characters
+* `<filename>`: a valid path in the current working directory
```
brian-brazil: You use `filepath` below

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,348 @@
...
+```
+global:
+  # ResolveTimeout is the time after which an alert is declared resolved
+  # if it has not been updated.
+  [ resolve_timeout: <duration> | default = 5m ]
+
+  # The default SMTP From header field.
+  [ smtp_from: <tmpl_string> ]
+  # The default SMTP smarthost used for sending emails.
+  [ smtp_smarthost: <tmpl_string> ]
```
brian-brazil: From the code this looks to be a plain string

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,348 @@
...
+
+
+## Email receiver `<email_config>`
+
+```
+# Whether or not to notify about resolved alerts.
+[ send_resolved: <boolean> | default = false ]
+
+# The email address to send notifications to.
+to: <string>
```
brian-brazil: tmpl_string

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,348 @@
...
+# The sender address.
+[ from: <email> | default = global.smtp_from ]
```
brian-brazil: also tmpl_string

content/docs/alerting/configuration.md
```diff
@@ -0,0 +1,348 @@
...
+      "labels": <object>,
+      "annotations": <object>,
+      "startsAt": "<rfc3339>",
+      "endsAt": "<rfc3339>"
+    },
+    ...
+  ]
+}
+```
+
```
brian-brazil: Two consecutive code blocks could be confusing, I'd put them the other way around so that the last above sentence can go between them.

fabxc:
All addressed

content/docs/alerting/alertmanager.md
```diff
@@ -6,196 +6,64 @@ nav_icon: sliders
...
+this cluster if that particular alert is firing.
+This prevents notifications for hundreds or thousands of firing alerts that
+are unrelated to the actual issue.
+
+Inhibitions are configured through the Alertmanager's configuration file.
+
+## Silences
+
+Silences are a straightforward way to simply mute alerts for a given time.
+A silence is configured based on matchers, just as the routing tree. Incoming
```
brian-brazil: s/as/like/

brian-brazil:
:+1:

PR #278, opened 2016-01-02
Add initial glossary

### LINKED ISSUE #177: Add a glossary
We've lots of terms, and I expect some of them (like bridge vs. collector vs exporter) are only in a few people's heads. We should have a glossary.

brian-brazil:
Fixes #177

### DIFF
--- a/content/docs/introduction/glossary.md
+++ b/content/docs/introduction/glossary.md
@@ -1,0 +1,81 @@
+---
+title: Glossary
+sort_rank: 8
+---
+
+# Glossary
+
+
+### Alert
+
+An alert is the outcome of an alerting rule in Prometheus that is
+actively firing. Alerts are sent from Prometheus to the Alertmanager.
+
+### Alertmanager
+
+The [Alertmanager](../../alerting/overview/) takes in alerts, aggregates them into
+groups, de-duplicates, applies silences, throttles, and then sends out
+notifications to email, Pagerduty, Slack etc.
+
+### Bridge
+
+A bridge is a component that takes samples from a client library and
+exposes them to a non-Prometheus monitoring system. For example the Python
+client can export metrics to Graphite.
+
+### Client library
+
+A client library is a library in some language (e.g. Go, Java, Python, Ruby)
+that makes it easy to directly instrument your code, write custom collectors to
+pull metrics from other systems and expose the metrics to Prometheus.
+
+### Collector
+
+A collector is a part of an exporter that represents a set of metrics. It may be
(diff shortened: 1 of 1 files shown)


RichiH:
A few comments:

* Nitpicky, but an Exporter is not always a binary.
* A target is always scraped via an exporter (as far as I know)
* Mentioning that PromQL is turing-complete would give most people a better idea of the power it has
* Cross-links between the various definitions in the glossary might be worthwhile. I can do that if you don't want to
* Personal preference, but I would tend towards using absolute, not relative, links in your cross-references as that allows moving around of glossary.md while all links need to be changed if targets move, either way.

brian-brazil:
> Nitpicky, but an exporter is not always a binary.

An exporter is always a binary.

> A target is always scraped via an exporter

Targets are what Prometheus scrapes.

> Mentioning that PromQL is turing-complete would give most people a better idea of the power it has

I'd rather not scare them off :)

> Cross-links between the various definitions in the glossary might be worthwhile. I can do that if you don't want to

That'd be appreciated.

> Personal preference, but I would tend towards using absolute, not relative, links in your cross-references as that allows moving around of glossary.md while all links need to be changed if targets move, either way.

We've currently a mix, and as we were talking about having next-release hosted somewhere then relative make more sense.

RichiH:
> An exporter is always a binary.

snmp_exporter is a script. Again, this is rather nitpicky.

> > Cross-links between the various definitions in the glossary might be worthwhile. I can do that if you don't want to
> That'd be appreciated.

Will do after you merge that in.

> We've currently a mix, and as we were talking about having next-release hosted somewhere then relative make more sense.

Sorry, I meant `[bar](../../foo/bar)` vs `[bar](/quux/foo/bar)`

beorn7:
:+1: 

Eventually, we should clarify terms like _metric_ ws. _metric family_ vs. _time series_ etc.

brian-brazil:
This is just a first pass with the more common confusions, we can add to it over time.

RichiH:
@beorn7 That addition should take you less than 15 minutes :)

beorn7:
@RichiH It's a bit more complicated than you might think…

RichiH:
@beorn7 For the three above, that seems unlikely. I might be missing context, though.

juliusv:
@RichiH I guess the challenge would be that this terminlogy is not used 100% consistently between the client libraries, wire formats, and Prometheus itself.

RichiH:
@juliusv While that increases the workload, that makes normalizing on common nomenclature even more desireable :)

juliusv:
Agreed :)

PR #279, opened 2016-01-05
functions.md: Split count() out into its own section

RichiH:


fabxc:
`count()` is not a function but an aggregator and [documented as such](https://github.com/prometheus/docs/blob/master/content/docs/querying/operators.md#aggregation-operators). The distinction is not immediately obvious, especially if there is no `by` clause involved.

RichiH:
Thanks for pointing that out.

I am not sure if that distinction makes sense from the end user PoV, but would you merge a PR which tries to make this disctinction clearer? Off-hand, I would make the two occurences of "aggregation operator" links to the appropriate section in the aggregator page.

fabxc:
Sure, that's obviously something the documentation should make clearer.

PR #281, opened 2016-01-05
operators.md: Switch order in operator precedence list

RichiH:
Merge it or close it, but lowest to highest is really weird to me in this context.


fabxc:
I agree. Let's wait a bit to give others a chance to voice their opinions.

brian-brazil:
Highest to lowest is more common.

PR #286, opened 2016-01-10
Add PowerDNS exporter

janeczku:


brian-brazil:
Thanks for the new exporter.

You should have a read of https://docs.google.com/document/d/1JapuiRbp-XoyECgl2lPdxITrhm5IyCUq9iA_h6jp3OY/edit to help your users get the most out of the exporter.

In particular:

 - Prometheus has a histogram type, you should use that for `answers_rtime_total`. It's also encouraged to put the unit as seconds, and put the unit in the metrics name. The cache hits should be merged into 0ms, and then add a separate counter for them. 
 - latency_avg_microseconds has inconsistent units with the rest, seconds are recommended. What time period is this gauge over? We strongly encourage leaving all rate calculations to be done on the Prometheus side, export the total time spent if you have it.
 - `qTypeLabelMap` doesn't seem to be used

janeczku:
@brian-brazil  Thanks Brian for your valuable recommendations! :smiley:

janeczku:
> Prometheus has a histogram type, you should use that for answers_rtime_total.

@brian-brazil  I looked into this:
The histogram type doesn't have a `Set()` method. That's why it cannot be used with the PowerDNS API who only exposes counters and not event streams.

brian-brazil:
When writing an exporter you shouldn't be using the usual metric types, rather use ConstMetric.

janeczku:
Right, but like i said, i don't have the samples for the response times, just the counts of samples for each bucket. Hence, i have no way of calculating the Sum for the ConstHistrogram metric.

brian-brazil:
We haven't run into this particular problem before, I'd suggest putting in a 0. The sum isn't needed for the histogram to work,

janeczku:
It isn't? Okay i will give ConstHistogram a try then!

juliusv:
@brian-brazil Wouldn't `NaN` make more sense if the sum is unknown?

brian-brazil:
Either could work.

juliusv:
Right, but `NaN` more correctly expresses that we don't know what it is. `0` doesn't say "absent data", it's just wrong?

PR #288, opened 2016-01-13
Note the required ordering of lines with timestamps

samdroid-apps:
When attempting to feed Prometheus timestamped metrics, the timestamps
were not adhered to when the lines were unordered.  This should be
documented to avoid this mistake.

### DIFF
--- a/content/docs/instrumenting/exposition_formats.md
+++ b/content/docs/instrumenting/exposition_formats.md
@@ -83,7 +83,8 @@ syntax (EBNF):
 `value` is a float, and timestamp an `int64` (milliseconds since epoch, i.e. 1970-01-01 00:00:00 UTC, excluding leap seconds), represented as required by the [Go strconv package](http://golang.org/pkg/strconv/) (see functions `ParseInt` and `ParseFloat`). In particular, `Nan`, `+Inf`, and `-Inf` are valid values.
 
 All lines for a given metric must be provided as one uninterrupted group, with
-the optional `HELP` and `TYPE` lines first.
+the optional `HELP` and `TYPE` lines first.   If the timestamp is provided,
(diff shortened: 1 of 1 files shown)


samdroid-apps:
CC @juliusv

grobie:
@beorn7

brian-brazil:
This is not an officially supported feature of Prometheus, I don't think we should be documenting this.

samdroid-apps:
Timestamps are documented in the spec. Maybe you should document it is not supported?

But either way, it may be helpful for people who are trying to use timestamps.

brian-brazil:
Timestamps are documented, the ability to pass multiple of them in one scrape for the same timeseries is not. No exporter should be exposing metrics in this manner.

beorn7:
I'm surprised that you can ingest multiple samples for the same time series in this way at all. (I thought it would only work with protobuf, and even that only by accident if parallel ingestion doesn't reorder metric families.)

To clarify, we should add that any name/label combination must be unique. Otherwise the behavior is undefined.

beorn7:
@samdroid-apps I have checked the code. What you did is currently indeed possible but only by accident. The correct documentation is that multiple occurrences of the exact same metric (i.e. same name and same label-name/label-value combination) will lead to undefined behavior.

I'll add a clarification to the protocol documentation.

BTW: The feature you are looking for doesn't exist yet, but it is tracked in https://github.com/prometheus/prometheus/issues/535 .

beorn7:
See https://github.com/prometheus/docs/pull/290

RichiH:
Wouldn't it then make sense to document that you may only ever expose a given, specific metric exactly once per scrape even if you include a timestamp?

If yes, I will gladly send a PR.

beorn7:
This is documented by now.

PR #295, opened 2016-01-19
Fix 2 typos

dabcoder:
Minor corrections to the FAQ.

### DIFF
--- a/content/docs/introduction/faq.md
+++ b/content/docs/introduction/faq.md
@@ -11,7 +11,7 @@ sort_rank: 5
 Prometheus is an open-source systems monitoring and alerting toolkit
 with an active ecosystem. See the [overview](/docs/introduction/overview/).
 
-### How does Prometheus compare against \[other monitoring system\]?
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

content/docs/introduction/faq.md
```diff
@@ -11,7 +11,7 @@ sort_rank: 5
 Prometheus is an open-source systems monitoring and alerting toolkit
 with an active ecosystem. See the [overview](/docs/introduction/overview/).
 
-### How does Prometheus compare against \[other monitoring system\]?
+### How does Prometheus compare against \[other monitoring systems\]?
```
juliusv: This wasn't actually a typo, but a placeholder (thus the square brackets). But I guess the way it's now is ok too.

grobie: It'd actually make more sense without the brackets now. It'd also be more readable imho.

juliusv: Pushed to master.

PR #296, opened 2016-01-20
Table of contents

### LINKED ISSUE #200: Table of contents on pages

grobie:
This adds a table of contents filter. After playing around with that and finding myself adding a toc to more and more pages, I finally just activated it everywhere. This should make the documentation pages consistent at least. We might want to deactivate it on 1-2 pages though, let me know if you think that's important.

Some sections have (unnecessary IMHO) long headers, especially several FAQ entries. It might be helpful for readability reasons to shorten the question and move information into the explanation instead @beorn7.

@brian-brazil @RichiH @juliusv 

Fixes #200.

screenshot from 2016-01-20 00-15-32
screenshot from 2016-01-20 00-15-21

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -72,7 +72,7 @@ scrape_configs:
 ```
 
 
-### Scrape configurations `<scrape_config>`
+### `<scrape_config>`
 
 A `scrape_config` section specifies a set of targets and parameters describing how
 to scrape them. In the general case, one scrape configuration specifies a single
@@ -195,7 +195,7 @@ Where `<scheme>` may be `http` or `https` and `<path>` is a valid URL path.
 regex `[a-zA-Z_][a-zA-Z0-9_-]`.
 
 
-### TLS configuration `<tls_config>`
+### `<tls_config>`
 
 A `tls_config` allows configuring TLS connections.
 
@@ -212,7 +212,7 @@ A `tls_config` allows configuring TLS connections.
 ```
 
 
-### Target groups `<target_group>`
+### `<target_group>`
 
 A `target_group` allows specifying a list of targets and a common label set for them.
 They are the canonical way to specify static targets in a scrape configuration.
@@ -231,7 +231,7 @@ Where `<host>` is a valid string consisting of a hostname or IP followed by a po
 number.
 
 
-### DNS-SD configurations `<dns_sd_config>`
+### `<dns_sd_config>`
 
 A DNS-SD configuration allows specifying a set of DNS record names which
 are periodically queried to discover a list of targets (host-port pairs). The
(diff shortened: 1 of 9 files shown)


brian-brazil:
:+1: Great!

For the FAQ I'd say we make the ToC take up the full width, rather than being a float. The questions are always going to be long.

RichiH:
Not tested, but looks good.

Richard

Sent by mobile; excuse my brevity.

denis-soundcloud:
:+1:

I wanted to get around to adding this, but I suppose you beat me to it!

beorn7:
I'm fine with either, making the FAQ TOC full width or try to keep FAQs shorter (with more explanation in the answer). Obviously, when I wrote those FAQ, I didn't optimize for them showing up in a TOC.

A remotely related problem I see is that the directory on the left side is longer than the page height, which is especially confusing if you are reading a page where the highlighted entry on the left is below the fold.

Now that we have TOCs, perhaps we should merge some of the sections with own entries on the left into larger pages. Or is there a Nanoc feature to condense sections I'm not looking at (e.g. only show "querying" on the left and not the subsections if I'm not reading anything in the "querying" section)?

fabxc:
:+1:

juliusv:
Woot :+1: Thanks @grobie!

content/docs/operating/configuration.md
```diff
@@ -399,7 +399,7 @@ Prometheus relabeling to control which instances will actually be scraped. Also
 by default all apps will show up as a single job in Prometheus (the one specified
 in the configuration file), which can also be changed using relabeling.
 
-### Zookeeper Serverset SD configurations `<serverset_sd_config>`
+### `<serverset_sd_config>`
```
juliusv: Deleting the full headline here removes the Zookeeper mention. Maybe at least mention Zookeeper in the paragraph below?

grobie: It is mentioned in the paragraph.

grobie: http://prometheus.io/docs/operating/configuration/#<serverset_sd_config>

grobie: The reason was that the TOC doesn't break and reads a lot cleaner now http://prometheus.io/docs/operating/configuration/

grobie: I wonder if we should remove the `<` `>` from the headers/links. Github (and I suspect others) will remove them from links as they look like a HTML tag.

juliusv: Ah, GitHub's collapsed code view played a trick on me, sorry.

Yeah, I think it's fine to remove the angled brackets.

denis-soundcloud: The filter that generates the ToC supports `data-nav-title` attributes on the headers, which will be used in the ToC rather than the full header title. The Nanoc site uses it on e.g. the [Troubleshooting](http://nanoc.ws/doc/troubleshooting/) page ([raw source](https://raw.githubusercontent.com/nanoc/nanoc.ws/master/content/doc/troubleshooting.md)).

It doesn’t seem like this works with RedCarpet, though. The Nanoc site uses Kramdown which supports it.

PR #297, opened 2016-01-20
Use light color in code blocks

grobie:
The current black code blocks are very dominant. In my personal view
they are too disruptive when scaning a page. This change makes them
similar to inline code examples.

@brian-brazil @fabxc 

screenshot from 2016-01-20 11-16-05

### DIFF
--- a/static/docs.css
+++ b/static/docs.css
@@ -224,8 +224,9 @@ a.sc-logo img {
 
 pre {
   font-family: "Courier New", Monaco, Menlo, Consolas, monospace;
-  background-color: #444;
-  color: #f8f8f2;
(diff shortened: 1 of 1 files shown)


brian-brazil:
:+1:

fabxc:
Totally agree. But the red makes it look like warning boxes, for which we might have other uses (or have already somewhere?).

grobie:
I see what you mean. The color of Prometheus is red though. I changed the
border to be grey. If someone wants to change the background-color of
inline+block code, please go ahead.

fabxc:
Cool, thanks

PR #300, opened 2016-01-25
Add company logos to the frontpage.

brian-brazil:
Social proof helps us gain users, and avoid misconceptions as to us being a one-company project.

This is what this looks like:
screenshot - 250116 - 14 35 48

I went with alphabetical for simplicity.


fabxc:
Great, thanks!
Those are all verified to be okay with showing up on our site?

juliusv:
Great! But the shapes, colors and sizes all seem like quite a jumble here. I wonder if we can make it all look a bit more uniform somehow. I can give it a try later too - I'm still finishing up the corresponding blog post.

brian-brazil:
The problems is that the aspect ratios are all over the place (Ericsson vs Robust Percaption at the extreme). I made them all a maximum of 256x128, and with the table it mostly looks okay.

fabxc:
For maximum visual appeal the way to go is probably a single image manually composed – that will be more difficult with links though.

brian-brazil:
We could drop the links, though some of these companies are lesser known and can't be identified by logo alone. I'm also expecting a reasonable amount of churn.

grobie:
The section takes up a bit too much space IMHO, I'd go with smaller logos. I'm always more leaning towards understatement though, so I might be wrong here.

In order to present the logos in a visually more uniform way, I guess all logos should be in a div with the same dimensions. That will at least make all rows the same height.

juliusv:
Yes, at the very least, the rows need to be the same height.

brian-brazil:
I've made all the rows the same height.

grobie:
Yeah, that's definitely better. The section still doesn't align well with the rest of the page, the title and logos are much bigger than the main part. If it were for me I'd shrink down the logos.

juliusv:
Yeah, already looks better. I don't like the h1-style title either. I'm looking into some options, like giving that part of the site another background color and changing the title somewhat.

juliusv:
So yeah, a different background color seems to be a bad idea, given that the logos don't all have transparent backgrounds and everything below the jumbotron doesn't span the whole width of the page anyways.

juliusv:
How about something in this direction:

logos

Which I did by:

```html
  <div class="container text-center">
    <div class="row">
      <hr>
      <div class="col-md-12">Trusted by:</div>
    </div>
    <div class="row logos">
      <!-- ... -->
```

brian-brazil:
That looks better, updated.

juliusv:
Awesome. How about maybe swapping the bottom-right uda100 logo with the one next to it so it doesn't squeeze so hard against the (also tall) SoundCloud logo?

brian-brazil:
Moved it two over so we can claim it's still sorted somehow.

juliusv:
Great, :+1: from my side!

grobie:
LGTM

discordianfish:
I could reach out to Docker if they are OK with showing up. Not sure though since it might look like an official endorsement or something.

We can definitely add Jodel though :)

juliusv:
@discordianfish That would be great. Let's collect enough logos to fill another row :)

PR #304, opened 2016-01-26
Include Python pushgateway in Pushing docs

bluecmd:
Currently only shell, Java and Go are listed. Python supports pushgateways as well but requires users to go to the client library page to discover about it.


grobie:
I'm fine with making it a list, but let's remove the empty lines then maybe?

bluecmd:
I wouldn't argue if the the items weren't multi line paragraphs, but in this case it looks quite dense when I remove the newlines. It's up to you.

PR #308, opened 2016-01-27
Add best pratice around operation counters

discordianfish:
After asking the same questions a dozen times, I thought we should add it to the docs.

Not sure about words and everything but better than nothing. Let me know if I should rephase/move.

### DIFF
--- a/content/docs/practices/instrumentation.md
+++ b/content/docs/practices/instrumentation.md
@@ -32,6 +32,8 @@ this category.
 The key metrics in such a system are the number of performed queries, errors,
 and latency. The number of in-progress requests can also be useful.
 
+For counting failed queries, see section [Failures](#failures) below.
+
 Online-serving systems should be monitored on both the client and server side.
 If the two sides see different behaviors, that is very useful information for debugging.
 If a service has many clients, it is also not practical for the service to track them


fabxc:
:+1:

content/docs/practices/instrumentation.md
```diff
@@ -32,6 +32,9 @@ this category.
 The key metrics in such a system are the number of performed queries, errors,
 and latency. The number of in-progress requests can also be useful.
 
+When exposing operation counters, create seperate metrics for total and failed number of
+requests. That makes calculating percental error rates easier.
```
brian-brazil: This doesn't really belong here, and is already present further down the page.

Ratios are also preferred to percentages.

discordianfish: Now I found it. Is Failures as Subsystem? Maybe that can be improved.
And maybe it should be made more explicit. As I said, I still get confused why not use labels but separate metrics and the docs don't really clarify that (to be fair, neither does my change.)

beorn7: I suggest to add the following at the end of the section (after the paragraph about counting queries):

For counting failed queries, see section [Failures](#failures) below.

content/docs/practices/instrumentation.md
```diff
@@ -32,6 +32,9 @@ this category.
 The key metrics in such a system are the number of performed queries, errors,
 and latency. The number of in-progress requests can also be useful.
 
+When exposing operation counters, create seperate metrics for total and failed number of
```
juliusv: s/seperate/separate/

discordianfish:
Ok, so I removed the section again and just pointed to the bits in 'Failures'. That should be enough I guess.

beorn7:
:+1:

PR #310, opened 2016-01-27
Rework storage documentation and add "rushed mode"

beorn7:
Since this also includes the information of suspended ingestion upon
too many chunks in memory, it can only be merged once this han been
implemented (I'm working on it next).

@fabxc

### DIFF
--- a/content/docs/operating/storage.md
+++ b/content/docs/operating/storage.md
@@ -150,6 +150,22 @@ rushed mode, the following strategies are applied to speed up persisting chunks:
 
 Prometheus leaves rushed mode once the score has dropped below 0.7.
 
+## Settings for very long retention time
+
+If you have set a very long retention time via the `storage.local.retention`
+flag (more than a month), you might want to increase the flag value
+`storage.local.series-file-shrink-ratio`.
+
+Whenever Prometheus needs to cut off some chunks from the beginning of a series
+file, it will simply rewrite the whole file. (Some file systems support “head
+truncation”, which Prometheus currently does not use for several reasons.) To
+not rewrite a very large series file to get rid of very few chunks, the rewrite
+only happens if at least 10% of the chunks in the series file are removed. This
+value can be changed via the mentioned `storage.local.series-file-shrink-ratio`
+flag. If you have a lot of disk space but want to minimize rewrites (at the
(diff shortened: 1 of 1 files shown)


beorn7:
@fabxc This could be merged now. Do you have comments?

fabxc:
You know that I'm not a fan of overly exhausting documentation – especially on such details.
I simply believe that it will keep many people from reading it in the first place, thus increasing the number of times it gets asked in IRC or the mailing list again and again.

But yeah on the docs :+1: 
We need a checklist to quickly figure out appropriate flags.

(I'm sure that it's totally important to understand what flags you are setting and why etc. – but that doesn't reflect usage reality.)

beorn7:
In my experience, those flags are only needed for "power users". Some of them do ask those questions or irc or by mail, but pointing them to this section usually helps, so it _has_ saved us the work of repeating the same thing over and over.

At some point, we want a "memory autopilot" (see https://github.com/prometheus/prometheus/issues/455 ), which would avoid users to run into this kind of issues in the first place.

fabxc:
Yes, the docs should be there as they are. Just wondering whether its possible to provide a TL;DR of it.

beorn7:
My guess it won't be helpful in most cases. Once you are in tweaking land, you need more context than just ruled of thumb. It may be we gain more confidence over time, but then we can implement them right into the "memory autopilot".

juliusv:
The default setting for `-storage.local.memory-chunks` actually hits non-power users quite often, if they try out Prometheus on a too small server and it OOMs, no matter how few time series they are ingesting.

beorn7:
But by reducing it, we would hit many other non-power users that have a very modest but not low number of time series.

The default settings are good for ~4GiB of RAM, i.e. it will run nicely on most servers and personal computers. I see it as the sweet spot.

A memory "auto pilot" will deal with all of that.

juliusv:
I just wanted to argue that this information is also relevant for the opposite of power users - thus making a TL;DR useful for a wider range of people.

And as I understand it, the memory "auto pilot" is nothing that is concretely happening, because nobody knows how to do it right.

brian-brazil:
How about we add a warning at startup if the machine has less than -storage.local.memory-chunks * 2KB? That should handle the small cases.

PR #311, opened 2016-01-27
exporters.md: Add knxd_exporter

RichiH:


RichiH:
Now with brains!

brian-brazil:
Thanks, this would probably fit better as a http server.

RichiH:
On Wed, Jan 27, 2016 at 10:45 PM, Brian Brazil <notifications@github.com> wrote:

> Thanks, this would probably fit better as a http server.

I know; this is shoddy, but it works. I plan to improve it over time
and might get around to implementing basic HTTP for this and other
exporters.


Richard

grobie:
Looks actually like a great candidate for the textfile collector of the
node_exporter.

brian-brazil:
Knxd can also work over IP (not just USB), so that implies it should to be an exporter.

brian-brazil:
http://www.robustperception.io/writing-a-jenkins-exporter-in-python/ should be a good starting point.

PR #312, opened 2016-01-29
Update sustainable real-world use case record

fabxc:
Less time series, less targets but a higher scrape frequency.

@beorn7 @juliusv


beorn7:
@fabxc You beat me by a few seconds. Still I like my version better without the "~". https://github.com/prometheus/docs/pull/313

beorn7:
Since the differences are cosmetic, I just merge mine and close this one.

PR #317, opened 2016-02-04
exporters.md: Add speedtest_exporter

RichiH:


RichiH:
Yes, this exporter is as shoddy as the first one. I am working on making a nice version in Mojolicious and will then switch everything over.

content/docs/instrumenting/exporters.md
```diff
@@ -65,6 +65,7 @@ hosted outside of the Prometheus GitHub organization.
    * [Rsyslog exporter](https://github.com/digitalocean/rsyslog_exporter)
    * [scollector exporter](https://github.com/tgulacsi/prometheus_scollector)
    * [SMTP/Maildir MDA blackbox prober](https://github.com/cherti/mailexporter)
+   * [speedtest exporter](https://github.com/RichiH/speedtest_exporter)
```
juliusv: Seems their capitalization is "Speedtest" on http://www.speedtest.net/about, so let's capitalize the S?

RichiH:
Does it break convention if I use the full name in the link and the shortname in the exporter name?

juliusv:
Nope, that's fine! :+1:

PR #322, opened 2016-02-08
exporters.md: Add openweathermap_exporter

RichiH:


fabxc:
The exporter exposes all TYPE and HELP lines first, followed by the metrics.
The text protocol format requires all lines for a metric to be groupted. So the TYPE and HELP lines must directly followed by the metric lines.

http://prometheus.io/docs/instrumenting/exposition_formats/#text-format-details

RichiH:
Meh, sorry; will fix.

Arguably, the pushgateway should reject that then. It does reject duplicate HELP or TYPE lines.

fabxc:
I would argue differently.

The producer of protocol output should do its very best to conform to the specification. The receiver should do its best to read that output even if it does not conform 100% as long as it is unambiguous and not prone to break with further protocol extensions.

That said, the output it produces is actually valid. A metric started by TYPE and HELP lines doesn't have to followed by actual metrics (because with dynamic label values you cannot know before one was instantiated). Which makes the first block valid.
The second block is valid because TYPE and HELP are optional.

Either way :+1: on listing it :)

RichiH:
So should I rewrite or leave as is? Both is fine.

Main motivation to not sort around too much initially was because I wanted to stay in linear time, but I realized I can still do that if I introduce one extra variable per metric.

fabxc:
For the top TYPE and HELP things to be properly associated with the metrics below they need to be grouped together. It's not invalid, but not what's intended either basically.

I fully understand that it seemed simpler and more readable to do it like this.
Guess we need a pearl client library ;)

juliusv:
On a side note, this is Postel's "Robustness Principle": https://en.wikipedia.org/wiki/Robustness_principle - which was accepted as dogma for a long time, but recently has come under criticism for causing interoperability and security problems: http://queue.acm.org/detail.cfm?id=1999945

So it's not that black and white anymore.

PR #323, opened 2016-02-09
Host DreamHack blog post images from repo.

juliusv:


juliusv:
(in case imgur ever deletes old images)

PR #324, opened 2016-02-11
Add video from my FOSDEM talk to the media page

brian-brazil:


juliusv:
Hmm, two totally different talks, but same title. This could be confusing to people? Not sure there's much that can be done about that though.

Otherwise :+1:

brian-brazil:
Well, we could remove yours ;)

juliusv:
No way :P

PR #327, opened 2016-02-19
Improve alerting exmaple based on severity

grobie:
It should be alertmanager's responsibility how to route alerts and which
receiver to use. Alerts should only provide all necessary information,
but not describe the receiver of an alert. As this might not be true in
all cases, it can quickly create confusion.

@fabxc @juliusv

### DIFF
--- a/content/docs/alerting/rules.md
+++ b/content/docs/alerting/rules.md
@@ -74,7 +74,7 @@ Examples:
     ALERT InstanceDown
       IF up == 0
       FOR 5m
-      LABELS { severity = "page" }
+      LABELS { severity = "critical" }
       ANNOTATIONS {
         summary = "Instance {{ $labels.instance }} down",
         description = "{{ $labels.instance }} of job {{ $labels.job }} has been down for more than 5 minutes.",


juliusv:
:+1:

brian-brazil:
Alert routing is an interplay between prometheus and alertmanager. I find it pretty useful to look at an alert in prometheus and know if it pages or not. "Critical" is more a value judgement, and may vary by team. E.g. a critical alert that must be resolved in the next year. Criticiality is not the same as urgency.

grobie:
Exactly because you can't look at alerts in prometheus and know how they get routed, alert definitions should not give that false impression. "My alert definition says page, but I was not paged."

fabxc:
That will always be a very opinionated thing.

We do crit/warn rather than explicitness in the label – some of our warnings have a fanout and that's slightly adjustable per team for secondary channels such as Slack. So one generalized term for the notification type isn't feasible.

Also, what @grobie said (I'm aware the counter argument is "then your setup is bad, and you should feel bad").

brian-brazil:
That means though that I have to dig through the AM config to figure out what the alert does on every single occasion, which is extra work. I don't want to have to memorise how tens to hundreds of teams have setup their alerting labels to figure out which do/don't page - I want to look at an alert and tell if it pages.

> "then your setup is bad, and you should feel bad"

Exactly :)

fabxc:
People new to Prometheus struggle a lot with the concept of routing being based on arbitrary labels. Having "page" in the example gives the impression that Alertmanager has awareness of what a page is compared to an email etc.

Let's not dig to deep into how the perfect setup should look like but think about clarity for this particular piece of beginner documentation.

For that I think "critical" is more abstract and brings the concept across a bit better. Thus :+1:

brian-brazil:
> Let's not dig to deep into how the perfect setup should look like but think about clarity for this particular piece of beginner documentation.

As with all examples we should be pushing beginners towards best practice, unless there's a big gap in clarity if we don't. I don't see a big gap here.

> For that I think "critical" is more abstract and brings the concept across a bit better. 

I disagree, in alerting terms "critical" has essentially no semantic meaning due to drift. When it comes to alerting theory, a key aspect of an alert is whether it pages or not and thus that should be clear from the alert definition.

matthiasr:
> I have to dig through the AM config to figure out what the alert does on every single occasion

You need to do this whatever the text in the label. That's okay, because when I look at an alert I don't need to know the phone number that is going to be called eventually – I need to know which _class_ this alert belongs in.

> alerting terms "critical" has essentially no semantic meaning
> "critical" is more abstract and brings the concept across a bit better

Without the underlying [theory](https://docs.google.com/document/d/199PqyG3UsyXlwieHaqbGiWVa8eMWi8zzAn0YfcApr8Q/edit), neither "page" nor "critical" mean _anything_. FWIW, what we call "critical" in SoundCloud lingo is pretty much equivalent to what is called "page" in Google lingo.

I don't think changing this to "critical" will make anything clearer to the uninitiated. But neither does keeping it as "page".

> we should be pushing beginners towards best practice

This doesn't work. Write down what the best practice _is_, and how the example relates to it. Which word to use to represent the concepts is really secondary if and only if the concepts are explained. Conversely, if we don't explain the concepts, no choice of labels can fix that.

brian-brazil:
> You need to do this whatever the text in the label. 

With good naming I don't need to automatically check the alertmanager config.

> That's okay, because when I look at an alert I don't need to know the phone number that is going to be called eventually – I need to know which class this alert belongs in.

The class of the alert and the final destination are two different things.

> WIW, what we call "critical" in SoundCloud lingo is pretty much equivalent to what is called "page" in Google lingo.

Your particular definition only works if you have a consistent "critical alerts <=> paging alert" definition across your entire organisation (which is quite a feat if you've managed it). This is not at all a safe assumption we can make. These are the Prometheus docs, not the Soundcloud docs.

I've seen "critical" have diverse meanings across monitoring systems and organisations. Alerts are often called "critical", however that's not always the way that the users view it. Accordingly I want to avoid that particular word, and "warning" for the same reason.

> I don't think changing this to "critical" will make anything clearer to the uninitiated. But neither does keeping it as "page".

I think "critical" makes it less clear for everyone. Page is a reasonably well understand term, and not just in technology operations. Using "page" makes the intent obvious, with "critical" the intent isn't obvious. 

I've seen alerts classified "page" across hundreds of teams with a pretty consistent meaning, alerts classified "critical" were always more muddled. I take from this that users think in terms of outcomes (page, email, ticket, bug) and any other approach just gets in the way.

> This doesn't work. Write down what the best practice is, and how the example relates to it. Which word to use to represent the concepts is really secondary if and only if the concepts are explained. 

It does work, it's a standard way to do documentation, and is what we currently do throughout our docs. Where there's a small aspect like this that's not directly relevant to what's being explained (here we merely need to show some label to demonstrate the syntax), we use something that's realistic and follows best practice, and don't dwell on it.

This ensures consistency across the docs, and that when users copy&paste they don't end up going down the wrong path. Explaining every aspect of every example as it comes up would make the docs unwieldy and unusable. Reference docs are for reference.

PR #328, opened 2016-02-20
Document without aggregation modifier

brian-brazil:
Without should be used by ~everyone as it doesn't drop job-level labels (e.g. env, region or datacenter) when aggregating so list it first.

### DIFF
--- a/content/docs/practices/rules.md
+++ b/content/docs/practices/rules.md
@@ -17,8 +17,8 @@ convention.
 Recording rules should be of the general form `level:metric:operations`.
 `level` represents the aggregation level and labels of the rule output.
 `metric` is the metric name and should be unchanged other than stripping
-`_total` off counters when using `rate()`. `operations` is a list of operations
-that were applied to the metric, newest operation first. 
+`_total` off counters when using `rate()` or `irate()`. `operations` is a list
+of operations that were applied to the metric, newest operation first.
 
 Keeping the metric name unchanged makes it easy to know what a metric is and
 easy to find in the codebase. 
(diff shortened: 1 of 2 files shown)


content/docs/querying/operators.md
```diff
@@ -175,15 +175,16 @@ vector of fewer elements with aggregated values:
...
 
-    <aggr-op>(<vector expression>) [by (<label list>)] [keep_common]
+    <aggr-op>(<vector expression>) [without|by (<label list>)] [keep_common]
 
-By default, labels that are not listed in the `by` clause will be dropped from
-the result vector, even if their label values are identical between all
-elements of the vector. The `keep_common` clause allows to keep those extra
-labels (labels that are identical between elements, but not in the `by`
-clause).
+`without` removes the listed labels from the result vector, all other
```
juliusv: s/all/while all/ (run-on sentence otherwise)

content/docs/querying/operators.md
```diff
@@ -195,7 +196,7 @@ If the metric `http_requests_total` had time series that fan out by
 `application`, `instance`, and `group` labels, we could calculate the total
 number of seen HTTP requests per application and group over all instances via:
 
-    sum(http_requests_total) by (application, group)
+    sum(http_requests_total) without (instance)
```
juliusv: How about adding that this is equivalent to `sum(http_requests_total) by (application, group)`?

Actually, would you say `without` is always preferable? I'm just thinking of situations where one wants to keep output labels constant even when additional labels are added to a metric. And also, `by` lists more explicitly what labels are actually still there after the aggregation. With `without`, I have to know what labels exist in the first place to know what labels an expression outputs when looking at it.

content/docs/querying/operators.md
brian-brazil: Done

content/docs/querying/operators.md
brian-brazil: For the vast majority of usage I think `without` is preferable in practice. The sort of labels you're not listing are the ones you usually don't care about in rules as they're not being matched on anything (they're more useful up in the AM and at high level aggregations).

fabxc:
Recording rule naming is a bit vague as it is. For `without` expressions it will now be even more non-obvious. We should add a paragraph/example there too.

content/docs/querying/operators.md
```diff
@@ -175,15 +175,16 @@ vector of fewer elements with aggregated values:
...
-    <aggr-op>(<vector expression>) [by (<label list>)] [keep_common]
+    <aggr-op>(<vector expression>) [without|by (<label list>)] [keep_common]
 
-By default, labels that are not listed in the `by` clause will be dropped from
-the result vector, even if their label values are identical between all
-elements of the vector. The `keep_common` clause allows to keep those extra
-labels (labels that are identical between elements, but not in the `by`
-clause).
+`without` removes the listed labels from the result vector, while all other
+labels remain the same. By default, labels that are not listed in the `by`
```
juliusv: It could be good to make the difference between `without` and `by` clearer here.

How about:

`without` removes the listed labels from the result vector by aggregating over them, while all other labels are preserved in the output. `by` does the opposite and drops labels that are not listed in the `by` clause, even if their label values are the identical between all elements of the result vector. [...]

content/docs/querying/operators.md
juliusv: Ok, I'll defer to your judgement here, though I'm not really sure about this yet from my own experience.

brian-brazil:
> Recording rule naming is a bit vague as it is. For without expressions it will now be even more non-obvious. We should add a paragraph/example there too.

That belongs over in the best practices.

content/docs/querying/operators.md
brian-brazil: `by` works okay when you know a-priori what all of your labels are, and don't need to work with any other label taxonomy. `without` makes things a bit more nuanced in terms of naming (you'll need to distinguish between target and instrumentation labels), but allows much easier sharing of rules and is generally less error prone.

fabxc: I'm aware of the benefits of `without` but also still have my doubts about the "always" part.
It just needs one inconsiderate change to a metric's label set to break the semantics of a lot of recording rules.

fabxc:
> That belongs over in the best practices.

Yes, that's what I was going for. Let's just do it right away with this PR.

content/docs/querying/operators.md
brian-brazil: Done

content/docs/querying/operators.md
brian-brazil: Yeah, that's a risk. Thankfully labels change pretty rarely (I think we've had one in the node exporter, and I'm expecting maybe one more) whereas every user has a different label taxonomy.

content/docs/querying/operators.md
```diff
@@ -175,15 +175,16 @@ vector of fewer elements with aggregated values:
...
+    <aggr-op>(<vector expression>) [without|by (<label list>)] [keep_common]
 
-By default, labels that are not listed in the `by` clause will be dropped from
-the result vector, even if their label values are identical between all
-elements of the vector. The `keep_common` clause allows to keep those extra
-labels (labels that are identical between elements, but not in the `by`
-clause).
+`without` removes the listed labels from the result vector, while all other
+labels are preserved the output. `by` does the opposite and drops labels that
+are not listed in the by clause, even if their label values are identical
```
juliusv: Backticks around `by`.

brian-brazil: Done

brian-brazil:
That should do the trick. I needed to make the example more sophisticated to fully emphasise it, and to clarify some of what was happening before.

content/docs/practices/rules.md
```diff
@@ -41,69 +41,78 @@ Instead keep the metric name without the `_count` or `_sum` suffix and replace
...
-This is to prevent recording rules without a `job` label, which may cause
-conflicts across different jobs.
+Always specify a `without` clause with the labels you are aggregating away.
+This is to preserve all the other labels such as `job`, which will avoid
+conflicts and give you more useful metrics and alerts.
 
 ## Examples
 
-Aggregating up requests per second:
+Aggregating up requests per second that has a 'path' label:
```
juliusv: that has a path label -> that have a `path` label (plural and backticks)

brian-brazil: requests per second is a singular metric, so "has" applies. Added the backticks.

juliusv:
:+1:

fabxc:
:+1:

PR #329, opened 2016-02-23
Update anomaly blog post for new AM and alert syntax

brian-brazil:


content/blog/2015-06-18-practical-anomaly-detection.md
```diff
@@ -83,8 +83,6 @@ ALERT InstanceLatencyOutlier
       >
         1
   FOR 30m
-  SUMMARY "{{$labels.instance}} in {{$labels.job}} is a latency outlier"
```
juliusv: Don't wanna keep at least one of these in the new format? To showcase Prometheus's alert message templating on the side a bit?

brian-brazil: They were only here previously because they were mandatory. This example is complicated enough without adding that too, and it's going to an automated system anyway.

juliusv:
Na gut :+1:

PR #332, opened 2016-02-24
Add Varnish exporter to 3rd party section

jonnenauha:
Created this over the weekend. Might be useful as Varnish is quite common part of web infra. I have tested it to work and to actually be useful with our company production instances.

P.S. Yey for Varnish turning 10 years ;) Maybe this is my way of celebrating
http://info.varnish-software.com/varnish-cache-10-years


juliusv:
:+1: Thanks!

PR #333, opened 2016-02-25
Remove misleading hints about multiple evaluation_interval

grobie:
There is only one global evaluation_interval at this time, so a "by
default" is misleading as it can't be overwritten anywhere.

@brian-brazil

### DIFF
--- a/content/docs/introduction/getting_started.md
+++ b/content/docs/introduction/getting_started.md
@@ -37,8 +37,7 @@ Prometheus configuration as a file named `prometheus.yml`:
 ```
 global:
   scrape_interval:     15s # By default, scrape targets every 15 seconds.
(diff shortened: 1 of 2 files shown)


content/docs/introduction/getting_started.md
```diff
@@ -37,8 +37,7 @@ Prometheus configuration as a file named `prometheus.yml`:
 ```
 global:
   scrape_interval:     15s # By default, scrape targets every 15 seconds.
-  evaluation_interval: 15s # By default, scrape targets every 15 seconds.
-  # scrape_timeout is set to the global default (10s).
```
grobie: As this is a getting started document, I found this comment not helpful. The a meaningful timeout is already chosen, no need to confuse new starters about it.

matthiasr:
:+1:

brian-brazil:
:+1:

PR #334, opened 2016-02-26
Clarified target_label behavior in relabel_config

johnw188:
I wasn't aware that relabel_config could be used to add a new label to a target. Updated documentation to clarify this fact.

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -579,7 +579,8 @@ the `replace`, `keep`, `drop` and `labelmap` actions. The regex is fully anchore
 
 * `replace`: Match `regex` against the concatenated `source_labels`. Then, set
   `target_label` to `replacement`, with match group references
-  (`${1}`, `${2}`, ...) in `replacement` substituted by their value.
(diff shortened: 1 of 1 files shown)


matthiasr:
Better to be explicit :+1:

content/docs/operating/configuration.md
```diff
@@ -579,7 +579,8 @@ the `replace`, `keep`, `drop` and `labelmap` actions. The regex is fully anchore
 
 * `replace`: Match `regex` against the concatenated `source_labels`. Then, set
   `target_label` to `replacement`, with match group references
-  (`${1}`, `${2}`, ...) in `replacement` substituted by their value.
+  (`${1}`, `${2}`, ...) in `replacement` substituted by their value. If `target_label`
+  exists its value will be updated, if not a new label will be added with the specified value.
```
brian-brazil: If we're mentioning that we should also mention that an empty result will remove a label.

juliusv:
@johnw188 Heya, could you address the outstanding comment and rebase on current master?

brian-brazil:
Stale PR, cleaning up.

PR #335, opened 2016-02-26
Remove wrong scrape_timeout from getting started examples

grobie:
Scrape timeout can't be higher than scrape interval, so this config is
just wrong. This change remove all mentions of the timeout, we chose a
reasonable default, if a user has different needs they can find it in
the configuration docs.


grobie:
One second, I'll also fix evaluation interval comments.

juliusv:
Oh damn, that was broken for a long time, eh?

:+1:

grobie:
Please have another look. I removed `evaluation_interval` from the basic example without any rules. I fixed the comment in the example which has rules.

juliusv:
:+1:

PR #349, opened 2016-03-16
Add initial PromCon 2016 site.

juliusv:
@fabxc @beorn7 @brian-brazil @grobie 

This is ready for a look, but please don't merge yet without further coordination.

### DIFF
--- a/content/promcon/2016/index.html
+++ b/content/promcon/2016/index.html
@@ -1,0 +1,111 @@
+---
+layout: plain
+---
+<div class="jumbotron">
+  <div class="container">
+    <h1><img src="/assets/prometheus_logo.svg" alt="Prometheus logo"> PromCon 2016</h1>
+    <p class="subtitle">The Prometheus conference — August 25 - 26 in Berlin</p>
+    <p>
+      <a class="btn btn-default btn-lg" href="https://docs.google.com/forms/d/1b7lq7-h8f0JRPThl6oc3Q70YsbzR__FcGNKbCvdetow/viewform" target="_blank" role="button">Submit Talk</a>
+      <a class="btn btn-primary btn-lg" href="https://docs.google.com/forms/d/17KakZ4dE_8d1D6lyGA9SIE15vAm9NIiOPhlDiVkM9ak/viewform" target="_blank" role="button">Become a Sponsor</a>
+      <a class="btn btn-primary btn-lg" href="https://docs.google.com/forms/d/12H6o3vvnnQUJSDV1mQ77EqpU9B64Jk70iFboHwb2y3M/viewform" target="_blank" role="button">Register</a>
+    </p>
+  </div>
+</div>
+
+<div class="container">
+  <div class="row">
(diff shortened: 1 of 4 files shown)


juliusv:
50% zoom screenshot of entire page:

promcon

juliusv:
@RichiH in case you're interested as well.

content/promcon/2016/index.html
```diff
@@ -0,0 +1,110 @@
...
+        <h2>Register to attend</h2>
+        <p>
+          Registration will open soon — fill out the form linked below to get
+          notified when the registration opens.
+        </p>
+        <p>
+          <a class="btn btn-primary btn-lg" href="https://docs.google.com/forms/d/12H6o3vvnnQUJSDV1mQ77EqpU9B64Jk70iFboHwb2y3M/viewform" target="_blank" role="button">Notify Me</a>
+        </p>
+        <p>
+          Admission will likely cost around $50, but we will provide exceptions
```
brian-brazil: We'll probably need this higher, it's better to come down in price than go up so I'd put it nearer €100.

beorn7:
This ignores the existence of PromCon 1. *sniff*

content/promcon/2016/index.html
juliusv: I changed it to "around $50 - $100"

juliusv:
@beorn7 :'( I know, but mentioning it here would only serve to confuse people.

brian-brazil:
I'd put it in as the second.

juliusv:
@brian-brazil The second conference? PromCon 1 was pretty unofficial and more of a tiny developer meetup. I don't think it makes sense to put it in the same category of event...

RichiH:
"PromCon16 is happening" can stand without mentioning if it's the first; that would simply leave the question open. If anyone asks, it's always possible to say "it's the second one, but the first one for the public".

juliusv:
I think "first" makes sense in this context, and hopefully helps to generate excitement to join. Anyways, we can always keep it as an inside joke that we were at the "real" first PromCon.

RichiH:
> Anyways, we can always keep it as an inside joke that we were at the "real" first PromCon.

_You_ can, _we_ can't *sniff*

grobie:
What about using promcon.io? Might help in twitter etc. where only the first part of a URL is visible.

juliusv:
@grobie Yeah, I was considering promcon.io or promcon.org as well - though, it would make sense to have a whole separate website then? I could still copy the design of prometheus.io, but without all the other menu items? That would be the fastest way to get there, at least, since we really should start the CFP soon...

Also, what do people think about the suggested closing date for the CFP (April 30). Too short, or ok?

RichiH:
It leaves almost 1.5 months, which sounds fine given the rather limited scope.

juliusv:
Cool, my thinking as well.

juliusv:
Closing in favor of separate site at promcon.io.

PR #353, opened 2016-03-20
Clarify requirements for histograms and summaries text exposition format

knyar:
It seems that [text_parse.go](https://github.com/prometheus/common/blob/master/expfmt/text_parse.go) *requires* histograms and summaries to have type declaration, even though `HELP` and `TYPE` lines are described as optional in the document.

### DIFF
--- a/content/docs/instrumenting/exposition_formats.md
+++ b/content/docs/instrumenting/exposition_formats.md
@@ -109,6 +109,7 @@ format. The following conventions apply:
 * Each bucket count of a histogram named `x` is given as a separate sample line with the name `x_bucket` and a label `{le="y"}` (where `y` is the upper bound of the bucket).
 * A histogram _must_ have a bucket with `{le="+Inf"}`. Its value _must_ be identical to the value of `x_count`.
 * The buckets of a histogram and the quantiles of a summary must appear in increasing numerical order of their label values (for the `le` or the `quantile` label, respectively).
+* Histogram and summary metrics _must_ have a comment defining their type (`# TYPE` line before the samples of a given metric).
 
 See also the example below.
 


brian-brazil:
This isn't required, they should end up as UNTYPED. Is something else happening?

knyar:
I believe values of the `le` label do not get correctly parsed if a metric is not declared as a histogram.

brian-brazil:
That doesn't sound right at all. Do you have an example export that's being incorrectly parsed?

beorn7:
@knyar You are right in a way. The text format parser will only recognize a proper histogram if it is marked as such. It doesn't just "guess" that a bunch of time series look like belonging to a histogram and bundles them.

However, the Prometheus server itself breaks a histogram in individual bucket time series, and they look exactly as the time series you get if you fail to declare your histogram as such in the text format. So right now, on the Prometheus server, it doesn't really matter.

The text format parser is not only for the Prometheus server, though. So the Pushgateway, for example, displays a proper histogram on its status page, but only if you declare it as such. Otherwise, you'll see all the individual `UNTYPED` time series.

In the not too far future, the Prometheus server and PromQL will be "type aware", which will for example simplify the syntax of quantile calculation and allow for warnings about nonsensical expressions. That will then be another reason to properly declare the metric types (but that's true for all types, not just histograms).

knyar:
Thanks a lot for clarifying, @beorn7. The typing system does not seem to get surfaced in any way currently, so it seems that I got a bit confused by the fact that if a type is provided, prometheus server re-writes the `le` labels - i.e. when I am exposing a set if metrics like this:
```
nginx_http_response_size_bytes_bucket{host="h1",le="0000010.0"} 1
nginx_http_response_size_bytes_bucket{host="h1",le="1000000.0"} 2
```
they get parsed and stored the following way if `TYPE` is provided:
```
nginx_http_response_size_bytes_bucket{host="host1",le="10"}  1
nginx_http_response_size_bytes_bucket{host="host1",le="1e+06"} 2
```
If not, they are just stored verbatim.

This does not actually seem to have any impact on `histogram_quantile` results.
I should have probably spent more time reading the code before opening this PR. Sorry for the fuss.

PR #360, opened 2016-03-28
Add a link to nginx-lua-prometheus

knyar:
While nginx-lua-prometheus is not an exporter, it feels like this section is the best fit for it.

### DIFF
--- a/content/docs/instrumenting/clientlibs.md
+++ b/content/docs/instrumenting/clientlibs.md
@@ -22,8 +22,9 @@ Unofficial third-party client libraries:
 
 * [Bash](https://github.com/aecolley/client_bash)
 * [Haskell](https://github.com/fimad/prometheus-haskell)
-* [Node.js](https://github.com/siimon/prom-client)
+* [Lua](https://github.com/knyar/nginx-lua-prometheus) for Nginx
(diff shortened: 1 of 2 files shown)


brian-brazil:
You should probably list this over in Client Libraries too with the caveat that it's Nginx only

knyar:
Sure, I've added a link to `clientlibs.md` as well.

content/docs/instrumenting/clientlibs.md
```diff
@@ -24,6 +24,7 @@ Unofficial third-party client libraries:
 * [Haskell](https://github.com/fimad/prometheus-haskell)
 * [Node.js](https://github.com/siimon/prom-client)
 * [.NET / C#](https://github.com/andrasm/prometheus-net)
+* [Lua](https://github.com/knyar/nginx-lua-prometheus) for Nginx
```
brian-brazil: L sorts before N

knyar: Moved. Also swapped `Node.js` and `.NET / C#`, since otherwise it's not immediately obvious that the list should be sorted alphabetically.

brian-brazil:
Thanks!

PR #365, opened 2016-03-30
Collapse inactive navigation menus

grobie:
screenshot from 2016-03-29 21-33-55

### DIFF
--- a/static/docs.css
+++ b/static/docs.css
@@ -90,6 +90,7 @@ a.sc-logo img {
 }
 
 .side-nav .fa {
+  width: 20px;
   color: #555;
 }
 
@@ -102,14 +103,18 @@ a.sc-logo img {
 }
 
 .side-nav .nav-header {
-  color: #e6522c;
-  text-transform: uppercase;
-  font-size: 16px;
   display: block;
   margin: 20px auto 15px auto;
+  font-size: 16px;
 }
 
-.side-nav .active {
+.side-nav .nav-header a, .side-nav .nav-header span {
+  color: #e6522c;
+  text-transform: uppercase;
+  text-decoration: none;
+}
+
+.side-nav ul.active li.active {
   border-left: 3px solid #e6522c;
   margin-left: -2px;
   font-weight: bold;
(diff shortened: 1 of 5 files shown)


juliusv:
Now that the collapsed top-level items are directly under each other, it's visually more obvious that their texts are not properly left-aligned, because the various font-awesome icons have different widths. Could you add `width: 20px;` to https://github.com/prometheus/docs/blob/master/static/docs.css#L92-L94?

I guess I like it otherwise. It breaks my single-plage Ctrl-F-ability though, which I use a lot. But maybe that's worth it.

brian-brazil:
Being able to search within a page is important for documentation.

beorn7:
For me, the too-long side bar was always very annoying. So I like this change.
Perhaps add a site-map for menu findability?

juliusv:
What I like about this is that lower sections like "Best Practices" aren't so easily overlooked anymore. I've heard once or twice that people never noticed that section...

@beorn7 What exactly do you mean with a site-map?

beorn7:
"A site map (or sitemap) is a list of pages of a web site accessible to crawlers or users. It can be either a document in any form used as a planning tool for Web design, or a Web page that lists the pages on a Web site, typically organized in hierarchical fashion." (Wikipedia)

grobie:
* the current menu is a huge issue on mobile
* My current approach is a performance penalty for users on slow connections, just to see the items of another menu, they have to load a new page. I'll change the implementation to be javascript only (e.g. it's essentially a no-op for people who deactivate javascript).
* The behavior will become more smooth, my quickly being able to click around in all menu items until opening the page one is interested in.
* If people feel strongly someone can add a keyboard shortcut to expand all menus.

juliusv:
@beorn7 Ah, I thought you meant Ctrl-F-style searchability for users. Ok.

static/docs.css
```diff
@@ -102,14 +102,26 @@ a.sc-logo img {
...
+}
+
+.side-nav .nav-header a {
+  color: #e6522c;
+  text-transform: uppercase;
+  text-decoration: none;
+}
+
+.side-nav ul {
+	display: none;
```
juliusv: You're using tabs for indentation here, while the rest of the file uses two spaces. Same below.

juliusv:
:+1: IMO after CSS fixes.

PR #372, opened 2016-04-02
Improve search engine result presentation

grobie:
- best to keep title length under 55 characters
- descrioption should not be longer than 150-160 characters


brian-brazil:
:+1:

fabxc:
:+1:

Issue #369, opened 2016-04-02
Missing cache buster


grobie:
All assets should have a cache buster (using the current commit version), so that cloudflare won't deliver cached outdated asset versions.

juliusv:
I'm on this...

juliusv:
@ddfreyne While working on this, I ran into a nanoc problem: nanoc classifies all of our static CSS files (including https://github.com/prometheus/docs/blob/master/static/docs.css) as binary items, so trying to run a textual filter on them (to change any contained `url(...)` sections to cache-busted URLs) gives me an error:

```
Nanoc::Errors::CannotUseTextualFilter: The “Nanoc::Filters::CacheBuster” filter cannot be used to filter the “/assets/docs.css/” item (rep “default”), because textual filters cannot be used on binary items.
```

How does nanoc determine what is a binary item, and can I change it somehow?

juliusv:
Discovered the reason. We're using the static datasource (https://github.com/prometheus/docs/blob/master/nanoc.yaml#L64-L66) to serve our static assets (CSS, JS, fonts, and images), and that automatically sets all managed files to binary: https://github.com/nanoc/nanoc/blob/3.7.3/lib/nanoc/data_sources/static.rb#L51

@ddfreyne any advice? I guess we should just not use the static datasource for items we want to filter textually? Another hacky workaround would be to exempt the jumbotron background image file from getting a cachebusting fingerprint, so we wouldn't even have to rewrite its URL in the CSS.

PR #374, opened 2016-04-04
Add cachebusting fingerprints to asset files.

juliusv:
docs.css had to be moved out of /static because the static datasource
treats all files as binary, and thus no textual filters can be applied
(necessary for rewriting the jumbotron background image URL).

Fixes https://github.com/prometheus/docs/issues/369


juliusv:
@grobie

grobie:
:+1: if it works.

juliusv:
Well, it works locally - let's give this a try ;)

PR #375, opened 2016-04-04
Fix blog sidenav.

juliusv:
This got broken in https://github.com/prometheus/docs/pull/365

### DIFF
--- a/static/docs.css
+++ b/static/docs.css
@@ -108,7 +108,7 @@ a.sc-logo img {
   font-size: 16px;
 }
 
-.side-nav .nav-header a, .side-nav .nav-header span {
+.side-nav .nav-header a, .side-nav .nav-header {
   color: #e6522c;
   text-transform: uppercase;
   text-decoration: none;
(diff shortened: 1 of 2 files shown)


juliusv:
@grobie

static/docs.css
```diff
@@ -108,7 +108,7 @@ a.sc-logo img {
   font-size: 16px;
 }
 
-.side-nav .nav-header a, .side-nav .nav-header span {
```
grobie: The second span is used to replace the normal title with a link if javascript is enabled. Disable javascript, and this will probably break it.

grobie: I also played around with replacing the the full `nav-header` span with a link, but that created some artifacts during page load, where titles disappeared for a handful of milliseconds.

grobie:
Oops. It might make sense to rename `docs.css` as I wasn't aware it's used outside of the docs as well.

static/docs.css
juliusv: Yep, I was worried about that too and checked - all good even if I delete your JS. Everything just is expanded then, but looks correct.

grobie: Oh right, I see how it works. the inner `span` will inherit the settings from the outer `span.nav-header`.

grobie:
:+1:

PR #377, opened 2016-04-04
Clarifying regex anchoring docs

pete0emerson:
This pull request expands the documentation about regexes being anchored.

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -600,7 +600,9 @@ prefix is guaranteed to never be used by Prometheus itself.
 
 `<regex>` is any valid [RE2 regular
 expression](https://github.com/google/re2/wiki/Syntax). It is required for
-the `replace`, `keep`, `drop` and `labelmap` actions. The regex is fully anchored.
+the `replace`, `keep`, `drop` and `labelmap` actions. The regex is automatically
+anchored on both ends to simplify the majority of use cases. In order to effectively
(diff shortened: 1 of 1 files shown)


content/docs/operating/configuration.md
```diff
@@ -600,7 +600,9 @@ prefix is guaranteed to never be used by Prometheus itself.
 
 `<regex>` is any valid [RE2 regular
 expression](https://github.com/google/re2/wiki/Syntax). It is required for
-the `replace`, `keep`, `drop` and `labelmap` actions. The regex is fully anchored.
+the `replace`, `keep`, `drop` and `labelmap` actions. The regex is automatically
+anchored on both ends to simplify the majority of use cases. In order to effectively
+unanchor the regex, use `.*<regex>.*`.
```
beorn7: I would prefer to keep the documentation as concise and short as possible. "Automatically" can be left out without any loss of meaning (1st rule of style: omit needless words). We don't really need to justify the design choice in the documentation.

So how about:

The regex is anchored on both ends. To un-anchor the rexex, use `.*<regex>.*`.

juliusv:
@pete0emerson ping :)

grobie:
@pete0emerson ping

jkemp101:
Bump, this is a good change.  I wasted some time this morning with the current wording.  The `.*<regex>.*` example is important to add context.

beorn7:
I have performed the suggested change directly in master. Closing this.

PR #387, opened 2016-04-11
add routing tree editor

stuartnelson3:
@fabxc same treatment as before, but I'm a bit unsure how best to style this. The more space the tree has the better, as complicated trees can quickly get too compacted to be useful. I'll send pictures of how it looks at the moment. The tree is appended at the bottom in its own element so it has the full width of the window, but then it is pretty low.


fabxc:
:shipit: 👍 

That's the best way to figure things out.

If it's working all good you should write a small blog post :) – better visibility than mailing list.

juliusv:
I guess we shouldn't really have a menu item that takes you away to a completely separate page though. Maybe link it from one of the existing Alertmanager docs instead, like from http://prometheus.io/docs/alerting/configuration/?

grobie:
Hmm. I'd not expect this in the docs to be honest. It's also confusing that the side bar suddenly disappears when selecting it from the menu.

I'm fine with putting it somewhere in prometheus.io, but it's not really documentation. We can link to it from the alerting docs though.

juliusv:
@grobie Race condition.

grobie:
Auto-merge as we both said the same.

content/docs/alerting/routing-tree-editor.html
```diff
@@ -0,0 +1,71 @@
+---
+title: Routing tree editor
+sort_rank: 4
+nav_icon: sliders
+---
+<style>
```
grobie: This is a HTML violation, the style belongs to the head section.

juliusv: Even better, put it in a CSS file :)

juliusv: This whole page could use a custom nanoc layout (similar to the landing page etc.) that would include all the right headers etc. If we decide that it shouldn't be part of the docs itself, that would make even more sense...

fabxc:
Yea, we can change it – I just want to get a live impression.
It should be linked from somewhere visible. Navbar is currently the easiest way for that.

grobie:
The navigation behavior has to be consistent. Either bring back the menu on the editor page or remove it from the menu. I still don't understand why this is in the docs, it has nothing to do with documentation.

fabxc:
Because prometheus/docs *is* our website. As simple as that.

juliusv:
I think @grobie meant http://prometheus.io/docs, not the name of this repo? That's at least what I would mean, if I were him :)

grobie:
Yes. As I wrote above already, I'm with having it under https://prometheus.io, but it doesn't make sense to me to have in under /docs.

brian-brazil:
I agree, /webtools or something would make sense as I suspect there'll be a handful of these over time.

fabxc:
Ah, the URL – sure, agreeing with that.

stuartnelson3:
Ok, so to change:

- Move the CSS to its own file
- Move the page to its own endpoint (`/webtools/routing-tree-editor`?). I'm find nesting it under `webtools`, but I don't want to have a separate page yet that lists the webtools since there's only the one.

Questions:

- How do we link to it from the main page? Should we go with what Julius mentioned, and link to it from `/docs/alerting/configuration/`? I am a fan of doing both that and adding one additional menu option at the top, `Webtools`, that right now links directly to this routing tree editor. Thoughts?

brian-brazil:
It should be a link from the text on one or more of the alerting pages.

fabxc:
I don't think that's very good placement. Realistically, most users just skim through documentation.
Someone looking for a link to a tool will not look through a full-text page for it.

brian-brazil:
We can't list everything on in the menu just because it might be useful to some users. I don't think this meets the threshold for a menu entry, and having it beside the docs of what it's being used with should be sufficient for users to discover it.

fabxc:
The main entries are now collapsed. There's no concern of to many entries if it stays <10 in each main point.

This could fit in "Alerting" or "Visualization". No harm done.
If it's getting more we have a menu entry for "tools" and a page with direct links to those.

juliusv:
It should be somewhere under "Alerting", but I don't think the page itself should be directly linked from the side menu. If you click a side menu item, you don't expect the whole menu structure and everything to disappear on you. If you do want a separate menu item for it (instead of linking it from the AM configuration docs page), maybe make an interstitial docs page that keeps the page structure, and link it from there, with some explanation?

fabxc:
I heard `_blank` is still a thing.

grobie:
The documentation menu items should only be links to documentation pages. Even with a collapsed menu, having too many entries (>7 is too many, probably lower) makes it hard to consume the documentation. This tool is a nice gimmick, but not essential to the documentation.

juliusv:
I agree with @grobie here.

fabxc:
^ @stuartnelson3 3:1 vote is pretty clear

My general 2ct.
Too many items are not at all our problem but structure.
There's little distinction between configuring/operating and general concepts of the Prometheus eco-system.
There's a main point "Operating" which refers to configuring Prometheus. Alertmanager's "operating" is under "Alerting".

Also, still no documentation versioning.

stuartnelson3:
While I think calling this a gimmick might be being a little harsh on the tool (I do think it is useful), I do agree that finding the right place for this is tricky.

For now, I propose an "announced dark launch" (yeah, oxymoronic): let's do the minimal, easy solution that we know is right (providing a link in the AM configuration doc), announce it on the mailing list, and gather feedback. If feedback is positive and/or worthwhile additional functionality is requested, we can look at where best to make this accessible on the website. If it's met with disinterest, it'll still be available for those that want it.

As far as routing, how about nested in the `/alerting` path, but without having a menu item in the sidebar, and keep the current sidebar-less layout.

Let's do emoji :heart: for "yes, this is a good plan", or :tada: for "no, this needs further discussion".

juliusv:
<3 in general, but routing-wise, I would put the actual tool outside of `/docs` completely, namely under `/webtools/foobar` or something like that.

I don't feel strongly on whether to provision its own menu item for it (with interstitial docs page) or not, but the plan to add it to the AM config doc sounds ok for now.

fabxc:
👍 

Also on `/webtools/alertroutes` or similar.

PR #389, opened 2016-04-12
Update routing tree location

stuartnelson3:
Routing tree editor is now only available as a
link from within the configuration docs for
alertmanager.

Code changes:
- Routing tree CSS is now in a CSS file
- The label set input now draws the route when
"enter" is pressed

@juliusv @fabxc


layouts/header.html
```diff
@@ -40,6 +40,7 @@
 
     <!-- Custom styles for this template -->
     <link href="/css/docs.css" rel="stylesheet">
+    <link href="/css/routing-tree-editor.css" rel="stylesheet">
```
fabxc: This is now used in every page. Don't know whether there's a decent way to do it differently with nanoc, so fine with me. 
Anyone bothered by that?

content/docs/alerting/configuration.md
```diff
@@ -12,6 +12,9 @@ While the command-line flags configure immutable system parameters, the
 configuration file defines inhibition rules, notification routing and
 notification receivers.
 
+An visual routing tree editor and label set matcher is available
+[here](/webtools/alerting/routing-tree-editor).
```
fabxc: I'm generally not a fan of "here" links.

Maybe `The [visual editor]() can assist in building routing trees` or similar?

content/docs/alerting/configuration.md
```diff
@@ -12,6 +12,9 @@ While the command-line flags configure immutable system parameters, the
 configuration file defines inhibition rules, notification routing and
 notification receivers.
 
+An visual routing tree editor and label set matcher is available
```
juliusv: s/An/A/

layouts/header.html
juliusv: We could of course add an `<%= if ... %>` check here if we really wanted. Like we do in https://github.com/prometheus/docs/blob/routing-tree-deux/layouts/header.html#L56.

But currently in docs.css we also have some styles that are only applicable for certain pages, so we're loading too much CSS already. It's not nice and we should optimize that, but I think it's small enough to still be ok for now.

content/docs/alerting/configuration.md
stuartnelson3: :+1:, fixes both comments

layouts/header.html
stuartnelson3: Yeah, the total file custom css size is 2kb, so the concern is if these are overwriting any other styles. I'm pretty sure they aren't interfering with any other styles, though.

juliusv:
:+1:

stuartnelson3:
Will wait for agreement from @fabxc :)

fabxc:
+1

PR #390, opened 2016-04-12
Move GitHub star button to "Open Source" section.

juliusv:
As @grobie pointed out, there was some feedback that the "star button looks very needy and falls so much out of the "nice design"". I tend to agree. Also there was no noticeable uptake in GitHub stars since adding the button. Let's move it to the "Open Source" section at the bottom, where it will still give useful information to people (number of stars we already have), but less prominent.

### DIFF
--- a/content/css/docs.css
+++ b/content/css/docs.css
@@ -23,12 +23,6 @@ body {
   margin-left: 12px;
 }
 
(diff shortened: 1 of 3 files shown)


fabxc:
👍

PR #392, opened 2016-04-13
Remove third-party blackbox exporter

syscrusher:
Repository says it is deprecated and links to the official blackbox exporter.


grobie:
Thanks!

PR #396, opened 2016-04-14
@juliusv simple update to doc page for link to puppet module

abuxton:


content/docs/introduction/install.md
```diff
@@ -103,3 +103,7 @@ Chef:
 SaltStack:
 
 * [bechtoldt/saltstack-prometheus-formula](https://github.com/bechtoldt/saltstack-prometheus-formula)
+
+Puppet:
```
brian-brazil: P sorts before S, so this should go up one

content/docs/introduction/install.md
```diff
@@ -103,3 +103,7 @@ Chef:
...
+
+* [https://forge.puppet.com/brutus777/prometheus](https://forge.puppet.com/brutus777/prometheus)
```
brian-brazil: The others link to github repos, does that also make sense here?

grobie:
@abuxton Can you fix the comments?

brian-brazil:
This is stale, and we already have a puppet example listed.

PR #397, opened 2016-04-14
Word-wrap exporter and clientlib guideline docs.

juliusv:
This is to keep it consistent with the rest of our docs and to make
subsequent diffs less crazy.

### DIFF
--- a/content/docs/instrumenting/clientlib_guidelines.md
+++ b/content/docs/instrumenting/clientlib_guidelines.md
@@ -5,15 +5,23 @@ sort_rank: 2
 
 # Client Library Guidelines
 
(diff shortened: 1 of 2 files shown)


juliusv:
@grobie

grobie:
I actually started to switch support groups here and stopped wrapping docs. I don't see an improvement in diffs with wrapped lines, if at all the opposite actually, but I'm fine with both. Feel free to merge for consistency.

juliusv:
Yeah, the diff situation can go both ways, depending on the diff viewer and the nature of changes.

Besides consistency with the rest, I currently find wrapped lines easier to read and edit in an editor (and `gq` nicely rewraps a selection in vim).

PR #398, opened 2016-04-14
Remove title case in exporter and clientlib guidelines.

juliusv:
We have opted for not using title case in headers in Prometheus docs.


juliusv:
@grobie

grobie:
:+1:

PR #399, opened 2016-04-14
Improve landing page layout on smaller screens.

juliusv:
- The jumbotron text was not correctly centered between 750px and 970px
  width. Fixed by removing unnecessary HTML.
- Feature items flipped from 4 per row to 1 per row, without any 2-per row
  variant for medium-sized screens. Added a 2-per-row variant.
- The feature items had a fixed height which looked too tall on the
  large end of XS widths (one long item per row). Removed the fixed
  height and let the height adjust itself automatically.

### DIFF
--- a/content/css/docs.css
+++ b/content/css/docs.css
@@ -88,7 +88,6 @@ body {
   font-family: 'Lato', sans-serif;
   font-weight: 300;
   cursor: pointer;
-  height: 200px;
 }
 
 .feature-item:hover {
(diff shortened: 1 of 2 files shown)


juliusv:
@discordianfish @grobie

juliusv:
There's still one glitch when having a screen width around 1090px: the "Many client libraries" text overflows its column a bit and collides with the next one. But that's already the case on the current site. Just activating line breaks in feature item h2s isn't a good solution because that breaks the other items too early. Maybe I'll just activate line-breaking for the really long ones. I'll think about that.

PR #400, opened 2016-04-15
Add missing "s" to {metric_,}relabel_config

mfournier:
Also fix references to anchors, which were already broken.


content/docs/operating/configuration.md
```diff
@@ -546,7 +546,7 @@ Where `<filename_pattern>` may be a path ending in `.json`, `.yml` or `.yaml`. T
 may contain a single `*` that matches any character sequence, e.g. `my/path/tg_*.json`.
 
 
-### `<relabel_config>`
+### `<relabel_configs>`
```
brian-brazil: This is singular, as it's only about one `relabel_config` rather than a list of them.

mfournier:
@brian-brazil ok, right ! Doh :-)

I'll send another PR to fix the anchors. Thanks !

PR #410, opened 2016-04-16
Rename "X guidelines" docs to "Writing X"

juliusv:
The meanings of "Client library guidelines" and "Exporter guidelines" are not
immediately clear. For example, it could mean the usage of client
libraries or exporters. I think "Writing client libraries" and "Writing
exporters" makes this clearer.

Alternative to "Writing": "Building"?


fabxc:
👍

PR #411, opened 2016-04-16
Simplify title of "Exporters and third-party integrations"

juliusv:
"Exporters and integrations" is easier to read in the menu, doesn't wrap
lines, and hopefully means the same thing to people.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -1,9 +1,9 @@
 ---
-title: Exporters and third-party integrations
+title: Exporters and integrations
 sort_rank: 4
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/exporters.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Exporters and third-party integrations
+title: Exporters and integrations
 sort_rank: 4
 ---
```
brian-brazil: Change the headline on the page too?

juliusv: Doh, of course. Done.

PR #412, opened 2016-04-17
Explain when (not) to use the Pushgateway and why

juliusv:


juliusv:
@brian-brazil

content/docs/instrumenting/pushing.md
```diff
@@ -26,6 +26,56 @@ class.
...
+of the Pushgateway as a metrics cache is fundamentally separate from the
+lifecycle of the processes that push metrics to it. Contrast this to
+Prometheus's usual pull-style monitoring: when an instance disappears
+(intentional or not), its metrics will automatically disappear along with it.
+When using the Pushgateway, this is not the case, and you would now have to
+delete any stale metrics manually or automate this lifecycle synchronization
+yourself.
+
+**Usually, the only valid use case for the Pushgateway is for monitoring
+service-level batch jobs which do not run long enough to be scraped reliably.**
```
brian-brazil: It's more about capturing state at the very end of the batch job. Scraping them during their life to monitor performance (presuming they run for a while) is recommended elsewhere.

content/docs/instrumenting/pushing.md
```diff
@@ -26,6 +26,56 @@ class.
 
  * For use from Ruby see the [Pushgateway documentation](https://github.com/prometheus/client_ruby#pushgateway).
 
+## Should I be using the Pushgateway?
```
brian-brazil: Should this be under best practices?

content/docs/instrumenting/pushing.md
```diff
@@ -26,6 +26,56 @@ class.
...
+For batch jobs that are related to a machine (such as automatic
+security update cronjobs or configuration management client runs), expose the
+resulting metrics using the [Node Exporter's](https://github.com/prometheus/node_exporter)
+textfile module instead of the Pushgateway.
+
+For batch jobs that neither belong to a specific machine nor to an entire
+service (i.e. they should expose metrics about individual instances of a job),
+consider converting them into long-running processes that can be scraped.
+
+If a firewall or NAT is preventing you from pulling metrics from targets,
```
brian-brazil: inbound firewall

A proper firewall would break both push and pull.

content/docs/instrumenting/pushing.md
```diff
@@ -26,6 +26,56 @@ class.
...
+metrics in the Pushgateway.
+
+For batch jobs that are related to a machine (such as automatic
+security update cronjobs or configuration management client runs), expose the
+resulting metrics using the [Node Exporter's](https://github.com/prometheus/node_exporter)
+textfile module instead of the Pushgateway.
+
+For batch jobs that neither belong to a specific machine nor to an entire
+service (i.e. they should expose metrics about individual instances of a job),
+consider converting them into long-running processes that can be scraped.
```
brian-brazil: or having the instance of the job expose the metric

content/docs/instrumenting/pushing.md
juliusv: Could be, not sure. I guess I'm fine with either, but I guess it'd fit into the "Instrumentation" best practices doc (and linked from here)?

content/docs/instrumenting/pushing.md
juliusv: Ok, changed to be about capturing final batch job state and also added a link at the end of this paragraph to the batch job instrumentation best practices.

content/docs/instrumenting/pushing.md
juliusv: Not sure what you mean here - isn't that what I'm already implying by making it a long-running process?

content/docs/instrumenting/pushing.md
juliusv: Added "inbound".

content/docs/instrumenting/pushing.md
brian-brazil: If there's an instance the batch job is associated with, then rather than making the batch job a new daemon beside the existing daemon it should either be folded in or take care of exposing the metrics.

content/docs/instrumenting/pushing.md
brian-brazil: Given how often this is coming up, it probably should get it's own section.

content/docs/instrumenting/pushing.md
juliusv: Ah, that would be even more exotic I guess. What I actually meant was that the batch job itself has multiple instances (not related to any instances of another job) and might have per-instance metrics. That kind of contraption should probably just be a long-running job, if it has any reason for existence at all... or maybe just remove this paragraph?

brian-brazil: That sounds like a service-level sharded batch job. Presuming the shards are static so that you can compare across runs, that's fine for the pgw.

Removing the paragraph is probably best to avoid confusion.

content/docs/instrumenting/pushing.md
```diff
@@ -26,6 +26,56 @@ class.
...
+
+If an inbound firewall or NAT is preventing you from pulling metrics from
+targets, consider moving the Prometheus server behind the network barrier as
+well. We generally recommend running Prometheus servers on the same network as
+the monitored instances.
```
brian-brazil: We should probably plug service discovery too. This paragraph is what's relevant to most users I'd suspect, so I'd have it up top.

juliusv:
Ok, will address the rest after some sleep...

juliusv:
@brian-brazil Take another look. I moved it over to a new section under "Best practices" and added/reorged it a tiny bit to fit. I'm still not plugging SD anywhere though, since I wasn't sure how to fit that in nicely. You're welcome to add that :)

brian-brazil:
:+1:

PR #421, opened 2016-04-30
Add user interview with showmax

brian-brazil:

PR #422, opened 2016-05-02
Added alts to images and explicitly formatted external links

bobek:
Hi @brian-brazil and @grobie,

many thanks for helping out with the article. I have made changes suggested by @grobie in https://github.com/prometheus/docs/pull/421 . Not sure, who you wanna format long alt texts (e.g. 80 chars wrap). I've checked other blog posts and it seems that you keep them as on-liners.

Cheers,

    Antonin


brian-brazil:
Thanks, I was just about to do this. Typesetting on a laptop while travelling is fun :)

PR #430, opened 2016-05-11
Adding BIG-IP exporter

cgunning:


content/docs/instrumenting/exporters.md
```diff
@@ -76,6 +76,7 @@ hosted outside of the Prometheus GitHub organization.
    * [SQL query result set metrics exporter](https://github.com/chop-dbhi/prometheus-sql)
    * [Ubiquiti UniFi exporter](https://github.com/mdlayher/unifi_exporter)
    * [Varnish exporter](https://github.com/jonnenauha/prometheus_varnish_exporter)
+   * [BIG-IP exporter](https://github.com/ExpressenAB/bigip_exporter)
```
brian-brazil: Please keep it alphabetical

brian-brazil:
Thanks for the exporter. Taking a quick peek, you've marked everything as a counter while many are actually gauges. If you don't want to distinguish you should use Untyped. Seconds and bytes are also the standard units for Prometheus.

cgunning:
Thanks for taking a look. I'll look over the types used and convert metrics to standard units.

cgunning:
I've now made some changes to my exporter and moved my exporter to be in the correct alphabetical order

brian-brazil:
Thanks!

PR #433, opened 2016-05-16
Aerospike exporter

alicebob:


alicebob:
maybe for @brian-brazil ?

brian-brazil:
Thanks.

Three comments:
  - You should use ConstMetrics rather than using GaugeVec/CounterVec. This handles things like disappearing namespaces better
  - Are there raw metric versions of ops per sec, percent free ram/disk etc.? If so those are preferred
  - The latency metrics would be best exposed in the form of a histogram, though it's a little odd as the rate has already been taken.

PR #434, opened 2016-05-16
List unofficial Erlang client

deadtrickster:
https://github.com/deadtrickster/prometheus.erl


brian-brazil:
Thanks!

Only a float API should be offered. In your histogram example you should have the units in seconds. `orders_summary` should be `orders`, don't put types in metric names.

PR #436, opened 2016-05-18
Deleted leftover "e.g."

Mingan:
@juliusv

### DIFF
--- a/content/docs/practices/naming.md
+++ b/content/docs/practices/naming.md
@@ -7,8 +7,8 @@ sort_rank: 1
 
 The metric and label conventions presented in this document are not required
 for using Prometheus, but can serve as both a style-guide and a collection of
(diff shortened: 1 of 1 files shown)


beorn7:
I think the `e.g.` is fine here as the following sections is not only about names, but also about semantics of labels.

Mingan:
You're right about that, but shouldn't `e.g.` link a concept and a list of instances of that concept?For example:

*Individual organizations may want to approach some of these practices, e.g. naming conventions, differently.*

beorn7:
I leave it to @juliusv to decide as he might be into this kind of subtleties. ;)

juliusv:
The current wording sounds correct to me, but you could maybe make it even clearer in the way you suggested, @Mingan.

juliusv:
:+1: thanks!

PR #437, opened 2016-05-18
Add yet another RabbitMQ Exporter

deadtrickster:
So the differences with listed one (https://github.com/kbudde/rabbitmq_exporter) are:
 - prometheus_rabbitmq_exporter is not standalone daemon but a plugin for RabbitMQ itself. 
 - it also exports Erlang VM stats.

Also I tried to stay compatible with rabbitmq_exporter as much as possible (probably added a couple of labels)

Not sure that 'RabbitMQ Management Plugin exporter' makes said differences clear though :-)

Thank you


grobie:
Plugins seem to be between standalone exporters and direct instrumentation. Maybe we should start a third list for this class?

brian-brazil:
This still classes an exporter, similar to the jmx exporter.

PR #439, opened 2016-05-20
Replace the deprecated third-party memcached_exporter by the official one

sdurrheimer:
@grobie


grobie:
Yeah, I wanted to do that once we have the release (for which I wanted to add back the slab + unread metrics). Happy to do this now though.

PR #441, opened 2016-05-24
Add Transloadit as a Prometheus user

kvz:
This change is [Reviewable](https://reviewable.io/reviews/prometheus/docs/441)


content/index.html
```diff
@@ -183,6 +183,9 @@
     <div class="col-md-2 col-sm-4 col-xs-6 logos">
       <a href="https://www.weave.works/"><img src="assets/company-logos/weave.png"/></a>
     </div>
+    <div class="col-md-2 col-sm-4 col-xs-6 logos">
+      <a href="https://transloadit.com/"><img src="assets/company-logos/transloadit.png"/></a>
```
brian-brazil: Keep it sorted please

brian-brazil:
Can you squash your commits please?

kvz:
Thanks for the pointers! Yes I could do that, but maybe it's just as easy to use the new squash & merge functionality? https://github.com/blog/2141-squash-your-commits. If you disagree or I'm overlooking something I'm happy to rebase

kvz:
Never mind, I squashed them already :)

grobie:
This will break our current grouping. We either need to find 5 more or kick another one out.

brian-brazil:
I don't think we should be too perfectionist on getting it even, there's a bit too much admin work there.

kvz:
I think flexbox could solve this by auto-aligning logos in a container. I'm by no means an expert on this subject, but I've played around a bit using the current (+ Transloadit) logo here and that seems to work okay: http://codepen.io/kvz/pen/vGooNv. Then the data does not have to consider the presentation anymore.

brian-brazil:
The existing bootstrap CSS should handle that, the question raised was more that we'll end up with one logo on it's own on the last line.

kvz:
If you check out the codepen preview you'll see that it arranges smaller logos on other lines so that this does not happen

brian-brazil:
Ah, that's pretty nice.

grobie:
@juliusv please merge or comment. I guess you feel most strongly about such things.

juliusv:
Either we go for the suggested (yet to be implemented) flexbox approach or we always manually curate a full row of logos. I think the suggested flexbox approach doesn't look too bad in the demo, so let's try it?

@kvz I'm not on Mac, but it looks like either you don't have the necessary OpenSSL headers installed or they can't be found. Maybe check out http://stackoverflow.com/questions/30818391/gem-eventmachine-fatal-error-openssl-ssl-h-file-not-found?

grobie:
For easier development I created #30 to provide a dev docker container.

beorn7:
@grobie When you say #30, you obviously mean #448. ;)

brian-brazil:
Is there anyone against flexbox? If not let's go with that.

beorn7:
I also have another company logo in the pipeline (www.qaware.de). I'll wait for the flexbox to be merged before I add that one.

grobie:
I'm fine with flexbox, as long as it still reads fine on mobile.

beorn7:
So who will implement the flexbox? We shouldn't let our logo donators wait for too long.

beorn7:
Is anybody owning the conversion to flexbox? This is totally not my area of expertise, but I'd like us to be able to add logos again. @juliusv @brian-brazil @kvz ?

kvz:
I'm having trouble getting the ruby thing to work on my machine. I checked the docker variant but that doesn't seem finished yet. I don't really feel comfortable publishing the flexbox code without having seen with my own eyes it works in the prometheus site. If someone could take (bits of) the code that I put up on codepen http://codepen.io/kvz/pen/vGooNv, and verify it works in the site, then I think we're done?

RichiH:
Just so this does not get lost:

If you want, FOSDEM can be added as well.

space.net could also be added if you're OK with that.

As per IRC, I will try and have a look at flexbox, but _no_ promises made.

juliusv:
@kvz Thanks! Hmm, I played around with it with different numbers of logos and different browser widths (and larger image widths and heights than the tiny ones in the codepen). After some tweaking I managed to get it to look good in many situations, but still odd in some. For example, this looks good:

good

But when resizing the browser to just a bit narrower, it ends up like this due to Bootstrap's responsive column resizing:

odd

This kind of thing does not happen with the current code. Maybe it is simpler to just require full rows after all :-/ Or does someone want to play more with this?

We'd have four new logos already (Transloadit, QAware, FOSDEM, space.net). If we can just find two more... :)

kvz:
I also did this for tus.io. I had to add a max-height, max-width and justify-content to get it behaving better. I linked the commit to this ticket for further inspection. I apologize about the whitespace crime.

beorn7:
For reference, the QAware folks told me they would like the URL http://www.qaware.de in their link. I also attach the logo here to be included by whomever tackles this issue.
qaware_logo_rgb_big

beorn7:
Conclusion from in-person discussion: Let's stick with six logos per row for now.

Waiting for two more to fill another row.

RichiH:
Resized to fit in with the rest:
spacenet_128x49

dominikschulz:
@juliusv Are you serious about finding two more users? I'd love to provide one ;)

juliusv:
@dominikh Sure, go ahead :)

dominikh:
@juliusv I'd love to, but I'm not @dominikschulz :P

juliusv:
Whoops - ETOOMANYDOMINIKS :)

beorn7:
Is this going anywhere?

@juliusv do you own this? I think leaving the logos stale for ages is the worst option. I'd prefer to not have logos at all over annoying organizations who want to be added but then stay in limbo forever. (Arguably, we are well known by now so we don't need logos anymore. We could link to case studies instead and have a separate page listing users, like kubernetes.io.)

I don't feel strongly about how it is done in detail. I only feel strongly that we should not leave open trivial PRs like this one forever. Whoever feels strongly about details like "6 logos per row" should also own the thing.

juliusv:
@beorn7 Yeah, it doesn't need to be strictly "6 logos per row", it'd just be great if our site didn't look crappy, and 6 logos per row is one way of achieving it :) Linking to case studies of some of the more well-known companies and getting rid of the general logo section completely seems like an option to me. Should I start doing that?

beorn7:
@juliusv I'm fine with anything you will do, as long as you do _something_. It's a really bad experience to keep people waiting for 8 months. That's my major concern right now. If you own the logo section, all good. If nobody own it, we have to do something about it.

brian-brazil:
Superseded by #705

PR #442, opened 2016-05-24
make helpful metrics more helpful

tamsky:
use the same description 'active series' that the rest of the text is using

### DIFF
--- a/content/docs/operating/storage.md
+++ b/content/docs/operating/storage.md
@@ -171,7 +171,7 @@ for 30% of required chunk removal.
 Out of the metrics that Prometheus exposes about itself, the following are
 particularly useful for tuning the flags above:
 
-* `prometheus_local_storage_memory_series`: The current number of series held
+* `prometheus_local_storage_memory_series`: The current number of active series held
   in memory.
 * `prometheus_local_storage_memory_chunks`: The current number of chunks held
   in memory.


beorn7:
"time series in memory" != "active time series".

The former is an exact technical term within Prometheus, while the latter is more vague and refers more to a time series that still gets samples appended.

Not sure if providing exact definitions and explanations here will help the reader or be more of a distraction.

juliusv:
Yeah, I the "active" is not well-defined, so I don't think this makes things clearer.

tamsky:
> Not sure if providing exact definitions and explanations here will help the reader or be more of a distraction.

As someone that tried to use this page to understand tune our storage settings, it would help me if  "exact definitions and explanations" were somewhere (vs nowhere) in the docs.

I gather that `prometheus_local_storage_memory_series` relates to a precisely defined real life quantity, I'm just not sure what.

If we can define that and put it in the docs, I'll go away happy.

beorn7:
`prometheus_local_storage_memory_series` is the number of series that have a representation in memory. The docstring of the metric is "The current number of series in memory."

I don't know how to make this more expressive without becoming very verbose. Open to suggestions, though.

tamsky:
There's this phrase I'm trying to get to the bottom of:  'active time series'.

It appears four times in the paragraph with heading "[Settings for high numbers of time series](https://github.com/prometheus/docs/blob/master/content/docs/operating/storage.md#settings-for-high-numbers-of-time-series)"

As someone unsure of what "too high" is regarding "high numbers of time series", it seems like this "active time series" *thing* is a possibly important concept.

I'm hoping to spark up a definition of this phrase/concept here, especially if my commit doesn't help.

beorn7:
My take on a definition: An active time series is one that has received a new sample from the last scrape cycle.

(In contrast, a time series could be represented in memory while not having received a sample in quite some time. It would be counted as "in memory" but not as "active".)

@tamsky Do you think you could work something like that into this PR so that it becomes clear without getting to verbose?

discordianfish:
I've got confused by pretty much the same: The difference between active timeseries, timeseries in memory and how rushed mode relates to throttled ingestion.

From what I got, a Prometheus instance running out of memory runs into throttled ingestion, which is the key problem (I don't get all the samples I want) which can have the following reasons:

1.) Storage can't keep up which leads to to-be-persisted chunks to pile up
2.) Many timeseries lead to many open head chunks

What I'm missing from the documentation is how persisting the samples works outside of rushed mode. Does it try to flush all samples to disk? Or has this operation a deadline?

Beside that it could be made a bit cleaner that throttled ingestion is the key problem, rushed mode an indicator that storage is too slow? (still not 100% clear, need to understand what Prom is doing in non-rushed mode) and the number of 'open' timeseries something that needs to fit into memory.

Maybe I can come up with a bit rewording after understanding the regular, non-rushed mode of persisting things.

beorn7:
> 2.) Many timeseries lead to many open head chunks

That's not the only possible reason. A query could pin older, already persisted chunks in memory, in which case too many chunks might be in memory, too. Ingestion is throttled if more chunks are waiting for persistence than configured, or if the number of chunks configured to be kept in memory is exceeded by more than 10% _for whatever reason_.

WRT non-rushed mode: The only "flip the switch" behavior is syncing after writes to series files if `-storage.local.series-sync-strategy=adaptive` (which is the default). In rushed mode, files are not sync'd, otherwise they are. The other thing is a more gradual one. So-called series maintenance is happening at a limited rate to batch up more chunks for writing. The higher the persistence-urgency score, the faster series maintenance happens. In rushed mode, series maintenance happens as fast as possible.

discordianfish:
Ah! Guess the series maintenance is what I missed. Maybe we should add that to the docs?

beorn7:
I'm just concerned that drowning the reader with internals might not be helpful in the end. Sure, if we could teach the reader everything about the storage, they could make the perfect decision in the end. It might just take a long while.
I'm happy to review suggestions for the docs. Perhaps somebody who is not a storage coder themself is in a better position to write concise yet helpful documentation.

brian-brazil:
This is stale, and the current docs seems correct to me.

PR #459, opened 2016-06-10
Add missing underscore to webhook_config variable

samwierema:
The webhook_config variable was missing an underscore in the documentation.


brian-brazil:
Thanks!

PR #460, opened 2016-06-14
configuration: Update scrape config

beorn7:
@sdurrheimer Now that we have started to update documentation: Need this documentation update, too.

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -78,8 +78,12 @@ A `scrape_config` section specifies a set of targets and parameters describing h
 to scrape them. In the general case, one scrape configuration specifies a single
 job. In advanced configurations, this may change.
 
-Targets may be statically configured via the `target_groups` parameter or
+Targets may be statically configured via the `static_configs` parameter or
 dynamically discovered using one of the supported service-discovery mechanisms.
+
+NOTE: Prior to v0.20, `target_groups` was used instead of `static_configs`.
+`target_groups` can still be used alternatively in v0.20 itself, but not in
+later versions.
 
 Additionally, `relabel_configs` allow advanced modifications to any
 target and its labels before scraping.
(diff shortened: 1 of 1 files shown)


beorn7:
Ping! Or anybody... @fabxc ?

content/docs/operating/configuration.md
```diff
@@ -220,13 +220,14 @@ A `tls_config` allows configuring TLS connections.
 ```
 
 
-### `<target_group>`
+### `<static_config>`
 
-A `target_group` allows specifying a list of targets and a common label set for them.
-They are the canonical way to specify static targets in a scrape configuration.
+A `static_confic` allows specifying a list of targets and a common label set
```
brian-brazil: We keep a note of the old name in the docs for a few months, to reduce the inevitable support hassles.

beorn7: Are you referring to documenting the backwards-compatibility, or do you want a note like "This option was called `target_group` up to 0.19?
In which case the same should be done for the file SD, no?

brian-brazil: The former, and yes that'd make sense.

beorn7: But we have no backwards-compatibility for the file SD.

Perhaps you mean "the latter"?

brian-brazil: Yes, I did mean the latter.

beorn7: Alright. :)
Updating now, so that we can then update the docs in master for 0.20.

beorn7:
@brian-brazil Updated. PHAL.

content/docs/operating/configuration.md
brian-brazil: Weren't you going to add a mention of the name pre-0.20?

brian-brazil:
:+1:

content/docs/operating/configuration.md
beorn7: Yes, but at a different place. `target_group` was never a keyword in the config (only `target_groups`).

PR #465, opened 2016-06-24
[docs] fix relabel default configs

Cofyc:
See https://github.com/prometheus/prometheus/blob/0.20.0/config/config.go#L92.

```
DefaultRelabelConfig = RelabelConfig{
	Action:      RelabelReplace,
	Separator:   ";",
	Regex:       MustNewRegexp("(.*)"),
	Replacement: "$1",
}
```


fabxc:
Thanks! 👍

Issue #466, opened 2016-06-27
JavaScript errors in Routing Tree Editor


martin-helmich:
I'm getting a few JavaScript errors when using the Routing Tree Editor (using Chrome 51.0.2704.103, 64 bit):

    routing-tree-cb07916790a.js:63 Uncaught ReferenceError: $ is not defined
        (anonymous function) @ routing-tree-cb07916790a.js:63

When providing a configuration and submitting, some follow-up errors occur:

    d3.v3.min-cbd1784140c.js:1 Error: <path> attribute d: Expected number, "MNaN,NaNCNaN,NaN …".
        a @ d3.v3.min-cbd1784140c.js:1
        (anonymous function) @ d3.v3.min-cbd1784140c.js:3
        Y @ d3.v3.min-cbd1784140c.js:1
        Aa.each @ d3.v3.min-cbd1784140c.js:3
        Aa.attr @ d3.v3.min-cbd1784140c.js:3
        update @ routing-tree-cb07916790a.js:205
        loadConfig @ routing-tree-cb07916790a.js:140
        (anonymous function) @ routing-tree-cb07916790a.js:55
        (anonymous function) @ d3.v3.min-cbd1784140c.js:1

From a quick look (not sure about this, though), I'm suspecting that the `routing-tree.js` and jQuery are loaded in the wrong order.

beorn7:
@stuartnelson3 Do you have an idea?

stuartnelson3:
The first bit:

```
routing-tree-cb07916790a.js:63 Uncaught ReferenceError: $ is not defined
    (anonymous function) @ routing-tree-cb07916790a.js:63
```

Despite jQuery being loaded before the routing-tree javascript, it appears to not be available. This is something I can look into immediately.

@martin-helmich would it be possible to get a sanitized copy of your config.yaml? That would be the easiest way to reproduce and debug the problem.

martin-helmich:
@stuartnelson3, sure, here you go (this is basically an adjusted and slightly reduced version of the example configuration):

```yaml
global:
  smtp_smarthost: "{{ smtp_host }}"
  smtp_from: 'alertmanager@mittwald.de'
  smtp_auth_username: ''
  smtp_auth_password: ''
  smtp_require_tls: false

templates:
- '/etc/alertmanager/template/*.tmpl'

route:
  group_by: ['alertname', 'cluster', 'service']
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 3h

  receiver: team-A

[...]

PR #467, opened 2016-06-27
Fix jQuery and yaml parsing errors

stuartnelson3:
~~TODO:~~

~~Still need to reproduce and fix the error when drawing certain routing trees. See the below issue for details.~~

Addresses https://github.com/prometheus/docs/issues/466


stuartnelson3:
Two nodes:
image

One node:
image

PR #468, opened 2016-06-28
add linkerd to directly-instrumented list

wmorgan:
linker exports Prometheus metrics directly, currently on `:9990/admin/metrics/prometheus`. (See e.g. https://github.com/BuoyantIO/linkerd/blob/master/admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala and https://github.com/BuoyantIO/linkerd/issues/396 for making this configurable.)

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -98,6 +98,7 @@ separate exporters are needed:
    * [Etcd](https://github.com/coreos/etcd)
    * [Kubernetes-Mesos](https://github.com/mesosphere/kubernetes-mesos)
    * [Kubernetes](https://github.com/GoogleCloudPlatform/kubernetes)
+   * [linkerd](https://github.com/BuoyantIO/linkerd)
    * [RobustIRC](http://robustirc.net/)
    * [SkyDNS](https://github.com/skynetservices/skydns)
    * [Weave Flux](http://weaveworks.github.io/flux/)
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -98,7 +98,7 @@ separate exporters are needed:
    * [Etcd](https://github.com/coreos/etcd)
    * [Kubernetes-Mesos](https://github.com/mesosphere/kubernetes-mesos)
    * [Kubernetes](https://github.com/GoogleCloudPlatform/kubernetes)
-   * [linkerd](https://github.com/BuoyantIO/linkerd)
+   * [Linkerd](https://github.com/BuoyantIO/linkerd)
    * [RobustIRC](http://robustirc.net/)
    * [SkyDNS](https://github.com/skynetservices/skydns)
    * [Weave Flux](http://weaveworks.github.io/flux/)


content/docs/instrumenting/exporters.md
```diff
@@ -98,6 +98,7 @@ separate exporters are needed:
    * [Etcd](https://github.com/coreos/etcd)
    * [Kubernetes-Mesos](https://github.com/mesosphere/kubernetes-mesos)
    * [Kubernetes](https://github.com/GoogleCloudPlatform/kubernetes)
+   * [linkerd](https://github.com/BuoyantIO/linkerd)
```
beorn7: Do you feel strongly about the lower-case "l"? Otherwise, I'd prefer a capital "L"  for consistency.

beorn7:
Great to hear. I could imagine @brian-brazil feels tempted now to give you some advice about the integration, but it's certainly fine to list it here.

brian-brazil:
This isn't directly instrumented, it looks to be an internal exporter that's working off one of the two twitter metrics systems. I don't think we've got a place for those yet on this page.

wmorgan:
@brian-brazil What's the distinction?

brian-brazil:
You're not using the Prometheus java client to instrument your code, you're using a different instrumentation system.

wmorgan:
Well that's true, but the heading for this section doesn't mention the official Prometheus Java client. It says "Some third-party software already exposes Prometheus metrics natively, so no separate exporters are needed:". I believe linkerd satisfies that description.

content/docs/instrumenting/exporters.md
wmorgan: We usually lower-case it regardless of whether it's at the beginning of a sentence or a list, but I'm honestly not too concerned about this. Will uppercase if you want.

beorn7: I'd prefer that, like we did with etcd above. ("Cadvisor" would not have worked, as a counter example. ;)

beorn7:
I agree that native instrumentation does not imply usage of an official client library.

The question is if the metrics you expose here are metrics _about_ linkerd or if those metrics are about other systems that linkerd interacts with.

wmorgan:
Ok, you two are saying different things.

@beorn7: to answer your question, linkerd exposes both metrics about its internal state and metrics about the systems it interacts with.

I've changed the capitalization. Happy to put it in another section or another page if you prefer.

beorn7:
> Ok, you two are saying different things.

That happens now and then. ;)

But in fact, while @brian-brazil takes a more formal approach by requiring usage of the client library, we both agree on the results: The exposed metrics should have proper "Promethean" semantics. Personally, I don't care if that's accomplished with or without an official client library. The client library makes it easy to support proper semantics, but you can create broken semantics with the client library and proper one without one.

I haven't looked into too much detail here, which is why I hinted that @brian-brazil might want to tell you about them.

I don't feel strongly in which category we link linkerd. In doubt, let's just put it under "Other third-party utilities". That can't be wrong.

brian-brazil:
We need a new category here I think. It doesn't fit in 3rd party utilities as linkerd doesn't appear to be an instrumentation tool. I imagine over time we'll find many things that are instrumented with some other library, but also can expose metrics in the Prometheus format.

> I haven't looked into too much detail here, which is why I hinted that @brian-brazil might want to tell you about them.

If you look at the linked scala file it's pretty clear. It's pulling from a Twitter `Registry`, sanitising metric names and manually writing the text format.

I'd like us to have a section for software that is doing Prometheus instrumentation in the recommended way, which is with our client libraries. "Directly instrumented software" seems like the right place for that. Right now the heading says "directly instrumented" and the copy says "exposes Prometheus metrics natively" which are different things which is where the confusion is.

Cadvisor should also move to the exporters as it's primarily proxying data from the kernel so it isn't directly instrumented.

beorn7:
Feel free to refactor the sections. I'll trust your judgement.

brian-brazil:
#749 rejigs this and add linkerd.

PR #473, opened 2016-07-05
Clarify what will be exposed in each histogram bucket

karora:
The ability to blacklist individual buckets is an important point in encouraging their use, and it is not obvious to the naive user.

### DIFF
--- a/content/docs/concepts/metric_types.md
+++ b/content/docs/concepts/metric_types.md
@@ -48,6 +48,10 @@ A _histogram_ samples observations (usually things like request durations or
 response sizes) and counts them in configurable buckets. It also provides a sum
 of all observed values.
 
+Since each observation bucket holds the total count of observations less than or
+equal to the the bucket's upper bound, any individual bucket can be blacklisted
+without affecting the validity of the resulting histogram.
+
 A histogram with a base metric name of `<basename>` exposes multiple time series
 during a scrape:
 


content/docs/concepts/metric_types.md
```diff
@@ -48,6 +48,10 @@ A _histogram_ samples observations (usually things like request durations or
 response sizes) and counts them in configurable buckets. It also provides a sum
 of all observed values.
 
+Since each observation bucket holds the total count of observations less than or
```
brian-brazil: This is an implementation detail, so would better go in the exposition format docs than the more user-facing docs.

karora: How would you phrase it so that it referred to the metrics I would see in a plain text version of the histogram information on a /metrics page that Prometheus is scraping?

What I should see in such case goes beyond being "an implementation detail" and becomes an interface which Prometheus has specific expectations about.  I imagine that implementations may simply maintain a counter for each bucket, and only add them up when they are being scraped.

This page is a conceptual discussion of the _histogram_ and should make it's characteristics clear - certainly after 40-odd years of seeing histograms as graphs this seemed odd, until the 'aha' moment when the point of wishing to blacklist buckets made the reasoning suddenly (blindingly) obvious.

beorn7: A couple of lines further down, it is mentioned that the counters are cumulative. "Cumulative histogram" is a real thing: https://en.wikipedia.org/wiki/Histogram#Cumulative_histogram
Perhaps it would be sufficient to stress the fact that the histograms in Prometheus are cumulative a bit more prominently and further up.

The implications of that are manifold, blacklisting is only one of them, so I would not go into that kind of detail here.

karora: A cumulative counter is, of course, something you expect of a counter.  To indicate that this is a **Cumulative Histogram**, and that that is actually a technical term (perhaps by linking to the Wikipedia page) would definitely clarify things.

It would be great if there were more detail explaining the manifold reasons _why_ a cumulative histogram is a good choice.  That was the one that Brian mentioned to me which made it super clear, but it'd be nice to know more about that, and so I shall now go read that Wikipedia article in the hope that I can learn more... :-)

beorn7: Yes, let's improve the wording and emphasis, and linking the term in wikipedia is also good.

However, we try to avoid justifications for technical decisions and design in the core documentation as that will blow them up immensely. This is more something we would have in a blog post, or perhaps in one of the less terse parts of the docs like the best-practice pages (it might have a place in https://prometheus.io/docs/practices/histograms/ ).

brian-brazil: I'd mention the blacklist offhand in the client library guidelines.

beorn7:
Let's get this done. @karora are you up for changing the PR so that it mentions the fact that the histogram is cumulative (with a link to wikipedia)?

karora:
I confess I got kind of lost on exactly what would meet your documentation standards and moved onto other things.  I'll close the issue if it's easier for you.

beorn7:
OK, let's close this, and I'll follow up with a PR that clarifies the cumulative nature of the histogram.

PR #475, opened 2016-07-07
update alert manager configuration fix typo

csawyerYumaed:
slack_config is an error in alert manager, it should be slack_configs

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -318,7 +318,7 @@ token: <string>
 [ expire: <duration> | default = 1h ]
 ```
 
-## Slack receiver `<slack_config>`
+## Slack receiver `<slack_configs>`
 
 Slack notifications are sent via [Slack webhooks](https://api.slack.com/incoming-webhooks).
 
@@ -341,6 +341,16 @@ channel: <tmpl_string>
 [ pretext: <tmpl_string> | default = '{{ template "slack.default.pretext" . }}' ]
 [ text: <tmpl_string> | default = '{{ template "slack.default.text" . }}' ]
 [ fallback: <tmpl_string> | default = '{{ template "slack.default.fallback" . }}' ]
+```
+
+example  slack_config receiver:
+
+```
+receivers:
+- name: 'slack'
+  slack_configs:
+   - api_url: 'https://hooks.slack.com/services/T00000000/B00000000/XXXXXXXXXXXXXXXXXXXXXXXX'
+     channel: '#Town Square'
 ```
 
 


brian-brazil:
The docs are correct, it's describing a single slack config.

csawyerYumaed:
relevant YAML:
receivers:
- name: 'mattermost'
  slack_config:
   - api_url: 'http://blahblah'
     channel: '#Town Square'

21:08 docker/3f250690b399[737]: time="2016-07-07T05:21:08Z" level=error msg="Loading configuration file failed: unknown fields in receiver config: slack_config" file="/etc/alertmanager/config.yml" source

brian-brazil:
Yes, that's what the docs say.

csawyerYumaed:
Sorry I'm confused.  The docs as I read them say use 'slack_config' (i.e. no s at the end).  however, alertmanager, the code seems to want the s at the end.  That's why I made this pull request, to add the s at the end. Or am I just a moron somehow (totally possible)?

csawyerYumaed:
So at the very least, the documentation is confusing (at least to me).  I think we can at least agree, the YAML file needs the s at the end, since that's the only way the code runs.... yes?

Are the docs not confusing to anyone else?

brian-brazil:
￼> the YAML file needs the s at the end, since that's the only way the code runs.... yes?

Yes, which is documented at https://prometheus.io/docs/alerting/configuration/#receiver-<receiver>

csawyerYumaed:
OK, I spent some time re-reading the documentation, since you kept saying that slack_configs is documented..  First I apologize this was definitely a drive-by patch, and I clearly didn't read the docs correctly.  You are correct, the word slack_configs does indeed show up, exactly once, in a block quote. the word slack_config however shows up 3 times, despite it not showing up in the YAML file at all, and is merely a placeholder for the actual config.

Overall I want the documentation to be easier to parse, when you are just trying to make it work.  Right now, you are correct, it *IS* documented, but it's pretty obtuse, with the slack_config (and all the other _config words on that page, it's very easy(at least for me) to assume that's the word you put in the YAML file....

If I spent some actual time re-working that whole page, adding in some words, and giving examples, and removing the word 'slack_config' (and replacing it with say... Slack Config (no _) so it's clear it's not tied to what you actually put in the YAML, would that be better looked upon?

If not that's fine, I was just trying to make other people's lives easier. Either way, let's get rid of this pull request, since it's also clearly incorrect with the rest of the documentation.

brian-brazil:
The thing is that above we call it `slack_config` so calling it anything else (particularly with spaces) will cause ambiguity when you consider how it's a list, and then we'd get to repeat the list semantics in each receiver type - which is also messy.

This is reference documentation, it's not the place for examples.

PR #478, opened 2016-07-08
Document count_values aggregator.

brian-brazil:


content/docs/querying/operators.md
```diff
@@ -194,12 +195,18 @@ or preserve distinct dimensions by including a `without` or `by` clause.
...
+`parameter` is only required for `count_values`,`topk` and `bottomk`. `without`
+removes the listed labels from the result vector, while all other labels are
+preserved the output. `by` does the opposite and drops labels that are not
+listed in the `by` clause, even if their label values are identical between all
+elements of the vector. The `keep_common` clause allows keeping those extra
+labels (labels that are identical between elements, but not in the `by`
+clause).
+
+`count_values` outputs one time series per unique sample value, with a
+label added with the label name given by the aggregation parameter and
```
beorn7: `with a label added with the label` is confusing to read. Perhaps break up the sentence:

`count_values` outputs one time series per unique sample value. Each series has an additional label. The name of that label is given by the aggregation parameter, and the label value is the unique sample value.

brian-brazil: That reads better alright, updated.

beorn7:
:+1:

PR #487, opened 2016-07-15
Document quantile and quantile_over_time

brian-brazil:
Goes with https://github.com/prometheus/prometheus/pull/1799

### DIFF
--- a/content/docs/querying/functions.md
+++ b/content/docs/querying/functions.md
@@ -325,6 +325,7 @@ over time and return an instant vector with per-series aggregation results:
 * `max_over_time(range-vector)`: the maximum value of all points in the specified interval.
 * `sum_over_time(range-vector)`: the sum of all values in the specified interval.
 * `count_over_time(range-vector)`: the count of all values in the specified interval.
+* `quantile_over_time(scalar, range-vector)`: the φ-quantile (0 ≤ φ ≤ 1) of the values in the specified interval.
 
 Note that all values in the specified interval have the same weight in the
 aggregation even if the values are not equally spaced throughout the interval.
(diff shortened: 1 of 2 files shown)


content/docs/querying/functions.md
```diff
@@ -336,3 +325,4 @@ over time and return an instant vector with per-series aggregation results:
 * `max_over_time(range-vector)`: the maximum value of all points under the specified interval.
 * `sum_over_time(range-vector)`: the sum of all values under the specified interval.
 * `count_over_time(range-vector)`: the count of all values under the specified interval.
+* `quantile_over_time(scalar, range-vector)`: the given quantile of the values under the specified interval.
```
beorn7: As we did in the code itself, should we change "under" to a less weird preposition like "in" or "within" or "contained in"?

beorn7:
:+1: but needs a rebase.

beorn7:
One more thing: If we just say "the given quantile" or "the k quantile", people will probably think about q-quantiles as in https://en.wikipedia.org/wiki/Quantile , i.e. integer numbers etc, e.g. "the 100-quantile is called percentile".

I haven't yet found a good name for the fractional quantiles we use throughout. Sometimes I see it called "φ quantile", which I also used in the Prometheus doc, e.g. in the famous paper https://www.cs.ucsb.edu/~suri/psdir/ency.pdf , but it's definitely not a widely known term. Not sure what's the best way here.

beorn7:
OK, checked our existing documentation, I think we should do it in the same way as for https://prometheus.io/docs/querying/functions/#histogram_quantile() , i.e. use the wording "φ-quantile (0 ≤ φ ≤ 1)". That should make clear what we mean and is also consistent throughout the docs.

content/docs/querying/functions.md

brian-brazil: Actually that was just an artifact of not being rebased.

beorn7:
Perhaps the next-release branch should be reset to the top of master first?

beorn7:
If you rebase _now_, things should be good.

And please consider what I said about the quantile terminology.

brian-brazil:
That's updated.

beorn7:
Thanks. :+1:

PR #488, opened 2016-07-18
Update Kubernetes SD config

fabxc:


brian-brazil:
The line listing the label values above also needs updating.

fabxc:
updated

PR #492, opened 2016-07-18
Add 1.0 announcement news bar.

juliusv:
Revert this after we feel that the news is not hot anymore.


juliusv:
@grobie

grobie:
Thanks!

juliusv:
Looks like this now:

prom

PR #495, opened 2016-07-19
Replaces deprecated `scrape_configs` in examples

benzimmer:
According to https://prometheus.io/docs/operating/configuration/ the `target_groups` field was replaced by `static_configs`.


beorn7:
:+1: Thanks for spotting and fixing.

PR #499, opened 2016-07-23
Blog post: "Pull does not scale - or does it?"

juliusv:
This change is [Reviewable](https://reviewable.io/reviews/prometheus/docs/499)

### DIFF
--- a/content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
+++ b/content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
@@ -1,0 +1,181 @@
+---
+title: Pull doesn't scale - or does it?
+created_at: 2016-07-23
+kind: article
+author_name: Julius Volz
+---
+
+Let's talk about a particularly persistent myth. Whenever there is a discussion
+about monitoring systems and Prometheus's pull-based metrics collection
+approach comes up, someone inevitably chimes in about how a pull-based approach
+just “fundamentally doesn't scale”. The given reasons are often vague or only
+apply to systems that are fundamentally different from Prometheus. In fact,
+having worked with pull-based monitoring at the largest scales, this claim runs
+counter to our own operational experience.
+
+We already have an FAQ entry about
+[why Prometheus chooses pull over push](/docs/introduction/faq/#why-do-you-pull-rather-than-push?),
+but it does not focus specifically on scaling aspects. Let's have a closer look
+at the usual misconceptions around this claim and analyze whether and how they
+would apply to Prometheus.
+
+## Prometheus is not Nagios
+
+When people think of a monitoring system that actively pulls, they often think
+of Nagios. Nagios has a reputation of not scaling well, in part due to spawning
+subprocesses for active checks that can run arbitrary actions on the Nagios
+host in order to determine the health of a certain host or service. This sort
+of check architecture indeed does not scale well, as the central Nagios host
+quickly gets overwhelmed. As a result, people usually configure checks to only
+be executed every couple of minutes, or they run into more serious problems.
+
+However, Prometheus takes a fundamentally different approach altogether.
+Instead of executing check scripts, it only collects time series data from a
+set of instrumented targets over the network. For each target, the Prometheus
+server simply fetches the current state of all metrics of that target over HTTP
+(in a highly parallel way, using goroutines) and has no other execution
+overhead that would be pull-related. This brings us to the next point:
+
+## It doesn't matter who initiates the connection
+
+For scaling purposes, it doesn't matter who initiates the TCP connection over
+which metrics are then transferred. Either way you do it, the effort for
+establishing a connection is small compared to the metrics payload and other
+required work.
+
+But a push-based approach could use UDP and avoid connection establishment
+altogether, you say! True, but the TCP/HTTP overhead in Prometheus is still
+negligible compared to the other work that the Prometheus server has to do to
+ingest data (especially persisting time series data on disk). To put some
+numbers behind this: a single big Prometheus server can easily store millions
+of time series, with a record of 800,000 incoming samples per second (as
+measured with real production metrics data at SoundCloud). Given a 10-seconds
+scrape interval and 700 time series per host, this allows you to monitor over
+10,000 machines from a single Prometheus server. The scaling bottleneck here
+has never been related to pulling metrics, but usually to the speed at which
+the Prometheus server can ingest the data into memory and then sustainably
+persist and expire data on disk/SSD.
+
+Also, although networks are pretty reliable these days, using a TCP-based pull
+approach makes sure that metrics data arrives reliably, or that the monitoring
+system at least knows immediately when the metrics transfer fails due to a
+broken network.
+
+## Prometheus is not an event-based system
+
+Some monitoring systems are event-based. That is, they report each individual
+event (an HTTP request, an exception, you name it) to a central monitoring
+system immediately as it happens. This central system then either aggregates
+the events into metrics (StatsD is the prime example of this) or stores events
+individually for later processing (the ELK stack is an example of that). In
+such a system, pulling would be problematic indeed: the instrumented service
+would have to buffer events between pulls, and the pulls would have to happen
+incredibly frequently in order to simulate the same “liveness” of the
+push-based approach and not overwhelm event buffers.
+
+However, again, Prometheus is not an event-based monitoring system. You do not
+send raw events to Prometheus, nor can it store them. Prometheus is in the
+business of collecting aggregated time series data. That means that it's only
+interested in regularly collecting the current *state* of a given set of
+metrics, not the underlying events that led to the generation of those metrics.
+For example, an instrumented service would not send a message about each HTTP
+request to Prometheus as it is handled, but would simply count up those
+requests in memory.  This can happen hundreds of thousands of times per second
+without causing any monitoring traffic. Prometheus then simply asks the service
+instance every 15 or 30 seconds (or whatever you configure) about the current
+counter value and stores that value together with the scrape timestamp as a
+sample. Other metric types, such as gauges, histograms, and summaries, are
+handled similarly. The resulting monitoring traffic is low, and the pull-based
+approach also does not create problems in this case.
+
+## But now my monitoring needs to know about my service instances!
+
+With a pull-based approach, your monitoring system needs to know which service
+instances exist and how to connect to them. Some people are worried about the
+extra configuration this requires on the part of the monitoring system and see
+this as an operational scalability problem.
+
+We would argue that you cannot escape this configuration effort for
+serious monitoring setups in any case: if your monitoring system doesn't know
+what the world *should* look like and which monitored service instances
+*should* be there, how would it be able to tell when an instance just never
+reports in, is down due to an outage, or really is no longer meant to exist?
+This is only acceptable if you never care about the health of individual
+instances at all, like when you only run ephemeral workers where it is
+sufficient for a large-enough number of them to report in some result. Most
+environments are not exclusively like that.
+
+If the monitoring system needs to know the desired state of the world anyway,
+then a push-based approach actually requires *more* configuration in total. Not
+only does your monitoring system need to know what service instances should
+exist, but your service instances now also need to know how to reach your
+monitoring system. A pull approach not only requires less configuration,
+it also makes your monitoring setup more flexible. With pull, you can just run
+a copy of production monitoring on your laptop to experiment with it. It also
+allows you just fetch metrics with some other tool or inspect metrics endpoints
+manually. To get high availability, pull allows you to just run two identically
+configured Prometheus servers in parallel. And lastly, if you have to move the
+endpoint under which your monitoring is reachable, a pull approach does not
+require you to reconfigure all of your metrics sources.
+
+On a practical front, Prometheus makes it easy to configure the desired state
+of the world with its built-in support for a wide variety of service discovery
+mechanisms for cloud providers and container-scheduling systems: Consul,
+Marathon, Kubernetes, EC2, DNS-based SD, Azure, Zookeeper Serversets, and more.
+Prometheus also allows you to plug in your own custom mechanism if needed.
+In a microservice world or any multi-tiered architecture, it is also
+fundamentally an advantage if your monitoring system uses the same method to
+discover targets to monitor as your service instances use to discover their
+backends. This way you can be sure that you are monitoring the same targets
+that are serving production traffic and you have only one discovery mechanism
+to maintain.
+
+## Accidentally DDoS-ing your monitoring
+
+Whether you pull or push, any time-series database will fall over if you send
+it more samples than it can handle. However, in our experience it's slightly
+more likely for a push-based approach to accidentally bring down your
+monitoring. If the control over what metrics get ingested from which instances
+is not centralized (in your monitoring system), then you run into the danger of
+experimental or rogue jobs suddenly pushing lots of garbage data into your
+production monitoring and bringing it down.  There are still plenty of ways how
+this can happen with a pull-based approach (which only controls where to pull
+metrics from, but not the size and nature of the metrics payloads), but the
+risk is lower. More importantly, such incidents can be mitigated at a central
+point.
+
+## Real-world proof
+
+Besides the fact that Prometheus is already being used to monitor very large
+setups in the real world (like using it to [monitor millions of machines at
+DigitalOcean](http://promcon.io/talks/scaling_to_a_million_machines_with_prometheus/)),
+there are other prominent examples of pull-based monitoring being used
+successfully in the largest possible environments. Prometheus was inspired by
+Google's Borgmon, which was (and partially still is) used within Google to
+monitor all its critical production services using a pull-based approach. Any
+scaling issues we encountered with Borgmon at Google were not due its pull
+approach either. If a pull-based approach scales to a global environment with
+many tens of datacenters and millions of machines, you can hardly say that pull
+doesn't scale.
+
+## But there are other problems with pull!
+
+There are indeed setups that are hard to monitor with a pull-based approach.
+A prominent example is when you have many endpoints scattered around the
+world which are not directly reachable due to firewalls or complicated
+networking setups, and where it's infeasible to run a Prometheus server
+directly in each of the network segments. This is not quite the environment for
+which Prometheus was built, although workarounds are often possible ([via the
+Pushgateway or restructuring your setup](/docs/practices/pushing/)). In any
+case, these remaining concerns about pull-based monitoring are usually not
+scaling-related, but due to network operation difficulties around opening TCP
+connections.
+
+## All good then?
+
+This article addresses the most common scalability concerns around a pull-based
+monitoring approach. With Prometheus and other pull-based systems being used
+successfully in very large environments and the pull aspect not posing a
+bottleneck in reality, the result should be clear: the “pull doesn't scale”
+argument is not a real concern. We hope that future debates will focus on
+aspects that matter more than this red herring.


juliusv:
@grobie @brian-brazil

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,175 @@
+---
+title: Pull doesn't scale - or does it?
+created_at: 2016-07-23
+kind: article
+author_name: Julius Volz
+---
+
+Let's talk about one particularly hard-to-kill myth again. Whenever there is a
```
beorn7: I suggest to use a word like "misconception" instead of "myth". As somebody who did serious academic work on myths, I always cringe if it is used as a synonym of "misconception". It's really like using "hacker" for "criminal that breaks into computer system for personal gain".

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,175 @@
...
+discussion about monitoring systems and Prometheus's pull-based metrics
+collection approach comes up, someone inevitably chimes in about how a
+pull-based approach just "fundamentally doesn't scale". When asked why they
```
beorn7: Typography nag: Use “…” instead of "…"

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,175 @@
...
+think this is the case, they either respond with vague fears or cite reasons
+which only apply to systems which are fundamentally different from Prometheus
```
beorn7: s/which/that/

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Pfft, those new-fangled non-ASCII quotes aren't on my keyboard. But ok :P

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,175 @@
...
+altogether, you say! True, but the TCP/HTTP overhead in Prometheus is still
+negligible compared to the other work that the Prometheus server has to do to
+ingest data (especially persisting time series data on disk). To put some
+numbers behind this: a single beefy Prometheus server can easily store millions
+of time series, with a record of 800,000 incoming samples per second (as
+measured with real production metrics data at SoundCloud). This allows you to
+monitor over 10,000 machines from a single Prometheus server. The scaling
+bottleneck here has never been related to pulling metrics, but usually to the
+speed at which the Prometheus server can ingest the data into memory and then
+sustainably persist it on disk/SSD.
```
beorn7: Not sure if you want to mention it, but the real bottleneck is the purging, not so much the writing. Only once purging kicks fully in, we hit the 800,000 limit.
I guess it would confuse people even more. But perhaps hint towards it like "sustainably persist and expire it on disk".

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,175 @@
...
+serious monitoring setups in any case: if your monitoring system doesn't know
+what the world *should* look like and which monitored service instances
+*should* be there, how would it be able to tell when an instance just never
+reports in, is down due to an outage, or really is no longer meant to exist?
+This is only acceptable if you never care about the health of individual
+instances at all, like when you only run ephemeral workers where it is
+sufficient for a large-enough number of them to report in some result. Most
+environments are not exclusively like that.
+
+If the monitoring system needs to know the desired state of the world anyways,
```
beorn7: s/anyway/anyways/ as discussed earlier.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Dunno, Google defines the second meaning of "myth" as "a widely held but false belief or idea." Seems legit to me, and is common usage.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Changed the quotes here and elsewhere.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Done.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,175 @@
...
+directly in each of the network segments. This is not quite the environment for
+which Prometheus was built, although workarounds are often possible ([via the
+Pushgateway or restructuring your setup](/docs/practices/pushing/)). In any
+case, these remaining concerns about pull-based monitoring are usually not
+scaling-related, but concern some other operational difficulty that pulling
+poses in certain environments.
+
+## All good then?
+
+We know that the debate about pulling versus pushing is a religious one and as
```
beorn7: A "religious" debate will just become more "religious" if you call them out as such. After having nicely demonstrated where pull vs. push matters and where not, perhaps it would have a more positive tone if you express the hope that the debate will in future be more about the aspects that matter and not about the red herring of scalability anymore?

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
beorn7: Yes, that's why I compared it to the word "hacker". I'm pretty sure the most common usage is the one for a criminal. It's your blog post, of course. I just tried to convey what a domain expert of myths feels if you use the word in the common way. Your decision if you like to make the experts feel that way or if the common usage is good enough for your standards.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: I changed the last paragraph - check it out.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Done.

beorn7:
Great post. Please consider all my comments as nits.

The one thing I really like to stress is missing, though: In a microservice or any multi-tiered architecture, it is fundamentally an advantage if your monitoring system uses the same method to discover targets to monitor as your service instance uses to discover its backends.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Hmm, I think I'm fine with using the "common person's" definition of "myth", but let's see what others say.

juliusv:
@beorn7 That's a great point. I took that sentence and stuck it at the end of the service discovery paragraph.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+Besides the fact that Prometheus is already being used to monitor very large
+setups in the real world (like using it to [monitor millions of machines at
+DigitalOcean](http://promcon.io/talks/scaling_to_a_million_machines_with_prometheus/)),
+there are other prominent examples of pull-based monitoring being used
+successfully in the largest possible environments. Prometheus was inspired by
+Google's Borgmon, which was (and partially still is) used within Google to
+monitor all its critical production services using a pull-based approach. Any
+scaling issues we encountered with Borgmon at Google were not due its pull
+approach either. If a pull-based approach scales to a global environment with
+many tens of datacenters and hundreds of thousands of machines, you can hardly say
```
grobie: This sounds a bit like Google runs less than a million of machines.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+which your production monitoring is reachable without reconfiguring all of your
+metrics sources.
+
+On a practical front, Prometheus makes it easy to configure the desired state
+of the world with its built-in support for a wide variety of service discovery
+mechanisms for cloud providers and container-scheduling systems: Consul,
+Marathon, Kubernetes, EC2, DNS-based SD, Azure, Zookeeper Serversets, and more.
+Prometheus also allows you to plug in your own custom mechanism if needed.
+In a microservice world or any multi-tiered architecture, it is also
+fundamentally an advantage if your monitoring system uses the same method to
```
grobie: Why? This is a claim without any explanation or example.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Accidentally DDoS-ing your monitoring
+
+Whether you use a pull-based or a push-based approach, any time-series database
+will fall over if you send it more samples than it can handle. However, we have
+made the experience that it's slightly more likely for a push-based approach to
+accidentally bring down your monitoring. If the control over what metrics get
+ingested from which instances is not centralized (in your monitoring system),
+then you run into the danger of experimental or rogue jobs suddenly pushing
+lots of garbage data into your production monitoring and bringing it down.
```
grobie: This sounds like there is any kind of protection in Prometheus against single rogue jobs sending too much data.

grobie: It might make sense to highlight manually changing scrape interval as well as removing a job completely explicitly here.

grobie:
Thanks a lot for writing this great article!

Given that this debate is already quite a heated one, I'd have tried to use a less lecturing tone in a few places. And I'd suggest to keep the discussion open and using a different final paragraph than telling the reader that they have been wrong. There are likely other concerns people have and we don't know about / you didn't address in the article. Instead of  hardening the fronts further by closing the post with "you're wrong", what about something like "Do you see other scalability issues with the pull model we haven't addressed here? Tell us in the comments!"?

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+fundamentally an advantage if your monitoring system uses the same method to
+discover targets to monitor as your service instances use to discover their
+backends.
+
+## Accidentally DDoS-ing your monitoring
+
+Whether you use a pull-based or a push-based approach, any time-series database
+will fall over if you send it more samples than it can handle. However, we have
+made the experience that it's slightly more likely for a push-based approach to
+accidentally bring down your monitoring. If the control over what metrics get
```
grobie: I don't really see why it'd be easier in push based to cause trouble. Conceptionally both systems appear to be in the same state. What I see is easier is to project against rogue clients due to having ultimate control over what gets ingested (metric relabeling) from whom. I'd suggest to change the sentence to `However, we have made the experience that it's easier to mitigate such situations in a pull-based approach ...`.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+created_at: 2016-07-23
+kind: article
+author_name: Julius Volz
+---
+
+Let's talk about one particularly hard-to-kill myth again. Whenever there is a
+discussion about monitoring systems and Prometheus's pull-based metrics
+collection approach comes up, someone inevitably chimes in about how a
+pull-based approach just “fundamentally doesn't scale”. When asked why they
+think this is the case, they either respond with vague fears or cite reasons
```
grobie: That's an example for the lecturing wording I meant by the way. If you want to convince proponents of the push model, I'd suggest to use a less alienating wording.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
brian-brazil: "again"?

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+subprocesses for active checks that can run arbitrary actions on the Nagios
+host in order to determine the health of a certain host or service. This sort
+of check architecture indeed does not scale well, as the central Nagios host
+quickly gets overwhelmed. As a result, people usually configure checks to only
+be executed every couple of minutes, or they run into more serious problems.
+
+However, Prometheus takes a fundamentally different approach altogether.
+Instead of executing check scripts, it only collects time series data from a
+set of instrumented targets over the network. For each target, the Prometheus
+server simply fetches the current state of all metrics of that target over HTTP
```
brian-brazil: I'd mention that this is threaded/uses go routines

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+required work.
+
+But a push-based approach could use UDP and avoid connection establishment
+altogether, you say! True, but the TCP/HTTP overhead in Prometheus is still
+negligible compared to the other work that the Prometheus server has to do to
+ingest data (especially persisting time series data on disk). To put some
+numbers behind this: a single beefy Prometheus server can easily store millions
+of time series, with a record of 800,000 incoming samples per second (as
+measured with real production metrics data at SoundCloud). This allows you to
+monitor over 10,000 machines from a single Prometheus server. The scaling
```
brian-brazil: We should probably state the assumptions here - 10s scrape interval and 700 time series.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+bottleneck here has never been related to pulling metrics, but usually to the
+speed at which the Prometheus server can ingest the data into memory and then
+sustainably persist and expire data on disk/SSD.
+
+Also, although networks are pretty reliable these days, using a
+TCP-based pull approach makes doubly sure that metrics data arrives reliably,
```
brian-brazil: Lots of push approaches use TCP, and the checksums in TCP aren't that great. I'm not sure what "doubly sure" is having as the two items.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+or that the monitoring system at least knows immediately when the metrics
+transfer fails due to a broken network.
+
+## Prometheus is not an event-based system
+
+Some monitoring systems are event-based. That is, they report each individual
+event (an HTTP request, an exception, you name it) to a central monitoring
+system immediately as it happens. This central system then either aggregates
+the events into metrics (StatsD is one example of this) or stores events
```
brian-brazil: I'd say the prime example

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+individually for later processing (the ELK stack is an example of that). In
+such a system, pulling would be problematic indeed: the instrumented service
+would have to buffer events between pulls, and the pulls would have to happen
+incredibly frequently in order to simulate the same “liveness” of the
+push-based approach and not overwhelm event buffers.
```
brian-brazil: This shouldn't make a difference for push vs. pull, same bytes on the wire and you'd need all those buffers anyway to handle the other end being flaky.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+counter value and stores that value together with the scrape timestamp as a
+sample. Other metric types, such as gauges, histograms, and summaries, are
+handled similarly. The resulting monitoring traffic is low, and the pull-based
+approach also does not create problems in this case.
+
+## But now my monitoring needs to know about my service instances!
+
+With a pull-based approach, your monitoring system needs to know which service
+instances exist and how to connect to them. Often people are worried about the
+extra configuration this requires on the side of the monitoring system.
```
brian-brazil: "on the monitoring system side"

it'd be "on the part of the monitoring system"

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+instances at all, like when you only run ephemeral workers where it is
+sufficient for a large-enough number of them to report in some result. Most
+environments are not exclusively like that.
+
+If the monitoring system needs to know the desired state of the world anyway,
+then a push-based approach actually requires *more* configuration in total. Not
+only does your monitoring system need to know what service instances should
+exist, but your service instances now also need to know how to reach your
+monitoring system. This is not only more required configuration, it also makes
+your monitoring setup less flexible. Now you cannot just run a copy of
```
brian-brazil: I'd frame this as a positive for pull rather than a negative for push.

I think it'd help convince people more to show the cool stuff you can do with pull (e.g. decentralisation) rather than saying what push can't do.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+In a microservice world or any multi-tiered architecture, it is also
+fundamentally an advantage if your monitoring system uses the same method to
+discover targets to monitor as your service instances use to discover their
+backends.
+
+## Accidentally DDoS-ing your monitoring
+
+Whether you use a pull-based or a push-based approach, any time-series database
+will fall over if you send it more samples than it can handle. However, we have
+made the experience that it's slightly more likely for a push-based approach to
```
brian-brazil: However in our experience it's...

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+
+There are indeed setups that are hard to monitor with a pull-based approach.
+A prominent example is when you have many endpoints scattered around the
+world which are not directly reachable due to firewalls or complicated
+networking setups, and where it's infeasible to run a Prometheus server
+directly in each of the network segments. This is not quite the environment for
+which Prometheus was built, although workarounds are often possible ([via the
+Pushgateway or restructuring your setup](/docs/practices/pushing/)). In any
+case, these remaining concerns about pull-based monitoring are usually not
+scaling-related, but concern some other operational difficulty that pulling
```
brian-brazil: concrete network operation difficulty that opening a tcp connection

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: I removed the "again". I had it there because we've been explaining this to people in many other places, but not in an official Prometheus blog post yet.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: I changed the wording - does it sound nicer?

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Done.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Done.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: I removed the "doubly" - or did you want to remove this section altogether?

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Done.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Well, at least StatsD just takes a fire-and-forget approach to this, so no buffers are needed there. Maybe that needs more clarification / differentiation?

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Done.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Changed the whole second half of this paragraph to be positively worded.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: I would guess because then you don't have to build two potentially inconsistent implementations of what your world looks like, and your monitoring system would see the world differently than your services? But @beorn7 can say more about why he proposed this sentence.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Done.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Although mitigation is the focus, I'd say it's still slightly less likely to happen to begin with because some random intern's first-day test job cannot accidentally send metrics to your monitoring system at all without someone configuring the monitoring system to do so.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: Well, at one point... but yeah, changed it to millions now.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: That's an odd string of words, but I guess I know what you meant - changed it to something that hopefully represents that.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
beorn7: I thought that would be obvious, but obviously it's not... ;)

Using the same method for discovering backends as for discovering monitoring targets ensures that you are monitoring what your backends are using.

If monitoring config and service discovery are separate, you might monitor targets that are not serving production traffic and/or the targets that _are_ serving production traffic might stay unmonitored.

beorn7:
BTW: I don't mind the lecturing tone. On the contrary, I enjoy it being factual. (That's why I suggested to remove the implication the debate of the others is "religious". We would like to make the debate more factual. A lecturing tone might not be the worst strategy to accomplish that.)

fabxc:
I'd like to give this a proper read before publishing. Will get to it in a
few hours.

Agreeing with grobie though that a lecturing tone is not going to convert
anyone who has a strong opinion here. Staying a bit more open doesn't mean
we cannot be factual.
(Haven't read yet to judge the tone myself though.)

juliusv:
@fabxc I hope I improved the tone already.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+server simply fetches the current state of all metrics of that target over HTTP
+(in a highly parallel way, using goroutines) and has no other execution
+overhead that would be pull-related. This brings us to the next point:
+
+## It doesn't matter who initiates the connection
+
+For scaling purposes, it doesn't matter who initiates the TCP connection over
+which metrics are then transferred. Either way you do it, the effort for
+establishing a connection is small compared to the metrics payload and other
+required work.
```
fabxc: Is that backed by measurements? FWIW, TCP connection initiation, ramp up etc. is not particularly fast. Still, this applies to push and pull as well.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+
+But a push-based approach could use UDP and avoid connection establishment
+altogether, you say! True, but the TCP/HTTP overhead in Prometheus is still
+negligible compared to the other work that the Prometheus server has to do to
+ingest data (especially persisting time series data on disk). To put some
+numbers behind this: a single beefy Prometheus server can easily store millions
```
fabxc: I would scratch "beefy". Not exactly a technical word and not every non-native speaker will know what you mean here.

fabxc:
Nice read.

I'd probably emphasize the DoS aspect even more. Not only makes our approach Prometheus actually scale well, no, it makes it scale heaps better than classical system where the network is spammed with a ton of events.
Arguably, an issue that can be circumvented in StatsD by pre-aggregating (setup overhead) or by pushing metric states in intervals. But there's no system out there that does it, right?

beorn7:
I would assume that most metrics-based push-driven monitoring systems don't suffer from the StatsD "one packet per event" fallacy. I think we shouldn't speculate too much about other systems but focus on what's the case in the Prometheus ecosystem, which in fact cannot suffer from random monitored clients deciding to send data to the monitoring server. Also, Prometheus has several protection mechanisms against run-away scrapes and too many samples. Those could be improved' and a push system can implement protection mechanisms, too, but I'd say we can safely claim that the protection is easier to get in a pull-based system.

WRT convincing those with a strong opinion: I believe we won't convert many of those anyway, and luckily, I don't think there are so many out there in the first place. Our most important audience are those deciding which monitoring system to use, and those should not be scared away by unfounded claims that pull doesn't scale. For those, a nice factual and concise explanation is best.

fabxc:
Sounds good, thanks 👍

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
grobie: I'd not rely on shared "obvious" knowledge in a discussion where you're trying to convince people with a different opinion. There should be an explanation for every argument so that it doesn't just become a claim which can be easily ignored.

We're constantly discussing the need for monitoring targets which are unhealthy and don't serve production traffic (see Kubernetes sd pull requests) and at SoundCloud we don't have a single monitoring configuration which is identical to what is used in services for exactly the reason to not only monitor what's serving production traffic.

To be honest, I'd just remove that last sentence in this paragraph.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+This is only acceptable if you never care about the health of individual
+instances at all, like when you only run ephemeral workers where it is
+sufficient for a large-enough number of them to report in some result. Most
+environments are not exclusively like that.
+
+If the monitoring system needs to know the desired state of the world anyway,
+then a push-based approach actually requires *more* configuration in total. Not
+only does your monitoring system need to know what service instances should
+exist, but your service instances now also need to know how to reach your
+monitoring system. A pull approach not only more requires less configuration,
```
grobie: Remove the `more`?

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
```diff
@@ -0,0 +1,178 @@
...
+In a microservice world or any multi-tiered architecture, it is also
+fundamentally an advantage if your monitoring system uses the same method to
+discover targets to monitor as your service instances use to discover their
+backends.
+
+## Accidentally DDoS-ing your monitoring
+
+Whether you use a pull-based or a push-based approach, any time-series database
+will fall over if you send it more samples than it can handle. However, in our
+experience it's slightly more likely for a push-based approach to accidentally
```
grobie: I haven't seen that it's harder with Prometheus to bring down monitoring. Why's that?

It's def. easier to recover from such scenarios with a central configuration place. As commented earlier already, I think this should only focus on the mitigation aspect.

grobie:
LGTM :+1:

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
beorn7: If you look at concrete implementations, you get implementations in the way. All the ways you can bring down a pull-based system are fundamentally possible and as likely for a push-based system. However, a push-based system has a failure case a pull-based system cannot have, namely a rogue client starting to send monitoring data.

This article is and should be about fundamental differences between push and pull, not about n details of existing implementations. Obviously, a well-done mature push-based system is harder to bring down than a makeshift pull-based one.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
beorn7: > I'd not rely on shared "obvious" knowledge in a discussion where you're trying to convince people with a different opinion. There should be an explanation for every argument so that it doesn't just become a claim which can be easily ignored.

Yes. My above comment was to suggest to add an explanation along the lines I have sketched in the comment.

> We're constantly discussing the need for monitoring targets which are unhealthy and don't serve production traffic (see Kubernetes sd pull requests) and at SoundCloud we don't have a single monitoring configuration which is identical to what is used in services for exactly the reason to not only monitor what's serving production traffic.

That's an implementation detail of our SD mechanism. Fundamentally, we are using the same mechanism to discover backends and to discover targets to monitor, even if we have modifiers on top to not server from unhealthy targets while still monitoring them. There is a huge difference to having completely separate paths for both.

> To be honest, I'd just remove that last sentence in this paragraph.

To be honest, I think it is _the_ most important argument in the whole push-vs.-pull debate. So all up for adding more explanation, but please don't remove the sentence.

grobie: I understand the purpose of this blog post to debunk statements about pull doesn't scale, but not to list general arguments for the pull based model.

Using service discovery without any kind of load balancer between services doesn't seem to be that common yet (unfortunately). One of the most common comments regarding the pull model I get is "I don't use any kind of service discovery yet and with the pull model I have to add this additional complexity". Putting the discussion about the benefits of sd aside, for setups without service discovery the sentence doesn't mean anything, if not actually the opposite of what you want to say.

This paragraph doesn't really contribute to the discussion about scalability concerns of the pull model. I'd much rather see a separate post about the benefits of service discovery and Prometheus' configuration strength than watering this post with an unrelated and not immediately clear argument.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
grobie: I'd remove "vague fears". I don't think it helps starting a discussion by classifying someone else's opinion as emotional fear.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
beorn7: There is a lot in the article that goes beyond the scalability question. I took it as the intention of the author to deliberately not restrict it to strictly scalability issues.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
juliusv: I have a vague fear that you were looking at the old revision of the text, where the initial comment was made on. Check the new one.

grobie: Yeah. Sorry. These new github links are confusing sometimes.

content/blog/2016-07-23-pull-does-not-scale-or-does-it.md
brian-brazil: Scalability includes operational scalability, as that's a common argument against pull.

juliusv:
Thanks everyone, that was more feedback from more people than I had expected :) I will get to all the points in detail in the next days.

fabxc:
Hey, can we publish this soon. I just had a meeting with an analyst where "pull doesn't scale by definition" came up again.

beorn7:
This is :+1: from my side. :ship:

juliusv:
Yep - I should finally find time to respond to the last comments and then
merge it later today or tomorrow :)

juliusv:
Addressed some comments and ignored some others :) Gonna post now...

PR #503, opened 2016-07-26
Clarify DNS service discovery relationship to DNS-SD

grobie:
It's a common misconception to call any DNS SRV record based discovery
approach "DNS-SD". The latter is a [ratified IETF standard][rfc6763]
which specifies a naming schema and lookup mechanism in order to provide
additional structured information. Prometheus only supports basic SRV (A, AAAA)
record lookups.

[rfc6763]: https://tools.ietf.org/html/rfc6763

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -309,18 +309,22 @@ feature to replace the special `__address__` label.
 
 ### `<dns_sd_config>`
 
-A DNS-SD configuration allows specifying a set of DNS record names which
-are periodically queried to discover a list of targets (host-port pairs). The
+A DNS-based service discovery configuration allows specifying a set of DNS
+domain names which are periodically queried to discover a list of targets. The
 DNS servers to be contacted are read from `/etc/resolv.conf`.
 
-During the [relabeling phase](#relabel_config), the meta
-label `__meta_dns_name` is available on each target and is set to the SRV
+This service discovery method only supports basic DNS A, AAAA and SRV record
+queries, but not the advanced DNS-SD approach specified in
+[RFC6763](https://tools.ietf.org/html/rfc6763).
+
+During the [relabeling phase](#relabel_config), the meta label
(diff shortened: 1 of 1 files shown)


grobie:
@fabxc @brian-brazil anyone?

content/docs/operating/configuration.md
```diff
@@ -309,18 +309,21 @@ feature to replace the special `__address__` label.
 
 ### `<dns_sd_config>`
 
-A DNS-SD configuration allows specifying a set of DNS record names which
-are periodically queried to discover a list of targets (host-port pairs). The
-DNS servers to be contacted are read from `/etc/resolv.conf`.
+A DNS configuration allows specifying a set of DNS domain names which are
```
juliusv: I'd still keep some reference to service discovery here, even if you don't want to mention DNS-SD verbatim. Like "DNS-based service discovery configuration"?

content/docs/operating/configuration.md
```diff
@@ -309,18 +309,21 @@ feature to replace the special `__address__` label.
...
+periodically queried to discover a list of targets. The DNS servers to be
+contacted are read from `/etc/resolv.conf`.
 
-During the [relabeling phase](#relabel_config), the meta
-label `__meta_dns_name` is available on each target and is set to the SRV
+This service discovery method only supports basic DNS queries, but not the
```
juliusv: Maybe "only supports basic DNS A-record and SRV-record queries" so that people don't incorrectly assume from this paragraph that SRV lookups aren't possible at all.

content/docs/operating/configuration.md
grobie: Updated.

content/docs/operating/configuration.md
grobie: Updated (and included AAAA).

juliusv:
:+1:

PR #505, opened 2016-07-30
Add grok_exporter

fstab:


brian-brazil:
Thanks!

I note that in your example the metrics names aren't quite following convention. Counters should end in _total, Gauges/Histograms/Summarys should have no suffix, and avoid mentioning things like by_user as that label could be aggregated away.

fstab:
Thanks, I fixed the example metric names in the documentation.

PR #513, opened 2016-08-11
website typos in opsgenie and example for alertmanager config

EvertMDC:


content/docs/alerting/configuration.md
```diff
@@ -376,7 +376,7 @@ channel: <tmpl_string>
 ```
 
 
-## OpsGenie receiver `<opsgenie_config>`
```
fabxc: This defines one instance of an OpsGenie config. Hence, the missing 's' is intentional.

EvertMDC: Does the instance refer to the code or to the configuration? Because without the 's' alertmanager sees it as a mistake.

Here's my working config:
```
receivers:
- name: 'opsgenie-guard-duty'
  opsgenie_configs:
  - api_key: <key>
```

alertmanager if I remove the 's':
> time="2016-08-11T10:11:38+02:00" level=error msg="Loading configuration file failed: unknown fields in receiver config: opsgenie_config" file="/config/alertmanager.yml" source="main.go:153"

fabxc: The receiver block is correctly specified here: https://prometheus.io/docs/alerting/configuration/#receiver-receiver

You see `<opsgenie_config>` in there, which is meant as a reference defined in this block.

EvertMDC: Ok I see, my bad.

PR #515, opened 2016-08-13
Update InfluxDB comparision

jackzampolin:


content/docs/introduction/comparison.md
```diff
@@ -76,43 +76,18 @@ geared towards slightly different use cases.
 The same scope differences as in the case of
 [Graphite](/docs/introduction/comparison/#prometheus-vs-graphite) apply here.
 
+Similar to the Graphite ecosystem, InfluxDB relies on separate open-source utilities for [data collection](https://github.com/influxdata/telegraf) and [alerting](https://github.com/influxdata/kapacitor).
```
grobie: That's also true for Prometheus.

content/docs/introduction/comparison.md
```diff
@@ -76,43 +76,18 @@ geared towards slightly different use cases.
...
-   * Storing time series with completely unbounded dimensionality.
-      * E.g. storing user IDs or email addresses in the key-value metadata
-        *vs.* storing bounded dimensionality like the HTTP method, HTTP handler
-        and instance ID.
+*Summary:* InfluxDB has a similar data model to Prometheus. Both support annotation of data with
+arbitrary key-value pairs (Prometheus's `labels` or InfluxDB's `tags`). In both systems these annotations 
+define a series, with the metadata stored only once. Storage density in InfluxDB is similar to Prometheus. Both 
+systems achieve compression of around ~2 bytes per float or integer value on disk.
+
+InfluxDB, unlike Prometheus, supports storing **all individual events**, not just a time series of values. By contrast, Prometheus stores just the cumulative count of HTTP requests differentiated by labels.
```
grobie: The second sentence is too specific. Prometheus can store any kind of data, not just HTTP requests counters.

content/docs/introduction/comparison.md
```diff
@@ -76,43 +76,18 @@ geared towards slightly different use cases.
...
-	  * [Logs and Metrics and Graphs, Oh My!](https://blog.raintank.io/logs-and-metrics-and-graphs-oh-my/)
-        describes the difference between event logging and metrics recording.
-   * Storing time series with completely unbounded dimensionality.
-      * E.g. storing user IDs or email addresses in the key-value metadata
-        *vs.* storing bounded dimensionality like the HTTP method, HTTP handler
-        and instance ID.
+*Summary:* InfluxDB has a similar data model to Prometheus. Both support annotation of data with
+arbitrary key-value pairs (Prometheus's `labels` or InfluxDB's `tags`). In both systems these annotations 
+define a series, with the metadata stored only once. Storage density in InfluxDB is similar to Prometheus. Both 
+systems achieve compression of around ~2 bytes per float or integer value on disk.
```
grobie: Prometheus can't store integers, so maybe either specify that it only applies to Influx or just say sample?

Prometheus is with 1.28 bytes/sample closer to one byte than two bytes. What's the value influxdb achieves here? I'm all up for removing the outdated 11x comparison, but given that's the Prometheus page we might want to keep stating the improvement.

grobie:
Thanks for updating the outdated comparison page! I guess we want to keep the 80 characters line wrap in the document.

content/docs/introduction/comparison.md
```diff
@@ -76,43 +76,18 @@ geared towards slightly different use cases.
...
-If you are only interested in tracking the development of existing named
-time series (for example, the cumulative count of HTTP requests with the method
-`POST` to the `/api/tracks` endpoint on the instance
-`http://1.2.3.4:12345/metrics`), Prometheus will require much less storage
-space than InfluxDB. Further, Prometheus indexes all time series dimensions for
-efficient filtering, while InfluxDB currently only indexes tables by row
-timestamps (issue to track adding column indexes:
-https://github.com/influxdb/influxdb/issues/582).
-
-Still, InfluxDB is better geared towards the following use cases:
```
brian-brazil: This section is still good

content/docs/introduction/comparison.md
```diff
@@ -65,7 +65,7 @@ also works well for many short-lived, frequently changing sets of time series.
 
 ## Prometheus vs. InfluxDB
 
-[InfluxDB](https://influxdata.com/) is a very promising new open-source time
+[InfluxDB](https://influxdata.com/) is an open-source time
```
brian-brazil: isn't it open-core now?

content/docs/introduction/comparison.md
fabxc: This sentence gives off a strong judging vibe via the "just", as if its for the worse when in fact it's a deliberate design decision with pros and cons to either side.

content/docs/introduction/comparison.md
fabxc: The sentence above was very clear I think. Seems mostly like a link plug.

content/docs/introduction/comparison.md
toddboom: Is the 1.28 bytes/sample fixed, or is it dependent on the shape of the data?

This thread might be out of date, but seems to imply that it's variable (as is InfluxDB):
https://groups.google.com/forum/#!topic/prometheus-developers/NhMJZLnqANg

juliusv: It's variable, as it depends on the shape of data you ingest, plus the configured chunk encoding. See: https://prometheus.io/docs/operating/storage/#chunk-encoding

juliusv: In general, I would probably just state that the two have comparable storage density (at least it shouldn't be a big factor in choosing one over the other). The exact numbers for each will depend on ingested data, and nobody did comparable benchmarks for both systems in recent years, as far as I'm aware.

brian-brazil:
The comparison doc was just updated.

PR #520, opened 2016-08-19
Update kubernetes link to use the new github org

alex:
The old one redirects, but better to be up to date I think.


sdurrheimer:
:+1:

PR #521, opened 2016-08-24
Add PHP client for Prometheus

samwierema:
For the internal service that we use (called [Pushprom](https://github.com/messagebird/pushprom)) we've created a PHP client. All components are open source, and we believe it could be of benefit to Prometheus users.

We've added differentiators between the two PHP clients as well.

### DIFF
--- a/content/docs/instrumenting/clientlibs.md
+++ b/content/docs/instrumenting/clientlibs.md
@@ -27,7 +27,8 @@ Unofficial third-party client libraries:
 * [Lua](https://github.com/knyar/nginx-lua-prometheus) for Nginx
 * [.NET / C#](https://github.com/andrasm/prometheus-net)
 * [Node.js](https://github.com/siimon/prom-client)
-* [PHP](https://github.com/Jimdo/prometheus_client_php)
+* [PHP](https://github.com/Jimdo/prometheus_client_php) via [Redis](https://github.com/antirez/redis)
+* [PHP](https://github.com/messagebird/pushprom-php-client) via [Pushprom](https://github.com/messagebird/pushprom)
 
 When Prometheus scrapes your instance's HTTP endpoint, the client library
 sends the current state of all tracked metrics to the server.


sdurrheimer:
LGTM, also we might want to mention the dogstatsd via statsd_exporter alternative one day.

brian-brazil:
We only list one client/exporter per system, so one of these should be removed.

beorn7:
@brian-brazil Do you prefer one of the two PHP solutions? They follow quite different approaches, so it might make sense to list both.

grobie:
What's the problem with listing both?

brian-brazil:
We've only ever listed one client/exporter per system. This means that our users don't have to try to evaluate them, which only a Prometheus developer can really do. Thus far it's been pretty obvious which of a selection was better.

samwierema:
First, we believe that both offerings can be of benefit to users and by providing the distinction users can choose for themselves what better fits their stack.

But, going with @brian-brazil's feedback, here's a case for our PHP offering:
* Pushprom is simpler: it is just a proxy to the Prometheus Go client, it's battle tested and core to Prometheus tools.
* No dependencies on external libraries (e.g. `ext-redis` for PHP).

marcelcorso:
maybe @bracki or @schnipseljagd have an opinion?

bracki:
IMHO the perfect library should work with a pure PHP installation without any additional dependencies.
We chose Redis only due to time constraints. We also believe that true pull metrics are crucial. We wanted to be able to peek into our applications via HTTP.
Maybe we should join forces on a pure PHP impl. I don't think this helps with this particular PR, though.
I'm impartial to what's being linked into the docs.
People should choose based on their requirements.

sdurrheimer:
I don't think we can really make a pure-php implementation that will work well because of the way PHP works.

IMO, the best options are dogstatsd + statsd_exporter OR pushprom-php-client + pushprom.
I'm not a big fan of adding redis in the loop, just to store state.

bracki:
FYI I've nearly completed work on `APCu` support [here(https://github.com/Jimdo/prometheus_client_php/pull/25).
This means we can store state just by enabling a common PHP extension, instead of requiring an extra binary (such as Redis or the Golang binary mentioned above).
If you got any further input please comment (esp. @marcelcorso @samwierema).

samwierema:
Just a ping to see what the status of this discussion is now.

brian-brazil:
This client doesn't appear to be notably better than the existing listed one, so this will stay as-is for now.

PR #524, opened 2016-08-29
Added NATS exporter to list of 3rd party exporters

bjflanne:
Lovoo (whose Jenkins exporter is listed) also have a NATS exporter which I have added


juliusv:
:+1: thanks!

PR #526, opened 2016-08-30
file_sd_config: Correct refresh_interval to 5m

mattbostock:
The actual default is 5 minutes:
https://github.com/prometheus/prometheus/blob/ee9abf4535d2a28feb14473716e9118b4e12741b/config/config.go#L106

This was changed in prometheus/prometheus@139f27bf8.

/cc @fabxc


brian-brazil:
Thanks!

PR #529, opened 2016-09-01
Add standard exports into the documentation

SimenB:
It seems weird to me that this list lives on an open Google Docs [document](https://docs.google.com/document/d/1Q0MXWdwp1mdXCzNRak6bW5LLVylVRXhdi7_21Sg15xQ/edit).

Even if it's work in progress, it should rather be marked as such instead of living somewhere else. This way, you have more control over it, and it makes it more official.
A benefit, other than the obvious standardizing, is discoverability. It took me quite some digging to find this, having Google index it might help.


beorn7:
I guess the reason why those are still in the Google doc is that they are not quite finalized yet.

But they should, obviously, so we could use this PR to do so.

I'll leave a few comments, but I guess @brian-brazil has the strongest opinions on this.

content/docs/instrumenting/writing_clientlibs.md
```diff
@@ -327,12 +327,7 @@ readable formats) if it can be implemented without a significant resource cost.
 ## Standard and runtime collectors
 
 Client libraries SHOULD offer what they can of the Standard exports, documented
-at
-[https://docs.google.com/document/d/1Q0MXWdwp1mdXCzNRak6bW5LLVylVRXhdi7_21Sg15xQ/edit](https://docs.google.com/document/d/1Q0MXWdwp1mdXCzNRak6bW5LLVylVRXhdi7_21Sg15xQ/edit)
-
-In addition, client libraries are ENCOURAGED to also offer whatever makes sense
-in terms of metrics for their language’s runtime (e.g. Garbage collection
-stats).
+at [standard exports](/docs/instrumenting/standard_exports).
```
beorn7: I'd vote for including it right here instead of having yet another section.

The client library guide is anyway long, so no harm in making it a bit longer.

content/docs/instrumenting/standard_exports.md
```diff
@@ -0,0 +1,35 @@
...
+This document covers what exports and collectors Prometheus client libraries
+should offer, with the aim of consistency across libraries, making the easy use
+cases easy and making it easier to provide tooling and consoles for common use
+cases.
+
+In addition, client libraries are ENCOURAGED to also offer whatever makes sense
+in terms of metrics for their language’s runtime (e.g. Garbage collection
+stats).
+
+These exports should have the prefix `process_`. If a language or runtime
```
beorn7: We need more clarity here. The unified process metrics are prefixed `process_` but the language specific metrics are prefixed `<language>_` or similar (e.g. `go_goroutines`, but note that JVM metrics are prefixed `hotspot_`).

content/docs/instrumenting/standard_exports.md
```diff
@@ -0,0 +1,35 @@
...
+doesn't expose one of the variables it'd just not export it. All memory values
+in bytes, all times in unixtime/seconds.
+
+## Should
+
+| Metric name                     | Description                                            | Format  |
```
beorn7: I'd replace `Description` by `Help string` and make the help string given here mandatory. (There is still some stalled work lying around to unify the help strings. We can as well do it right now. But that requires to check where we have already consensus, and where we have deviations.)

content/docs/instrumenting/standard_exports.md
```diff
@@ -0,0 +1,35 @@
...
+| `process_open_fds`              | Number of open file descriptors.                       | Integer |
+| `process_max_fds`               | Maximum number of open file descriptors.               | Integer |
+| `process_virtual_memory_bytes`  | Virtual memory size in bytes.                          | bytes   |
+| `process_resident_memory_bytes` | Resident memory size in bytes.                         | bytes   |
+| `process_heap_bytes`            | Process heap size in bytes.                            | bytes   |
+| `process_start_time_seconds`    | Start time of the process since unix epoch in seconds. | Seconds |
+
+## May
+
+TODO: IDK the names of GC ones...
```
beorn7: That's so language dependent that I wouldn't even try to unify it. Modulo the right prefix, see above.

content/docs/instrumenting/writing_clientlibs.md
```diff
@@ -327,12 +327,7 @@ readable formats) if it can be implemented without a significant resource cost.
 ## Standard and runtime collectors
 
 Client libraries SHOULD offer what they can of the Standard exports, documented
-at
-[https://docs.google.com/document/d/1Q0MXWdwp1mdXCzNRak6bW5LLVylVRXhdi7_21Sg15xQ/edit](https://docs.google.com/document/d/1Q0MXWdwp1mdXCzNRak6bW5LLVylVRXhdi7_21Sg15xQ/edit)
-
-In addition, client libraries are ENCOURAGED to also offer whatever makes sense
-in terms of metrics for their language’s runtime (e.g. Garbage collection
-stats).
```
grobie: What's the reason to remove this paragraph? We could even explicitly link to client_golang's go_* metrics/go collector.

beorn7: The reason is that it is also contained in the export description.

But see my suggestion to inline the whole standard export description right here.

grobie:
:+1: Thank you!

All our documents were on Google once (client library guide lines, exporter, exposition formats, etc.). We simply haven't ported this yet, but the presented process metrics were the ones we agreed upon end of 2014.

content/docs/instrumenting/writing_clientlibs.md
grobie: SGTM.

content/docs/instrumenting/writing_clientlibs.md
grobie: Github needs to get better with showing comments in real-time. Makes all sense.

content/docs/instrumenting/standard_exports.md
SimenB: Agreeing on a naming convention could still be nice though, or at least provide some sort of baseline for naming. Might even be enough to say "Java called it this, and Go called it that". Just provide some structure, to keep it more the same across implementations.

content/docs/instrumenting/writing_clientlibs.md
SimenB: This was mostly to separate it from clients.

If some things are standardized, it's also relevant for scrapers and consoles, not just for the ones writing clients.

Say I wanna write a generic template for metric visualizations, it'd be nice to know what some standard metrics are named, but I don't care about what conventions for other metrics are, I just want to be able to scrape some default ones.

content/docs/instrumenting/standard_exports.md
SimenB: I agree!
Ideas on how to formulate that?

content/docs/instrumenting/standard_exports.md
SimenB: Deffo! Then description can maybe be more about what it is, in case it's not clear (I had to google the different memory stuff for instance).

I copied the descriptions here from the Go implementation (first one I found, but java seems to have the same)

content/docs/instrumenting/writing_clientlibs.md
SimenB: Yeah, it's moved over to the new article

content/docs/instrumenting/standard_exports.md
brian-brazil: The wording here seems fine to me, as we're just talking about the standard exports - not runtime metrics.

content/docs/instrumenting/standard_exports.md
brian-brazil: Unit, not format - but it's already implicit from the name

content/docs/instrumenting/standard_exports.md
brian-brazil: This doesn't belong here as these aren't consistent across languages let alone runtimes.

content/docs/instrumenting/writing_clientlibs.md
brian-brazil: I'd inline here. The client library and exporter guides are already a bit off to have in the instrumentation section.

content/docs/instrumenting/standard_exports.md
grobie: The current wording above make it sound like standard exports include runtime metrics. It might make sense to improve the wording of the definition then.

SimenB:
Ok, updated.

Changes made (in separate commit) is:

* Inlined it back into the original file
* Changed the 2 headings as suggested

Did not really change any phrasing, as I think you probably can come up with better ones anyway 😄

content/docs/instrumenting/writing_clientlibs.md
```diff
@@ -327,17 +327,32 @@ readable formats) if it can be implemented without a significant resource cost.
...
+| `process_open_fds`              | Number of open file descriptors.                       | Integer |
+| `process_max_fds`               | Maximum number of open file descriptors.               | Integer |
+| `process_virtual_memory_bytes`  | Virtual memory size in bytes.                          | bytes   |
+| `process_resident_memory_bytes` | Resident memory size in bytes.                         | bytes   |
+| `process_heap_bytes`            | Process heap size in bytes.                            | bytes   |
+| `process_start_time_seconds`    | Start time of the process since unix epoch in seconds. | Seconds |
+
+In addition, client libraries are ENCOURAGED to also offer whatever makes sense
+in terms of metrics for their language’s runtime (e.g. Garbage collection
+stats).
```
brian-brazil: I'd append "with an appropriate prefix such as go_, hostspot_" to separate them from the standard exports.

SimenB:
@brian-brazil Done 😄

content/docs/instrumenting/writing_clientlibs.md
```diff
@@ -327,17 +327,32 @@ readable formats) if it can be implemented without a significant resource cost.
...
+These exports should have the prefix `process_`. If a language or runtime
+doesn't expose one of the variables it'd just not export it. All memory values
+in bytes, all times in unixtime/seconds.
+
+## Metrics
+
+| Metric name                     | Help string                                            | Unit    |
+| ------------------------------- | ------------------------------------------------------ | ------- |
+| `process_cpu_seconds_total`     | Total user and system CPU time spent in seconds.       | Seconds |
+| `process_open_fds`              | Number of open file descriptors.                       | Integer |
```
brian-brazil: The unit of open file descriptors is file descriptors.

SimenB:
@beorn7 @grobie @brian-brazil any other comments? It'd be great to land this so we can point to it 😄 
Or if stuff is basically set, just some tweaking necessary (as in, the current metrics are fine, none will be added/removed) that's fine as well.

content/docs/instrumenting/writing_clientlibs.md
```diff
@@ -327,17 +327,32 @@ readable formats) if it can be implemented without a significant resource cost.
...
+| `process_cpu_seconds_total`     | Total user and system CPU time spent in seconds.       | Seconds          |
+| `process_open_fds`              | Number of open file descriptors.                       | File Descriptors |
+| `process_max_fds`               | Maximum number of open file descriptors.               | Integer          |
+| `process_virtual_memory_bytes`  | Virtual memory size in bytes.                          | bytes            |
+| `process_resident_memory_bytes` | Resident memory size in bytes.                         | bytes            |
+| `process_heap_bytes`            | Process heap size in bytes.                            | bytes            |
+| `process_start_time_seconds`    | Start time of the process since unix epoch in seconds. | Seconds          |
+
+In addition, client libraries are ENCOURAGED to also offer whatever makes sense
+in terms of metrics for their language’s runtime (e.g. Garbage collection
```
beorn7: Garbage -> garbage

content/docs/instrumenting/writing_clientlibs.md
```diff
@@ -327,17 +327,32 @@ readable formats) if it can be implemented without a significant resource cost.
...
+doesn't expose one of the variables it'd just not export it. All memory values
+in bytes, all times in unixtime/seconds.
+
+## Metrics
+
+| Metric name                     | Help string                                            | Unit             |
+| ------------------------------- | ------------------------------------------------------ | ---------------  |
+| `process_cpu_seconds_total`     | Total user and system CPU time spent in seconds.       | Seconds          |
+| `process_open_fds`              | Number of open file descriptors.                       | File Descriptors |
+| `process_max_fds`               | Maximum number of open file descriptors.               | Integer          |
```
beorn7: Unit here should also be "File Descriptors".

On a more general note, I'd write all the units with lower case, not just the bytes.

beorn7:
Just nits left from my side.

We can iterate on larger additions or improvements later. For now, it's great to have this in the real docs.

SimenB:
@beorn7 sounds great! I'll fix it as soon as I get to my laptop (half hour). Want me to squash, or do you use GH's squash?

juliusv:
@SimenB Thanks! And that was my first time trying out GitHub's squash feature!

PR #530, opened 2016-09-01
Fix broken links

glung:


grobie:
Both links work perfectly well in our documentation page: https://prometheus.io/docs/querying/rules/ Please prefer reading the docs there and not the raw markdown files on github.

PR #532, opened 2016-09-01
List unofficial Elixir client

deadtrickster:
Link: https://github.com/deadtrickster/prometheus.ex

Thanks!

### DIFF
--- a/content/docs/instrumenting/clientlibs.md
+++ b/content/docs/instrumenting/clientlibs.md
@@ -22,6 +22,7 @@ Unofficial third-party client libraries:
 
 * [Bash](https://github.com/aecolley/client_bash)
 * [Common Lisp](https://github.com/deadtrickster/prometheus.cl)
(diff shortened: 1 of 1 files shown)


grobie:
Thanks!

brian-brazil:
Thanks, I'd suggest a read of the client library guidelines in particular the section around Children. Users should be passing around Counter objects, not specifying the name of the counter on every call.

https://prometheus.io/docs/instrumenting/writing_clientlibs/#overall-structure

deadtrickster:
@brian-brazil hey, glad you found a minute to take a look at the library. Erlang/Elixir are pure function (c) i.e. without any state/object/etc. You can store state in a process loop or so called Erlang Term Storage, basically awesomely cool hash-table/bag. I'm using ETS to store metrics, each object here is a tuple {Key, ....}
My keys are {Registry, Name, LabelValues} this is why they always there. I tried to overcome this with as many shortcuts as possible e.g.

```erlang
```

PR #539, opened 2016-09-08
Added Docker Hub exporter to list of third-party exporters

Rucknar:
Adding details of a simple Docker Hub exporter for tracking statistics about pulls/stars etc

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -48,6 +48,7 @@ hosted outside of the Prometheus GitHub organization.
    * [Ceph exporter](https://github.com/digitalocean/ceph_exporter)
    * [CouchDB exporter](https://github.com/gesellix/couchdb-exporter)
    * [Django exporter](https://github.com/korfuri/django-prometheus)
+   * [Docker Hub exporter](https://github.com/infinityworksltd/docker-hub-exporter)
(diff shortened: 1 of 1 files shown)


grobie:
Please have a look at https://prometheus.io/docs/practices/naming/#metric-names. Counters should end on `_total` and not `_count`.

Also have a look at https://prometheus.io/docs/instrumenting/writing_exporters/ including the section about port numbers.

Rucknar:
Both very valid points!

I've pushed an update to the repo to resolve the counter issue and to move the tracking of stars to be a gauge from a counter. With regards to the port number, i've updated it to a new port and listed that on the wiki.

Let me know if there is anything else i've missed.

juliusv:
@Rucknar Having the stars as a gauge is correct, but then please also omit the `_total` suffix (we only use that on counters, to distinguish them from gauges). Also, the convention is to use the plural of whatever you're counting / measuring:

* `docker_hub_pulls_total`
* `docker_hub_stars`

juliusv:
(you may also want to update the sample metrics section in the README.md)

juliusv:
Merging this for now...

Rucknar:
Thanks @juliusv 

I've made the recommended adjustments above and pushed.

PR #555, opened 2016-09-16
Remove examples of time metrics not using seconds

grobie:


SimenB:
Is there a rule for all time based stuff to be reported in seconds?

juliusv:
It's a general recommendation and one we try to implement in all the official repos at least. Unfortunately, there's still some remnants from olden times (like in the Go client library - `http_request_duration_microseconds`), but those should be fixed over time. The rationale is that things are easier to work with if everything is in base units. That doesn't only go for time, but also for bytes and other units. I don't think we've written that down in any official doc yet though.

brian-brazil:
It's in both the client library and exporter guidelines.

RichiH:
And will be part of the RFC as SHOULD.

Richard

Sent by mobile; excuse my brevity.

juliusv:
Right, that doesn't cover normal users of direct instrumentation though. I would have expected it here: https://prometheus.io/docs/practices/naming/

Will send a PR...

juliusv:
Here it is: https://github.com/prometheus/docs/pull/574

SimenB:
Great, thanks for the clarification!

PR #558, opened 2016-09-19
Added GitHub Exporter to documentation

Rucknar:
As per title, similar to the recent Docker Hub Exporter, likely to be extended with more metrics in the coming weeks. Any feedback appreciated, happy to amend if i've strayed from guidelines anywhere.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -49,6 +49,7 @@ hosted outside of the Prometheus GitHub organization.
    * [CouchDB exporter](https://github.com/gesellix/couchdb-exporter)
    * [Django exporter](https://github.com/korfuri/django-prometheus)
    * [Docker Hub exporter](https://github.com/infinityworksltd/docker-hub-exporter)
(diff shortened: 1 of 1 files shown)


RichiH:
I was suprised by this PR and then realized I never added a PR to add https://github.com/RichiH/github_exporter - I don't have to have mine in that list, but am wondering how to deal with this in the generic case. Add both? First come?

juliusv:
Yeah, hard to say. @RichiH's has more metrics, but has no full README.md and is written in Perl. @Rucknar's is written in Python (a bit more palatable nowadays), uses the official Python client library, has a README.md, but fewer metrics. Still the better base to build upon I would say.

juliusv:
@Rucknar I'll merge this one for now...

RichiH:
Fine by me; @Rucknar , can you add my metrics to yours?

Rucknar:
Shall do @RichiH , i'll definitely be adding in the timestamps and boolean's. Not sure i'll be keen to implement things such as repo description however as it feels like that might be a poor fit to the prometheus data structure.

PR #564, opened 2016-09-23
querying examples: fix proctype vs proc metric key

filippog:
Looks like `proctype` is mentioned only there, should be `proc` to match the examples.


brian-brazil:
Thanks!

PR #570, opened 2016-09-27
Added ProxySQL exporter to the list.

roman-vynar:
Added ProxySQL exporter to the list.
[ProxySQL](https://github.com/sysown/proxysql) is a high performance proxy, currently for MySQL and forks.


juliusv:
:+1: Thanks!

PR #571, opened 2016-09-27
Update AlertManager SMTP TLS configuration options

sgissi:
@juliusv  Update AlertManager configuration options related to [https://github.com/prometheus/alertmanager/pull/512]

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -67,7 +67,7 @@ global:
   [ smtp_auth_username: <string> ]
   [ smtp_auth_password: <string> ]
   [ smtp_auth_secret: <string> ]
-  # The default SMTP TLS requirement
+  # The default SMTP TLS requirement.
   [ smtp_require_tls: <bool> | default = true ]
 
   # The API URL to use for Slack notifications.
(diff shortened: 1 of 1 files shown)


content/docs/alerting/configuration.md
```diff
@@ -67,6 +67,8 @@ global:
   [ smtp_auth_username: <string> ]
   [ smtp_auth_password: <string> ]
   [ smtp_auth_secret: <string> ]
+  # The default SMTP TLS requirement
```
juliusv: Period at the end.

juliusv:
:+1: besides nit

sgissi:
Thanks, added the missing trailing period.

@fabxc @juliusv The docs are published only after a new release of AlertManager? Just to make sure users don't try to use a feature that only exists in master.

juliusv:
@sgissi Ah, good point! For Prometheus releases, we have the `next-release` doc branch, but I don't believe we have one for the Alertmanager yet. Maybe best to let this sit here then until @fabxc or @brancz bake the next Alertmanager release.

sgissi:
With the new AM released we should merge this. Should I mention in the comment line that it applies to AlertManager >= 0.5.0 to avoid luring the users of the packaged versions into using an invalid configuration?

brancz:
As far as I understand the typical process is to have the docs reflect the latest released stable version, but I think there are some mentions of previous versions of Prometheus as well. Wdyt about this @juliusv @fabxc @brian-brazil ?

fabxc:
Yea, we generally just refer to the latest stable version.
We hopefully figure out a general approach to versioned docs eventually. Mentioning exceptions across versions is a nightmare we don't want to get into.

PR #576, opened 2016-09-30
Mark EC2 SD as stable.

brian-brazil:
This has been around a good while, and there's no indications that
we'll be changing any aspects of it so let's make it stable.

@fabxc @juliusv


juliusv:
:+1:

PR #577, opened 2016-09-30
naming.md: Specify SI base units

RichiH:


content/docs/practices/naming.md
```diff
@@ -26,7 +26,7 @@ A metric name...
  * <code><b>http</b>\_request\_duration\_seconds</code>
    (for all HTTP requests)
 * ...must have a single unit (i.e. do not mix seconds with milliseconds, or seconds with bytes).
-* ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers).
+* ...should use SI base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers).
```
brian-brazil: The kg is a SI base unit and the byte is not covered by SI units so this isn't really accurate.

RichiH:
I was unsure if including the kg/g distinction would cause confusion, but you are right about bytes. Would you merge a PR which simply `s/meter/gram/`? That way it's embedded in the docs and avoids mentioning SI.

RichiH:
Reminder-poke. I can also just close if you prefer.

brian-brazil:
There's a pending comment

PR #578, opened 2016-10-02
[WIP] introduce categories to exporterlist

cherti:
As @juliusv mentioned once in #prometheus, the list of external exporters has grown quite a bit. This attempts to increase overview by introducing categories for the third party exporters.

The chosen categories are a first proposal and might still be subject to change, depending on the feedback, any of which is greatly appreciated!

PR #579, opened 2016-10-05
Add prombolt to third-party utilities

mdlayher:
Adds my [Bolt](https://github.com/boltdb/bolt) collector package to other third party utilities.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -139,6 +139,7 @@ practices.
    * Clojure: [prometheus-clj](https://github.com/soundcloud/prometheus-clj)
    * Go: [go-metrics instrumentation library](https://github.com/armon/go-metrics)
    * Go: [gokit](https://github.com/peterbourgon/gokit)
+   * Go: [prombolt](https://github.com/mdlayher/prombolt)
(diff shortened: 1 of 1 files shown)


juliusv:
Thanks! I just merged https://github.com/prometheus/docs/pull/578, which introduces categories to the exporter list. Could you rebase ontop of the latest master?

mdlayher:
@juliusv updated.  I think this still technically qualifies as a "third party utility" since it's a package you add to your applications to enable bolt instrumentation.  Let me know if you think otherwise!

content/docs/instrumenting/exporters.md
```diff
@@ -139,6 +139,7 @@ practices.
    * Clojure: [prometheus-clj](https://github.com/soundcloud/prometheus-clj)
    * Go: [go-metrics instrumentation library](https://github.com/armon/go-metrics)
    * Go: [gokit](https://github.com/peterbourgon/gokit)
+   * Go: [prombolt](https://github.com/mdlayher/prombolt)
```
grobie: This should go to the Databases section I believe, as it is an exporter for bolt?

brian-brazil:
Thanks!

content/docs/instrumenting/exporters.md
brian-brazil: It appears to be a library rather than an exporter.

Issue #584, opened 2016-10-07
Fix parsing of versions in download page generator


grobie:
Alertmanager started to use pre-release versions containing dots like `x.y.z-beta.0`. The additional dot in the pre-release part is not handled correctly.

https://github.com/prometheus/docs/blob/9fe5a55d5c98fce3494804f570297b9f00aaef0d/lib/helpers/download.rb#L103-L111

PR #585, opened 2016-10-07
Fix download page problems

grobie:
The download page currently doesn't parse correctly pre-release assets containing dots. A recent javascript (dependency?) also broke the dropdown.

@juliusv @beorn7


grobie:
Fixes #584

beorn7:
:+1: (but neither a Ruby expert nor a JS expert)

PR #587, opened 2016-10-08
Add GCE SD docs

dominikschulz:
This PR adds documentation for the GCE SD, including the change from https://github.com/prometheus/prometheus/pull/2063

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -180,6 +180,10 @@ ec2_sd_configs:
 file_sd_configs:
   [ - <file_sd_config> ... ]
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

Technically the status should go into next-release, but given we were lacking docs it's better to get the rest in.

dominikschulz:
Sorry, got that messed up.

juliusv:
This does not actually document the `gce_sd_config` section fields itself?

See https://prometheus.io/docs/operating/configuration/#gce_sd_config

dominikschulz:
Indeed, let me fix that.

PR #596, opened 2016-10-13
Add lmsensors exporter to exporters list

mdlayher:
I put this in the hardware section since I guess reading sensor data counts as "hardware"?  Feel free to suggest alternate categories.


brian-brazil:
Thanks for the PR. Hwmon support was recently added to the node_exporter, so it'd best best not to confusion by including this exporter in the list.

mdlayher:
Neat, I had no idea.  Will have to update my node_exporter and check that out.

PR #597, opened 2016-10-14
Fix typos.

filippog:
Also a couple of non-typos: its vs it's and publically vs publicly.

I'm not attached to publically vs publicly though the latter seems more common.


brian-brazil:
Thanks!

PR #599, opened 2016-10-16
Document remote write config.

### LINKED ISSUE #582: Document remote write relabeling

### LINKED ISSUE #583: Document remote write configuration

brian-brazil:
Fixes #582
Fixes #583

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -99,6 +99,3 @@ DEPENDENCIES
   rb-inotify
   redcarpet
   rspec
-
-BUNDLED WITH
-   1.11.2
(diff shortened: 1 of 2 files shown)


juliusv:
:+1: Since this configuration is only useful in combination with something Prometheus-external, there should be another document at some point (or even embed it here if it's short?) that explains the remote write payload format and how to make use of it. Maybe just don't close issue https://github.com/prometheus/docs/issues/583 completely yet?

brian-brazil:
As it stands I think the demo receiver covers it mostly. It's an experimental feature, I think digging through the code for the few users who want it is okay for now.

juliusv:
Hmm, maybe link to the demo app then at least?

juliusv:
:+1: Thanks

PR #603, opened 2016-10-18
Add prometheus-cpp to client libs

jupp0r:


brian-brazil:
This client has a number of notable violations of the guidelines (see the mailing list), there's better C++ clients out there so this one shouldn't be listed.

PR #610, opened 2016-10-25
Move prometheus-clj to client libraries

grobie:
@brian-brazil While a wrapper, it's still a client library and not an exporter, no?


brian-brazil:
It's in the correct place already. Users should not be using this instead of the Java client, as it's not following some of the guidelines.

grobie:
So it should be moved to client libraries, but only once it has been fixed. Which might not ever happen.

brian-brazil:
It should never be moved to client libraries, as it's not a client library.

PR #612, opened 2016-10-25
Add JustWatch as a Prometheus user

dominikschulz:
As suggested in https://github.com/prometheus/docs/pull/441#issuecomment-249054860 I'd like to add another proud Prometheus user to the frontpage.


brian-brazil:
Superseded by #705

PR #615, opened 2016-10-27
Added cloudmonitor exporter to list of exporters

rickardkarlsson:


content/docs/instrumenting/exporters.md
```diff
@@ -96,6 +96,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [BIND exporter](https://github.com/digitalocean/bind_exporter)
    * [Blackbox exporter](https://github.com/prometheus/blackbox_exporter) (**official**)
    * [BOSH exporter](https://github.com/cloudfoundry-community/bosh_exporter)
+   * [Cloudmonitor exporter](https://github.com/ExpressenAB/cloudmonitor_exporter)
```
brian-brazil: I think this would classify as a monitoring system.

rickardkarlsson:
@brian-brazil I updated your suggested changes.

Regarding the labels, it would be difficult to get some interesting metrics without having the current labels. I therefore wish to keep the labels as is.

brian-brazil:
> Regarding the labels, it would be difficult to get some interesting metrics without having the current labels. I therefore wish to keep the labels as is.

Be that as it may, labels aren't free. If you need that level of granularity and thus cardinality then logs are likely more appropriate.

PR #618, opened 2016-11-03
Fix formatting of k8s-SD config docs

juliusv:
The redcarpet markdown renderer doesn't interpret the newlines within
bulleted lists in the same way as GitHub, so the layout was broken.

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -518,69 +518,81 @@ Kubernetes SD configurations allow retrieving scrape targets from
 [Kubernetes'](http://kubernetes.io/) REST API and always staying synchronized with
 the cluster state.
 
-One of the following `role`s can be configured to discover targets:
+One of the following `role` types can be configured to discover targets:
 
-* **node**: Discovers one target per cluster node with the address defaulting
-  to the Kubelet's HTTP port.
-  The target address defaults to the first existing address of the Kubernetes
-  node object in the address type order of `NodeInternalIP`, `NodeExternalIP`, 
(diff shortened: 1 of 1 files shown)


juliusv:
@fabxc @brancz @brian-brazil

grobie:
It seems to be simpler to just use subheadings here, which would also give anchor links for free.

juliusv:
@grobie Changed it to subheadings. The only drawback is that it's not 100% obvious when the last subheading section ends (that is, when the roles section is over), because it's not indented.

fabxc:
👍

grobie:
Given the sections above end all on a list of bullet points, I think it's clear enough.

:+1:

content/docs/operating/configuration.md
```diff
@@ -518,69 +518,77 @@ Kubernetes SD configurations allow retrieving scrape targets from
...
-    from underlying pods), the following labels are attached:
-    * `__meta_kubernetes_endpoint_ready`: Set to `true` or `false` for the endpoint's ready state.
-    * `__meta_kubernetes_endpoint_port_name`: Name of the endpoint port.
-    * `__meta_kubernetes_endpoint_port_protocol`: Protocol of the endpoint port.
-  * If the endpoints belong to a service, all labels of the `role: service` discovery are attached.
-  * For all targets backed by a pod, all labels of the `role: pod` discovery are attached.
+One of the following `role` types can be configured to discover targets:
+
+#### `node`
+The `node` role discovers one target per cluster node with the address defaulting
```
grobie: My markdown linter always tells me to leave an empty line after a heading.

PR #619, opened 2016-11-04
Change `rpc_durations_seconds` -> `rpc_durations_microseconds`

iamjarvo:
The example go service seems to output `rpc_durations_microseconds`
instead of `rpc_durations_seconds`

I was following the getting started guide and I had to use microseconds instead. Here is the output of the example go service `rpc_durations_microseconds_count{service="uniform"} 2479`

### DIFF
--- a/content/docs/introduction/getting_started.md
+++ b/content/docs/introduction/getting_started.md
@@ -203,7 +203,7 @@ scrape_configs:
 
 Go to the expression browser and verify that Prometheus now has information
 about time series that these example endpoints expose, such as the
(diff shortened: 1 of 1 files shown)


beorn7:
Thanks for the PR. However, the mistake is in the example app. It should output seconds and not microseconds. This is https://github.com/prometheus/client_golang/issues/245 . Feel free to fix it over there in client_golang. It was on my plate anyway.

iamjarvo:
Thanks for the feedback, I'll fix it there

PR #623, opened 2016-11-16
Add script exporter to exporters list

jkassemi:


brian-brazil:
Hmm, I thought we already had one of these listed.

Similar to the blackbox exporter, the duration should be exposed as a gauge for just that scrape and the module passed in as a url parameter. The SNMP exporter shows how to do this correctly, the blackbox exporter hasn't been converted yet.

jkassemi:
@brian-brazil, ty. 

My use case involves checking the result of a curl request against several services with client side authentication for the purposes of verifying connectivity (arguably something I should work into the blackbox-exporter):

```
- name: service_a
   script: curl --key a --cert a.pem http://a.example.com/

- name: service_b
   script: curl --key b --cert b.pem http://b.example.com/
```

I'm interested in the results of all configured scripts, and by collecting them all it avoids defining a mapping to each one in the prometheus configuration. What's the overall rationale behind passing the module parameter? Is it something that I can support via a regular expression to maintain the current "collect all the things" functionality?

brian-brazil:
That'd be accepted in the blackbox exporter, if it's not already supported.

Exporters should leave scheduling and service discovery up to Prometheus. There's no exporter currently doing it that I know of, but there's nothing fundamentally wrong with asking for multiple modules at once.

PR #625, opened 2016-11-21
Added Gluster exporter in storage section

ofesseler:
I wrote an exporter for gluster fs, its still beta but in active development. 
Check it out at https://github.com/ofesseler/gluster_exporter


brian-brazil:
Thanks for sharing. This does not yet appear to be a usable exporter, so it'd be a bit premature to use it.

I'd suggest collecting data at each scrape and using ConstMetrics rather than direct instrumentation. The code would benefit from a `go fmt` too.

ofesseler:
Thanks for looking at my exporter. Indeed it is premature...

In the latest "release" I added ConstMetrics and  used go fmt... nice tool... . 

I'll update this thread if i think its more appropriate to use it.

brian-brazil:
That's looking usable now.

You should remove the `node` label, Prometheus already has that in the `instance` label.

PR #626, opened 2016-11-21
faq.md: 1.0.0 has been released ;)

RichiH:


content/docs/introduction/faq.md
```diff
@@ -40,14 +40,10 @@ Python, and Ruby.
...
-Prometheus server reaches version 1.0.0. These breaking changes will be pointed
-out in release announcements for components that already have a proper release
-process (like the Prometheus server) or communicated clearly otherwise. After
-releasing version 1.0.0, breaking changes will be indicated by increments of
-the major version. See also the documentation for [semantic
-versioning](http://semver.org/), which we are following.
+Within any given major version, we will only introduce fully backwards
+compatible changes. Any breaking changes will be indicated by increments
+of the major version. See also the documentation for
+[semantic versioning](http://semver.org/), which we are following.
```
RichiH: Valid point; given that this is just removing 1.0.0, I am not sure if keeping an up to date list there would make sense. Referring to a different place that lists them might make sense.

beorn7: The only public and official source I'm aware of is https://prometheus.io/blog/2016/07/18/prometheus-1-0-released/#fine-print
Obviously, this will not be updated, but perhaps we can point to it from here for now until we have a place to document stability guarantees in a more robust way.

brian-brazil:
@RichiH Could you update per @beorn7's comment?

Per discussion last week, we're not actually following semver to the letter so that wording should be weakened.

beorn7:
This is now sup.rseeded by https://github.com/prometheus/docs/pull/702/

PR #635, opened 2016-12-06
Add note on hiring technical writers

fabxc:
We can add the PromCon one back once we found someone. But this seems rather important now.

@alexsomesan @brancz

### DIFF
--- a/content/index.html
+++ b/content/index.html
@@ -13,7 +13,7 @@ layout: jumbotron
 </div>
 
 <div class="newsbar">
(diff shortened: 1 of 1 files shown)


brancz:
Can't think of something better, then "get paid for it" (which sounds like an m-turks style micro payment) because I'm unsure whether "We're hiring" is ok from legal perspective. But if there are no objections I'm fine with this 👍 . Maybe we can ask the CNCF what wording they would prefer.

RichiH:
"We are looking for freelance workers interested working on this"?

Adding a rough indication of compensation levels might help, or not.


Richard

fabxc:
Updated

brancz:
lgtm 👍

alexsomesan:
👍 nice. fingers crossed now.

RichiH:
+1

PR #637, opened 2016-12-07
Correcting bucket for 95th percentile

shredder12:
Please correct me if I am wrong but with most requests at 220ms, the true histogram should return 200-300ms bucket for 95th percentile.

cc @juliusv


brian-brazil:
Thanks!

PR #639, opened 2016-12-08
add rust-prometheus to exporters.md

brunoqc:


brian-brazil:
This is already listed as an instrumentation client.

brunoqc:
Sorry about that.

PR #640, opened 2016-12-11
Add AWS ECS exporter to the list of exporters on docs

slok:
This AWS exporter takes metrics from ECS API


brian-brazil:
Thanks!

That's certainly one of the best tested exporters I've seen. Watch that you only use the _total suffix on counters, not gauges.

slok:
You are right! 

I've updated the exporter fixing those gauges. Thank you!

PR #643, opened 2017-01-06
Add nginx-vts-exporter

hnlq715:
@juliusv @mpalmer @alex

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -63,7 +63,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Apache exporter](https://github.com/neezgee/apache_exporter)
    * [HAProxy exporter](https://github.com/prometheus/haproxy_exporter) (**official**)
    * [Nginx metric library](https://github.com/knyar/nginx-lua-prometheus)
-   * [Nginx vts exporter](https://github.com/hnlq715/nginx-vts-exporter)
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/exporters.md
```diff
@@ -63,6 +63,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Apache exporter](https://github.com/neezgee/apache_exporter)
    * [HAProxy exporter](https://github.com/prometheus/haproxy_exporter) (**official**)
    * [Nginx metric library](https://github.com/knyar/nginx-lua-prometheus)
+   * [Nginx vts exporter](https://github.com/hnlq715/nginx-vts-exporter)
```
beorn7: Capitalize acronyms: VTS

beorn7:
Thanks.

Note that you are not following https://prometheus.io/docs/instrumenting/writing_exporters/#collectors , i.e. you should use `NewConstMetric` to create metrics on the fly rather then setting and resetting them all in each scrape cycle. This will also allow you to use counters (rather than gauges) where appropriate.

(We can still link the exporter from the Prometheus site, so no need to do it now. But please consider it for later.)

hnlq715:
@beorn7 Thanks. I'll consider the point you figured out.

hnlq715:
I've refactored nginx-vts-exporter with NewConstMetric to create metrics 😄

PR #648, opened 2017-01-14
Fix for overlapping text in feature item headings.

joegaffey:
Only happens in certain window sizes.


juliusv:
Today I learned about `text-overflow: ellipsis;`. Thanks :)

joegaffey:
No prob. 😁

PR #651, opened 2017-01-17
Overview exporters: remove Ganglia, add Graphite

toriagibbs:
There is no Ganglia exporter listed in the [exporters doc](instrumenting/exporters.md) and support for gmond has been deprecated from the Node exporter, so I think we should remove Ganglia from the list here. I replaced with Graphite, but open to leaving it blank or picking some other example.

(cc @juliusv for real this time)


juliusv:
:+1: Thanks! Graphite is fine to list, as would be any number of other examples, but this should do it.

PR #652, opened 2017-01-19
Add Triton SD docs

richardkiene:
This commit adds documentation for the recently added Triton SD as requested by @fabxc in https://github.com/prometheus/prometheus/pull/2307.

PR #653, opened 2017-01-19
Add Triton SD docs

richardkiene:
This commit adds documentation for the recently added Triton SD as requested by @fabxc in prometheus/prometheus#2307.

@brian-brazil I closed https://github.com/prometheus/docs/pull/652 and applied your review comments to this.

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -200,6 +200,10 @@ nerve_sd_configs:
 serverset_sd_configs:
   [ - <serverset_sd_config> ... ]
 
+# List of Triton service discovery configurations.
+triton_sd_configs:
+  [ - <triton_sd_config> ... ]
+
 # List of labeled statically configured targets for this job.
 static_configs:
   [ - <static_config> ... ]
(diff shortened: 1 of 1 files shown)


content/docs/operating/configuration.md
```diff
@@ -703,6 +707,54 @@ paths:
...
+
+```
+# The information to access the Triton discovery API
+
+# The account to use for discovering new target containers
+account: <string>
+
+# The dns suffix which should be applied to target containers. This is often the
+# same value as endpoint
+dns_suffix: [ <string> ]
```
brian-brazil: This is not optional

content/docs/operating/configuration.md
```diff
@@ -703,6 +707,54 @@ paths:
...
+
+# The Triton discovery endpoint (e.g. 'cmon.us-east-3b.triton.zone')
+endpoint: [ <string> ]
```
brian-brazil: Nor this

content/docs/operating/configuration.md
```diff
@@ -703,6 +707,54 @@ paths:
...
+
+See below for the configuration options for Triton discovery:
+
+```
+# The information to access the Triton discovery API
+
+# The account to use for discovering new target containers
+account: <string>
+
+# The dns suffix which should be applied to target containers. This is often the
```
brian-brazil: It seems a bit weird to me that this needs to be configured. Shouldn't this be determined automatically?

content/docs/operating/configuration.md
```diff
@@ -703,6 +707,54 @@ paths:
...
 
+### `<triton_sd_config>`
+CAUTION: Triton SD is in beta: breaking changes to configuration are still
+likely in future releases.
+
+[Triton](https://github.com/joyent/triton) SD configurations allow retrieving
+scrape targets from [Container Monitor](https://github.com/joyent/rfd/blob/master/rfd/0027/README.md)
+discovery endpoints.
+
+Prometheus will periodically check the discovery endpoint(s) for running
```
brian-brazil: No need for this line.

content/docs/operating/configuration.md
```diff
@@ -703,6 +707,54 @@ paths:
...
+
+The following meta labels are available on targets during relabeling:
+
+* `__meta_triton_machine_id`: the UUID of the target container
+* `__meta_triton_machine_alias`: the alias of the target container
+* `__meta_triton_machine_image`: the target containers image type
+* `__meta_triton_machine_server_id`: the server UUID for the target container
+
+
+See below for the configuration options for Triton discovery:
```
brian-brazil: Not needed

richardkiene: This appears to be the header for every other SD configuration options section, so it might appear somewhat inconsistent if it is removed. Am I missing something that makes the Triton section different?

content/docs/operating/configuration.md
richardkiene: I'll admit that this setting does leak a bit of configuration detail from a separate system inside of Triton. However, I chose to expose this setting over adding additional complexity. In the future it is entirely possible that this could be automatically discovered.

content/docs/operating/configuration.md
richardkiene: Okay

content/docs/operating/configuration.md
richardkiene: Good catch, thanks!

content/docs/operating/configuration.md
richardkiene: Good catch, thanks!

content/docs/operating/configuration.md
brian-brazil: It's inconsistent by the looks of it.

content/docs/operating/configuration.md
```diff
@@ -703,6 +707,51 @@ paths:
...
+
+```
+# The information to access the Triton discovery API
+
+# The account to use for discovering new target containers
+account: <string>
+
+# The dns suffix which should be applied to target containers. This is often the
+# same value as endpoint
```
brian-brazil: I'd mention these the other way around so you're not referencing endpoint before it's mentioned.

richardkiene: No problem, will do.

content/docs/operating/configuration.md
richardkiene: Okay, I'll remove it.

brian-brazil:
Thanks!

PR #660, opened 2017-02-16
Add pingdom and kemp exporters to exporters list

JosephSalisbury:
We've made a few exporters @giantswarm - i thought i'd add them <3


brian-brazil:
Thanks!

I've a few suggestions for improvement. You should use ConstMetrics rather than direct instrumentation. Data should be fetched per-scrape. Paused should not be a label in pingdom, that's a boolean metric. Similarly resolution is a gauge metric.

JosephSalisbury:
@brian-brazil cheers for the suggestions, always good to get fresh eyes - i'll get on improving them as soon as

thanks! <3

PR #663, opened 2017-02-18
Updated roadmap per Promcon discussion.

brian-brazil:
This is per the notes from Promcon.
I decided to leave in the client library/exporter section,
but altered to be more hand wavy.

### DIFF
--- a/content/docs/introduction/roadmap.md
+++ b/content/docs/introduction/roadmap.md
@@ -11,13 +11,20 @@ features and current work, see the issue trackers for the various repositories,
 for example, the [Prometheus
 server](https://github.com/prometheus/prometheus/issues).
 
+### New storage engine
+
+A new more efficient storage engine is in development. It will offer reduced
+resource usage, better inverted indexes, and generally allow Prometheus to
+scale further.
+
 ### Long-term storage
 
 Currently Prometheus has support for storing samples on local disk, as well as
-experimental support for writing data into OpenTSDB and InfluxDB. We plan to
-improve long-term storage support, although the details are not determined yet.
-This might include read-back support from OpenTSDB/InfluxDB in Prometheus, or
-support for other long-term storage backends.
+experimental support for sending data into remote systems via a generic mechanism.
+Example receivers exist for a [variety of TSDBs](https://github.com/prometheus/prometheus/tree/master/documentation/examples/remote_storage).
+
+We plan to add read-back support from other TSDBs via a generic mechanism in Prometheus,
+such as [Cortex](https://github.com/weaveworks/cortex).
 
 GitHub issue: [#10](https://github.com/prometheus/prometheus/issues/10)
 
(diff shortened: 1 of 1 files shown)


content/docs/introduction/roadmap.md
```diff
@@ -11,13 +11,20 @@ features and current work, see the issue trackers for the various repositories,
 for example, the [Prometheus
 server](https://github.com/prometheus/prometheus/issues).
 
+### New storage engine
+
+A new more efficient storage engine is in development, offering reduced
```
SuperQ: Slightly long sentance.  Mabye `A new, more efficient, storage engine is in development.  It will offer reduced...`

brian-brazil: Updated

content/docs/introduction/roadmap.md
```diff
@@ -38,16 +45,22 @@ GitHub issue: [#398](https://github.com/prometheus/prometheus/issues/398)
...
 client libraries and in the exposition format, but not persisted or
 utilized in the Prometheus server. We plan on making use of this
-metadata in the future. For example, we could suggest automatic rates
-over counters, warn users if they take the rate of a gauge, or display
-metric documentation strings. Some metric types, like the upcoming
-[server-side histograms](https://github.com/prometheus/prometheus/issues/480),
-could also be stored and processed in a more efficient way.  The
-details of this are still to be determined.
+metadata in the future. The first step is to aggregate this data in-memory
+in Prometheus and provide it via an experimantal API endpoint.
```
beorn7: As I explicitly stated in a recent discussion, my understanding was that complex sample types to store histograms more efficiently are not off the table. It is also a problem that is relevant in various way for a typical major Prometheus user, as I can witness it in my day-to-day work.

My objection is completely ignored by this PR.

content/docs/introduction/roadmap.md
```diff
@@ -11,13 +11,20 @@ features and current work, see the issue trackers for the various repositories,
...
+scale further.
+
 ### Long-term storage
 
 Currently Prometheus has support for storing samples on local disk, as well as
-experimental support for writing data into OpenTSDB and InfluxDB. We plan to
-improve long-term storage support, although the details are not determined yet.
-This might include read-back support from OpenTSDB/InfluxDB in Prometheus, or
-support for other long-term storage backends.
+experimental support for sending data into remote systems via a generic mechanism.
```
beorn7: As discussed earlier, the term "generic" is confusing here. It is meant as "not specific to the remote system to send to", but it can easily be understood as "sending a generic representation of the sample data" (while it in fact sends the current internal representation, which is subject to change).

content/docs/introduction/roadmap.md
```diff
@@ -11,13 +11,20 @@ features and current work, see the issue trackers for the various repositories,
...
+Example receivers exist for a [variety of TSDBs](https://github.com/prometheus/prometheus/tree/master/documentation/examples/remote_storage).
+
+We plan to add read-back support from other TSDBs via a generic mechanism in Prometheus,
```
beorn7: See above about "generic".

content/docs/introduction/roadmap.md
beorn7: Typo: experimantal → experimental

beorn7:
@brian-brazil Please see my comments above.

content/docs/introduction/roadmap.md
brian-brazil: I can't really see how that'd be misinterpreted that way as it's talking about the mechanism, and that it's experimental covers that we're likely going to be changing things anyway.

content/docs/introduction/roadmap.md
brian-brazil: This is per notes from Promcon, as we forgot to update this doc at that time with the outcome of the discussion. The only thing we agreed on (and even that was tentative) was to aggregate the data, expose it via an API; once that's done we'll see how it's used and go from there.

I'll send a PR for the typo.

beorn7: We didn't agree on _removing_ the item from the roadmap. It's now on the agenda of the next dev summit. It would send the wrong signal to remove the item right now.

content/docs/introduction/roadmap.md
beorn7: We agreed that it is confusing enough so that we purged the word "generic" from the code base.

It's not overly confusing here, but I'd still avoid the term as it has many connotations.

(But to put things into perspective: This is a nit compared to removing the single one most needed feature from the roadmap without consulting the other core devs.)

PR #664, opened 2017-02-21
Fix for clipped descenders in feature headings

joegaffey:


brian-brazil:
I'm not seeing clipped descenders on the front page. What browser and resolution are you on?

juliusv:
Right, I see this problem too on Chrome on Linux, any window size: one or two pixels of the `g`s in the item headlines are cut off. Not sure why it doesn't look right by default, but this PR fixes it for me.

PR #667, opened 2017-02-23
Community: Add section to discourage 1:1 support queries

beorn7:
I find myself quite often in the situation where I have to tell people
what I'm adding here. Even if not all will read it, it is then easy to
point people to it.

@brian-brazil @fabxc @juliusv Does this make sense?

### DIFF
--- a/content/community.html
+++ b/content/community.html
@@ -27,6 +27,14 @@ title: Community
     <p>
       <strong>Issue tracker:</strong> We use the GitHub issue tracker for the various <a href="http://github.com/prometheus">Prometheus repositories</a>
     </p>
+    <p>
+      <em>Please do not ask individual project members for
+      support. Use the channels above instead, where the whole
+      community can help you and benefit from the solutions
+      provided. If community support is insufficient for your
+      situation, please refer to the Commercial Support section
+      below.</em>
+    </p>
 
     <h1>Contributing</h1>
     <p>


brian-brazil:
:+1: 

I try to convert them into customers, but that's not for everyone :)


Do we also want to add something about not using the issue tracker for support requests? That has been becoming more common, distracting from actual issues.

beorn7:
Yes, I thought so, too. Issue tracker for bugs, mailing list for questions.

content/community.html
```diff
@@ -27,6 +27,14 @@
     <p>
       <strong>Issue tracker:</strong> We use the GitHub issue tracker for the various <a href="http://github.com/prometheus">Prometheus repositories</a>
     </p>
+    <p>
+      <em>Please refrain from soliciting individual project members
```
matthiasr: Maybe use use less fancy language here? "do not ask"?

content/community.html
```diff
@@ -27,6 +27,14 @@
...
+      for support. Use the channels above instead, where you will find
+      more potential helpers, and the whole community can benefit from
```
matthiasr: "where the whole community can help you and benefit from the solutions provided"

beorn7: I like it.

content/community.html
beorn7: Makes sense.

beorn7:
I'll work in MR's suggestions, but wait for @fabxc and/or @juliusv to get quorum.

juliusv:
:+1: to everything, including MR's comments.

PR #674, opened 2017-03-06
Add justwatchcom/sql_exporter

dominikschulz:
This commit adds the SQL exporter to the list of database exporters.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -37,6 +37,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Redis exporter](https://github.com/oliver006/redis_exporter)
    * [RethinkDB exporter](https://github.com/oliver006/rethinkdb_exporter)
    * [SQL query result set metrics exporter](https://github.com/chop-dbhi/prometheus-sql)
+   * [SQL exporter](https://github.com/justwatchcom/sql_exporter)
    * [Tarantool metric library](https://github.com/tarantool/prometheus)
 
 ### Hardware related


brian-brazil:
To keep things sane we only allow one of a given type of integration, and there's already one of these here. Are there notable features of this exporter over the existing one?

dominikschulz:
Yes, there are. I'm preparing a small blog post highligning the key differences. I'll update the PR once it's published.

dominikschulz:
We tried to explain our motivation and reasoning in this blog post: https://www.justwatch.com/blog/post/announcing-prometheus-sql-exporter/

TL;DR: The existing SQL exporter(s) weren't flexible enough and/or (slightly) broken. We tried to create a slim and clean exporter that includes the most important configurations for e.g. Kubernetes, Prometheus (Alerts, Recording Rules) and Grafana. Something I'd like to see for every exporter out there.

If this doesn't convince you to include this project in the docs feel free to close this PR.

grobie:
Looks good to me.

beorn7:
At a first glance, it looks good to me, too.

@dominikschulz let's way a day or two for objections, but otherwise, I'd say remove the other SQL exporter in this PR and leave yours in.

brian-brazil:
The justwatch exporter only does postgres, whereas the existing one does anything there's support for. The existing one also keeps SD within Prometheus, so I'd be overall for sticking with the current one.

beorn7:
Then it sounds the use cases are sufficiently different so that we could just have both in the list, perhaps with a little explanation of the different areas of usage.

brian-brazil:
As far as I can see, the existing exporter can do everything the justwatch one does. There's no major advantage that I can see that justifies breaking the rule of one of one type of integration.

beorn7:
@dominikschulz seems to disagree and wrote a whole blog post to make his point. A more detailed explanation why there is no advantage in their exporter would be welcome.

brian-brazil:
The blog post merely states how it works, it doesn't compare other exporter.

As it happens, the prostgresql_exporter also shares this functionality.

dominikschulz:
The main point of the post was to show what the exporter does. We also tried to explain why we saw the need for another SQL exporter without being to picky about the existing exporters. We're very glad they exist and we certainly learned some things from these projects. Unfortunately they didn't feel like something we'd like to use in our day by day work.

IMHO many existing exporters are rather hard to use because they don't provide enough information (recording rules, alerts, dashboards, ...) to get you started. We tried to do better. I don't know if we succeeded, but this was our humble goal. If you think we did, you should include this exporter in the docs.

The question should be rather about being helpful to users looking for help, not so much about rules.

brian-brazil:
We appreciate you contributing to the community and sharing what works with you.

We need to consider the broader context, if a new user comes and looks at this list and finds 5-6 different exporters for the same functionality they are unlikely to be very happy about having to evaluate them. Particularly as they're new users and will not be aware of the subtleties around exporter design.

Thus far we've been looking at exporter quality, featureset and popularity; and encouraging where possible contribution to existing projects.

Given this exporter is a subset of two different already listed exporters in terms of general functionality, and doesn't appear from a quick look to have features that'd justify supplanting either of them I don't think it'd make sense to list as-is. If for example it grew to do general postgres stats, handled multiple types of databases, or rectified some notable lack then it'd be in the running to replace one of the existing entries.

beorn7:
Perhaps @SuperQ or @discordianfish have an opinion here?

Just from a very high level perspective, if a reasonably involved Prometheus user like Justwatch felt the two listed exporters weren't good enough, there seems to be added value to the users to have their alternative, too. I don't have the capacity, though, to do the research for a more qualified line of argument.

Perhaps @SuperQ or @discordianfish have an opinion here?

I'd just feel much better if we rejected the listing based on more than one strong "no". Otherwise, it smells like https://github.com/prometheus/docs/pull/266 , where a contribution was rejected because of a single strong "no" by Brian, accompanied by many weak "yes" votes, of which nobody had the energy to make their point more solidly. The contributor got frustrated and we lost a potentially better value to users.

discordianfish:
I actually thought about this list a few times already and concluded that right now we don't have a way to vet that properly, so I'd keep the more liberal approach (as we IMO did in the past). So yes, I would accept this.

On the long run it would be great to have some 'registry' where people can add their own exporters and crowdsource vetting of them, similar to grafana.net.

brian-brazil:
> I actually thought about this list a few times already and concluded that right now we don't have a way to vet that properly, so I'd keep the more liberal approach 

I read through every new exporter and client library that wants to be added to this list.

> (as we IMO did in the past)

We never did this.

discordianfish:
It's arguably at least as good as other SQL exporters, overlap or not. That's reason enough for my to accept this. So I vote for merging this.

SuperQ:
There is a bunch of overlap, but I think this one may be worth listing, or replacing the existing one.

The main reason in my opinion is that the current SQL exporter requires a separate SQL service, where this one has no external dependencies.  This is a big plus.

brian-brazil:
> The main reason in my opinion is that the current SQL exporter requires a separate SQL service, where this one has no external dependencies. This is a big plus.

If you want that, use the postgres_exporter.

SuperQ:
That is not the same thing.

It looks like the SQL exporter mentioned here supports MySQL now.

juliusv:
:+1: from my side too. I count 5x :+1: vs. 1x :-1:. Will merge unless there's a new argument surfacing today.

brian-brazil:
> Will merge unless there's a new argument surfacing today.

That still leaves the question of which one we remove. The existing sql exporter supports more types of sql databases, and better follows the exporter guidelines.

I'd want to see an argument that this new exporter is significantly better than what's already there to justify usurping it. Such an argument has not been presented.

juliusv:
I'm for having both, due to their different tradeoffs, as mentioned e.g. by @SuperQ above. The majority of others seem to agree.

brian-brazil:
By the standards that have been applied up to this point, this exporter doesn't meet the bar. And I don't see that this case is anywhere near special enough that we should change the rules.


Thus far we've had the rule that it's one listing for a given type of exporter/client library. We have ignored implementation preferences and tradeoffs, rather trying to encourage the community to work on one exporter for a given use that follows the best practices.

By letting in this exporter we're saying that it's a) okay to leave it up to users to evaluate exporters and b) that exporters that violate best practices should have the same standing as exporters that follow them, which makes a) even harder for users.

This question should be looked at in from a high level, not just in the context of one particular exporter. What's being proposed here is to list for example all 5+ elasticsearch exporters that are out there, which I don't think is at all a win for the community.

juliusv:
I'm for changing our rules then. I'd be for adding exporters when:

- they don't blatantly violate Prometheus best practices (by our judgement) and are not really crappy in other ways
- they have sufficiently relevant tradeoffs such that one is better for some users, and the other better for others

I'm not saying we should add exporters that we think are bad in terms of best practices overall. Do you think the exporter being added here is too violating of best practices?

> What's being proposed here is to list for example all 5+ elasticsearch exporters that are out there, which I don't think is at all a win for the community.

I have not studied each of these in detail, but the question is whether that many of them pass the criteria defined above.

brian-brazil:
> they don't blatantly violate Prometheus best practices (by our judgement) and are not really crappy in other ways

The exporter in question violates best practices compared to what's already listed, in that it doesn't leave service discovery up to Prometheus.

If an exporter came along that followed best practices better than what's there (e.g. a new redis exporter that followed the 1:1 rule, as the rest was improved) with no new notable downsides then I'd replace it.

Conversely if an exporter comes along with new best practice violations, it'd have to have major other benefits to counteract that and be the one we list.

> they have sufficiently relevant tradeoffs such that one is better for some users, and the other better for others

I have a much higher standard here than you I think.

To me this particular exporter is a minor variant on what's already listed. Everyone is going to claim their tradeoffs are so important that their way is the only way, from an objective standpoint that's very rarely the case.

beorn7:
> Everyone is going to claim their tradeoffs are so important that their way is the only way, from an objective standpoint that's very rarely the case.

In this case, it was not only the author of the exporter but also @SuperQ , who expressed his opinion the Justwatch exporter has "a big plus".

@SuperQ , in view of the discussion above, do you still think the Justwatch exporter has benefits that justify listing it despite its shortcomings it terms of best practices?

If yes, my suggestion is to list it _in addition_ to the existing ones, with a very short explanation when to use which (e.g. "doesn't require a separate SQL service").

juliusv:
@beorn7 agreed, I think @SuperQ's point of not having to run yet another separate process is a relevant distinction, and not made by the author of the exporter. There's also mention in the blog post that existing exporters didn't perform to spec, which would be interesting to get more info about.

To address the other point:

> The exporter in question violates best practices compared to what's already listed, in that it doesn't leave service discovery up to Prometheus.

That is a valid point. I'd recommend changing that in this exporter, though I personally would not see that as a blocker for listing it. I wonder if the motivation for specifying the targets in the exporter config file was to allow persistent connections to the DB (I'm seeing the `keepalive` option), but I guess that should also be possible in a dynamic way (like it's generally done for HTTP + keepalive).

brian-brazil:
> There's also mention in the blog post that existing exporters didn't perform to spec, which would be interesting to get more info about.

I'd in general like to encourage improvement of existing exporters over fracturing of the ecosystem. Having to run two processes seems unnecessary for example, so an embedded option by default would seem like an idea.

> I personally would not see that as a blocker for listing it

We've others doing this right now (e.g. redis), however there's only one game in town when it comes to redis exporters.

> I wonder if the motivation for specifying the targets in the exporter config file was to allow persistent connections to the DB (I'm seeing the keepalive option), but I guess that should also be possible in a dynamic way (like it's generally done for HTTP + keepalive).

That seems orthogonal to me. Sharing DB connections between scrapes would have to be done with a lot of care in any case.

brian-brazil:
I've just added a reference to https://github.com/prometheus/prometheus/wiki/Default-port-allocations from this page. As we're not going to list all exporters in existence here, I'd suggest grabbing a port number (9237 is currently next) and adding yourself there.

SuperQ:
@brian-brazil Please do not close issues without coming to a consensus.

dominikschulz:
@brian-brazil Did what you suggested. Thanks

Sorry if this led to more discussions than I intended.

If I would have been aware of a rule on the number of exporters in the docs I would have discussed this elsewhere before opening an PR here.

Feel free to close this PR if you like.

We'll try to address your critique and make this the best SQL exporter around.

brian-brazil:
> We'll try to address your critique and make this the best SQL exporter around.

That'd be great. The existing listed one is far from perfect user-wise, so there's certainly scope for an exporter that follows the exporter best practices and is easier for users.

SuperQ:
@dominikschulz The discussion has been useful, so don't feel bad about it.

Both this and the currently listed exporter are not really exporters, but report generators.  This makes them slightly outside the norm for Prometheus.

brian-brazil:
Obsoleted by #928

dominikschulz:
@brian-brazil I've justed noticed #928 and I still wonder why you prefer https://github.com/free/sql_exporter over this one? Is it the more idiomatic on-scrape data collection or were there other issues?

brian-brazil:
> Is it the more idiomatic on-scrape data collection or were there other issues?

That is in the exporter guidelines, and from memory the other exporter was quite full featured compared to the various other candidates.

dominikschulz:
I don't see the difference, yet. But I'll review the other exporter and the guidelines and try to improve ours acordingly. Thanks for the feedback anyway!

PR #677, opened 2017-03-10
Add Transmission exporter

metalmatze:
I've created a transmission exporter that exports every metric from their RPC endpoint.
Just open up a Issue or PR on the project itself if there's anything you want to change. Thanks!


brian-brazil:
Thanks!

I don't think the labels on the free space metric make sense. Files_total is a gauge so shouldn't have the total_suffix.I'd also be very careful with the per-torrent metrics, they're likely to have high churn and thus cause issues.

PR #678, opened 2017-03-12
Document custom query timeouts.

juliusv:


juliusv:
@Gouthamve @fabxc

content/docs/querying/api.md
```diff
@@ -71,6 +71,7 @@ URL query parameters:
 
 - `query=<string>`: Prometheus expression query string.
 - `time=<rfc3339 | unix_timestamp>`: Evaluation timestamp. Optional.
+- `timeout=<duration>`: Evaluation timeout. Optional. Defaults to value of `-query.timeout` flag.
```
fabxc: Defaults and also maxes to that one.

juliusv: Ooohhhh... good point. Hmm. Is that actually the behavior we want? I can imagine most users wanting to set a *higher* than default timeout for certain queries that take a long time, but which they still really want to get an answer to.

brian-brazil: As an admin of a Prometheus server, I'd want this as a hard limit to avoid queries of sloth.

juliusv: @brian-brazil I think I'd want a default of 1 or 2 minutes, then per-query configurable timeouts, but which are still capped at something reasonable. So that'd require two flags (default query limit, absolute max limit).

juliusv: I mean, when does it happen that someone actually wants to set a *lower* than default timeout for a query?

brian-brazil: If they know the result is of no use to them if it's not fast.

fabxc:
I think what we have here is a good solution addressing a good chunk of the
problem.
We can discuss other rules for a second iteration later.

juliusv:
Ok, I updated the description.

I really like the ability to set a custom timeout, I'm just wondering if someone really has a use case for setting it lower, not higher, than the default.

content/docs/querying/api.md
grobie: :+1 to default and maximum.

PR #685, opened 2017-03-24
Add DigitalOcean exporter

metalmatze:
Prometheus exporter for various metrics about your DigitalOcean droplets, volumes, snapshots & networks and much more, written in Go.


brian-brazil:
Thanks!

Both of your status metrics are a bit weird. Usually you either use 1/0 to indicate the account is active OR have a 1 on the current status as a label. A mix of both is going to be difficult to deal with.

metalmatze:
Thanks for that valuable feedback. I will look into that, just like I will look into the transmission-exporter comments you gave me. :+1:

metalmatze:
I changed `digitalocean_account_status{status="active"} 1` to `digitalocean_account_active 1`. If there was another metric, didn't find know you were talking about, please let me know. Thanks! :)

PR #686, opened 2017-03-25
Installation: mention Puppet module

mika:
The puppet prometheus module works fine with the current version of prometheus.


RichiH:
LGTM.

Richard

default.

juliusv:
:+1: Thanks!

PR #687, opened 2017-03-26
Removing FlowDock reference

leecalcote:
FlowDock integration is not currently supported in .5 version of AM.

See https://github.com/prometheus/alertmanager/issues/236


brian-brazil:
Thanks!

PR #689, opened 2017-03-27
Please add MQTT blackbox exporter to list

frittentheke:
This exporter, written in Go, tests MQTT messaging round trips (publish/subscribe on same topic).


brian-brazil:
Thanks!

Counters should end in _total by convention, and don't forget to mention units such as seconds in your latency metrics.

PR #691, opened 2017-04-02
Add remote_read docs and update remote_write ones

juliusv:
Next step: add example code for remote read, point more precisely to
examples and specs for each.

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -81,16 +81,11 @@ alerting:
 
 # Settings related to the experimental remote write feature.
 remote_write:
-  [ url: <string> ]
-  [ remote_timeout: <duration> | default = 30s ]
-  tls_config:
-    [ <tls_config> ]
-  [ proxy_url: <string> ]
-  basic_auth:
-    [ username: <string> ]
-    [ password: <string> ]
-  write_relabel_configs:
-    [ - <relabel_config> ... ]
+  [ - <remote_write> ... ]
+
+# Settings related to the experimental remote read feature.
+remote_read:
+  [ - <remote_read> ... ]
 ```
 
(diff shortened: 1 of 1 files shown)


content/docs/operating/configuration.md
```diff
@@ -995,16 +990,77 @@ Where `<scheme>` may be `http` or `https` and `<path>` is a valid URL path.
...
+url: <string>
+
+# Timeout for requests to the remote write endpoint.
+[ remote_timeout: <duration> | default = 30s ]
+
+# List of remote write relabel configurations.
+write_relabel_configs:
+  [ - <relabel_config> ... ]
+
+# Sets the `Authorization` header on every remote write request with the
```
brian-brazil: Is there a better option to copy&pasting these everywhere?

juliusv: I was wondering this too, but the problem is that these are embedded fields in the config, and we'd have to make up another notation to describe that, making our config notation even more complex and confusing. So maybe just copy&paste is the least bad.

brian-brazil: As long as we can keep it updated everywhere, as this is going to end up in more files than this (AM, blackbox and consul will have them too).

brian-brazil:
:+1:

content/docs/operating/configuration.md
juliusv: We can also consider making them non-embedded at some point, maybe for 2.0? Not sure if that is worth it though.

grobie:
We could of course throw in even more ruby and generate our docs from some sets of definitions and embed them where needed. It'd make the source a lot harder to read though.

juliusv:
@grobie Yeah, that way only we would have to deal with a special construct at least, not the user...

PR #698, opened 2017-04-06
We already call these write receivers, not adapters.

brian-brazil:
@juliusv Forgot we'd already named these.

### DIFF
--- a/content/docs/introduction/glossary.md
+++ b/content/docs/introduction/glossary.md
@@ -90,15 +90,15 @@ A remote read endpoint is what Prometheus talks to when doing a remote read.
 Remote write is a Prometheus feature that allows sending ingested samples on the
 fly to other systems, such as long term storage.
 
-### Remote Write Adapter
-
-Not all systems directly support remote write. A remote write adapter sits
-between Prometheus and another system, converting the samples in the remote
-write into a format the other system can understand.
-
 ### Remote Write Endpoint
 
 A remote write endpoint is what Prometheus talks to when doing a remote write.
+
+### Remote Write Receiver
+
+Not all systems directly support remote write. A remote write receiver sits
+between Prometheus and another system, converting the samples it receives in
+the remote write into a format the other system can understand.
 
 ### Silence
 


juliusv:
What would we call a process that does full read+write then? Wouldn't it be better to call it all an "adapter", and thus also call the write path an "adapter" (or part of it), to be less confusing?

brian-brazil:
Thus far we've managed to come up with a different name for every type of data path. I'd worry if we call both "adapter" then we can't tell if someone says e.g. "mysql adapter" whether they're reading or writing.

juliusv:
Hm ok - what do we call a process that does both then? Still adapter? Read-write adapter?

brian-brazil:
I'm not sure to be honest, we're kinda running out of names. "receiving adapter" would be the first one that jumps to mind.

juliusv:
"receiving adapter" sounds like it only receives. How about adapter for both, but just call it "Read adapter", "Write adapter", or "Read/write adapter"? Keeps it simple... it's possible that people won't always use the prefix, but if we make the rules too complicated, they won't name it right anyway :)

brian-brazil:
"receiver" is already in use by a few implementations, as the name has been out there for a while, so probably best not to change it.

Could do with "full adapter" or similar. I imagine the main support load will be with things on the read rather than the write side.

juliusv:
> "receiver" is already in use by a few implementations

I was trying to google this, but only found our demo write receiver, which we can rename. Or are there others? At the same time, that made me notice that we already have the term "receiver" in Alertmanager as a different concept, so that can actually cause confusion.

So that's why I think the "adapter" with read/write qualifiers might still be best?

RichiH:
Loaning from networking, what about receiver, transmitter, and bidi (bi-directional)?

brian-brazil:
> Or are there others?

I found two I think (and I know of one other that you won't find publically), and I've had a blog post up for a good while now too.

> At the same time, that made me notice that we already have the term "receiver" in Alertmanager as a different concept, so that can actually cause confusion.

I recently realised that, there's not much overlap so hopefully that'll work out.

> transmitter, and bidi (bi-directional)?

A read adapter is bi-directional.

juliusv:
> I found two I think (and I know of one other that you won't find publically), and I've had a blog post up for a good while now too.

I don't think such few existing cases should block us from choosing the best possible name for the many years to come.

@RichiH receiver/transmitter has the problem that a read-only adapter usually both receives *and* transmits (to the backend). And "transmitter" doesn't sound like something you can query.

I'm yet to be convinced that we have any term better than "read|write|read/write" adapter"... also, this PR currently wants to define a "Remote Write Receiver" to be a bridge/adapter between Prometheus and a final remote system, while in common English, "receiver" could also mean a system which directly supports remote writes (like Cortex). Something that receives remote writes, that is...

brian-brazil:
> while in common English, "receiver" could also mean a system which directly supports remote writes (like Cortex). Something that receives remote writes, that is...

Yeah, that was my original thinking. We now have endpoint for that too.

PR #705, opened 2017-04-10
Added new company logos

Conorbro:
Added the 4 new logos mentioned in #441  and #612 to prometheus.io homepage.

Any objections @beorn7 @RichiH @dominikschulz ?

screen shot 2017-04-10 at 14 51 12

PR #707, opened 2017-04-11
Add Transloadit logo

kvz:
This was introduced in https://github.com/prometheus/docs/pull/441 already, but it that got superseded by https://github.com/prometheus/docs/pull/705, which featured an improved version of my flex grid code, but no longer our logo.

Re-adding here based on the new work, as we'd still love to be featured as happy users : )


brian-brazil:
Thanks!

Issue #438, opened 2016-05-19
Telegram bot


hryamzik:
I've finally published my [telergam reporting bot](https://github.com/inCaller/prometheus_bot), please feel free to point it in alert manager docs if you find it useful.

vtolstov:
thanks!

brian-brazil:
Please send a PR if you'd like it added.

PR #708, opened 2017-04-11
Added telegram bot info

hryamzik:
Following #438.


brian-brazil:
These are the configuration docs, so this doesn't belong here. I'd suggest adding a new page for these.

brian-brazil:
Closing in favour of #810 which includes this.

PR #714, opened 2017-04-14
Added JavaMelody Exporter

fschlag:


brian-brazil:
Thanks!

You should really use [Custom Collectors](https://github.com/prometheus/client_java#custom-collectors) for everything rather than implementing your own registry variant and trying to make direct instrumentation work for something it's not designed for.

fschlag:
Hi @brian-brazil, thanks for the hint and merge. Didn't know about Custom Collectors yet. I'll try to fix this in the future.

PR #716, opened 2017-04-15
Render tarball sha256 checksum if available for release

grobie:
@juliusv @sdurrheimer @brian-brazil 

Requires:
* merging of prometheus/promu#64
* generation of sha256sums.txt for all repositories
* rebuild of the docs page

screenshot from 2017-04-15 01-45-11

### DIFF
--- a/content/css/docs.css
+++ b/content/css/docs.css
@@ -233,8 +233,9 @@ footer {
 }
 
 .downloads .checksum {
-  color: #aaa;
-  font-style: italic;
+  font-family: monospace;
+  font-size: .6em;
+  vertical-align: bottom;
 }
 
 /* Docs-related styles. */
--- a/content/download.html
+++ b/content/download.html
@@ -29,11 +29,6 @@ title: Download
       the <a href="/docs/introduction/getting_started/">installation instructions</a>.
     </p>
 
-    <div class="alert alert-info" role="alert">
-      <strong>Work in progress!</strong>
-      We will provide more precompiled binary versions as well as checksums soon.
-    </div>
-
     <div class="panel panel-default download-selection">
       <div class="panel-body">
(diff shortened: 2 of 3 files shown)


RichiH:
If we touch that anyway, what about going full paranoia and using two hash
families?

SHA2-512 and SHA3-512 would be the obvious choices.


Richard

default.

juliusv:
:+1:

grobie:
There are a couple of reasons I'm not in favor of this idea. First a few
minor ones: It was already hard to squeeze squeeze the 64 characters of the
hex checksum in the table, sha512 would create real layout issues. When it
comes to sha3, there wasn't a sha3 checksum generator installed on any of
the 4 Linux boxes I checked. Package managers like homebrew mostly support
sha256 but nothing else.

My main concern is the message we'd send by offering such long/strong hash
sums. Our build process is really not that secure, users have to trust four
companies (Docker, Quay / CoreOS, CircleCI and Github) and the integrity of
Docker images itself we use for building.

I don't think stronger checksums add any additional security, but rather
give a false sense of security (I'm already worried about).

On Sat, Apr 15, 2017 at 5:18 AM Richard Hartmann < notifications@github.com
> wrote:

> If we touch that anyway, what about going full paranoia and using two hash
>
?
>

> SHA2-512 and SHA3-512 would be the obvious choices.
>
>
> Richard
>
> Sent by mobile; Excuse my brevity and the wall of text Gmail appends by
>
default.

sdurrheimer:
Wouldn't be better to keep `.zip` support ? I'm currently thinking it might be better to switch windows releases files to zip files given that it is natively supported in contrary of tarballs.

grobie:
Done.

sdurrheimer:
Thank you !

PR #717, opened 2017-04-18
Add Tinyproxy exporter

igzivkov:
Metrics for https://tinyproxy.github.io/ from its internal status page.


brian-brazil:
Thanks!

By convention counters should end in _total. connections_bad_total would also be more typical than bad_connections_total as that keeps all the connection-related metrics together.

igzivkov:
Thanks, I'll fix those as soon as possible.

PR #718, opened 2017-04-18
Add Lustre exporter

joehandzik:
Signed-Off-By: Joe Handzik <joseph.t.handzik@hpe.com>


brian-brazil:
This seems to all be stats from /proc. Is this something that works with a stock kernel such that it'd make sense to put in with the Node exporter, or are there other stats in the future that won't be from /proc?

You may wish to double-check your imports in `lustre_exporter.go`, and reading through https://prometheus.io/docs/instrumenting/writing_exporters/#naming

joehandzik:
@brian-brazil I had considered putting it all in the node_exporter, but we wanted to leave ourselves the option of including non-proc stats (which we are planning to do at some point) and we felt the Lustre use case was niche enough to justify a separate exporter regardless. 

Good catch on the imports, thanks (converted the repo owner today). That should all be fixed now. We're aware that we're not compliant with the metric names yet, but we'll figure out the right way to bring our naming inline when we start tagging releases.

PR #723, opened 2017-04-26
Added pgbouncer_exporter

larseen:
Ive added a new pgbouncer exporter, not sure how it should be added given that there already is an pgbouncer_exporter in the documentation. However the one I wrote is somewhat simpler to deploy and exports more stats.


brian-brazil:
We only list one of each type of exporter, and usually only replace an existing one if there's a notably better replacement.

Have you talked to the author of the existing exporter to see if their can take on your improvements?

I'd also suggest a read of https://prometheus.io/docs/instrumenting/writing_exporters/ in particular around naming

mdlayher:
@brian-brazil should this be closed?

PR #724, opened 2017-04-27
Added VictorOps configuration to docs

Conorbro:


content/docs/alerting/configuration.md
```diff
@@ -408,6 +408,30 @@ api_key: <string>
...
 ```
+## `<victor_ops_config>`
+
+VictorOps notifications are sent out via the [VictorOps API](https://portal.victorops.com/public/api-docs.html#/)
+
+```
+# The API key to use when talking to the VictorOps API.
+api_key: <string>
+
+# The VictorOps webhook URL.
```
brian-brazil: This seems more likely to be an API URL.

content/docs/alerting/configuration.md
```diff
@@ -408,6 +408,30 @@ api_key: <string>
...
+[ api_url: <string> ]
+
+# A key used to map the alert to a team.
+[ routing_key: <string> ]
+
+# Describes the behavior of the alert.
```
brian-brazil: You should mention the possible values, keeping in mind the munging we do.

content/docs/alerting/configuration.md
```diff
@@ -408,6 +408,30 @@ api_key: <string>
 # Comma separated list of tags attached to the notifications.
 [ tags: <tmpl_string> ]
 ```
+## `<victor_ops_config>`
+
+VictorOps notifications are sent out via the [VictorOps API](https://portal.victorops.com/public/api-docs.html#/)
```
brian-brazil: That's not the API we're using: https://github.com/prometheus/alertmanager/blob/master/config/config.go#L284

content/docs/alerting/configuration.md
```diff
@@ -408,6 +408,30 @@ api_key: <string>
...
+
+```
+# The API key to use when talking to the VictorOps API.
+api_key: <string>
+
+# The VictorOps API URL.
+[ api_url: <string> ]
```
brian-brazil: This is defaulted to the global setting

PR #727, opened 2017-05-01
Fix broken link in `alerting/rules.md` to `querying/rules.md`

math4tots:
There was a typo in `content/docs/alerting/rules.md` resulting in a broken link. I think this is probably what you guys meant?

@brian-brazil


brian-brazil:
The current link works for me

grobie:
Please note that the markdown links are interpreted in the context of the generated website, they don't work in Github. The link on https://prometheus.io/docs/alerting/rules/ is correct.

PR #734, opened 2017-05-08
Note extrapolation done in the rate, delta, and increase

cubranic:
As suggested in prometheus/prometheus#2683.

### DIFF
--- a/content/docs/querying/functions.md
+++ b/content/docs/querying/functions.md
@@ -89,8 +89,9 @@ month for each of the given times in UTC. Returned values are from 28 to 31.
 `delta(v range-vector)` calculates the difference between the
 first and last value of each time series element in a range vector `v`,
 returning an instant vector with the given deltas and equivalent labels.
-The delta is interpolated to cover the full time range, so that it is 
-possible to get non-integer result even if the samples are all integers.
+The delta is extrapolated to cover the full time range as specified in
+the range vector selector, so that it is possible to get a non-integer
+result even if the sample values are all integers.
 
 The following example expression returns the difference in CPU temperature
 between now and 2 hours ago:
@@ -204,9 +205,10 @@ equivalent labels.
 `increase(v range-vector)` calculates the increase in the
 time series in the range vector. Breaks in monotonicity (such as counter
 resets due to target restarts) are automatically adjusted for. The
-increment is interpolated to cover the full time range, so that it is
-possible to get non-integer result even if counter increments only by
-integers.
+increase is extrapolated to cover the full time range as specified
+in the range vector selector, so that it is possible to get a
+non-integer result even if a counter increases only by integer
+increments.
 
 The following example expression returns the number of HTTP requests as measured
 over the last 5 minutes, per time series in the range vector:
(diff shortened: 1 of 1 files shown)


content/docs/querying/functions.md
```diff
@@ -89,7 +89,8 @@ month for each of the given times in UTC. Returned values are from 28 to 31.
 `delta(v range-vector)` calculates the difference between the
 first and last value of each time series element in a range vector `v`,
 returning an instant vector with the given deltas and equivalent labels.
-The delta is interpolated to cover the full time range.
+The delta is interpolated to cover the full time range, so that it is
```
juliusv: interpolated -> extrapolated

full time range -> full time range as specified in the range vector selector

content/docs/querying/functions.md
```diff
@@ -89,7 +89,8 @@ month for each of the given times in UTC. Returned values are from 28 to 31.
 `delta(v range-vector)` calculates the difference between the
 first and last value of each time series element in a range vector `v`,
 returning an instant vector with the given deltas and equivalent labels.
-The delta is interpolated to cover the full time range.
+The delta is interpolated to cover the full time range, so that it is 
+possible to get non-integer result even if the samples are all integers.
```
juliusv: to get -> to get a

samples -> sample values

content/docs/querying/functions.md
```diff
@@ -202,7 +203,10 @@ equivalent labels.
 
 `increase(v range-vector)` calculates the increase in the
 time series in the range vector. Breaks in monotonicity (such as counter
-resets due to target restarts) are automatically adjusted for.
+resets due to target restarts) are automatically adjusted for. The
+increment is interpolated to cover the full time range, so that it is
```
juliusv: increment -> increase

Same comments as above apply.

content/docs/querying/functions.md
```diff
@@ -202,7 +203,10 @@ equivalent labels.
...
+possible to get non-integer result even if counter increments only by
```
juliusv: if counter increments only by integers -> if a counter increases only by integer increments

content/docs/querying/functions.md
```diff
@@ -212,8 +216,9 @@ increase(http_requests_total{job="api-server"}[5m])
 ```
 
 `increase` should only be used with counters. It should be used primarily for
-human readability. Use `rate` in recording rules so that increases are tracked
-consistently on a per-second basis.
+human readability and is really just a syntactic sugar around `rate` times the
```
juliusv: Avoid words like "really" in reference docs. Better:

"It is syntactic sugar for `rate(v)` multiplied by the number of seconds under the specified range vector time window. It should be used primarily for human readability."

content/docs/querying/functions.md
```diff
@@ -299,7 +304,9 @@ regression](http://en.wikipedia.org/wiki/Simple_linear_regression).
 
 `rate(v range-vector)` calculates the per-second average rate of increase of the
 time series in the range vector. Breaks in monotonicity (such as counter
-resets due to target restarts) are automatically adjusted for.
+resets due to target restarts) are automatically adjusted for. Also, the
+calculation extrapolates to the ends of the time range, allowing for missed
+scrapes or imperfect alignment of scrape cycle with the range's time period.
```
juliusv: scrape cycle -> scrape cycles

juliusv:
Thank you, left some comments!

cubranic:
OK, incorporated.

juliusv:
:+1: Thanks!

PR #741, opened 2017-05-12
Add Tincho to the list of consultants

TheTincho:
I was told that it was OK to directly sending a PR for this. Please, let me know if it is not.


grobie:
:+1:

p.s. @juliusv You should get https for your page, it's triggering my OCD.

juliusv:
@grobie I'm all https-ed up since a while now - I just need to update the link here :)

TheTincho:
Thanks!

PR #744, opened 2017-05-15
Update writing_exporters.md

tback:
Add note on choosing port numbers for external applications. 

Discussion on prometheus-users is in https://groups.google.com/d/topic/prometheus-users/S1ZzJMkR1tk/discussion

I don't care much about the exact wording. I only care about sparing future developers unnecessary thoughts.

### DIFF
--- a/content/docs/instrumenting/writing_exporters.md
+++ b/content/docs/instrumenting/writing_exporters.md
@@ -461,6 +461,9 @@ putting your username and WIP is fine.
 This is a registry to make our users’ lives a little easier, not a commitment
 to develop particular exporters.
 
+For exporters on internal applications we recommend using ports outside of the
+range of default ports. Choosing ports above 9500 should be sufficient. 
+
 ## Announcing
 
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/writing_exporters.md
```diff
@@ -461,6 +461,9 @@ putting your username and WIP is fine.
 This is a registry to make our users’ lives a little easier, not a commitment
 to develop particular exporters.
 
+For exporters on internal applications we recommend using ports outside of the
```
brian-brazil: Users in such a situation will likely be following their internal port number policy, I think this text would only confuse them into thinking they needed to follow the Prometheus suggested ports rather than their own org's policy.

tback: ok, I still believe some kind of note would be helpful for people in my seat, but I don't care enough to push any further.

brian-brazil: The thing to do would be to soften the language above and make it clear the port list is just a suggestion.

PR #745, opened 2017-05-15
added Naemon exporter

Griesbacher:
This is an exporter for the Naemon Monitoringsystem. It is using the NEB Interface which is basically introduced by Nagios, so it should work with them, too. But there is still a strange bug which is currently preventing this.
Therefore I would only put Naemon in the title, because that is the system it is properly working with.


brian-brazil:
Based on the docs I'm not quite sure this is an exporter, how does it get its data?

Only counters should end in `_total`.

Griesbacher:
It's a module, which gets loaded by the core at the start. It uses callbacks and public variables to get its data. Similar to Livestatus or ModGearman, just is one is written in Go and therefore uses CGo.

So the countervector shouldn't use _total? I can change that.

SuperQ:
@Griesbacher `_total` should only be used for values that go up, or are reset to 0 (but not very often).  If the value goes up or down, it should not have `_total` and should be exported as a gauge.

Griesbacher:
Sorry, I've missed the gauge values but they are fixed.

brian-brazil:
Thanks!

PR #746, opened 2017-05-15
Update writing_exporters.md

tback:
Add note about choosing port numbers for exporters on internal
applications. This reflects that the port range is a mere suggestion.

Discussion on prometheus-users is in
https://groups.google.com/d/topic/prometheus-users/S1ZzJMkR1tk/discussion

Replaces https://github.com/prometheus/docs/pull/744

### DIFF
--- a/content/docs/instrumenting/writing_exporters.md
+++ b/content/docs/instrumenting/writing_exporters.md
@@ -459,7 +459,7 @@ preferably before publicly announcing it. If you’re not ready to release yet,
 putting your username and WIP is fine.
 
 This is a registry to make our users’ lives a little easier, not a commitment
-to develop particular exporters. For exporters on internal applications we
+to develop particular exporters. For exporters for internal applications we
 recommend using ports outside of the range of default port allocations.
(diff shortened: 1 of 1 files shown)


brian-brazil:
I suggested altering the above language to make it clearer that the port numbers are a suggestion, not adding a new paragraph that brings internal org polices into it.

brian-brazil:
You're still adding a paragraph, and we don't recommend any particular port range - ours or otherwise.

content/docs/instrumenting/writing_exporters.md
```diff
@@ -459,7 +459,8 @@ preferably before publicly announcing it. If you’re not ready to release yet,
 putting your username and WIP is fine.
 
 This is a registry to make our users’ lives a little easier, not a commitment
-to develop particular exporters.
+to develop particular exporters. For exporters on internal applications we
```
brian-brazil: s/on/for/

brian-brazil:
Thanks!

PR #749, opened 2017-05-17
Seperate out directly instrumented vs merely exposing.

brian-brazil:
Directly instrumented applications use a Prometheus client library that
follows our guidelines, and can be used as examples of software doing
that.

Software that uses something else to do instrumentation/metrics but
also exposes at one of our formats gets its own category.

Recategorise things accordingly, updating links as needed and
adding some more I know of.
I'm taking Quobyte's word for it that they're directly instrumented,
as it's commercial software and thus I can't check.
kubernetes-mesos has no instrumentation that I can find.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -120,6 +120,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [BIND exporter](https://github.com/digitalocean/bind_exporter)
    * [Blackbox exporter](https://github.com/prometheus/blackbox_exporter) (**official**)
    * [BOSH exporter](https://github.com/cloudfoundry-community/bosh_exporter)
+   * [cAdvisor](https://github.com/google/cadvisor)
    * [Dovecot exporter](https://github.com/kumina/dovecot_exporter)
    * [Jenkins exporter](https://github.com/lovoo/jenkins_exporter)
    * [Kemp LoadBalancer exporter](https://github.com/giantswarm/prometheus-kemp-exporter)
@@ -141,20 +142,24 @@ list](https://groups.google.com/forum/#!forum/prometheus-developers).  We are
 happy to give advice on how to make your exporter as useful and consistent as
 possible.
 
-## Directly instrumented software
+## Software exposing Prometheus metrics
 
-Some third-party software already exposes Prometheus metrics natively, so no
+Some third-party software exposes metrics in the Prometheus format, so no
 separate exporters are needed:
 
-   * [cAdvisor](https://github.com/google/cadvisor)
-   * [Doorman](https://github.com/youtube/doorman)
-   * [Etcd](https://github.com/coreos/etcd)
-   * [Kubernetes-Mesos](https://github.com/mesosphere/kubernetes-mesos)
-   * [Kubernetes](https://github.com/kubernetes/kubernetes)
+   * [Collectd](https://collectd.org/wiki/index.php/Plugin:Write_Prometheus)
+   * [Doorman](https://github.com/youtube/doorman) (**direct**)
+   * [Etcd](https://github.com/coreos/etcd) (**direct**)
+   * [Kubernetes](https://github.com/kubernetes/kubernetes) (**direct**)
+   * [Linkerd](https://github.com/BuoyantIO/linkerd)
+   * [Netdata](https://github.com/firehol/netdata)
+   * [Quobyte](https://www.quobyte.com/) (**direct**)
    * [RobustIRC](http://robustirc.net/)
-   * [Quobyte](https://www.quobyte.com/)
-   * [SkyDNS](https://github.com/skynetservices/skydns)
-   * [Weave Flux](http://weaveworks.github.io/flux/)
+   * [SkyDNS](https://github.com/skynetservices/skydns) (**direct**)
+   * [Telegraf](https://github.com/influxdata/telegraf/tree/master/plugins/outputs/prometheus_client)
+   * [Weave Flux](https://github.com/weaveworks/flux)
+
+The software marked *direct* is also directly instrumented with a Prometheus client library.
 
 ## Other third-party utilities
 


grobie:
What is the goal of this separation?

brian-brazil:
So we can distinguish which software is following the Prometheus way of doing things and we can point at as examples, versus things that happen to produce our format. The current docs conflate instrumentation and exposition.

grobie:
I don't see the primary use case of this page to see examples for best practices and find code examples, but to see which software of my stack can already be monitored. In that context it doesn't matter whether something uses a client library or some other way, as long as the exposition format interface is fulfilled. I'd rather remove the current separations than introducing more.

If you feel very strongly about pointing out that some of the listed items don't follow our best practices, we can also just add a label / icon to their name, similar to the `official` we already have.

brian-brazil:
To me this is a very core distinction. One is much likely to follow all of our best practices and be a good experience, the other is not.

I think it's important we clearly list which software that's directly instrumented.

content/docs/instrumenting/exporters.md
```diff
@@ -143,18 +144,26 @@ possible.
 
 ## Directly instrumented software
 
-Some third-party software already exposes Prometheus metrics natively, so no
-separate exporters are needed:
+Some third-party software directly uses a Prometheus client library and exposes
```
beorn7: Perhaps use a wording like the following to emphasize that this is more about instrumentation than exposition:
"Some third-party software is directly instrumented using a Prometheus client library and exposes..."

RichiH:
After some deliberation, I agree with @grobie that this seems to be
the wrong place to make this distinction.

If we are aware of bad implementations, instrumented, translated, or
exported, this should be addressed separately.

content/docs/instrumenting/exporters.md
brian-brazil: That's a good idea, updated.

brian-brazil:
To me the point of this list is to show off software that's following the Prometheus ideal, it got muddled though. Fixing this has been on my backlog for a while, as it blocked another PR.

brian-brazil:
> If we are aware of bad implementations, instrumented, translated, or
exported, this should be addressed separately.

How do you suggest addressing them? The only lever we have would be to remove them from this page, which would then be a very bare page.

RichiH:
You seem to imply that (almost) none of the software listed atm has metric exposition of good quality.

If that is the case, it should be moved to the bottom in a "Partial or broken support" section and removed completely after X time.

If the metrics are of good quality, I doubt the users care too much about how the metrics have been created.

brian-brazil:
> You seem to imply that (almost) none of the software listed atm has metric exposition of good quality.

Few fully follow our guidelines. Exporters are always a bit murky though. That's not the case with instrumented code.

> If that is the case, it should be moved to the bottom in a "Partial or broken support" section and removed completely after X time.

I don't let ones that are completely unsable in.

> If the metrics are of good quality, I doubt users care too much about how said metrics have been created.

Directly instrumented are generally far better quality, as there's a whole pile of mistakes they can't make.

This list has more than one purpose, the purpose I care about here is showing off how many projects are doing things the Prometheus Way so that ultimately we get better metrics for users. Hopefully one day that'll dominate this page.

juliusv:
@brian-brazil We had this discussion a week ago or so, where I was also surprised that with "directly instrumented" you only meant things that used exactly our client library, vs. something else to expose it. I had always thought that "directly instrumented" just meant it had some form of whitebox instrumentation that in the end produced Prometheus metrics, without needing an exporter.

In general I agree with @grobie and @RichiH that the primary focus of this page should be on the user-visible aspect of components, not how they are implemented internally. We can still separate out components with bad metrics, as @RichiH suggested.

grobie:
I'd just add an icon after the name, like an exclamation mark which shows a
tooltip like "This exporter doesn't follow the project's best practices".

brian-brazil:
> I had always thought that "directly instrumented" just meant it had some form of whitebox instrumentation that in the end produced Prometheus metrics, without needing an exporter.

I believe I've always been consistent on this terminology. Direct instrumentation means Prometheus direct instrumentation.

I've also always been clear on the distinction between instrumentation and exposition. Instrumentation says nothing about exposition.

> We can still separate out components with bad metrics, as @RichiH suggested.

Components with good metrics are the exception, not the rule - especially when talking about exporters. There's also various other things exporters commonly get wrong, such as polling, using direct instrumentation rather than ConstMetrics or doing SD.

This a big reason I want to call out the handful of applications that are doing the things the way we want them to. I don't believe having a big Your Doing It Wrong on the majority of entries on this page would be productive.

RichiH:
There does not seem to be a direct, hard relation between "good metrics" and "Prometheus direct instrumentation", to me.

The page already links to best practices etc, I am not sure if much more can be done there.

What about
* moving instrumentation above exporters
* introducing symbols for
    * official
    * good metrics
    * shitty metrics
* wording `docs/instrumenting/writing_exporters.md` more strongly to emphasize why Prometheus libraries are best if you start out anyway

FTR, I forced^wconvinced several teams at work to toss out their own implementations and use the official libraries instead. But that was in the context of best practices, not in the context of what they could use to get ready-made metrics into Prometheus.

brian-brazil:
> There does not seem to be a direct, hard relation between "good metrics" and "Prometheus direct instrumentation", to me.

There's more to it than the names and structure of the metrics. The newest addition to the list seems to be pulling and possibly storing some metrics in Redis for example.

I want somewhere we can point to for applications doing things the Prometheus way, other than the instrumentation of Prometheus&co itself.

> FTR, I forced^wconvinced several teams at work to toss out their own implementations and use the official libraries instead. But that was in the context of best practices, not in the context of what they could use to get ready-made metrics into Prometheus.

My belief is that having a list you can point to of other applications doing that would help such convincing.

It took us longer than it should have to have an example of a correctly implemented exporter, we should avoid the same friction for direct instrumentation.

grobie:
> I want somewhere we can point to for applications doing things the
Prometheus way, other than the instrumentation of Prometheus&co itself.

I don't see this page being the right place for that.

brian-brazil:
> I don't see this page being the right place for that.

If that was the case we should remove the directly instrumented apps (and only those) as they are neither exporters nor integrations.

cherti:
> possibly storing some metrics in Redis

So a stateless-worker-thread based architecture with a database backend for sync cannot be instrumented directly at all with canonical Prometheus-metrics at all?

brian-brazil:
> Just for understanding: so a stateless-worker-thread based architecture with a database backend for sync cannot be instrumented directly at all with canonical Prometheus-metrics at all?

It can, just use the Prometheus python client which keeps things entirely in memory. For Python as long as you're not using multiple processes (e.g. unicorn, multiprocessing), everything should just work.

cherti:
> For Python as long as you're not using multiple processes (e.g. unicorn, multiprocessing), everything should just work.

And if it is using that? If the workerthreads are completely decoupled, synced only via the database keeping the application's state, and potentially die faster than the scrape interval?

brian-brazil:
This is discussion for elsewhere, but if they're threads it's not a problem as the state is kept in memory per-process.

brian-brazil:
To try and break this deadlock, how would keeping this list but adding an annotation for direct instrumentation sound?

I'd like to keep any annotations positive rather than negative.

grobie:
SGTM.

I was thinking about using an icon from https://getbootstrap.com/components/
which shows an explanation like "Follows Prometheus best practices" when
hovering over it, just to reduce the amount of text on the page. We could
also do the same for the current *official* label. I don't mind the details
too much though.

brian-brazil:
I'll just do `**direct**` or similar for now I think. "Follows Prometheus best practices" is a higher bar, and particularly with exporters that gets very messy.

cherti:
how about "with official libraries"?
This would not open the discussion/confusion about what exactly is "direct" (see the comment of @juliusv), would imply something in the direction of "follows upstreams best practices" as in "they are at least already using the officially provided libraries", but would not imply that the metrics that come from that must necessarily follow Prometheus' best practices (as you can still screw up the metric naming even if you use official libraries).

brian-brazil:
> how about "with official libraries"?

Not all the libraries are official, so we get into how closely the client libraries guidelines are being followed (the ones we've listed are largely okay) and which of those are relevant to instrumentation.

cherti:
true thing. I was actually wondering how e.g. C++-applications would be classified as there is no official library.
"with officially supported library" or something like that?

brian-brazil:
The key word isn't "direct", it's "instrumentation". Producing output "exposition". Direct means it is in the code, not going via an exporter.

cherti:
"Direct means it is in the code, not going via an exporter."
By that definition, self-cooked prometheus-metrics-code, baked directly into the application, is also "direct"ly instrumented, only with a different (unofficial) library (or "library" if it's really hardcoded in there). That's the confusion I mean.

brian-brazil:
> By that definition, self-cooked prometheus-metrics-code, baked directly into the application, is also "direct"ly instrumented, only with a different (unofficial) library (or "library" if it's really hardcoded in there).

That's a case I've pondered and am not sure about, I can imagine it coming up in C/C++ cases. I figure we'll cross that bridge when we come to it.

juliusv:
Yeah, that was my confusion as well. There's nothing in the phrase "direct instrumentation" itself that suggests whether an official or unofficial instrumentation library was used. You might have used it consistently in the past, but I probably never paid much attention to the exact phrase :) The phrase itself (without consulting any Prometheus-specific glossary or something) just means that there is instrumentation directly in the code.

brian-brazil:
The "with a Prometheus client library" is silent :)

Also: https://prometheus.io/docs/introduction/glossary/#direct-instrumentation

juliusv:
> Also: https://prometheus.io/docs/introduction/glossary/#direct-instrumentation

See, so even our glossary for it doesn't make the distinction based on *how* you add instrumentation to the code... thus I don't think we should tie that phrase to the specific use of our own Prometheus client library.

brian-brazil:
It's implicit, you can do instrumentation with many libraries but when you're on prometheus.io I think it's pretty obvious we mean the Prometheus ones.

juliusv:
> but when you're on prometheus.io I think it's pretty obvious we mean the Prometheus ones.

As you can see from this thread, it isn't obvious at all to multiple people who are even familiar with Prometheus. When I hear "direct instrumentation", my mind tells me: someone put instrumentation directly into their code (no matter what lib), and that finally enables a program to expose Prometheus (and potentially other) metrics by itself (as opposed to requiring an exporter). So I don't think we should use the term "direct instrumentation" without further qualification to mean "using an official Prometheus client library".

As others have said, I think in 90% of cases, a user visiting this page wants to find an answer to the question, "I want to get Prometheus metrics out of X, can it be done already?". Not "I want to find examples of software using specifically the Prometheus client library.". It's fine if you want to indicate that specifically in addition as you proposed, but just a `**direct**` marker is too misleading for the above reasons. It should be clear from plain English that the annotation is about using a Prometheus client library.

brian-brazil:
> So I don't think we should use the term "direct instrumentation" without further qualification to mean "using an official Prometheus client library".

That was always what I was proposing.

PTAL.

juliusv:
>> So I don't think we should use the term "direct instrumentation" without further qualification to mean "using an official Prometheus client library".
>
> That was always what I was proposing.

I meant that the `(**direct**)` marker is not clear by itself, but since there's an explanation of it at the bottom of the section, that's ok for now. I'd prefer if we found an equally brief clearer term for it in the future though (`**using official instrumentation library**` is a bit long).

So :+1:

cherti:
How to make a distinction between good external libraries (official libraries only include Go, Python, Java, Ruby), which @brian-brazil would classify as "using official library" if I understand correctly and good or bad external, but baked-in libraries? Or in other words: if a baked-in library would be factored out and would make it to the client-libraries-page, would that change the classification of the instrumented software in this context?

brian-brazil:
> Or in other words: if a baked-in library would be factored out and would make it to the client-libraries-page, would that change the classification of the instrumented software in this context?

As I said above, let's worry about it when it happens. I've yet to see any embedded instrumentation library that came close to being a Prometheus library, and I'd generally hope users would use something existing rather than writing their own.

PR #762, opened 2017-06-01
Add speedtest exporter (golang)

nlamirault:


brian-brazil:
We only list one of each type of exporter, it appears in this case that yours is the better one. Can you update to just your?

You should put the units in your metric names, and ensure they're base second and bytes/second units.

PR #763, opened 2017-06-01
Added OpenStack service discovery docs.

cgroschupp:
Documentation for prometheus/prometheus#2701

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -176,6 +176,10 @@ dns_sd_configs:
 ec2_sd_configs:
   [ - <ec2_sd_config> ... ]
 
+# List of OpenStack service discovery configurations.
+openstack_sd_configs:
+  [ - <openstack_sd_config> ... ]
+
 # List of file service discovery configurations.
 file_sd_configs:
   [ - <file_sd_config> ... ]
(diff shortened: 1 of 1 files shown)


content/docs/operating/configuration.md
```diff
@@ -399,6 +403,65 @@ region: <string>
...
+# control panel to discover your account's username. In Identity V3, either
+# UserID or a combination of Username and DomainID or DomainName are needed.
+[ username: <string> ]
+[ userid: <string> ]
+
+# Password for the Identity V2 and V3 APIs. Consult with your provider's
+# control panel to discover your account's preferred method of authentication.
+[ password: <string> ]
+
+# At most one of DomainID and DomainName must be provided if using Username
```
juliusv: I'd spell the field names the same way in the comments as they are named in the YML below (snake_case, not camelCase). Same for the field names below in further sections.

juliusv: (also, comments in sections above, not just below)

content/docs/operating/configuration.md
```diff
@@ -399,6 +403,65 @@ region: <string>
...
+# Some require both. Your provider's authentication policies will determine
+# how these fields influence authentication.
+[ project_name: <string> ]
+[ project_id: <string> ]
+
+# Refresh interval to re-read the instance list.
+[ refresh_interval: <duration> | default = 60s ]
+
+# The port to scrape metrics from. If using the public IP address, this must
+# instead be specified in the relabeling rule.
```
juliusv: Could you elaborate a bit on this (why? background?)?

juliusv:
:+1: Looks good besides nits, thanks!

brancz:
As this is completely new, I think we should explicitly mark this as a beta features to allow potential changes.

juliusv:
@brancz Good catch! Yes, this should be marked as beta, like some of the other SDs.

content/docs/operating/configuration.md
juliusv: @cgroschupp This comment was still open, but I think I get it now: by default, the private IP + configured port are prepopulated into the `__address__` label, but if you want to use the public IP, you'll need to manually use relabeling to use the `__meta_openstack_public_ip` meta label and also specify the port in there. Fine to merge this for now I think, but could be good to explain this further at some point.

juliusv:
:+1:

content/docs/operating/configuration.md
brian-brazil: This comment shouldn't be here, these are reference docs not user guides.

juliusv: @brian-brazil There should be a specification how this port is used though, right? (that by default it is put into the `__address__` label together with the private IP)

Or maybe somewhere outside of the config snippet we should explain that by default, the private IP is used to form the address together with the specified port.

brian-brazil: The bit around the public IP is more suitable for a user guide. We should of course document each field's behaviour, and this one is similar to fields in other SDs.

cgroschupp: Sorry, I overlooked the comment

PR #771, opened 2017-06-19
Fix documentation: The correct key is 'api_url'

michas2:
See https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L238


brian-brazil:
Thanks!

PR #774, opened 2017-06-21
Marathon SD: Add port def/mapping labels

mattbostock:
These labels were added in:
https://github.com/prometheus/prometheus/pull/2506

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -716,6 +716,8 @@ The following meta labels are available on targets during [relabeling](#relabel_
 * `__meta_marathon_image`: the name of the Docker image used (if available)
 * `__meta_marathon_task`: the ID of the Mesos task
 * `__meta_marathon_app_label_<labelname>`: any Marathon labels attached to the app
+* `__meta_marathon_port_definition_label_<labelname>`: the port definition labels
+* `__meta_marathon_port_mapping_label_<labelname>`: the port mapping labels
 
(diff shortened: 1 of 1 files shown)


content/docs/operating/configuration.md
```diff
@@ -716,6 +716,8 @@ The following meta labels are available on targets during [relabeling](#relabel_
 * `__meta_marathon_image`: the name of the Docker image used (if available)
 * `__meta_marathon_task`: the ID of the Mesos task
 * `__meta_marathon_app_label_<labelname>`: any Marathon labels attached to the app
+* `__meta_marathon_port_definition_label_<labelname>`: any labels attached to the port definitions
```
brian-brazil: "The labels of the port defintions" would be in line with the rest of the doc

mattbostock:
Is `master` the correct branch for this change as the labels being documented here are already in released versions of Prometheus?

content/docs/operating/configuration.md
mattbostock: Would 'the port definition labels' read better?

brian-brazil:
Yes, master is correct

content/docs/operating/configuration.md
brian-brazil: I'm not too bothered either way

mattbostock:
Amended.

brian-brazil:
Thanks!

PR #780, opened 2017-07-02
Add CRG to directly instrumented software

brian-brazil:


RichiH:
@grobie You should be ashamed for breaking this pristine and important UTF8-art; escaping is important:

¯\\_(ツ)_/¯

As to the actual PR: I think we (still) list everything and everyone who has integrations, so LGTM.

RichiH:
¯\\\_(ツ)_/¯

brian-brazil:
I am obliged to point out that the correct way to signal confusion when using this software is to repeatedly tap both shoulders.

> As to the actual PR: I think we (still) list everything and everyone who has integrations, so LGTM.

Yes, at some point when this directly instrumented list gets too long we'll remove it. But we're far from there yet.

PR #782, opened 2017-07-05
Document new PromConsole.Graph parameter colorScheme

alin-amana:
Added a link into the Rickshaw source code because I couldn't find any documentation, not sure whether that's acceptable.


brian-brazil:
Thanks!

PR #783, opened 2017-07-05
Correct example run command

c3st7n:
The example in the docs caused "prometheus: error: unknown short flag '-c'" error, as double `-` needed for `config.file` argument.


grobie:
You're using the alpha version of Prometheus 2.0. The docs describe Prometheus 1.0. You could open the change against the dev-2.0 branch though, in case it hasn't been fixed there.

Please note that we'll rework our docs before 2.0 release to allow for versioned documentation.

c3st7n:
Okay, my mistake. Looks like it is the same on the `dev-2.0` branch, so will change it to point to that (although I understand it will all be reviewed anyway).

PR #784, opened 2017-07-05
Correct example run command

c3st7n:
The example in the docs caused "prometheus: error: unknown short flag '-c'" error.


brian-brazil:
Thanks!

PR #786, opened 2017-07-09
adding instaclustr-exporter to the list of APIs exporter

fcgravalos:
Adding InstaClustr Exporter: 

https://github.com/fcgravalos/instaclustr_exporter

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -87,12 +87,12 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Docker Cloud exporter](https://github.com/infinityworksltd/docker-cloud-exporter)
    * [Docker Hub exporter](https://github.com/infinityworksltd/docker-hub-exporter)
    * [GitHub exporter](https://github.com/infinityworksltd/github-exporter)
+   * [InstaClustr exporter](https://github.com/fcgravalos/instaclustr_exporter)
    * [Mozilla Observatory exporter](https://github.com/Jimdo/observatory-exporter)
    * [OpenWeatherMap exporter](https://github.com/RichiH/openweathermap_exporter)
    * [Pagespeed exporter](https://github.com/foomo/pagespeed_exporter)
    * [Rancher exporter](https://github.com/infinityworksltd/prometheus-rancher-exporter)
    * [Speedtest.net exporter](https://github.com/RichiH/speedtest_exporter)
-   * [InstaClustr exporter](https://github.com/fcgravalos/instaclustr_exporter)
 
 ### Logging
    * [Google's mtail log data extractor](https://github.com/google/mtail)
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/exporters.md
```diff
@@ -92,6 +92,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Pagespeed exporter](https://github.com/foomo/pagespeed_exporter)
    * [Rancher exporter](https://github.com/infinityworksltd/prometheus-rancher-exporter)
    * [Speedtest.net exporter](https://github.com/RichiH/speedtest_exporter)
+   * [InstaClustr exporter](https://github.com/fcgravalos/instaclustr_exporter)
```

brian-brazil:
_count should only be used with Summaries/Histograms, not Gauges.

The standard units for latency in Prometheus are seconds.

_percentile95 would be clearer than _percentile

Do you really need to identify a node by all of it's ID, public IP and private IP?

fcgravalos:
Hi Brian,

Thanks for the review! I will take care of it tomorrow morning (CET time).

About nodes, they don't have a name,  they might be in different VPCs with the same IP space and they might not have public ip. So the tuple: (rack, private-ip) fully identifies a node but I find convenient adding extra information that makes sense for clusters operators.

Does that make sense to you? Do you have any other suggestion?

brian-brazil:
How we generally handle such annotations is to have another metric with a name like `cassandra_node_info` with all labels not needed for identity.

I'd wonder similarly about clusterId/clusterName.

brian-brazil:
Thanks!

fcgravalos:
Thanks for the reply Brian.

Sorry if my question sounds stupid or newbie:

Let's say I use the Id for clusters and nodes and I add cassandra_cluster_info and cassandra_node_info 1-value metrics. And now I want to filter values by clusterName, IPs etc. Would that mean that the queries will be much more complicated?

For instance, I want an alarm and a dashboard in Grafana with a cluster name and status using cassandra_cluster_running metric. Now, it's just: IF up{cassandra_cluster_running{clusterName: "mycluster"}}. Would it be much more complicated with your approach?

brian-brazil:
Yes, it's a bit more complicated however it makes other uses easier as they don't have to remove other labels.

https://www.robustperception.io/exposing-the-software-version-to-prometheus/ has an explanation of how it works in the most general case, if you just want to filter it's `cassandra_cluster_metric * on(instance, clusterId) cassandra_cluster_info{cluster_name="foo"}`

fcgravalos:
Ok thanks!! I'll check that out!

And also (and last question, I promise), what's the rule when querying an API? For instance, counts as you said normally are summaries. But in my case, the count is already provided by the API (it's a monitoring API), so I'm not really counting anything. Does that still make sense being a summary or the Gauge approach is sufficient?

Let me know if I should ask better these questions in IRC.

Thanks in advance Brian!

brian-brazil:
In this case everything will be a Gauge. It'd be a Summary if you could extract the raw data for the _sum and _count.

The thing is more that nodes_count should be just nodes.

fcgravalos:
Ok, makes sense. I will rename it then! Thanks Brian!

PR #787, opened 2017-07-10
Restore and clarify metric types on roadmap

fabxc:
A first pass on reverting the deletion of the item and extended by explanations of what first class metric types mean in different layers of the stack.

@beorn7 @brian-brazil

### DIFF
--- a/content/docs/introduction/roadmap.md
+++ b/content/docs/introduction/roadmap.md
@@ -48,6 +48,23 @@ utilized in the Prometheus server. We plan on making use of this
 metadata in the future. The first step is to aggregate this data in-memory
 in Prometheus and provide it via an experimental API endpoint.
 
+### First-class metric type support
+
+Prometheus breaks up notions of first class metric types such as Histograms or Summaries
+into a flattened time series based representation in various places. Prometheus' database
+and query language are both time-series oriented. The exposition formats provide different
+degrees of first-class metrics.
+  
+The protocol buffer exposition format provides full first-class metrics. The text format
+reduces them into a time series representation and merely provieds typing hints. Providing
+stronger typing for the text format is one potential goal of the [OpenMetrics intiative](https://github.com/RichiH/OpenMetrics).
+Adapting such first-class metrics in the storage layer may provide significant performance
+benefits through a reduced number of time series and specialized compression algorithms.
+This can likely be implemented as transparently to the user.  
+Stronger typing in the query language may enhance expressiveness and prevent many cases of
+semantically invalid queries. It is unclear whether PromQL can be adapted to
+support true metric types or whether a new query language should be developed.
+
 ### Prometheus metrics format as a standard
 
 We intend to submit a cleaned up version of our format for standardization


content/docs/introduction/roadmap.md
```diff
@@ -48,6 +48,23 @@ utilized in the Prometheus server. We plan on making use of this
...
+### First-class metric type support
+
+Prometheus breaks up notions of first class metric types such as Histograms or Summaries
+into a flattened time series based representation in various places. Prometheus' database
+and query language are both time-series oriented. The exposition formats provide different
+degrees of first-class metrics.
+  
+The protocol buffer exposition format provides full first-class metrics. The text format
+reduces them into a time series representation and merely provieds typing hints. Providing
+stronger typing for the text format is one potential goal of the [OpenMetrics intiative](https://github.com/RichiH/OpenMetrics).
```
brian-brazil: @RichiH Do we want to mention this publicly yet?

content/docs/introduction/roadmap.md
```diff
@@ -48,6 +48,23 @@ utilized in the Prometheus server. We plan on making use of this
...
 in Prometheus and provide it via an experimental API endpoint.
 
+### First-class metric type support
+
+Prometheus breaks up notions of first class metric types such as Histograms or Summaries
+into a flattened time series based representation in various places. Prometheus' database
+and query language are both time-series oriented. The exposition formats provide different
+degrees of first-class metrics.
+  
+The protocol buffer exposition format provides full first-class metrics. The text format
```
brian-brazil: The text format also provides first class metrics, it's how Prometheus interprets them and the general data model that does the flattening.

The text and proto formats are (mostly) equivalent, and proto is a distraction from the actual issues. 

I'd remove this sentence.

content/docs/introduction/roadmap.md
```diff
@@ -48,6 +48,23 @@ utilized in the Prometheus server. We plan on making use of this
...
+reduces them into a time series representation and merely provieds typing hints. Providing
+stronger typing for the text format is one potential goal of the [OpenMetrics intiative](https://github.com/RichiH/OpenMetrics).
+Adapting such first-class metrics in the storage layer may provide significant performance
+benefits through a reduced number of time series and specialized compression algorithms.
```
brian-brazil: ...for histograms

content/docs/introduction/roadmap.md
```diff
@@ -48,6 +48,23 @@ utilized in the Prometheus server. We plan on making use of this
...
+This can likely be implemented as transparently to the user.  
+Stronger typing in the query language may enhance expressiveness and prevent many cases of
```
brian-brazil: I'm not sure about "strong", we already don't permit e.g. using strings as ranges, or ranges as instant vectors.

Supporting first class metrics in the query language and data model...

content/docs/introduction/roadmap.md
```diff
@@ -48,6 +48,23 @@ utilized in the Prometheus server. We plan on making use of this
 metadata in the future. The first step is to aggregate this data in-memory
 in Prometheus and provide it via an experimental API endpoint.
 
+### First-class metric type support
+
+Prometheus breaks up notions of first class metric types such as Histograms or Summaries
```
brian-brazil: breaks up doesn't feel right here. Maybe: Prometheus handles concepts of...

content/docs/introduction/roadmap.md
```diff
@@ -48,6 +48,23 @@ utilized in the Prometheus server. We plan on making use of this
...
+degrees of first-class metrics.
+  
+The protocol buffer exposition format provides full first-class metrics. The text format
+reduces them into a time series representation and merely provieds typing hints. Providing
+stronger typing for the text format is one potential goal of the [OpenMetrics intiative](https://github.com/RichiH/OpenMetrics).
+Adapting such first-class metrics in the storage layer may provide significant performance
+benefits through a reduced number of time series and specialized compression algorithms.
+This can likely be implemented as transparently to the user.  
+Stronger typing in the query language may enhance expressiveness and prevent many cases of
+semantically invalid queries. It is unclear whether PromQL can be adapted to
```
brian-brazil: It can be adapted, the question is more can it be sanely adapted. I'd suggest "Is is unclear whether PromQL can be adapted in a clean fashion"

content/docs/introduction/roadmap.md
```diff
@@ -48,6 +48,23 @@ utilized in the Prometheus server. We plan on making use of this
...
+support true metric types or whether a new query language should be developed.
```
brian-brazil: s/should/would need to be/

beorn7:
Thanks for the write-up.

I'm not sure if it makes sense to go into that depth of technical detail. This will stir up a lot of discussion to get consensus on. I would have a few remarks, but they are much different and partially in contradiction with Brian.

I suggest to not have this discussion right now. I'd keep it more high level in the roadmap, also joining the _Server-side metric metadata support_ item into one with the 1st class metric item, and then link to an umbrella GH issue, where in turn we can list sources (e.g. the various discussions and documents we already have).

Depending on the course of action actually taken, implementing a separate API endpoint to serve metadata might be a distraction.

Suggested headline: _Support for metric types and metadata_

beorn7:
I think it doesn't make any sense to keep this stale PR open.

For the record: The roadmap as we have it on the website right now has no lazy consensus (I explicitly objected against it in #663) and it was never formally voted upon. It's therefore invalid. It would be nice to have a valid roadmap again. But I leave that to those that found it acceptable to put an invalid roadmap on the website in the first place.

PR #789, opened 2017-07-17
Clarify that Marathon server is selected at random

mattbostock:
Both I and a colleague had assumed that you could specify multiple
Marathon masters, where each master is responsible for a different group
of tasks. The Marathon service discovery selects servers at random:

https://github.com/prometheus/prometheus/blob/2a9ca394dd92dec53efe3cb3add75e346f201cc7/discovery/marathon/marathon.go#L186

Make that clear in the documentation.

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -722,6 +722,8 @@ See below for the configuration options for Marathon discovery:
 
 ```
 # List of URLs to be used to contact Marathon servers.
+# A server will be selected randomly; it is assumed that all servers are
+# responsible for the same tasks.
 # You need to provide at least one server URL, but should provide URLs for
 # all masters you have running.
 servers:


brian-brazil:
If this is added here then it also need to be added to most other SDs, as they all work in the same way.

juliusv:
@brian-brazil The only other SD type I see that takes a list of servers to talk to is Zookeeper SD (am I missing any?). If that really also only uses one of the provided servers, that should indeed be documented as well.

brian-brazil:
That does appear to be it, I had thought there were more.

I'm always a bit wary of listing every misconception users have had in docs, we have to presume some basic level knowledge at some point or the docs will end up overtaken by such items.

I'd suggest changing the sentence above to say "a cluster of Marathon servers". The sentence below can can be trimmed after the comma, as the extra usage suggestion is probably what caused the confusion. Putting usage advice in reference docs has to be done with care.

juliusv:
@brian-brazil If you list multiple servers and it uses only one, and the field for that just says `servers`, that really seems kind of unclear though. I think it totally makes sense to document the exact meaning of the field. I totally wouldn't have gotten that (though admittedly I know neither Zookeeper or Marathon, but still, even if I knew there could be multiple replicated servers, it might not be clear that it's about that vs. specifying servers from different clusters).

xperimental:
I'd also rather go with the mention of a "Marathon cluster" and rewording of the original documentation:

```
# List of URLs to be used to contact Marathon cluster.
#
# At least one master of the cluster has to be provided but
# it is recommended to list all masters of the cluster to be able
# to list tasks in case of a master failure. The listed servers
# are expected to be part of the same Marathon cluster and
# one will be picked at random.
```

brian-brazil:
That sounds good, I'd still cut the last sentence (e.g. have you checked its random?). The rest makes it clear it's about a single cluster, whereas the previous docs were a little ambiguous.

mattbostock:
@brian-brazil Please see the link in my commit message regarding it being random.

brian-brazil:
Is that random per request, per iteration, or per Prometheus run? I don't think we should specify this level of detail in the API docs, as then that could become a breaking change later on. This should remain an implementation detail IMHO.

mattbostock:
Thanks @brian-brazil. I focused on the randomness because I found it surprising to see the targets flip-flop periodically, but agree that the main thing is to explain how the list of servers are treated. Will update.

brian-brazil:
Did you forget to push your latest change?

xperimental:
> have you checked its random?

It might not be the best randomness (just the normal `rand` package), but that is not important here. Maybe using random here is actually also not the best way to pick a server at random, because one is usually the active leader (see also the comment in [the code](https://github.com/prometheus/prometheus/blob/76c9a0d931ba2d912a88e3b69749ec7b415ba47c/discovery/marathon/marathon.go#L282)). If you like I could also take a stab at the "make it better" hinted there, but the current approach works.

juliusv:
@mattbostock Your text seems fine to me. I don't know enough about Marathon to judge how beneficial it would be to always query the active leader, will defer to other peoples' judgement for that.

robsonpeixoto:
IMO it's important to make clear that the `servers` list is for one cluster and that it's selected in a random fashion.
When I have configured Prometheus to scrape the Marathon I need to read the code to figure out how `servers` is used.

brian-brazil:
The thing is if we mention that this is random on this page, then that's a documented part of our API and we can basically never change it to be better. Thus we should focus on the cluster aspect rather than the exact load balancing implementation.

xperimental:
The point about making the randomness permanent is a good one, maybe we should omit it then and just point out that it is supposed to be one cluster.

> how beneficial it would be to always query the active leader

The non-leader masters proxy all requests they receive to the current leader, so it is a bit less efficient than directly hitting the leader, but it is supported. In the case of Prometheus I don't see a major drawback of just randomly selecting a master, because it is not hitting the API that often.

brian-brazil:
I'm proposing #878 to resolve this.

brian-brazil:
This has been handled over in the 2.0 docs.

PR #790, opened 2017-07-18
Update PagerDuty link in Alertmanager config

stuartnelson3:
PagerDuty has a page specifically for integrating
with Prometheus/Alertmanager, which seems like a
good idea to link to from our documentation.

Link: https://www.pagerduty.com/docs/guides/prometheus-integration-guide/

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -295,7 +295,8 @@ room_id: <tmpl_string>
 
 ## `<pagerduty_config>`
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
We are using that API though, I think we should add a link rather than replacing the existing one.

stuartnelson3:
Ah sure, makes sense. Updated. Wording is up to you if you want to change it, I'm mostly interested in providing the direct link to readers.

brian-brazil:
There's similar docs for Victorops and Opsgenie, Conor and I talked to them over the past while to make Prometheus an official integration.

PR #791, opened 2017-07-20
Add a PHP-FPM exporter to the list of exporters

andrewhowdencom:
Currently, there is a large set of exporters available for a lot of different software tools.
However, the popular runtime "PHP-FPM", used for executing PHP script through the
FastCGI protocol, is not listed.

Various exporters currently exist, and the author has not tested any (including the
suggested) extensively. However, a quick review of the available options shows that,
in the authors limited review, this appears to be well constructed, thought out and
documented.

This commit adds the exporter to the canonical list for easy discovery. 

Design Notes:

Alternatives considered were:
- https://github.com/peakgames/php-fpm-prometheus/ (no releases, less docs)
- https://github.com/blablacar/phpfpm-prometheus-exporter (no docs)

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -119,6 +119,9 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [SNMP exporter](https://github.com/prometheus/snmp_exporter) (**official**)
    * [StatsD exporter](https://github.com/prometheus/statsd_exporter) (**official**)
 
+### Interpreters 
+   * [PHP-FPM exporter](https://github.com/bakins/php-fpm-exporter)
+
 ### Miscellaneous
    * [BIG-IP exporter](https://github.com/ExpressenAB/bigip_exporter)
    * [BIND exporter](https://github.com/digitalocean/bind_exporter)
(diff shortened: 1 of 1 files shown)


andrewhowdencom:
Yo @bakins what do you think. I haven't got around to testing stuff, but your exporter looked pretty nice. Would you mind if it was included?

content/docs/instrumenting/exporters.md
```diff
@@ -119,6 +119,9 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [SNMP exporter](https://github.com/prometheus/snmp_exporter) (**official**)
    * [StatsD exporter](https://github.com/prometheus/statsd_exporter) (**official**)
 
+### Interpreters
```
brian-brazil: This falls under misc

andrewhowdencom: Will attempt to fix upstream before this gets merged. Leaving as is for now, will kill it in 7 days if there hash been no update.

brian-brazil: I'm happy to merge as soon as this is in the right place on the page.

andrewhowdencom:
Amended. I'll chase the metrics renaming up in a separate stream of work

brian-brazil:
Thanks!

bakins:
@andrewhowdencom Is fine with me. I wrote it as none of the others quite did what I wanted. 

@brian-brazil I can rename metrics to match the conventions. Right now, I think they are just named the same as php-fpm exposes them.  I'm not yet running this in production, so it's easy enough to change.

brian-brazil:
The general thinking is that if you're listing the names for everything anyway in your exporter, you may aswell make them in line with guidelines. For a more automated conversion, there'd not be much point.

PR #793, opened 2017-07-20
Update getting_started docs

timrs2998:
Prometheus uses two dashes instead of one dash for its arguments (latest darwin download)

@brian-brazil


brian-brazil:
These docs refer to Prometheus 1.x, which still use one dash.

timrs2998:
Awesome thanks!

Issue #413, opened 2016-04-18
Provide logos


grobie:
There are many cases were people want to embed the prometheus logo somewhere, e.g. in third-party integrations. We should provide a logo pack somewhere in different resolutions. 

@juliusv

ahus1:
Ideas: 
  * add logo section to introduction/media
  * add a link to the [cloudnative store](https://store.cncf.io/collections/prometheus/products/prometheus-decals) - although I have to say shipping takes very long and is expensive
  * add a link to [stickermule](https://www.stickermule.com) so people can order stickers themselves
  * add entry to FAQ

I'm happy to provide a pull request given some comments on the options above.

jan--f:
Hello from the bug scrub!

We could perhaps simply link to https://github.com/cncf/artwork/blob/main/examples/graduated.md#prometheus-logos or https://github.com/cncf/artwork/tree/main/projects/prometheus?

PR #795, opened 2017-07-22
derived two orange logos from the original grey logo

ahus1:
The original logo in SVG format was grey only. I add the grey logo colored in orange (same color as used on website). I also add an inverted logo that should work with die cut stickers.


grobie:
:+1:

juliusv:
Hmm, we have some variations in https://github.com/cncf/artwork/tree/master/prometheus - wonder if these would be better to put there too?

ahus1:
@juliusv - I'd like to tackle #413 next to link the logo download. For this it would be nice to have them in the same repo. But the logo should be at cncf as well. I can prepare a PR for CNCF once we agree what to do here.

juliusv:
@ahus1 Ok, sounds good to me :)

PR #798, opened 2017-07-26
nit: right associative example

math4tots:
See: https://prometheus.io/docs/querying/operators/#binary-operator-precedence

`2 * 3 ^ 2` being equivalent to `2 * (3 ^ 2)` doesn't actually demonstrate that `^` is right associative. It just follows from operator precedence. The significance of right associativity is better exemplified by something of same precedence level as `^`.


brian-brazil:
Thanks!

PR #801, opened 2017-07-31
Correct the version reference in the webhook spec

andrewhowdencom:
The version expressed by the AlertManager was changed in the following
commit: 

  https://github.com/prometheus/alertmanager/commit/3269bc39e121452abdce24db1d3d9003d99184d1

This updates the documentation to reflect that change. 

Design Notes:

  This wasn't checked against the full spec. The author has judged that
  this change was to the values of the spec, rather than the keys.

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -459,7 +459,7 @@ endpoint:
 
 ```
 {
-  "version": "3",
+  "version": "4",
   "groupKey": <number>     // key identifying the group of alerts (e.g. to deduplicate)
   "status": "<resolved|firing>",
   "receiver": <string>,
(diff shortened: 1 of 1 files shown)


content/docs/alerting/configuration.md
```diff
@@ -459,7 +459,7 @@ endpoint:
 
 ```
 {
-  "version": "3",
+  "version": "4",
   "groupKey": <number>     // key identifying the group of alerts (e.g. to deduplicate)
```
brian-brazil: The group key is no longer a number, it's a string.

andrewhowdencom:
Amended. Not sure what your commit guidelines look like; there's a joke/inappropriateness in the last commit. 

Besides that, addressed.

brian-brazil:
Thanks!

PR #803, opened 2017-08-07
List base units for metric families

pilhuhn:
Explicitly list base units for some measurement families.

Inspired by https://github.com/RichiH/OpenMetrics/issues/17

### DIFF
--- a/content/docs/practices/naming.md
+++ b/content/docs/practices/naming.md
@@ -26,7 +26,7 @@ A metric name...
  * <code><b>http</b>\_request\_duration\_seconds</code>
    (for all HTTP requests)
 * ...must have a single unit (i.e. do not mix seconds with milliseconds, or seconds with bytes).
-* ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers).
+* ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). See below for a list of base units.
 * ...should have a suffix describing the unit, in plural form. Note that an accumulating count has `total` as a suffix, in addition to the unit if applicable.
  * <code>http\_request\_duration\_<b>seconds</b></code>
  * <code>node\_memory\_usage\_<b>bytes</b></code>
@@ -61,3 +61,24 @@ pairs represents a new time series, which can dramatically increase the amount
 of data stored. Do not use labels to store dimensions with high cardinality
 (many different label values), such as user IDs, email addresses, or other
 unbounded sets of values.
+
+
+## Base units
+
+Prometheus does not have any units hard coded. For better compability, base
+units should be used. The following lists some metrics families with their base unit.
+The list is not exhaustive.
+
+| Family | Base unit | Remark | 
+| -------| --------- | ------ |
+| Time   | seconds   |        |
+| Temperature | celsius | Celsius is the most common one encountered in practice |
+| Length | meters | |
+| Bytes  | bytes | | 
+| Bits   | bytes | |
+| Percent | ratio(*) | Values are 0-1. <br/> *) Usually 'ratio' is not used as suffix, but rather A\_per\_B. Exceptions are e.g. disk\_usage\_ratio  |
+| Voltage | volts | |
+| Electric current | amperes | | 
+| Energy | joules | |
+| Weight | grams | |
+ 


content/docs/practices/naming.md
```diff
@@ -61,3 +61,22 @@ pairs represents a new time series, which can dramatically increase the amount
...
+units should be used. The following lists metrics families with their base unit:
+
+| Family | Base unit | Remark | 
+| -------| --------- | ------ |
+| Time   | seconds   |        |
+| Temperature | celsius |     |
+| Length | meter | |
+| Bytes  | bytes | | 
+| Bits   | bytes | |
+| Percent | percent | Values are 0-1 |
```
brian-brazil: ratio is the base unit

content/docs/practices/naming.md
```diff
@@ -61,3 +61,22 @@ pairs represents a new time series, which can dramatically increase the amount
...
+| Voltage | volt | |
+| Electric current | ampere | | 
+| Power | watts | |
```
brian-brazil: This is unclear, in the cases we've seen it has tended to be a counter in joules.

content/docs/practices/naming.md
```diff
@@ -61,3 +61,22 @@ pairs represents a new time series, which can dramatically increase the amount
...
+| Weight | gramms | |
```
brian-brazil: Only one m

content/docs/practices/naming.md
```diff
@@ -61,3 +61,22 @@ pairs represents a new time series, which can dramatically increase the amount
 of data stored. Do not use labels to store dimensions with high cardinality
 (many different label values), such as user IDs, email addresses, or other
 unbounded sets of values.
+
+
+## Base units
+
+Prometheus does not have any units hard coded. For better compability, base
```
brian-brazil: It should be made clear that this is non-exhaustive

content/docs/practices/naming.md
```diff
@@ -61,3 +61,22 @@ pairs represents a new time series, which can dramatically increase the amount
...
+units should be used. The following lists metrics families with their base unit:
+
+| Family | Base unit | Remark | 
+| -------| --------- | ------ |
+| Time   | seconds   |        |
+| Temperature | celsius |     |
```
brian-brazil: This needs a remark as to why it's not Kelvin

content/docs/practices/naming.md
pilhuhn: Watt is Joule per second. Watt is a unit of Power, while Joule is a unit of Energy.

We can also just leave it out.

content/docs/practices/naming.md
```diff
@@ -61,3 +61,22 @@ pairs represents a new time series, which can dramatically increase the amount
...
+| Length | meter | |
```
beorn7: "meters" for consistent pluralization?

content/docs/practices/naming.md
```diff
@@ -61,3 +61,22 @@ pairs represents a new time series, which can dramatically increase the amount
...
+| Bytes  | bytes | | 
+| Bits   | bytes | |
+| Percent | percent | Values are 0-1 |
+| Voltage | volt | |
```
beorn7: volts? see above.

content/docs/practices/naming.md
```diff
@@ -61,3 +61,22 @@ pairs represents a new time series, which can dramatically increase the amount
...
+| Electric current | ampere | |
```
beorn7: amperes? see above.

content/docs/practices/naming.md
beorn7: Let's just put energy (in joules) in for now and see how use cases with power develop?

pilhuhn:
@beorn7 You are probably right about all those 's', but for me they look so unusual and ugly, as in German Volt is Volt in singular and plural :-/

grobie:
This is the base unit list we check for in `promtool`: https://github.com/prometheus/prometheus/blob/1ea9ab601e936de5bb164edb9de2a3988a199fb7/util/promlint/promlint.go#L235-L246 We should sync the documentation and promtool suggestions. @mdlayher

mdlayher:
I'm in favor for sure. I picked the units I did because they are the SI base units, IIRC. That said, I'd be shocked if someone has ever used candela as a unit with Prometheus. :)

pilhuhn:
@grobi I see your list has kelvin(s) and also meters/metres - for the former I'd then also allow volt/volts and ampere/amperes. 
Vs meter/metre : I think this is one of the cases I want to avoid as the exception needs to be encoded in tools. Similar for the non -s versions of kelvin, volt and ampere (and as this was a french guy why not ampère? :-) )

content/docs/practices/naming.md
```diff
@@ -61,3 +61,26 @@ pairs represents a new time series, which can dramatically increase the amount
...
+| Length | meters | |
+| Bytes  | bytes | | 
+| Bits   | bytes | |
+| Percent | ratio | Values are 0-1 |
+| Voltage | volts | |
+| Electric current | amperes | | 
+| Power | watts | |
+| Energy | joules | |
+| Weight | grams | |
+| Data transfer | bytes\_per\_second | |
```
brian-brazil: A counter with bytes is the way to transfer this, bytes_per_second should be quite rare (basically only interface speed limits).

content/docs/practices/naming.md
```diff
@@ -61,3 +61,26 @@ pairs represents a new time series, which can dramatically increase the amount
...
+The list is not exhaustive.
+
+| Family | Base unit | Remark | 
+| -------| --------- | ------ |
+| Time   | seconds   |        |
+| Temperature | celsius | Celsius is the most common one encountered in practice |
+| Length | meters | |
+| Bytes  | bytes | | 
+| Bits   | bytes | |
+| Percent | ratio | Values are 0-1 |
```
brian-brazil: ratio is the base unit, but we should mention that there's no suffix for it in the metric name. It'd be A_per_B

pilhuhn: Not sure I get the A_per_B part. E.g. for disk usage it is common to say 85% which would be `disk_usage=0.85` . Here using a _percent suffix may make it easier for people to differentiate from disk_usage_bytes=123456 (absolute number).
But in both cases I fail to see `disk_usage_current_to_full = 0.85`

brian-brazil: A_per_B is the rule naming convention for ratios. _percent would be incorrect, though I see an argument for disk_usage_ratio

pilhuhn:
Is this list good now as a starter? More units can always be added later

content/docs/practices/naming.md
```diff
@@ -61,3 +61,25 @@ pairs represents a new time series, which can dramatically increase the amount
...
+| -------| --------- | ------ |
+| Time   | seconds   |        |
+| Temperature | celsius | Celsius is the most common one encountered in practice |
+| Length | meters | |
+| Bytes  | bytes | | 
+| Bits   | bytes | |
+| Percent | ratio(*) | Values are 0-1. <br/> *) Usually 'ratio' is not used as suffix, but rather A\_per\_B. Exceptions are e.g. disk\_usage\_ratio  |
+| Voltage | volts | |
+| Electric current | amperes | | 
+| Power | watts | |
```
brian-brazil: I'd exclude this, as it's usually expressed in terms of joules

pilhuhn: Not sure I agree, but it is removed now.

brian-brazil:
Thanks!

rightfold:
Kilograms is usually the base unit for mass, Newtons is the base unit for weight.

PR #804, opened 2017-08-07
update copyright year

cstyan:
update year in copyright footer from 2016 -> 2017

### DIFF
--- a/layouts/container_footer.html
+++ b/layouts/container_footer.html
@@ -2,6 +2,6 @@
 
 <footer>
   <p class="pull-left">
-    &copy; Prometheus Authors 2016
+    &copy; Prometheus Authors <%= Date.today.year %>
   </p>
 </footer>


layouts/container_footer.html
```diff
@@ -2,6 +2,6 @@
 
 <footer>
   <p class="pull-left">
-    &copy; Prometheus Authors 2016
+    &copy; Prometheus Authors 2017
```
grobie: Alright, let's solve this issue once and for all with this PR: `<%= Date.today.year %>`.

cstyan: good idea! I can make that change

out of curiosity, what templating language are we using?

grobie: We use https://nanoc.ws/, a ruby static site generator. All HTML files are run through an [ERB filter](https://github.com/prometheus/docs/blob/master/Rules#L43).

cstyan:
@grobie updated

test.rb
```diff
@@ -0,0 +1,2 @@
+require 'date'
```
grobie: What's the purpose of this file?

cstyan:
removed `test.rb`, did not mean to stage that file :)

grobie:
Thank you!

PR #805, opened 2017-08-10
clarify gce sd config when meta data is returning urls as values

mkm85:
Fixes issue https://github.com/prometheus/prometheus/issues/3042

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -531,11 +531,6 @@ The following meta labels are available on targets during [relabeling](#relabel_
 * `__meta_gce_tags`: comma separated list of instance tags
 * `__meta_gce_zone`: the GCE zone URL in which the instance is running
 
-For meta labels returning an URL the format is as follows:
-* `__meta_gce_network`: `https://www.googleapis.com/compute/v1/projects/<project>/global/networks/<network>`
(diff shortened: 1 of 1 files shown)


content/docs/operating/configuration.md
```diff
@@ -523,13 +523,18 @@ The following meta labels are available on targets during [relabeling](#relabel_
...
 * `__meta_gce_private_ip`: the private IP address of the instance
 * `__meta_gce_project`: the GCP project in which the instance is running
 * `__meta_gce_public_ip`: the public IP address of the instance, if present
-* `__meta_gce_subnetwork`: the subnetwork of the instance
+* `__meta_gce_subnetwork`: the subnetwork URL of the instance
 * `__meta_gce_tags`: comma separated list of instance tags
-* `__meta_gce_zone`: the GCE zone in which the instance is running
+* `__meta_gce_zone`: the GCE zone URL in which the instance is running
+
+For meta labels returning an URL the format is as follows:
```
brian-brazil: This may change due to no action on our part, so this shouldn't be documented here. We don't want GCE to cause what counts as a breaking change in Prometheus.

mkm85: I agree. I still think it's most intuitive if it does not return urls in the first place.

brian-brazil: You are free to take that up with GCE. Please remove this section.

brian-brazil:
Thanks!

PR #806, opened 2017-08-11
Add note about long term metric storage

SamSaffron:
I feel the current comparison is **highly** misleading. After reading this page an argument is made that Prometheus is superior and most solutions out there are inferior. 

Long-term metrics, by-design, are not handled by prometheus. You can capture a couple of weeks of data, but not years, there are no downsampling options and memory usage is quite high. 

I feel is is EXTREMELY important to highlight this design decision here so sysadmins of the world do not get confused.

### DIFF
--- a/content/docs/introduction/comparison.md
+++ b/content/docs/introduction/comparison.md
@@ -5,6 +5,13 @@ sort_rank: 4
 
 # Comparison to alternatives
 
+> **Important note**
+> Prometheus focuses on monitoring the "here-and-now" it is designed to capture
+> data for a limited number of days. If you need long-term trends you will need
+> to consider using a remote data adapter and a time series database that supports
+> downsampling metrics and dropping fidelity over time such as Graphite Carbon or 
+> InfluxDB
+
 ## Prometheus vs. Graphite
 
 ### Scope


RichiH:
I have 48 million time series in a time frame of 8+ months in my main
system, sample rates 5, 60, or 120s. It works.

Generic read/write already exists and will solve the long term storage
problems.



Richard

default.

SamSaffron:
Hi Richard, 

The official messaging is very different

https://www.robustperception.io/using-the-remote-write-path/ 

"Prometheus’s local storage isn’t intended as a long term data store, rather as more of an ephemeral cache. This is as storing effectively unbounded amounts of time series data would require a distributed storage system, whose reliability characteristics would not be what you want from a monitoring system." 

Also see: 

https://github.com/prometheus/prometheus/issues/10

RichiH:
This messaging will be overhauled very soon, anyway.

Unless you are setting `-storage.local.target-heap-size`, Prometheus will take as much RAM as it can get; that's in the docs, though.

beorn7:
> Unless you are setting `-storage.local.target-heap-size`, Prometheus will take as much RAM as it can get; that's in the docs, though.

Not quite right. -storage.local.target-heap-size` is always set, as it has a default value of 2GiB.

@SamSaffron In different news, the retention time of your Prometheus server should have very little impact on your RAM usage.

SamSaffron:
@beorn7 however there is not simple mechanism of defining retention per metric and downsampling like carbon or influx support, unless I am missing something, will post memory stats on our Discourse collector, but it got so huge we had to wind it back to 2 weeks so stuff did not explode.

brian-brazil:
On the PR itself, this document already mentions that the systems which are multi-node (which isn't all of them) are better for long term storage.

SamSaffron:
@brian-brazil I don't think that is fair, both graphite and influx can run single node with pretty straight forward config and downsample which allows you to carry years of data on a metric without sweating on storage. I don't see how multi-node is directly related here, downsampling is the missing feature.

brian-brazil:
We see multi node as the core feature for indefinite long term storage, as otherwise you're limited to whatever you can fit on a single machine. Downsampling only gets you so far in terms of long term storage, and has other issues technically depending on how you implement it.

Downsamlping isn't presently mentioned in the doc, we could add a brief mention that Graphite&Influx support it natively and you can do it by hand with Prometheus.

beorn7:
Closing because the debate is very outdated.

PR #810, opened 2017-08-14
Add a page to list all the other types of integrations

brian-brazil:
We've needed one of these for a while, I've seeded it from PRs and ones I know of offhand.

The angle brackets in the links to the config files aren't working, does anyone happen to know what's up there?

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -498,6 +498,10 @@ Each target has a meta label `__meta_filepath` during the
 [relabeling phase](#relabel_config). Its value is set to the
 filepath from which the target was extracted.
 
+There is a list of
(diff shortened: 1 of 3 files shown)


content/docs/operating/configuration.md
```diff
@@ -498,6 +498,10 @@ Each target has a meta label `__meta_filepath` during the
 [relabeling phase](#relabel_config). Its value is set to the
 filepath from which the target was extracted.
 
+There is a list of
+[integrations](/docs/operating/configuration/#<file_sd_config>) with this
```
mdlayher: `s/configuration/integrations/`?

brian-brazil: Did you get this the wrong way around?

PR #811, opened 2017-08-15
Add FreeBSD to list of software exposing metrics

skreuzer:
Document that FreeBSD now ships with the prometheus_sysctl_exporter(8) daemon which exposes sysctl values in Prometheus's format.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -159,6 +159,7 @@ separate exporters are needed:
    * [CRG Roller Derby Scoreboard](https://github.com/rollerderby/scoreboard) (**direct**)
    * [Doorman](https://github.com/youtube/doorman) (**direct**)
    * [Etcd](https://github.com/coreos/etcd) (**direct**)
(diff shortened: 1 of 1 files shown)


RichiH:
Reading https://reviews.freebsd.org/D8792 I am wondering if this is truly direct.

brian-brazil:
This is not direct. I think to be clear we should say FreeBSD kernel.

skreuzer:
> This is not direct. I think to be clear we should say FreeBSD kernel.

Ah, I misunderstood what direct means. I also renamed the link to 'FreeBSD Kernel'

brian-brazil:
Thanks! If there's official docs at some point it'd be good to switch the link to them.

matthiasr:
Is [this link](https://www.freebsd.org/cgi/man.cgi?query=prometheus_sysctl_exporter&apropos=0&sektion=8&manpath=FreeBSD+12-current&arch=default&format=html) suitable? man pages are very much official.

RichiH:
It's definitely better, yes.

RichiH:
Can you publish a new PR?

PR #815, opened 2017-08-21
Update link for FreeBSD kernel exposition

matthiasr:
as discussed in #811, the manpage is core documentation for BSDs. Link it instead of the discussion around adding this.

/cc @RichiH


RichiH:
This is limited to FreeBSD 12-current, but there does not seem to be a way to point to always-latest...

PR #816, opened 2017-08-28
extended the Naemon exporter with Nagios

Griesbacher:
The Naemon exporter is now also capable of exporting from Nagios Systems.


content/docs/instrumenting/exporters.md
```diff
@@ -114,7 +114,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [JavaMelody exporter](https://github.com/fschlag/javamelody-prometheus-exporter)
    * [JMX exporter](https://github.com/prometheus/jmx_exporter) (**official**)
    * [Munin exporter](https://github.com/pvdh/munin_exporter)
-   * [Naemon exporter](https://github.com/Griesbacher/Iapetos)
+   * [Naemon / Nagios exporter](https://github.com/Griesbacher/Iapetos)
```
brian-brazil: I'd put the Nagios first, as that's the more popular system

Griesbacher:
I've changed the order.
It was primarily developed for Naemon that's the reason for the original ordering.

PR #828, opened 2017-09-07
Add aws-health-exporter to the list of exporters

smon:
We developed a prometheus exporter for the AWS Health API and would like to add it to the list of exporters. 

https://github.com/Jimdo/aws-health-exporter

Please have a look @brian-brazil.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -85,6 +85,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
 
 ### APIs
    * [AWS ECS exporter](https://github.com/slok/ecs-exporter)
+   * [AWS Health exporter](https://github.com/Jimdo/aws-health-exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

You should use ConstMetrics rather than direct instrumentation with exporters. See https://www.robustperception.io/setting-a-prometheus-counter/ for an example.

A _count suffix indicates a summary or histogram, it shouldn't be on a gauge.

smon:
That was really quick, thanks! 

Regarding ConstMetrics - as events with the same set of labels can occur in one scrape I always got the `was collected before with the same name and label values` error. Therefore using and reseting the gauge seems to be easier than aggregating the events manually beforehand. 
Any other way around that error while still using ConstMetrics?

brian-brazil:
Ah, that makes things more fun. What you might want to do there is create a new gauge per scrape and collect it. That ways scrapes can happen concurrently, and you don't need the locking.

smon:
I'll give that a try. Thanks!

PR #830, opened 2017-09-09
added new ec2 discovery role_arn option

jamiemoore:
@grobie Please let me know if there are any changes needed.  Thanks.

### DIFF
--- a/content/docs/operating/configuration.md
+++ b/content/docs/operating/configuration.md
@@ -396,6 +396,9 @@ region: <string>
 # Named AWS profile used to connect to the API.
 [ profile: <string> ]
 
+# AWS Role ARN, an alternative to using AWS API keys.
+[ role_arn: <string> ]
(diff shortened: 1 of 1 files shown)


content/docs/operating/configuration.md
```diff
@@ -396,6 +396,9 @@ region: <string>
 # Named AWS profile used to connect to the API.
 [ profile: <string> ]
 
+# AWS Role ARN, an alternative to using AWS API keys.
+[ role_arn: <arn:aws:iam::account-id:role/role-name> ]
```
grobie: The part enclosed in the carrot brackets is the type of the field, not an example string. I think it's well enough defined by AWS that we don't need a further example here, I'd just use a plain `<string>` type here.

jamiemoore: Done.

PR #831, opened 2017-09-11
Correct flag in Getting Started guide

applegrain:
I was going through the Getting Started guide and this bit me. The `config.file` flag should be followed by double dashes, not a single dash, in order to work right.

@brian-brazil

### DIFF
--- a/content/docs/introduction/getting_started.md
+++ b/content/docs/introduction/getting_started.md
@@ -67,7 +67,7 @@ Prometheus build directory and run:
 ```language-bash
 # Start Prometheus.
 # By default, Prometheus stores its database in ./data (flag -storage.local.path).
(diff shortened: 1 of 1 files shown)


grobie:
Well, in general the reason the docs are for v1.x while only Prometheus v2.x including the available beta release requires double dashes.

On the other hand, while go's stdlib flag libraries prints flags only with a single dash, double dashes are accepted as well. Therefore, maybe let's merge this change to avoid further confusion. 

:+1:

applegrain:
Ok - whatever you think is best. I just downloaded latest prometheus to go through the guide and didn't catch the fact that is was for v1.x.

applegrain:
Closing this PR because it seems like this is more related to properly versioning docs instead of fixing this one small thing.

PR #832, opened 2017-09-13
Added reference to SQS exporter

jmal98:
I have created a simple AWS SQS exporter for tracking queue information.

https://github.com/jmal98/sqs_exporter


brian-brazil:
Thanks!

The attribute name should go in the metric name, and the queue name in a label.

jmal98:
@brian-brazil I have updated the format per your comment.

Thanks for the feedback.

PR #837, opened 2017-09-21
Add a note about changes in Prometheus 2.0 rules file format

mvasilenko:
@brian-brazil will you accept such note until 2.0 docs will be finished?


brian-brazil:
Our docs cover 1.0.

PR #840, opened 2017-09-30
Add swagger-stats to the list of Other utilities

sv2:
Propose to add [swagger-stats](https://github.com/slanatech/swagger-stats) to the list of Other third-party utilities. swagger-stats is node.js module that exposes Prometheus metrics for REST API in node.js app.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -172,7 +172,7 @@ separate exporters are needed:
    * [SkyDNS](https://github.com/skynetservices/skydns) (**direct**)
    * [Telegraf](https://github.com/influxdata/telegraf/tree/master/plugins/outputs/prometheus_client)
    * [Weave Flux](https://github.com/weaveworks/flux)
-
+   
 The software marked *direct* is also directly instrumented with a Prometheus client library.
 
 ## Other third-party utilities
@@ -189,3 +189,5 @@ practices.
    * Go: [prombolt](https://github.com/mdlayher/prombolt)
    * Java/JVM: [Hystrix metrics publisher](https://github.com/ahus1/prometheus-hystrix)
    * Python-Django: [django-prometheus](https://github.com/korfuri/django-prometheus)
+   * Node.js: [swagger-stats](https://github.com/slanatech/swagger-stats)
+   


brian-brazil:
This would go on the exporters&integrations page, down the bottom. Does the existing node.js client not offer this?

sv2:
Yes, I am proposing to add it to exporters&integrations page, down the bottom to the section  "Other third-party utilities". Swagger-stats is different from node.js client - it is not generic client library. It itself calculates and provides metrics for API as defined in Swagger(OpenAPI) specification (like number of request with label path='/api/{param}'). Thank you for your consideration !

brian-brazil:
Is this code using the listed node.js client library? I don't want to confuse users by adding things that don't do so.

sv2:
No, it does not use node.js client library. It calculates and exposes metrics itself, following guidelines in Prometheus docs, of course. The main reason is to minimize dependencies and required changes for the application that uses it. As swagger-stats is a library itself, I din't want to make an assumption whether application already uses client or not, and which client it is. If there is more preferred way to handle such scenario, would appreciate guidance. Thanks !

brian-brazil:
The preferred way to handle such things is to use the listed client library and provide a custom collector for it. That's how this is handled elsewhere, as it allows multiple libraries within an application to be instrumented as one.

sv2:
OK, understood. Will change it to work this way. Thank you!

sv2:
Just released an update to swagger-stats module, addressing your recommendation. Now listed Prometheus client library for node.js  - [prom-client](https://www.npmjs.com/package/prom-client) - is used to expose Prometheus metrics (https://github.com/slanatech/swagger-stats/issues/20).
Thank you!

brian-brazil:
What you've done is add prom-client as a potential output, not use prom-client. https://github.com/slanatech/swagger-stats/blob/44de212ad6c847da013c5c1267536348c69362bc/lib/swsHistogram.js for example should no longer exist, as it'd be provided by prom-client.

sv2:
OK. I've kept this as it's also used to show histograms on embedded UI. I'll make a change to remove it and use other implementation for internal needs. Thank you!

sv2:
Released an update (v0.93.0) to address last comments. Now swagger-stats always require prom-client library, and uses prom-client to for all Prometheus metrics. Old code has been removed.  Thank you!

brian-brazil:
Thanks!

sv2:
Thanks a lot for your help !

PR #841, opened 2017-10-02
Add Loodse to the list of users

metalmatze:
My new company is using Prometheus too, so I'd like to add us to the landing page! :blush:  :+1:


brian-brazil:
Thanks!

PR #843, opened 2017-10-04
Update AlertManager new notifiers recommendation

josedonizetti:
There are several PRs to add new notifiers on AlertManager, and usually, the team will answer that webhooks should be used for now, while a more flexible solution is planned. This PR updates the doc to reflect this opinion.

PRs of new notifiers:
https://github.com/prometheus/alertmanager/pull/391
https://github.com/prometheus/alertmanager/pull/600
https://github.com/prometheus/alertmanager/pull/601
https://github.com/prometheus/alertmanager/pull/1021
https://github.com/prometheus/alertmanager/pull/885
https://github.com/prometheus/alertmanager/pull/1003
https://github.com/prometheus/alertmanager/pull/1012

@stuartnelson3

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -214,9 +214,7 @@ source_match_re:
 
 Receiver is a named configuration of one or more notification integrations.
 
-__Other receiver implementations available in version 0.0.4 of Alertmanager
-are not implemented yet. We are gladly accepting any contributions to add them
-to the new implementation.__
+__We're not actively adding new receivers, we recommend implementing custom notification integrations via the [webhook](/docs/alerting/configuration/#webhook_config) receiver.__
 
 ```
 # The unique name of the receiver.


stuartnelson3:
@fabxc @brancz @brian-brazil

content/docs/alerting/configuration.md
```diff
@@ -214,9 +214,7 @@ source_match_re:
 
 Receiver is a named configuration of one or more notification integrations.
 
-__Other receiver implementations available in version 0.0.4 of Alertmanager
-are not implemented yet. We are gladly accepting any contributions to add them
-to the new implementation.__
+__We're not actively adding notifiers at the moment as we try to figure out something more flexible. For the time being, we recommend implementing custom notifiers via the webhook notifier.__
```
brancz: I'm not sure how much it's about being on the search for something more flexible, the jsonnet approach would definitely be more flexible but I also see it being more error prone and can easily get out of sync as users would need to update those themselves separate from an Alertmanager update.

I think this is primarily because we can't take on the additional maintenance burden, therefore we chose to support only the major providers out of the box, and additional need to be added through the webhook integration.

My 2 cents, but happy to hear what others have to say.

brian-brazil: My thoughts are in line with @brancz

stuartnelson3: cool by me

josedonizetti: Wonder if the best action is just to remove this section of the documentation. What do you think?

brian-brazil: I'd leave it given the frequency of requests, but remove the "as we try to figure out something more flexible. For the time being,"

PR #845, opened 2017-10-04
Fix docker volume link in install guide page

wiennat:
Update data volume container link in Using Docker section to Using Volumes page in Docker Document.


wiennat:
Close this since it is already fixed by other PR.

PR #849, opened 2017-10-09
Add correct defaults to Alertmanager route config

josedonizetti:
Don't think we should have those open phrases "Usually ~X". I've changed it to have the correct defaults. 

https://github.com/prometheus/alertmanager/blob/master/dispatch/route.go#L31


brian-brazil:
I think the guidance on these values is okay to keep, users have a tendency to set them well outside these which causes them problems.

josedonizetti:
@brian-brazil Keep both, or close this?

brian-brazil:
The defaults should be documented.

josedonizetti:
Tried to be consistent with the notations for time (h, m, s) like it is used elsewhere in the docs.

brian-brazil:
Thanks!

PR #850, opened 2017-10-10
include solr exporter

fabriziofortino:


brian-brazil:
Does this produce metrics that the jmx exporter can't?

fabriziofortino:
It's a lightweight / non-obtrusive way to read metrics from the Solr MBean handler (https://lucene.apache.org/solr/guide/6_6/mbean-request-handler.html).

brian-brazil:
We try hard to not list things which overlap, so it's easier for end users to choose. As this is mBean based that means it (and more) is already covered by the JMX exporter, so this isn't something we'd list.

I'd suggest looking at submitting a PR for a jmx exporter config to cover Solr.

PR #863, opened 2017-10-11
Update documentation of the victorops alertmanager receiver

jperville:
This PR updates the documentation of the victorops alertmanager receiver, after the discussion in https://github.com/prometheus/alertmanager/issues/1018.

Summary:
 - the `message_type` attribute is a template string (or will be once https://github.com/prometheus/alertmanager/pull/1038 is merged)
 - the `state_message` attribute is also template string

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -429,14 +429,14 @@ api_key: <secret>
 # A key used to map the alert to a team.
 [ routing_key: <string> ]
 
-# Describes the behavior of the alert (Critical, Acknowledgement, Info, Recovery).
-[ message_type: <string> ]
+# Describes the behavior of the alert (CRITICAL, WARNING, INFO).
+[ message_type: <tmpl_string> | default = 'CRITICAL' ]
 
 # Contains summary of the alerted problem.
 [ entity_display_name: <string> | default = '{{ template "victorops.default.entity_display_name" . }}' ]
 
 # Contains long explanation of the alerted problem.
-[ state_message: <string> | default = '{{ template "victorops.default.state_message" . }}' ]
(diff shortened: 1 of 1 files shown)


content/docs/alerting/configuration.md
```diff
@@ -429,14 +429,14 @@ api_key: <secret>
 # A key used to map the alert to a team.
 [ routing_key: <string> ]
 
-# Describes the behavior of the alert (Critical, Acknowledgement, Info, Recovery).
-[ message_type: <string> ]
+# Describes the behavior of the alert (CRITICAL, WARNING, INFO, ACKNOWLEDGEMENT or RECOVERY).
+[ message_type: <tmpl_string> | default = 'CRITICAL' ]
```
stuartnelson3: The default message type when the alert is firing is `CRITICAL`, if the alert is resolved then it is hardcoded to send `RECOVERY`.

There's no support for the `ACKNOWLEDGEMENT` message, as it's not in the white list and will be over-written to `CRITICAL` when the alert is firing, or be changed to `RECOVERY` when the alert is resolved.

We should either remove acknowledgement from this list and note that recovery is automatically sent, or add acknowledgement to the whitelist in alertmanager.

jperville: Sure, let's remove ACKNOWLEDGEMENT from the list, since it is not used by alertmanager and there is a whitelist in the code.

jperville:
I just updated to remove mention of ACKNOWLEDGEMENT and RECOVERY, since the only values that are whitelisted by the code are CRITICAL, WARNING and INFO (see https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L845-L849).

jperville:
Thanks @stuartnelson3

PR #869, opened 2017-10-16
Add send_resolved doc to pushover, and victorops

josedonizetti:
I've added send_resolved to both pushover, and victorops.

1) Those are the only two not showing this option, form all other receivers
2) The send_resolved default is true for both, and for me make them even more important to document


brian-brazil:
Thanks!

PR #870, opened 2017-10-17
Add link_names option to slack_config

josedonizetti:
https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L223


josedonizetti:
I'll recreate this PR against the alertmanager repo now that docs will be moved.

PR #871, opened 2017-10-17
Add note and message doc to opsgenie_config

josedonizetti:
https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L300
https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L306

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -405,8 +405,12 @@ api_key: <secret>
 # The host to send OpsGenie API requests to.
 [ api_host: <string> | default = global.opsgenie_api_host ]
 
+# Alert text limited to 130 characters.
+[ message: <tmpl_string> ]
+
 # A description of the incident.
 [ description: <tmpl_string> | default = '{{ template "opsgenie.default.description" . }}' ]
+
 # A backlink to the sender of the notification.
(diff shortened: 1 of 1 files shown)


content/docs/alerting/configuration.md
```diff
@@ -405,8 +405,12 @@ api_key: <secret>
 # The host to send OpsGenie API requests to.
 [ api_host: <string> | default = global.opsgenie_api_host ]
 
+# Alert text limited to 130 characters
```
brian-brazil: There's no truncation done in the code if that's an actual API limit.

Comments should end in a full stop.

josedonizetti: @brian-brazil Got this text from the api documentation itself, where they say the limit exists. Should I remove it?
https://docs.opsgenie.com/docs/alert-api-deprecated

brian-brazil: It's more that we should implement it on the alertmanager side, as we do for some other limited fields.

josedonizetti: @brian-brazil Gotcha! Gonna create an issue there and work on it.

brian-brazil:
Thanks!

PR #872, opened 2017-10-17
Fix field type on victorops_config

josedonizetti:
https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L881
https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L883


brian-brazil:
Thanks!

PR #873, opened 2017-10-18
Add syntax link for filter option in gce_sd_config 

esloho:
The filter option in gce_sd_config has a particular syntax with some mandatory elements and it even accepts regular expressions. It would be useful to add a link to the syntax reference (I miss it myself when trying to use it and spent too much time looking for this information).


brian-brazil:
Thanks!

PR #876, opened 2017-10-19
Add kafka_exporter to exporters

danielqsj:
A faster and easier-to-use kafka exporter.

Compared with current two exporters: `Kafka consumer group exporter` and `Kafka ZooKeeper exporter`.
kafka_exporter get metrics directly from kafka cluster, which is more faster and easier-to-use.
And metrics of kafka_exporter contains all metrics of the above two exporters.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -58,8 +58,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
 
 ### Messaging systems
    * [Beanstalkd exporter](https://github.com/messagebird/beanstalkd_exporter)
-   * [Kafka consumer group exporter](https://github.com/kawamuray/prometheus-kafka-consumer-group-exporter)
-   * [Kafka ZooKeeper exporter](https://github.com/cloudflare/kafka_zookeeper_exporter)
+   * [Kafka exporter](https://github.com/danielqsj/kafka_exporter)
    * [NATS exporter](https://github.com/nats-io/prometheus-nats-exporter)
    * [NSQ exporter](https://github.com/lovoo/nsq_exporter)
    * [Mirth Connect exporter](https://github.com/vynca/mirth_exporter)


brian-brazil:
We try to avoid duplication in this list, so if this is a true superset then it should replace the existing two exporters. Having one is also a better experience for users. 
This is the first time we've had a superset like this, so I want to be sure it's better. @kawamuray @prymitive

Port 9206 is already taken, 9308 is the next free one. Please register at https://github.com/prometheus/prometheus/wiki/Default-port-allocations

There should be a reference in the readme to the jmx exporter for the metrics kafka itself exposes.

Should you be accepting multiple addresses for brokers, rather than just one, for redundancy?

danielqsj:
@brian-brazil Thanks. Port change and multiple addresses support will be implemented in next release.

@kawamuray Thanks for your work about `Kafka consumer group exporter`, but `kafka-consumer-groups.sh` is too slow with the cost of jvm. Directly get metric from kafka cluster is far more faster. In my test, the time spent is reduced from the minute level to the second level.

@prymitive This exporter is using [sarama](https://github.com/Shopify/sarama) which is based on [Kafka Protocol](http://kafka.apache.org/protocol.html) I also have considered about what you said. There are three reasons I choose to get metrics directly from kafka.

1. [Sarama](https://github.com/Shopify/sarama) has a good [mechanism ](https://github.com/Shopify/sarama/wiki/Cluster-metadata) to try best to get correct metadata from healthy brokers. By the way, some metrics will collected from [leader broker](https://github.com/Shopify/sarama/blob/630b33cbf6f5c5af999918be52bac32cf52d0176/client.go#L563) which is definitely correct.

1. Split brain may also happen in zookeeper cluster. So if `Kafka ZooKeeper exporter` is configured to unhealthy zk node, it will also report bogus data. If some zk nodes are unhealthy, directly connecting to kafka cluster may be a better way to get correct metrics if zk cluster is still working. So these problem will not be completely solved by connecting to zk instead of kafka.

1. **Most important**, Kafka consumers offset storage mechanism in Zookeeper will be [deprecated](https://kafka.apache.org/documentation/#impl_zkconsumers) in a future release. So correct metrics from kafka directly is better in my option.

brian-brazil:
Sounds like there's no strong objections. Can you change this PR to replace the existing two?

danielqsj:
Done, thanks for your review. :)

brian-brazil:
Thanks!

PR #878, opened 2017-10-20
Tweak marathon wording around clustering.

brian-brazil:
The current wording has led several users to believe this
works across multiple clusters, which is not the case.


xperimental:
Sounds good to me. It's good that it is a bit shorter, but I don't know if everyone will pick up on the "in a cluster" part.

brian-brazil:
Reference documentation tends to be concise, we can always see if higher than usual confusion arises from this.

PR #879, opened 2017-10-20
Clarify what are valid alert names.

### LINKED ISSUE #636: Clarify naming restrictions for recording and alerting rules
They follow the same restrictions as metric names https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
but we don't mention that anywhere explicitly, AFAIK.

brian-brazil:
Fixes #636 

@beorn7 

The current recording rule docs say "time series name" , and the docs already cover the syntax for those. That bit is all getting rewritten soon for 2.0 anyway.


beorn7:
:+1:

brian-brazil:
I've just written the docs for 2.0 and ensured this is covered for both.

PR #880, opened 2017-10-22
The promtool ships with the server

jamtur01:
You could they could grab the `promtool` standalone with `go get` but that assumes they have Go installed and setup. I think it's more likely they have the Prometheus server download.

### DIFF
--- a/content/docs/querying/rules.md
+++ b/content/docs/querying/rules.md
@@ -17,11 +17,11 @@ process. The changes are only applied if all rule files are well-formatted.
 
 ## Syntax-checking rules
 To quickly check whether a rule file is syntactically correct without starting
-a Prometheus server, install and run Prometheus's `promtool` command-line
-utility tool:
+a Prometheus server you can run Prometheus's `promtool` command-line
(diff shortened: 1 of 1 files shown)


content/docs/querying/rules.md
```diff
@@ -17,11 +17,11 @@ process. The changes are only applied if all rule files are well-formatted.
 
 ## Syntax-checking rules
 To quickly check whether a rule file is syntactically correct without starting
-a Prometheus server, install and run Prometheus's `promtool` command-line
-utility tool:
+a Prometheus server you can run Prometheus's `promtool` command-line
+utility tool. The `promtool` binary ships with the Prometheus server
+download.
```
brian-brazil: mentioning the tarball would be clearer, as we also ship docker images

brian-brazil:
For Prometheus 2.0 we're in the middle of moving this document over to the Prometheus repo, would you mind making this change against https://github.com/prometheus/prometheus/blob/master/docs/configuration/recording_rules.md ?

PR #881, opened 2017-10-22
Add note to clarify range vector selector

bngsudheer:
So that the user's don't have confusion when following the documentation.


brian-brazil:
Thanks for the PR, documentation on the expression browser does not belong in this page. I'd suggest changing the expression browser's page instead.

beorn7:
Closing for lack of followup.

PR #883, opened 2017-10-24
Fixing run example

edmz:
The parameters needs double dash otherwise I get:

    Error parsing commandline arguments: unknown short flag '-c'

in linux at least.


brian-brazil:
This example is for Prometheus 1.x, for which it is correct.

edmz:
I am sorry, I thought I was editing the v2 docs.

PR #884, opened 2017-10-26
Remove me from community.html, replace with Kausal

tomwilkie:


content/community.html
```diff
@@ -78,7 +78,7 @@
       <li><a href="https://juliusv.com">Julius Volz</a> (independent contractor)</li>
       <li><a href="https://container-solutions.com">Container Solutions</a></li>
       <li><a href="https://tincho.org">Martín Ferrari</a> (independent contractor)</li>
-      <li><a href="mailto:tom.wilkie@gmail.com">Tom Wilkie</a> (independent contractor)</li>
+      <li><a href="https://kausal.co">Kausal</a> - hosted Prometheus, support and consulting.</li>
```
brian-brazil: To be consistent the services shouldn't be listed here, as none of the others list them.

grobie: So whoever wants to add them, is free to do so?

brian-brazil: We could do that, I'm a little wary of getting into having to police this. For example what if someone decides to call their service "Premium Consulting" in order to game this list? I think the safest thing would be just a link to wherever on the org's website.

tomwilkie: We hold the keys to this repo, so we can just say no to PRs like that.  I only added it as the paragraph above doesn't cover everything Kausal is doing:

> ...third-party companies and individuals who provide support or consulting services for Prometheus...

We could change that to say:

> ...third-party companies and individuals who provide products or services related to Prometheus...

If you like?

brian-brazil: That makes sense to me, it needs updating with adding Kausal in any case.

PR #886, opened 2017-10-27
fix broken link

miqui:
fix broken link


grobie:
That link seems to work fine? Please don't use absolute urls including hostname, that breaks dev/test setups.

josedonizetti:
Yup, works properly.

PR #887, opened 2017-10-30
Documented global defaults for email_config

pracucci:


brian-brazil:
:+1: 

@grobie Is it safe to merge this? It's alertmanager docs.

PR #888, opened 2017-10-31
Replace github.com/fabxc/jiralerts with github.com/alin-sinpalean/jiralert

alin-sinpalean:
As a followup to https://github.com/prometheus/alertmanager/pull/1012 and the discussion there, I am adding JIRAlert as the webhook handler for JIRA.


brian-brazil:
Thanks!

PR #891, opened 2017-11-02
Update security docs for 2.0 secure by default flags.

brian-brazil:


grobie:
This page was moved in the versioned-docs branch. Having an entire section for a single page doens't sound right.

brian-brazil:
Where did you put Integrations? This would be similar general operation stuff.

grobie:
There should be only one place where we document flags. This page should
only mention that admin + lifecycle endpoints are dialed by default and
link to the Prometheus config documenting the flags.

brian-brazil:
One of the main purposes of this page is to act as a comprehensive one-stop security audit checklist for Prometheus. Thus exact flag and configuration field names are mentioned, so that they've nice and easy to then grep in config files.

Think of it as a guide - we provide full examples in guides rather than letting the user figure it out from the reference docs.

PR #892, opened 2017-11-02
Add link to webhook receiver implementations

alin-amana:
Might be useful for readers of the documentation to check for existing webhook implementations before they start rolling their own.

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -215,7 +215,7 @@ source_match_re:
 
 Receiver is a named configuration of one or more notification integrations.
 
-__We're not actively adding new receivers, we recommend implementing custom notification integrations via the [webhook](/docs/alerting/configuration/#webhook_config) receiver.__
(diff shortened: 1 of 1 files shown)


content/docs/alerting/configuration.md
```diff
@@ -215,7 +215,7 @@ source_match_re:
 
 Receiver is a named configuration of one or more notification integrations.
 
-__We're not actively adding new receivers, we recommend implementing custom notification integrations via the [webhook](/docs/alerting/configuration/#webhook_config) receiver.__
+__We're not actively adding new receivers, we recommend implementing custom notification integrations via the [webhook](/docs/alerting/configuration/#webhook_config) receiver. Check [this list](https://prometheus.io/docs/operating/integrations/#alertmanager-webhook-receiver) for already implemented webhook receiver integrations.__
```
brian-brazil: This whole paragraph should really be moved to the Alertmanager readme/contribution docs rather than in the user-facing docs, as we're long past the switch from 0.0.4 by now.

There's also already a link down in the webhook section.

alin-amana: Didn't notice the link at the bottom. Sounds very reasonable.

PR #895, opened 2017-11-04
A first guide

jamtur01:
Added a draft of a "first steps" guide. Very rough, could be expanded or
contracted and very open to feedback. I used some of the browser material from the Getting Started on the assumption that some of that would be truncated. 

I added the scaffolding for guides as an example of how they might work but not attached to their menu placement, nav icon, etc, etc.


jamtur01:
@brian-brazil As discussed - just wanted to get something down for comment before investing too much time into it. Feedback very much appreciated.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
+---
+title: First steps
+sort_rank: 3
+---
+
+# First steps with Prometheus
+
+Welcome to Prometheus! Prometheus is a monitoring platform that collects metrics from monitored targets by scraping metrics HTTP endpoints on these targets. This guide will show you to how to install, configure and monitor our first resource with Prometheus. You'll download, install and run Prometheus. You'll also download and install an exporter, plugins that expose time series data on hosts and services. Our first exporter will be the Node Exporter. This exporter exposes host-level metrics like CPU, memory, and disk.
```
brian-brazil: Exporters aren't plugins/

brian-brazil: Exporters aren't plugins, Prometheus doesn't have plugins currently

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
...
+
+Prometheus configuration is [YAML](http://www.yaml.org/start.html). The Prometheus download comes with a sample configuration in a file called `prometheus.yml` that's a good place to get started.
+
+We've stripped out the comments in the example file to make it more succinct.
+
+```language-yaml
+global:
+  scrape_interval:     15s
+  evaluation_interval: 15s
+  external_labels:
```
brian-brazil: I'd not mention external labels here, that's a deployment thing

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Starting Prometheus
+
+To start Prometheus with our newly created configuration file, change to the directory containing the Prometheus binary and run:
+
+```language-bash
+./prometheus --config.file=prometheus.yml
+```
+
+Prometheus should start up. You should also be able to browse to a status page about itself at `http://localhost:9090`. Give it a couple of seconds to collect data about itself from its own HTTP metrics endpoint.
```
brian-brazil: With this configuration it'll take 30s to have useful information

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
...
+```
+prometheus_target_interval_length_seconds
+```
+
+This should return a number of different time series (along with the latest value
+recorded for each), all with the metric name
+`prometheus_target_interval_length_seconds`, but with different labels. These
+labels designate different latency percentiles and target group intervals.
+
+If we were only interested in the 99th percentile latencies, we could use this
```
brian-brazil: I'd keep away from summary quantiles in this, as they're not aggregatable. They're also quantiles, not percentiles.

I'd suggest finding a simple gauge.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Using the graphing interface
+
+To graph expressions, navigate to `http://localhost:9090/graph` and use the "Graph" tab.
+
+For example, enter the following expression to graph the per-second rate of all
+storage chunk operations happening in the self-scraped Prometheus:
+
+```
+rate(prometheus_local_storage_chunk_ops_total[1m])
```
brian-brazil: This is a 1.x metric that's not in 2.x

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+You can experiment with the graph range parameters and other settings.
+
+## Installing the Node Exporter
+
+Collecting metrics from Prometheus alone isn't a good representation of Prometheus' capabilities. So let's use the Node Exporter to monitor our first resource. We're going to monitor a Linux host called `node.example.com` but you could monitor any Linux or OS X host. There's also [a WMI exporter](https://github.com/martinlindhe/wmi_exporter) for Microsoft Windows hosts too.
```
brian-brazil: We try to avoid conjunctions like "isn't", apparently they're harder to read.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
...
+
+Let's start the node exporter now on our Linux host.
+
+```language-bash
+./node_exporter
+```
+
+The Node Exporter's metrics are available on port `9100` on the host at the `/metrics` path. In our case this is:
+
+`http://node.example.com:9100/metrics`
```
brian-brazil: localhost?

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
...
+
+We will configure Prometheus to scrape this new target. To achieve this, add the following job definition to the `scrape_configs` section in our `prometheus.yml`:
+
+```
+scrape_configs:
+  - job_name: 'prometheus'
+    static_configs:
+      - targets: ['localhost:9090']
+  - job_name:  'node-example'
+    scrape_interval: 5s
```
brian-brazil: Varying scrape intervals inside a Prometheus is to be discouraged. It's an advanced topic that shouldn't be mentioned in a getting started, as we already have too many users over-engineering their setups in this way.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
...
+    static_configs:
+      - targets: ['node.example.com:9100']
+        labels:
+          group: 'nodes'
```
brian-brazil: You already have the job label, don't add a redundant label with the same information

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
...
+## Configuring Prometheus to monitor the host
+
+We will configure Prometheus to scrape this new target. To achieve this, add the following job definition to the `scrape_configs` section in our `prometheus.yml`:
+
+```
+scrape_configs:
+  - job_name: 'prometheus'
+    static_configs:
+      - targets: ['localhost:9090']
+  - job_name:  'node-example'
```
brian-brazil: This is just `node` by convention

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,185 @@
...
+    scrape_interval: 5s
+    static_configs:
+      - targets: ['node.example.com:9100']
+        labels:
+          group: 'nodes'
+```
+
+Our new job is called `node-example`. It scrapes a static target, `node.example.com` on port `9100`. You would replace this name with the name or IP address of the host you're monitoring. We also add a label, `group` with a value of `nodes`, to our time series.
+
+Now we restart or `SIGHUP` our Prometheus instance to activate our new job.
```
brian-brazil: SIGHUP won't work for Windows users, they'll need to activate the lifecycle API and hit the http endpoint. I'd stick with a restart here for simplicitly.

content/docs/guides/first_steps.md
jamtur01: This was more a consequence of using the example config. You open to editing that from that example?

content/docs/guides/first_steps.md
jamtur01: This is lifted from the getting started guide. I'll trim it.

content/docs/guides/first_steps.md
jamtur01: Fixed.

content/docs/guides/first_steps.md
jamtur01: See https://github.com/prometheus/prometheus/pull/3409.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,178 @@
...
+    static_configs:
+      - targets: ['localhost:9090']
+  - job_name:  'node'
+    static_configs:
+      - targets: ['localhost:9100']
+        labels:
+          role: 'monitoring'
+```
+
+Our new job is called `node`. It scrapes a static target, `localhost` on port `9100`. You would replace this name with the name or IP address of the host you're monitoring. We also add a label, `role` with a value of `monitoring`, to our time series.
```
roidelapluie: Maybe clarify that you add role: monitoring because the server is a monitor server; and that label names and called could be anything

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,178 @@
...
+
+Now we restart our Prometheus server to activate our new job.
+
+Go to the expression browser and verify that Prometheus now has information
+about the time series that this endpoint exposes, you'll see a collection of new metrics, such as the `node_cpu` metric.
+
+More?
```
roidelapluie: It took me a while to "discover" the up metric. Maybe it deserves some space here. At least to help people debugging their first install.

content/docs/guides/first_steps.md
jamtur01: Good point.

content/docs/guides/first_steps.md
jamtur01: Yeah - me too. :)

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,178 @@
...
+To start Prometheus with our newly created configuration file, change to the directory containing the Prometheus binary and run:
+
+```language-bash
+./prometheus --config.file=prometheus.yml
+```
+
+Prometheus should start up. You should also be able to browse to a status page about itself at `http://localhost:9090`. Give it about 30 seconds to collect data about itself from its own HTTP metrics endpoint.
+
+You can also verify that Prometheus is serving metrics about itself by
+navigating to its own metrics endpoint: `http://localhost:9090/metrics`.
```
brian-brazil: This would be more useful as a link.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,178 @@
...
+```
+scrape_configs:
+  - job_name: 'prometheus'
+    static_configs:
+      - targets: ['localhost:9090']
+  - job_name:  'node'
+    static_configs:
+      - targets: ['localhost:9100']
+        labels:
+          role: 'monitoring'
```
brian-brazil: This shouldn't be here, target label taxonomy is a nuanced thing that doesn't belong in a getting started guide. Blindly adding target labels will cause more problems for new users than anything else.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,178 @@
...
+
+`http://localhost:9100/metrics`
+
+You can browse to this URL to see the metrics being exposed.
+
+We now need to tell Prometheus about our new target.
+
+## Configuring Prometheus to monitor the host
+
+We will configure Prometheus to scrape this new target. To achieve this, add the following job definition to the `scrape_configs` section in our `prometheus.yml`:
```
brian-brazil: You should probably make it clearer that there's only one scrape_configs. This could be read as appending this whole block to the existing file.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,178 @@
...
+        labels:
+          role: 'monitoring'
+```
+
+Our new job is called `node`. It scrapes a static target, `localhost` on port `9100`. You would replace this name with the name or IP address of the host you're monitoring. We also add a label, `role` with a value of `monitoring`, to our time series. We're using this label to identify the role of the host being as being our monitoring server. You could add any label you want to the time series to help identify its origin.
+
+Now we restart our Prometheus server to activate our new job.
+
+Go to the expression browser and verify that Prometheus now has information
+about the time series that this endpoint exposes, you'll see a collection of new metrics, such as the `node_cpu` and `up` metrics. The `up` metric is useful and can be used to track the status of the node.
```
brian-brazil: Where would you see this collection of new metrics?

content/docs/guides/first_steps.md
jamtur01: Okay - given that it is a very common way to categorize monitoring targets in almost every monitoring tool I think you need to introduce it early. :) I'll pull this out of here though. Nothing I've read elsewhere raises these risks so perhaps a label-specific guide might be useful? Or an expansion of one of the pages that already talks about labels?

jamtur01:
Okay - updated based on feedback. I'd be interested in folk's thoughts on what more to include, if anything.

content/docs/guides/first_steps.md
```diff
@@ -154,25 +154,20 @@ We now need to tell Prometheus about our new target.
...
 Now we restart our Prometheus server to activate our new job.
 
 Go to the expression browser and verify that Prometheus now has information
-about the time series that this endpoint exposes, you'll see a collection of new metrics, such as the `node_cpu` and `up` metrics. The `up` metric is useful and can be used to track the status of the node.
+about the time series that this endpoint exposes. Navigate to
+http://localhost:9090/graph and use the dropdown next to the "Execute" button to see a list of metrics this server is collecting. In the list you'll see a number of metrics prefixed with `node_`, that have been collected by the Node Exporter by our `node` job. For example, you can see the node's CPU usage via the `node_cpu` metric. 
 
-More?
\ No newline at end of file
+One useful metric to look for is the `up` metric. The `up` metric can be used to track the status of the target. If the metric has a value of `1` then the scrape of the target was successful, if `0` it failed. This can help give you an indication of the status of the target. You'll see two `up` metrics, one for each target we're scraping: the Prometheus server and the metrics exported by the Node Exporter.
```
roidelapluie: LGTM

content/docs/guides/first_steps.md
brian-brazil: https://www.robustperception.io/target-labels-are-for-life-not-just-for-christmas/

jamtur01: Love the title! :)

brian-brazil: I do try :)

content/docs/guides/first_steps.md
brian-brazil: This is still here

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,173 @@
...
+
+Our new job is called `node`. It scrapes a static target, `localhost` on port `9100`. You would replace this name with the name or IP address of the host you're monitoring. 
+
+Now we restart our Prometheus server to activate our new job.
+
+Go to the expression browser and verify that Prometheus now has information
+about the time series that this endpoint exposes. Navigate to
+http://localhost:9090/graph and use the dropdown next to the "Execute" button to see a list of metrics this server is collecting. In the list you'll see a number of metrics prefixed with `node_`, that have been collected by the Node Exporter by our `node` job. For example, you can see the node's CPU usage via the `node_cpu` metric. 
+
+One useful metric to look for is the `up` metric. The `up` metric can be used to track the status of the target. If the metric has a value of `1` then the scrape of the target was successful, if `0` it failed. This can help give you an indication of the status of the target. You'll see two `up` metrics, one for each target we're scraping: the Prometheus server and the metrics exported by the Node Exporter.
```
brian-brazil: s/the metrics exported by//

the metrics aren't the target.

grobie:
I don't think we need to version guides. Given they will usually touch multiple components, having them in prometheus/docs makes sense to me. I think that's what @fabxc and @juliusv thought as well.

brian-brazil:
Ah yes, this is a node exporter guide now too. Makes sense.

Should we be looking to remove the existing guide from the prometheus repo then?

grobie:
You mean the migration guide? First I was confused seeing that in the docs, given we had this discussion about keeping general guides in docs. But then for the migration guide it kinda makes sense, given it'll be a different one for each major release. On the other hand, too many places might be confusing.

I'm slightly leaning towards leaving migration guides in repos, but everything else in docs.

brian-brazil:
I meant the Getting Started guide: https://prometheus.io/docs/prometheus/latest/getting_started/

> I'm slightly leaning towards leaving migration guides in repos, but everything else in docs.

I agree on migration guides. I think that guides specific to a repo should live in the repo, but more high level/overall stuff can go in docs.

grobie:
Ah yes. That makes sense! I think we moved it because of different flags and such, but I'd just say guides should reflect the latest version and that's it.

jamtur01:
I added a quick summary just in case anyone wants to deploy this with 2.0.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,175 @@
...
+The Prometheus monitoring server
+
+. . .
+```
+
+Before starting Prometheus, let's configure it. 
+
+## Configuring Prometheus
+
+Prometheus configuration is [YAML](http://www.yaml.org/start.html). The Prometheus download comes with a sample configuration in a file called `prometheus.yml` that's a good place to get started.
```
roidelapluie: Missing a . after `prometheus.yml`,

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,175 @@
...
+  - job_name: 'prometheus'
+    static_configs:
+      - targets: ['localhost:9090']
+```
+
+There are three blocks of configuration in the example configuration file: `global`, `rule_files`, and `scrape_configs`. 
+
+The `global` block controls the Prometheus server's global configuration. We have three options present. The first, `scrape_interval`, controls how often Prometheus will scrape targets. You can override this for individual targets. In this case the global setting is to scrape every 15 seconds. The `evaluation_interval` option controls how often Prometheus will evaluate rules. Prometheus uses rules to create new time series and to generate alerts. The last option, `external_labels`, adds labels to time series or alerts that leave the Prometheus server.
+
+The `rule_files` block specifies the location of any rules we want the Prometheus server to load. For now we've got no rules.
```
roidelapluie: (# indicates a comment in Yaml)

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,175 @@
...
+
+The last block, `scrape_configs`, controls what resources Prometheus monitors. Since Prometheus also exposes data about itself as an HTTP endpoint it can scrape and monitor its own health. In the default configuration there is a single job, called `prometheus`, which scrapes the time series data exposed by the Prometheus server. The job contains a single, statically configured, target, the `localhost` on port `9090`. Prometheus expects metrics to be available on targets on a path of `/metrics`. So this default job is scraping via the URL:
+
+`http://localhost:9090/metrics`
```
roidelapluie: make that a link?

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,175 @@
...
+
+For a complete specification of configuration options, see the
+[configuration documentation](/docs/operating/configuration).
+
+## Starting Prometheus
+
+To start Prometheus with our newly created configuration file, change to the directory containing the Prometheus binary and run:
+
+```language-bash
+./prometheus --config.file=prometheus.yml
```
roidelapluie: prometheus.yml is default. maybe it is not needed to specify it here.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+Prometheus should start up. You should also be able to browse to a status page about itself at http://localhost:9090. Give it about 30 seconds to collect data about itself from its own HTTP metrics endpoint.
+
+You can also verify that Prometheus is serving metrics about itself by
+navigating to its own metrics endpoint: http://localhost:9090/metrics.
```
roidelapluie: how is that link rendered? does it have a . ?

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,175 @@
...
+./node_exporter
+```
+
+The Node Exporter's metrics are available on port `9100` on the host at the `/metrics` path. In our case this is:
+
+`http://localhost:9100/metrics`
+
+You can browse to this URL to see the metrics being exposed.
+
+We now need to tell Prometheus about our new target.
```
roidelapluie: s/new target/exporter/ ?

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Configuring Prometheus to monitor the host
+
+We will configure Prometheus to scrape this new target. To achieve this, add a new job definition to the `scrape_configs` section in our `prometheus.yml`:
+
+```
+  - job_name:  'node'
```
roidelapluie: two spaces and unneeded '

roidelapluie:
Loogs very good. Small comments but nothing blocking.

content/docs/guides/first_steps.md
jamtur01: Not sure I understand? That's one sentence.

content/docs/guides/first_steps.md
jamtur01: Yeah I'd prefer to leave it in given we showed folks the file earlier.

content/docs/guides/first_steps.md
jamtur01: It's rendered correctly.

content/docs/guides/first_steps.md
roidelapluie: oh yes my bad

content/docs/guides/first_steps.md
roidelapluie: ok

roidelapluie:
Thanks @jamtur01 I have no further comment.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,173 @@
...
+
+scrape_configs:
+  - job_name: 'prometheus'
+    static_configs:
+      - targets: ['localhost:9090']
+```
+
+There are three blocks of configuration in the example configuration file: `global`, `rule_files`, and `scrape_configs`. 
+
+The `global` block controls the Prometheus server's global configuration. We have three options present. The first, `scrape_interval`, controls how often Prometheus will scrape targets. You can override this for individual targets. In this case the global setting is to scrape every 15 seconds. The `evaluation_interval` option controls how often Prometheus will evaluate rules. Prometheus uses rules to create new time series and to generate alerts. The last option, `external_labels`, adds labels to time series or alerts that leave the Prometheus server.
```
brian-brazil: external_labels is no longer here

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,173 @@
...
+```
+count(http_requests_total)
+```
+
+For more about the expression language, see the
+[expression language documentation](/docs/querying/basics/).
+
+## Using the graphing interface
+
+To graph expressions, navigate to `http://localhost:9090/graph` and use the "Graph" tab.
```
brian-brazil: Sometimes you use backticks on urls, other times you don't. It would be good to be consistent.

content/docs/guides/first_steps.md
```diff
@@ -0,0 +1,173 @@
+---
+title: First steps
```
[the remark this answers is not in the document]
jamtur01: Would you like to replace the existing getting started or brand it some other way?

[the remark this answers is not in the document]
brian-brazil: The existing getting started already moved over to the Prometheus repo, which we'll now want to remove. You should be able to slot directly into the current introduction section.

jamtur01:
@brian-brazil Okay done.

brian-brazil:
Looks like the getting_started in this repo is defunct.

Thanks!

PR #902, opened 2017-11-08
Use escaped #<remote_read> and #<remote_write> in integration links.

rektide:
Neither of these anchor tags work atm.

Escape the links in the same way known-working FAQ.md does.
See the link to configuration#<scrape_config>:
https://prometheus.io/docs/introduction/faq/#why-don-t-the-prometheus-server-components-support-tls-or-authentication-can-i-add-those


brian-brazil:
Thanks!

PR #905, opened 2017-11-09
Redirect docs/prometheus/latest/querying/rules

roidelapluie:
Fixes prometheus/prometheus#3451


grobie:
I think we should fix the pages in question instead. Can you update them to point to recording_rules please?

roidelapluie:
@grobie done

PR #908, opened 2017-11-13
Fix Alertmanager link

povesteam:
The broken link is currently at the bottom of [this page](https://prometheus.io/docs/alerting/rules/)


brian-brazil:
Actually this page should be deleted, it has moved to under Prometheus.

povesteam:
Are you sure?

Go to https://prometheus.io/, click on `DOCS` (top), then `ALERTING` (left) and `Alerting rules`, and you will land there https://prometheus.io/docs/alerting/rules/

It's not a big issue, though. 😄

brian-brazil:
That link would need to be updated to https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/

PR #912, opened 2017-11-15
fix examples documentation

rdemachkovych:
@brian-brazil 
Fix examples documentation.
If used examples from documentation receive the error below:
Checking '/etc/alertmanager/alertmanager.yml'  FAILED: yaml: unmarshal errors:
  line 5: cannot unmarshal !!seq into config.plain


brian-brazil:
Thanks!

PR #917, opened 2017-11-18
Improve documentation for email sending

bjoernhaeuser:
This is my try to work on prometheus/alertmanager#384

I tried to make the example concise as possible. 

What do you think? Is this a way of going forward?

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -63,15 +63,19 @@ global:
 
   # The default SMTP From header field.
   [ smtp_from: <tmpl_string> ]
-  # The default SMTP smarthost used for sending emails.
+  # The default SMTP smarthost used for sending emails, including port number.
+  # Port number usually is 25, or 587 for SMTP over TLS (sometimes referred to as STARTTLS).
+  # Example: smtp.example.org:587
   [ smtp_smarthost: <string> ]
   # The default hostname to identify to the SMTP server.
   [ smtp_hello: <string> | default = "localhost" ]
-  # SMTP authentication information.
   [ smtp_auth_username: <string> ]
+  # SMTP Auth using LOGIN and PLAIN.
   [ smtp_auth_password: <secret> ]
-  [ smtp_auth_secret: <secret> ]
+  # SMTP Auth using PLAIN.
   [ smtp_auth_identity: <string> ]
+  # SMTP Auth using CRAM-MD5. 
+  [ smtp_auth_secret: <secret> ]
   # The default SMTP TLS requirement.
   [ smtp_require_tls: <bool> | default = true ]
 


content/docs/alerting/configuration.md
```diff
@@ -64,10 +64,16 @@ global:
...
   [ smtp_from: <tmpl_string> ]
   # The default SMTP smarthost used for sending emails.
+  # Provide the correct port for the smarthost:
+  # 25 for servers not providing TLS / SSL (set smtp_require_tls to "false" 
+  # 587 for STARTTLS servers
+  # 465 for SMTPS severs
+  # example: smtp.example.org:587
   [ smtp_smarthost: <string> ]
-  # SMTP authentication information.
+  # SMTP authentication information for SMTP Auth using PLAIN / LOGIN
```
brian-brazil: Comments should end with a full stop.

content/docs/alerting/configuration.md
```diff
@@ -64,10 +64,16 @@ global:
   # The default SMTP From header field.
   [ smtp_from: <tmpl_string> ]
   # The default SMTP smarthost used for sending emails.
+  # Provide the correct port for the smarthost:
```
brian-brazil: This is takes up a bit much space for reference docs.  I'd mention that the smarthost must include a port, and quickly run down the ports rather than give them a line each.

content/docs/alerting/configuration.md
```diff
@@ -64,10 +64,16 @@ global:
...
+  # 25 for servers not providing TLS / SSL (set smtp_require_tls to "false" 
+  # 587 for STARTTLS servers
```
brian-brazil: I don't believe that's the name of the protocol.

content/docs/alerting/configuration.md
brian-brazil: I don't think this is accurate, you need identity, username and password for gmail.

content/docs/alerting/configuration.md
bjoernhaeuser: STARTTLS is the name of the SMTP command: https://en.wikipedia.org/wiki/Opportunistic_TLS

bjoernhaeuser:
@brian-brazil I changed the config. Can you take another look please?

content/docs/alerting/configuration.md
brian-brazil: So it's not the name of the protocol in question, and thus not the name of the port

content/docs/alerting/configuration.md
```diff
@@ -64,12 +64,17 @@ global:
   # The default SMTP From header field.
   [ smtp_from: <tmpl_string> ]
   # The default SMTP smarthost used for sending emails.
+  # Provide the correct port for the smarthost (25 / 587 for SMTP over TLS (STARTTLS), 465 for deprecated SMTPS). 
+  # Example: smtp.example.org:587
   [ smtp_smarthost: <string> ]
-  # SMTP authentication information.
+  # Set the username if required by the server.
```
brian-brazil: This seems redundant, the field name already tells me this.

content/docs/alerting/configuration.md
bjoernhaeuser: I changed the explanation by a bit to make this distinction clear

content/docs/alerting/configuration.md
```diff
@@ -64,12 +64,17 @@ global:
   # The default SMTP From header field.
   [ smtp_from: <tmpl_string> ]
   # The default SMTP smarthost used for sending emails.
+  # Provide the correct port for the smarthost (25 / 587 for SMTP over TLS (STARTTLS), 465 for deprecated SMTPS).
```
brian-brazil: Is this field for the smarthost or the port? I can see this confusing users.

"The default SMTP smarthost used for sending emails, including port number." would be clearer.

brian-brazil: SMTPS is also SMTP over TLS, so this is a bit confusing.

content/docs/alerting/configuration.md
```diff
@@ -64,12 +64,17 @@ global:
   # The default SMTP From header field.
   [ smtp_from: <tmpl_string> ]
   # The default SMTP smarthost used for sending emails.
+  # Provide the correct port for the smarthost (25 / 587 for SMTP over TLS (STARTTLS), 465 for deprecated SMTPS). 
+  # Example: smtp.example.org:587
   [ smtp_smarthost: <string> ]
-  # SMTP authentication information.
+  # Set the username if required by the server.
   [ smtp_auth_username: <string> ]
+  # SMTP authentication information for SMTP Auth using LOGIN.
```
brian-brazil: This is incorrect, it's also used by PLAIN.

Any documentation here needs to be 100% correct and precise as it is reference documentation, not just what works in some setups.

content/docs/alerting/configuration.md
bjoernhaeuser: I am not sure how to properly write this. What do you think about removing the mentioning of `SMPTS` completely, because its actually not really in use anymore?

brian-brazil: If it's not what the user is after, there's no real need to mention it

bjoernhaeuser: Removed. :)

content/docs/alerting/configuration.md
```diff
@@ -63,13 +63,17 @@ global:
...
   # The default SMTP From header field.
   [ smtp_from: <tmpl_string> ]
-  # The default SMTP smarthost used for sending emails.
+  # The default SMTP smarthost used for sending emails, including port number.
+  # Port number should be: 25 / 587 for SMTP over TLS (sometimes referred to as STARTTLS). 
+  # Example: smtp.example.org:587
   [ smtp_smarthost: <string> ]
-  # SMTP authentication information.
   [ smtp_auth_username: <string> ]
+  # SMTP authentication information for SMTP Auth using LOGIN and PLAIN.
```
brian-brazil: This is stuttering, we're saying SMTP auth 3 times

bjoernhaeuser:
Removed stuttering and rebased to current master.

bjoernhaeuser:
My next try :) 

If this is not good enough I would say I am not good enough to write documentation for prometheus.

content/docs/alerting/configuration.md
```diff
@@ -63,13 +63,17 @@ global:
 
   # The default SMTP From header field.
   [ smtp_from: <tmpl_string> ]
-  # The default SMTP smarthost used for sending emails.
+  # The default SMTP smarthost used for sending emails, including port number.
+  # Port number usually is 25 or 587 for SMTP over TLS (sometimes referred to as STARTTLS).
```
brian-brazil: Comma after 25

RichiH: Oxford commas are usually omitted if you only have two short items.

brian-brazil: This isn't an Oxford comma

stuartnelson3:
@brian-brazil what remains to get this merged?

brian-brazil:
Just looks like conflicts need resolving.

bjoernhaeuser:
Rebased to current master.

PR #919, opened 2017-11-19
Prevent regex matches on undefined labels

garretruh:
In the routing tree editor, given the following `config.yml`:
```yaml
route:
  receiver: A
  routes:
    - receiver: B
      match_re:
        app: .+
    - receiver: C
      match:
        system: foo
```

A label set of `{system="foo"}` is shown as routing to receiver B when it'll actually route to receiver C.

Root of the problem is that in `matcher.value.test(v)`, undefined values of `v` are cast to a string `'undefined'`. This resolves the issue by explicitly checking for undefined label values.

@brian-brazil

### DIFF
--- a/static/routing-tree.js
+++ b/static/routing-tree.js
@@ -118,6 +118,9 @@ function matchLabel(matcher, labelSet) {
   var v = labelSet[matcher.name];
 
   if (matcher.isRegex) {
+    // Prevent regex matches on undefined labels.
+    if (undefined === v) return false;
+    
     return matcher.value.test(v)
   }
 
--- a/static/routing-tree.js
+++ b/static/routing-tree.js
@@ -119,8 +119,8 @@ function matchLabel(matcher, labelSet) {
 
   if (matcher.isRegex) {
     // Prevent regex matches on undefined labels.
-    if (undefined === v) return false;
-    
+    if (undefined === v) v = '';
+
     return matcher.value.test(v)
   }
 


Conorbro:
Looks good to me.

static/routing-tree.js
```diff
@@ -118,6 +118,9 @@ function matchLabel(matcher, labelSet) {
   var v = labelSet[matcher.name];
 
   if (matcher.isRegex) {
+    // Prevent regex matches on undefined labels.
+    if (undefined === v) return false;
```
Conorbro: Rather than return false, I'd change "undefined" to an empty string.

For example if 'app: .+' was changed to 'app: .*', matching an empty label value would result in no route to B being drawn.

garretruh:
Interesting.

I had assumed the regex matching operated the same way as PromQL (where a query of `up{app=~".*"}` doesn't match a metric without the `app` label like `up{system="foo"}`), but it doesn't look like that's the case.

static/routing-tree.js
garretruh: `v` is now rewritten to an empty string

garretruh:
> I had assumed the Alertmanager regex matching operated the same way as PromQL (where a query of up{app=~".*"} doesn't match a metric without the app label like up{system="foo"}), but it doesn't look like that's the case.

Apparently prometheus/tsdb#228 resolves that aberrant PromQL behavior

PR #921, opened 2017-11-20
Update rules.md

kuba--:
update link to alertmanager (/docs/alerting/alertmanager)


brian-brazil:
This page should be deleted, and references to it updated as it's in the prometheus docs now.

kuba--:
ok, but the link to alertmanager on the web page: https://prometheus.io/docs/alerting/rules/
is still broken.

PR #928, opened 2017-11-24
Add SQL Exporter to list of database exporters

free:


brian-brazil:
We try to avoid duplicates in this list, so this would have to replace the existing listed exporter and there's also another contender.

In term of best practices an exporter should either talk to one local instance, or work like snmp/blackbox do and take the target from a URL parameter. This is as service discovery is the responsibility of Prometheus, and the features here with SD and labels make this a mini-monitoring system rather than an exporter. If you wish to replace the existing exporter, you'll need to switch to one of the standard models.

In relation to timeouts, you can simplify that a good bit by using the `X-Prometheus-Scrape-Timeout-Seconds` header. See the blackbox exporter for how to use it.

You appear to be creating metrics called `up` and `scrape_duration_seconds`. These are produced internally by Prometheus, please prefix them. Given the nature of this exporter, failing rather than using an `up` metric is probably best.

free:
I've switched the exporter to the single instance model. I kept around the option of using the jobs and targets configuration for development purposes (I'm using it for that myself, because of its faster turnaround time). If you feel it's not clear enough, I could label them "production deployment" and "testing deployment" in the documentation, or something along those lines.

Regarding timeouts, I didn't know about `X-Prometheus-Scrape-Timeout-Seconds`, thanks for pointing it out. I'll definitely make use of it.

The `up` and `scrape_duration_seconds` metrics are only created in "dev mode", there are no such metrics when run against a single target. The exporter will actually return HTTP 500 if it cannot ping its single target, as there are no useful metrics to return in that case.

brian-brazil:
The mentioning of the "dev mode" at all is an issue, as users frequently get misconceptions about the Prometheus deployment model. Preferably that code wouldn't exist, as it's not really needed for development or testing.

free:
OK, went back and dropped all references to "dev" or "hub" mode from the documentation and all example configuration files. I also implemented support for retrieving Prometheus' scrape timeout from `X-Prometheus-Scrape-Timeout-Seconds`.

Regarding this PR specifically, I've replaced the existing SQL exporter, instead of adding to the list. It's obviously up to you if you think it makes sense: this exporter adheres more closely to Prometheus best practices (deployment model, on-demand collection) and it's more flexible, but it is as yet less battle tested than the alternatives.

brian-brazil:
Thanks!

One small thing, I'd suggest defaulting the type to Gauge as virtually all metrics produced will be gauges.

PR #931, opened 2017-12-02
Added unofficial Clojure client library

jamtur01:


brian-brazil:
We try to avoid duplicates in this list, and we already have one JVM client library.

jamtur01:
I find most Clojure folks will use a wrapper over a native client library because it's more idiomatic. I don't see that as duplication.

brian-brazil:
It's considered a duplication, and quickly looking at that library I'm not sure it's following the client library guidelines. 

Something that was actually a light wrapper or provided utilities for http would be fine (we list those elsewhere), this is some mix of wrapper and client library.

jamtur01:
Okay.

PR #932, opened 2017-12-02
Fix label matching in routing tree editor

paystee:
Hi @brian-brazil,

Label matching in routing tree editor is not consistent with Alertmanager:
- Unset labels must match empty string. [Source][1].
- Regexps must be anchored with start / end and wrapped in a non-capturing group. [Source][2].

Fixes #919.

[1]: https://github.com/prometheus/alertmanager/blob/v0.11.0/types/match.go#L69-L81
[2]: https://github.com/prometheus/alertmanager/blob/v0.11.0/config/config.go#L486-L498

### DIFF
--- a/static/routing-tree.js
+++ b/static/routing-tree.js
@@ -159,7 +159,7 @@ function massage(root) {
     for (var key in root.match_re) {
       var o = {};
       o.isRegex = true;
-      o.value = new RegExp(root.match_re[key]);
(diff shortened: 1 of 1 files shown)


brian-brazil:
@Conorbro Can you look at this?

Conorbro:
Looks good to me. From what I tested, the routing tree is routing the same way as the Alertmanager would.

PR #934, opened 2017-12-03
Added some file-based discovery integrations

jamtur01:
Also tightened language in preceding paragraph a little.

Unsure of the criteria for inclusion so let me know if these are okay.

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -19,10 +19,14 @@ these categories.
 
 ## File Service Discovery
 
-For service discovery mechanisms not natively supported by Prometheus,
-[file-based service discovery](/docs/operating/configuration/#<file_sd_config>) provides an interface for integrating.
+For service discovery mechanisms that are not natively supported by Prometheus,
+[file-based service discovery](/docs/operating/configuration/#<file_sd_config>) provides an interface for integration.
 
+ * [Amazon ECS](https://github.com/teralytics/prometheus-ecs-discovery)
+ * [Apache CloudStack](https://github.com/tsuru/prometheus-cloudstack-discovery)
+ * [Docker Cloud](https://github.com/rayhou/prometheus-file-sd)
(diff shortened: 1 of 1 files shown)


brian-brazil:
I'd generally prefer that the authors of these ask to add them, so that we know that they want to be listed and feedback (if any) can be provided. That way any obstacles to being a generally usable integration can get resolved, and our users are pointed at quality integrations 

For example from a quick look 3 of these do not atomically write the file, 2 of them expose secrets in argv and there appears to be some deployment-specific hardcoding going on too.

jamtur01:
Where are the guidelines published for integrations?

brian-brazil:
https://prometheus.io/docs/instrumenting/exporters/ and https://prometheus.io/docs/instrumenting/writing_clientlibs/ are the ones written

jamtur01:
@brian-brazil If I was to jot down some notes on writing file discovery scripts where would you like that. I had in mind something like:

```markdown
* Make your file discovery configurable, don't hardcode options. But preferably also ensure your file discovery will work automatically with its default configuration, for example, the default configuration options assume the default installation state of the source.
* Don't expose secrets like API keys or passwords in configuration. Instead, rely on secret stores or [the environment](https://12factor.net/config).
* Operations on the files to which you output your targets should be atomic.
```

brian-brazil:
I'm not sure we have a place for that yet. It hasn't really come up much. Most SD integrations would be ad-hoc and organisation specific, so there's no real rules beyond the atomic write.

jamtur01:
Okay. I'll just park these here. My primary concern was principle of least surprise: "I wrote a thing ... sorry, the thing doesn't meet the project's criteria ... there are criteria?" :)

brian-brazil:
We have docs for SD generally (https://github.com/prometheus/prometheus/tree/master/discovery), but those requirements are likely quite a bit harder than what we'll end up using here.

PR #936, opened 2017-12-04
List concourse integration

jmcarp:
As of the latest release, concourse ci can export metrics to prometheus: https://github.com/concourse/atc/blob/master/metric/emitter/prometheus.go.


brian-brazil:
This doesn't classifies as direct as you are not using instrumentation inline with your code. It can be listed as software exposing Prometheus metrics though.

PR #937, opened 2017-12-06
Adding AppOptics storage adapter

leecalcote:


brian-brazil:
Thanks!

I'd suggest grabbing a port number off the exporter list.

PR #939, opened 2017-12-10
Add conntrack_exporter to exporters

dbendelman:
Uses libnetfilter_conntrack for direct access to the Linux kernel's connection table and exports network connections like this:

```
# HELP conntrack_open_connections How many open connections are there to the remote host?
# TYPE conntrack_open_connections gauge
conntrack_open_connections{host="10.0.1.5:3306"} 49
```

This exporter was created because we had trouble pinning down why sometimes connections from an app to redis were timing out, and when we used this exporter to graph the network connections on a redis instance we saw that instead of 3-4 stable connections our apps were creating far more and they were intermittent. Saved the day for us, and it wasn't possible to do this with node_exporter's `tcpstat` because the latter sums up all connections from everywhere and doesn't break them down by remote host.

This exporter also has almost zero overhead. We ran it in production and there was no noticeable increase in CPU or RAM usage, so we have plans to roll it out on all internal boxes.

Small writeup why this is complementary to node_exporter's `tcpstat` is [here](https://github.com/hiveco/conntrack_exporter#doesnt-prometheus-already-monitor-the-systems-connections-natively).

This doesn't go against Prometheus's guidelines against label explosion [here](https://prometheus.io/docs/practices/naming/#labels) because as mentioned in [this FAQ](https://github.com/hiveco/conntrack_exporter#should-i-run-this-on-a-server-that-listens-for-external-internet-connections), it's supposed to be run on an internal, non-public facing server. Therefore the number of metric:label pairs created should be very reasonable.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -54,6 +54,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [IPMI exporter](https://github.com/lovoo/ipmi_exporter)
    * [knxd exporter](https://github.com/RichiH/knxd_exporter)
    * [Node/system metrics exporter](https://github.com/prometheus/node_exporter) (**official**)
+   * [conntrack network connection exporter](https://github.com/hiveco/conntrack_exporter)
    * [Ubiquiti UniFi exporter](https://github.com/mdlayher/unifi_exporter)
 
 ### Messaging systems


brian-brazil:
Thank you for sharing your exporter. This exporter is useful only in highly limited circumstances. On a machine running a web browser or talking out to SaaS services this is likely to have cardinality issues, with high levels of churn. For the use cases listed event logging such as via an iptables LOG target or profiling with tcpdump are likely a better fit.
I'm afraid we're not willing to list this on this page as the risk of unintentionally taking out Prometheus is greater than the potential benefits.

dbendelman:
While I understand your decision @brian-brazil, I'd like to respectfully offer counter-points to your stated rationale. Having actually used this in production, and wishing I had something like it in previous projects, I think this fills a pretty important niche among existing exporters.

> This exporter is useful only in highly limited circumstances.

Quoting from the exporter's [README](https://github.com/hiveco/conntrack_exporter/#uses):

```
Uses:
* Operations: monitor when a critical link between your microservices is broken.
* Security: monitor which remote hosts your server is talking to.
* Debugging: correlate intermittently misbehaving code with strange connection patterns.
```

This is just meant to give a taste and obviously not exhaustive, and I think all of these are far from exotic. Apologies that my PR description was confusingingly only highlighting one rare use case, I was just sharing the reason why this exporter was created. Aside from the list above, this exporter can also do other cool things like allow Prometheus to alert when some specific external host attempts to connect, break down node egress by LAN vs WAN in a graph, etc. It's all up to the creativity of the user.

> On a machine running a web browser or talking out to SaaS services this is likely to have cardinality issues, with high levels of churn.

Running a web server (exposed to the public Internet) is exactly the kind of usage we advise against in the readme, and we clearly explain why. Not sure what other safety net we can offer the end user without throwing out the baby with the bathwater :)

Talking out to SaaS services is actually a perfect use case for this exporter: usually the number of endpoints connected to is small and constant (one per each SaaS service the app uses) so we're okay regarding cardinality, and it's extremely valuable to have visibility into connection activity for those outside services.

> For the use cases listed event logging such as via an iptables LOG target or profiling with tcpdump are likely a better fit.

I agree those would be alternative ways to debug intermittent connection issues, but it's hard to argue with the power of a graph even in this use case:

conntrack_open_connections

And of course, as I write above there's many use cases for this, including some that tcpdump and iptables won't help with.

I think the Prometheus team was correct in your decision to make the `tcpstat` collector disabled by default, and wrote next to it `Warning: the current version has potential performance issues in high load situations`. There shouldn't be anything in the default installation that allows a typical user to shoot themselves in the foot. However even though it only exports basic TCP state totals, `tcpstat` was still considered valuable enough to warrant it being included inside the official system exporter. I believe the same logic applies to this exporter, which shows even deeper visibility and is complementary to `tcpstat`: it's extremely useful in a large variety of use cases, but as a third party exporter it isn't part of a standard Prometheus installation (so it's "disabled by default") and can have a similar ominous warning next to it in the docs:

> Warning: has potential performance issues in high load situations that can affect the stability of your Prometheus server; see [README](https://github.com/hiveco/conntrack_exporter/#should-i-run-this-on-a-server-that-listens-for-external-internet-connections) for details on how to use this exporter safely.

I can update my PR if you'll consider merging with such a change. If you'd also like me to add more info to the warning in the README as well, I don't mind doing that too, just let me know.

P.S. Thanks for an excellent monitoring platform! :)

PR #940, opened 2017-12-11
Add new phpfpm exporter to list

Lusitaniae:
Reasons to include an additional exporter for phpfpm:
- It scrapes the status page via unix sockets. (The alternative scrapes via http)
- Automatic builds with release to GH, Quay and Docker.
- Ability to monitor multiple phpfpm pools.
- Ability to include script metrics (such as opcache metrics)

@brian-brazil Let me know what you think


brian-brazil:
We try to avoid duplicates in this list, so to replace the existing exporter you'd need to follow the existing exporter guidelines at least as well, have a superset of feature and have a community community.

It is expected that it's one exporter per thing being monitored, so monitoring multiple pools isn't in line with our guidelines.

brian-brazil:
It's been a while, but from a quick peek things don't seem to have changed much, so I don't think there's a need to change the one we currently list.

PR #942, opened 2017-12-14
Added pgbouncer exporter to exporters list

pracucci:
I've added to the list of exporter another exporter for pgbouncer (PostgresSQL connections pooler) whose main difference for the already-existing one is supporting multiple pgbouncer backends.

---

#### Why supporting multiple pgbouncer instances?

PgBouncer is a single thread application and so can only saturate a single CPU core on high load. In high load environments, it's common practice to run multiple pgbouncer processes per machine (one for each CPU core) and load balance traffic among them. Ideally we want to run a single exporter instance per machine, capable to monitor 1+ pgbouncer(s) running on the machine itself.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -41,6 +41,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [MySQL server exporter](https://github.com/prometheus/mysqld_exporter) (**official**)
    * [OpenTSDB Exporter](https://github.com/cloudflare/opentsdb_exporter)
    * [PgBouncer exporter](http://git.cbaines.net/prometheus-pgbouncer-exporter/about)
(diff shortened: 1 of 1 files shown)


brian-brazil:
We try to avoid duplicates in this list. If you want to replace the existing one you'd have to at the least follow the exporter guidelines at least as well. Doing service discovery in the exporter is not in line with the client library guidelines, so this can't replace the one we already list.

pracucci:
@brian-brazil Thanks for your quick feedback. The use case is to actually get the exporter scrape the metrics from multiple pgbouncers running on the same host, so there's no service discovery done by the exporter at all. What's your take on this?

brian-brazil:
That you have a config file listing hosts is service discovery.

pracucci:
It's a config file where you specify the DSN of each pgbouncer (because it's the way the postgresql driver works), and each DSN is localhost with a different port. If having a config file with two local ports is service discovery, then I agree with you.

Anyway, it's not a big deal not having the exporter listed there.

PR #945, opened 2017-12-22
add osquery_exporter to the list of exporters

zwopir:
add osquery_exporter to the list of exporters. The exporter exports metrics retreived via osquery (https://osquery.io) as prometheus metrics.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -120,6 +120,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Nagios / Naemon exporter](https://github.com/Griesbacher/Iapetos)
    * [New Relic exporter](https://github.com/jfindley/newrelic_exporter)
    * [NRPE exporter](https://github.com/robustperception/nrpe_exporter)
+   * [Osquery exporter](https://github.com/zwopir/osquery_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks! A few comments to improve things a little:

https://github.com/zwopir/osquery_exporter/blob/master/collector/osquery_collector.go#L90 - these metrics would be better exposed as Gauges just about the particular scrape using ConstMetrics, like the node exporter does it. `X_query_duration_seconds` would be the more usual name for that metric, as it includes the unit.

https://github.com/zwopir/osquery_exporter/blob/master/collector/osquery_collector.go#L124 - there's race conditions here. Concurrent scrapes can interfere and you'll usually return the metrics about the previous scrape as the collects are before the `wg.Wait`. Switching to ConstMetric gauges per above will fix this.

In the example on https://github.com/zwopir/osquery_exporter#counter-and-gauges the metric name has a `_count` suffix, this suffix should only be used with Summary/Histogram metrics. Generally, virtually all metrics something like this produces will be Gauges and I'd suggest defaulting accordingly.

Usually you wouldn't put `_exporter` in the metric names.

In the example on https://github.com/zwopir/osquery_exporter#countervecs-and-gaugevecs you generally shouldn't put the name of the label in the metric name, as it won't make sense if the label is aggregated away.

zwopir:
thanks for merging and the comments to improve the exporter! I'll rework the exporter accordingly

PR #949, opened 2018-01-08
Add Docker Daemon to software exposing Prometheus metrics

fhemberger:


brian-brazil:
Thanks!

fhemberger:
Wow, that was quick: :grinning: Thanks!

PR #953, opened 2018-01-23
Add softether_exporter

dalance:
Adding SoftEther VPN server (http://www.softether.org) exporter to list


brian-brazil:
Thanks!

I think you have your counters and gauges a bit mixed up. Counters should end in _total, and Gauges should not.

In addition you are using direct instrumentation for an exporter, which is not the best way to go about it. Instead you should use a custom collector and only create the metric families on each scrape. See https://www.robustperception.io/writing-json-exporters-in-python/ for example.

dalance:
Thank you for your advise.
I'll change metrics name, and learn about custom collector.

PR #954, opened 2018-01-25
docs/alerting: updating pagerduty_url default

noqcks:
It has been changed to use the Pagerduty v2 API by default in Alertmanager.

https://github.com/prometheus/alertmanager/blob/master/config/config.go#L324


brian-brazil:
Thanks!

PR #957, opened 2018-01-27
provide link to our dashboards to make initial prometheus use easier.

burton-scalefastr:
Adding a reference to our dashboard changes so we can share with the community.

### DIFF
--- a/content/docs/visualization/grafana.md
+++ b/content/docs/visualization/grafana.md
@@ -79,3 +79,21 @@ You must currently manually edit the downloaded JSON files and correct the
 `datasource:` entries to reflect the Grafana data source name which you
 chose for your Prometheus server.  Use the "Dashboards" → "Home" → "Import"
 option to import the edited dashboard file into your Grafana install.
+
+### Functioning Dashboards for Prometheus 2.0
+
+While many of the dashboards from the Grafana shared dashboards do work out of
+the box, many don't have unified labels and many were designed for Prometheus
+1.x and don't yet support Prometheus 2.x.
+
+Scalefastr (which uses Prometheus as part of their infrastructure) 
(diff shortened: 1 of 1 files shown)


brian-brazil:
We don't currently curate a list dashboards, I'd suggest uploading them to grafana.net.

burton-scalefastr:
Right.  This is why we wanted to provide a set of curated dashboards - so that new Prometheus users could get up and running.

We plan on maintaining these moving forward so new people can just start using them.

brian-brazil:
You're free to provide them, but I'm not sure it's appropriate to link from Prometheus.io to dashboards that presume one particular non-standard label hierarchy. PMMM isn't something we link either for example.

burton-scalefastr:
What's PMMM?

I agree about the labels. It's one of the issues we wanted to solve.  There is some community effort around container labeling which I think could help on standardization.

I'd like to get the link from prometheus.io so that it's easier for new members in the community to discovery this so that they don't have to repeat the work we did.

If you think it's more appropriate maybe it might be better to have a 'community contribution' section. 

I'm open to both scenarios. I just don't want the work we're doing to languish as we want to establish a community around it.

One of the problems with Grafana is that the dashboards are orphaned... We want to help solve that.

brian-brazil:
If you want to curate dashboards that work for you you're free to do so, however it's way too early to link anything like this from prometheus.io as recommended dashboards and certainly not on this page.

What you could do is one of our user interviews on the blog, which could include a link to these.

burton-scalefastr:
Thanks. Who should I talk to about a user interview.

I guess we're ok with either way - I just don't really want these to languish.

brian-brazil:
I handle those as it happens, I've just sent you an email to start that process.

PR #960, opened 2018-02-02
correct number of options in global config block

tkellen:
I'm digging into prometheus for the first time today by reading all of the official docs. I noted that the number of options mentioned as being present in the global config block seems to be incorrect. Have I missed something or is this actually wrong?

Thanks so much for making this project available to the world, I'm really excited to start using it 💃 !


brian-brazil:
Thanks!

There was previously `external_labels` in this example too, but it was removed for simplicity.

PR #967, opened 2018-02-08
Add apache yarn and apache airflow exporters

DemonTPx:
I could not find exporters for [Apache Airflow](https://airflow.apache.org/) and [Apache YARN](https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/YARN.html), so we decided to create it ourselves and publish the exporters on our github and the docker hub.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -129,7 +129,6 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
 
 ### Miscellaneous
    * [Apache Airflow exporter](https://github.com/PBWebMedia/airflow-prometheus-exporter)
-   * [Apache Hadoop YARN exporter](https://github.com/PBWebMedia/yarn-prometheus-exporter)
    * [BIG-IP exporter](https://github.com/ExpressenAB/bigip_exporter)
    * [BIND exporter](https://github.com/digitalocean/bind_exporter)
    * [Bitbucket exporter](https://github.com/AndreyVMarkelov/prom-bitbucket-exporter)
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -128,6 +128,8 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [StatsD exporter](https://github.com/prometheus/statsd_exporter) (**official**)
 
 ### Miscellaneous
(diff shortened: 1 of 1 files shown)


brian-brazil:
Does the YARN exporter do anything more than JMX would provide?

The Airflow one looks like a subset of the SQL exporter.

DemonTPx:
About the YARN exporter, it doesn't do more than the JMX exporter. However, we chose not to use the JMX exporter because it seems quite a bit more difficult to setup on an Amazon EMR cluster. This exporter can just be added on independently.

About the Airflow one, it does look similar. The main difference probably is that the Airflow one caches log event counts in memory and only queries new log events and increases the cached counts. Querying this table would be quite slow, since it typically contains a lot of records.

brian-brazil:
We try to avoid duplicates in this list, and the JMX already covers YARN and far more.

In relation to the Airflow exporter there's duplication with the SQL exporter, and I'm also a bit concerned at its scalability given it is doing a full table scan at startup.

DemonTPx:
I didn't know about the json exporter. That would have saved me some time :/

Airflow doesn't keep its metrics anywhere else (eg: in memory or in another table). The race condition I didn't consider and you are right. That might be fixable by running the query and updating the counts in a separate thread once in a while or maybe by using a mutex... I scrape it once every 10 seconds, so I don't worry about that in my setup, but I understand that isn't really a good reason to leave it like this.

SuperQ:
Rather than a background thread, I would probably just wrap it in a mutex.  This would make initial startup with a very large `log` a little easier to deal with.

But, if Airflow has any actual log files of this data, it may be easier to use something like [mtail](https://github.com/google/mtail) to generate metrics.

DemonTPx:
I've used a mutex to fix the race condition problem: https://github.com/PBWebMedia/airflow-prometheus-exporter/blob/master/collector.go#L155

I'll remove the YARN exporter.

SuperQ:
I still think the airflow exporter doesn't match up with exporter best practices.  While it's still OK to use, I don't think we want to list it officially.

DemonTPx:
Which specific best practice(s) doesn't it match up with?

gouthamve:
Given Airflow itself exports metrics and there are other exporters that export these metrics via Airflow w/o touching the db, is it okay to not list this @DemonTPx?

gouthamve:
I'm closing this. Please feel free to reopen if you think it is in error.

DemonTPx:
That's ok. :)

PR #973, opened 2018-02-13
Add instructions for adding a company logo

juliusv:


juliusv:
@caniszczyk :)

beorn7:
@juliusv do we still want this?

Now that "Prometheus is mainstream", perhaps we don't even need a logo page? We could do more something like the case study page of K8s or like the one on go.dev .  I mean, we don't need to do it right now, but perhaps we shouldn't solicit adding more logos.

brian-brazil:
I agree, the reasons why we added this originally no longer apply (when did you last hear "SoundCloud Prometheus") so we don't need additions and can probably remove it.

juliusv:
Yup, doing the k8s-like case studies list would be cool for me too. I'll close this for now.

PR #835, opened 2017-09-19
Oracle DB Exporter

sathieu:

PR #975, opened 2018-02-17
adding patterdale as an oracle db exporter

tjheslin1:
I created this project to expose metrics for a set of Oracle databases.

The oracle jdbc driver must be provided at runtime due to Oracle's policies.

Furthermore we wanted to run oracle specific queries.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -40,6 +40,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [MSSQL server exporter](https://github.com/awaragi/prometheus-mssql-exporter)
    * [MySQL server exporter](https://github.com/prometheus/mysqld_exporter) (**official**)
    * [OpenTSDB Exporter](https://github.com/cloudflare/opentsdb_exporter)
+   * [Oracle DB exporter](https://github.com/tjheslin1/Patterdale)
    * [PgBouncer exporter](http://git.cbaines.net/prometheus-pgbouncer-exporter/about)
    * [PostgreSQL exporter](https://github.com/wrouesnel/postgres_exporter)
    * [ProxySQL exporter](https://github.com/percona/proxysql_exporter)


brian-brazil:
Hi,

We try to avoid duplication in this list, and this appears to be duplicating the existing SQL exporter. In addition it appears to be performing service discovery and scheduling which are not the responsibility of an exporter. I'd also have cardinality concerns with your example queries, as it appears to be performing event logging. I would recommend reading https://prometheus.io/docs/instrumenting/writing_exporters/

At present I am planning on accepting #835 once @sathieu indicates it has actually been tested, as it is more in line with our guidelines.

sathieu:
@brian-brazil I won't test oracle_exporter any time soon, sorry. Anyone else?

tjheslin1:
Hi. Thanks @brian-brazil for having a look. I'll have another look over the docs.

I'm not sure that I see exporter doing any service discovery, could you explain on that?
It expects a jdbcUrl for the database it will be querying. As per the prometheus docs it would be advisable to run an instance per Oracle database.

As for scheduling, there is the `cacheDuration` property which will hold the result of database scrapes for that amount of time. This is to prevent any databases being denial of serviced due to a something scraping the metrics page being too eager.

The example query is from a user. I could swap it out for a better recommend query.

I wasn't aware of @sathieu implementation! It looks good, I wasn't aware of the [Oracle Instant Client](http://www.oracle.com/technetwork/database/database-technologies/instant-client/overview/index.html), I'll defintely have a look at that.

I'm happy for @sathieu 's implementation to be accepted. The only feature I see Patterdale having is that you can define your own queries and metrics.

RichiH:
Caching in an exporter is something of an antipattern. As there are still
valid use cases for it, I started, but never finished,
https://github.com/RichiH/prometheus_scrape_cache . Especially the issue of
stale data vs timestamps vs storage v3 is not trivial so it's best to do
this in one canonical place.

On Feb 17, 2018 21:43, "Thomas Heslin" <notifications@github.com> wrote:

Hi. Thanks @brian-brazil <https://github.com/brian-brazil> for having a
look. I'll have another look over the docs.

I'm not sure that I see exporter doing any service discovery, could you
explain on that?
It expects a jdbcUrl for the database it will be querying. As per the
prometheus docs it would be advisable to run an instance per Oracle
database.

As for scheduling, there is the cacheDuration property which will hold the
result of database scrapes for that amount of time. This is to prevent any
databases being denial of serviced due to a something scraping the metrics
page being too eager.

The example query is from a user. I could swap it out for a better
recommend query.

I wasn't aware of @sathieu <https://github.com/sathieu> implementation! It
looks good, I wasn't aware of the Oracle Instant Client
<http://www.oracle.com/technetwork/database/database-technologies/instant-client/overview/index.html>,
I'll defintely have a look at that.

I'm happy for @sathieu <https://github.com/sathieu> 's implementation to be
accepted. The only feature I see Patterdale having is that you can define
your own queries and metrics.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/prometheus/docs/pull/975#issuecomment-366469998>, or mute
the thread

brian-brazil:
> As per the prometheus docs it would be advisable to run an instance per Oracle database.

The example config file indicates otherwise, it lists more than one database.

> As for scheduling, there is the cacheDuration property which will hold the
result of database scrapes for that amount of time.

There are cases for caching, but doing it in a blanket fashion for this sort of exporter would be a concern. If you think you need it that indicates you need to reconsider the expense of the data you're fetching.

> The only feature I see Patterdale having is that you can define your own queries and metrics.

That duplicates the SQL exporter. Whether it is appropriate for a RDMS exporter to offer this feature is unresolved, however I would expect an exporter for a specific database to provide primarily metrics about the database's performance.

tjheslin1:
Thanks @RichiH. Your scrape cache is an interesting read and a great idea, I'll pass this on to those who asked for the cache in the first place.

Thanks @brian-brazil. I appreciate that the example config in its current state goes against the best practices. For caching the `cacheDuration` can be set to 0, which is the default if a value is not provided.

I'll definitely take on board your advice should I write another exporter. For now, if my implementation was still to be considered I can update the configuration to disable caching, show only one database connection provided more appropriate examples of queries.

Otherwise feel free to close the issue.

sathieu:
Just FYI, it's not my implementation, I just submitted a PR to add it to the list.

RichiH:
@tjheslin1 No worries. Feel free to poke me on irc.freenode.net if you or the others have any questions.

PR #978, opened 2018-02-21
Add alternative PHP-FPM exporter

estahn:


brian-brazil:
We try to avoid duplicates in this list in order to make this easier for users, and to try to rally the community around one solution. Does this exporter have major advantages that'd justify replacing the existing one?

estahn:
@brian-brazil Makes sense and no major difference. Will close this pull request.

PR #980, opened 2018-02-21
Add Gearman exporter

njam:
@brian-brazil ping

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -59,6 +59,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
 
 ### Messaging systems
    * [Beanstalkd exporter](https://github.com/messagebird/beanstalkd_exporter)
+   * [Gearman exporter](https://github.com/bakins/gearman-exporter)
    * [Kafka exporter](https://github.com/danielqsj/kafka_exporter)
    * [NATS exporter](https://github.com/nats-io/prometheus-nats-exporter)
    * [NSQ exporter](https://github.com/lovoo/nsq_exporter)


content/docs/instrumenting/exporters.md
```diff
@@ -66,6 +66,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [MQTT blackbox exporter](https://github.com/inovex/mqtt_blackbox_exporter)
    * [RabbitMQ exporter](https://github.com/kbudde/rabbitmq_exporter)
    * [RabbitMQ Management Plugin exporter](https://github.com/deadtrickster/prometheus_rabbitmq_exporter)
+   * [Gearman exporter](https://github.com/bakins/gearman-exporter)
```
brian-brazil: Keep it sorted.

njam:
Thanks for the feedback! I will fix it and then get back to you.

njam:
`gearman_jobs_queued` could be a bit misleading, as the metric actually represents all non-finished jobs, the ones queued for execution as well as the ones currently being executed.
I would suggest the following names:
- gearman_running_jobs
- gearman_total_jobs (while I don't like "total" very much, it *is* well established [from gearman](https://github.com/gearman/gearmand/blob/e2d76cf631414964decc2f22006a666e1c0e2424/libgearman-server/text.cc#L167))
- gearman_workers

@brian-brazil wdyt?

brian-brazil:
> gearman_total_jobs (while I don't like "total" very much, it is well established from gearman)

Everything in metrics is some form of total, it rarely adds anything to a metric name.

_jobs and _jobs_running would be the usual way to organise that.

njam:
@brian-brazil I went with your suggestion. Could you take another look?

brian-brazil:
You still have the version label on `up`, that should be moved to a different metric.

PR #981, opened 2018-02-21
add skipper

szuecs:
add skipper as software exposing prometheus metrics


content/docs/instrumenting/exporters.md
```diff
@@ -185,6 +185,7 @@ separate exporters are needed:
    * [Pretix](https://pretix.eu/)
    * [Quobyte](https://www.quobyte.com/) (**direct**)
    * [RobustIRC](http://robustirc.net/)
+   * [Skipper](https://github.com/zalando/skipper) (**direct**)
```
brian-brazil: This code is not directly instrumented with client_golang, there's a centralized metrics class and a facade in the way

szuecs: Ok, then I did not understood the description.

brian-brazil:
Thanks!

PR #982, opened 2018-02-22
Add gitlab runner to Software exposing Prometheus metrics

oussemos:
Add gitlab runner in Software exposing Prometheus metrics

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -177,6 +177,7 @@ separate exporters are needed:
    * [Doorman](https://github.com/youtube/doorman) (**direct**)
    * [Etcd](https://github.com/coreos/etcd) (**direct**)
    * [FreeBSD Kernel](https://www.freebsd.org/cgi/man.cgi?query=prometheus_sysctl_exporter&apropos=0&sektion=8&manpath=FreeBSD+12-current&arch=default&format=html)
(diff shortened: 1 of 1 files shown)


brian-brazil:
https://gitlab.com/gitlab-org/gitlab-runner/blob/master/executors/docker/machine/provider.go#L431

Most of this seems to be custom collectors, this is the only direct instrumentation I can find and it is procedurally generating metric names which is a no-no. That should be a label.

My idea is that directly instrumented code listed here provides prime examples of what Prometheus instrumentation is meant to look like.

content/docs/instrumenting/exporters.md
```diff
@@ -189,6 +189,7 @@ separate exporters are needed:
    * [SkyDNS](https://github.com/skynetservices/skydns) (**direct**)
    * [Telegraf](https://github.com/influxdata/telegraf/tree/master/plugins/outputs/prometheus_client)
    * [Weave Flux](https://github.com/weaveworks/flux)
+   * [Gitlab runner](https://docs.gitlab.com/runner/monitoring/README.html) (**direct**)
```
brian-brazil: Keep it sorted.

oussemos: Done

tmaczukin:
@brian-brazil Thanks for noticing this. We're currently in the process of cleaning-up the metrics naming in GitLab Runner, see https://gitlab.com/gitlab-org/gitlab-runner/issues/3114.

PR #984, opened 2018-03-06
Add iis exporter to the list

mateusz-jablonski94:
Hi,

We created this exporter to collect metrics from websites hosted on IIS (amount of processed HTTP requests, response times, request sizes etc.).
It watches a given directory with IIS logs, always finds the newest file and exposes live metrics for Prometheus.


brian-brazil:
We try to avoid duplicates in this list, and this appear to duplicate the existing log tailers and also possibly the WMI exporter.

mateusz-jablonski94:
As far as I can see WMI exporter provides general IIS metrics (not for specific website, endpoint etc.)
Seems like grok_exporter is pretty similar, but it watches only a single file. There is no support for watching directories or using wildcards (like: "directory/*.log"). IIS usually creates 1 new log file per day
and I think it makes grok_exporter impossible to use in that case.

brian-brazil:
If you feel the standard exporter in this space is missing a feature, you should request it.

Issue #986, opened 2018-03-12
Add Atlassian Bamboo exporter to exporter section


AndreyVMarkelov:
Add https://github.com/AndreyVMarkelov/bamboo-prometheus-exporter. This one is free from marketplace

PR #987, opened 2018-03-12
#986: add bamboo exporter

AndreyVMarkelov:
@brian-brazil please review


brian-brazil:
Thanks!

plan_key would be the usual label name rather than planKey.

PR #988, opened 2018-03-13
Document smtp_hello parameter

simonpasquier:
Follow up of https://github.com/prometheus/alertmanager/pull/892

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -65,6 +65,8 @@ global:
   [ smtp_from: <tmpl_string> ]
   # The default SMTP smarthost used for sending emails.
   [ smtp_smarthost: <string> ]
+  # The default hostname to identify to the SMTP server.
+  [ smtp_hello: <string> | default = "localhost" ]
   # SMTP authentication information.
   [ smtp_auth_username: <string> ]
   [ smtp_auth_password: <secret> ]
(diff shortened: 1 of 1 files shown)


content/docs/alerting/configuration.md
```diff
@@ -65,6 +65,8 @@ global:
   [ smtp_from: <tmpl_string> ]
   # The default SMTP smarthost used for sending emails.
   [ smtp_smarthost: <string> ]
+  # The default hostname to identify to the SMTP server.
```
brian-brazil: I think we should mention the default behaviour of using localhost here somehow.

content/docs/alerting/configuration.md
```diff
@@ -65,6 +65,9 @@ global:
   [ smtp_from: <tmpl_string> ]
   # The default SMTP smarthost used for sending emails.
   [ smtp_smarthost: <string> ]
+  # The default hostname to identify to the SMTP server. If not specified,
```
brian-brazil: We usually do this as `| default = "localhost"`, and it might be worth making that explicit over in the am code.

simonpasquier: Makes sense

content/docs/alerting/configuration.md
```diff
@@ -301,6 +306,8 @@ to: <tmpl_string>
 
 # The HTML body of the email notification.
 [ html: <tmpl_string> | default = '{{ template "email.default.html" . }}' ]
+# The text body of the email notification.
+[ text: <tmpl_string> ]
```
simonpasquier: This parameter has been added in https://github.com/prometheus/alertmanager/pull/934.

PR #992, opened 2018-03-22
Update security.md

RichiH:
Some minor cleanups and clarifications

### DIFF
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -14,13 +14,15 @@ attack vectors that some configurations may enable.
 
 As with any complex systems it is not possible to guarantee that there are no
 bugs. If you find a security bug, please file it in the issue tracker of the
-relevant component.
+relevant component. If you prefer to report security issues in secret, you may
+email the non-public [Prometheus team mailing list](mailto:prometheus-team@googlegroups.com).
 
 ### Prometheus
 
 It is presumed that untrusted users have access to the Prometheus HTTP endpoint
 and logs. They have access to all time series information contained in the
-database, plus a variety of operational/debugging information.
+database, plus a variety of operational/debugging information, including all
+targets and alerts.
 
 It is also presumed that only trusted users have the ability to change the
 command line, configuration file, rule files and other aspects of the runtime
@@ -34,10 +36,10 @@ that service discovery system.
 
 Scraped targets may be run by untrusted users. It should not by default be
 possible for a target to expose data that impersonates a different target.  The
-`honor_labels` option removes this protection, as can certain relabelling
-setups.
+`honor_labels` option removes and certain relabelling setups remove this
+protection.
 
-As of Prometheus 2.0, the `--web.enable-admin-api` flag controls access to the
+As of Prometheus 2.x, the `--web.enable-admin-api` flag controls access to the
 administrative HTTP API which includes functionality such as deleting time
 series. This is disabled by default. If enabled, administrative and mutating
 functionality will be accessible under the `/api/*/admin/` paths. The
@@ -84,7 +86,8 @@ configuration file.
 Any user with access to the Pushgateway HTTP endpoint can create, modify and
 delete the metrics contained within. As the Pushgateway is usually scraped with
 `honor_labels` enabled, this means anyone with access to the Pushgateway can
-create any time series in Prometheus.
+create any time series in Prometheus, including impersonating other targets or
+generating alerts.
 
 ## Exporters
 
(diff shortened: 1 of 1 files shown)


content/docs/operating/security.md
```diff
@@ -14,13 +14,15 @@ attack vectors that some configurations may enable.
...
-relevant component.
+relevant component. If you prefer to report security issues in secret, you may
+email the non-public [Prometheus team mailing list](mailto:prometheus-team@googlegroups.com).
 
 ### Prometheus
 
 It is presumed that untrusted users have access to the Prometheus HTTP endpoint
 and logs. They have access to all time series information contained in the
-database, plus a variety of operational/debugging information.
+database, plus a variety of operational/debugging information, including all
```
brian-brazil: There's far more than that exposed - I don't think we should try to list it all.

content/docs/operating/security.md
```diff
@@ -84,7 +86,8 @@ configuration file.
 Any user with access to the Pushgateway HTTP endpoint can create, modify and
 delete the metrics contained within. As the Pushgateway is usually scraped with
 `honor_labels` enabled, this means anyone with access to the Pushgateway can
-create any time series in Prometheus.
+create any time series in Prometheus, including impersonating other targets or
```
brian-brazil: This isn't possible.

content/docs/operating/security.md
```diff
@@ -14,13 +14,15 @@ attack vectors that some configurations may enable.
 
 As with any complex systems it is not possible to guarantee that there are no
 bugs. If you find a security bug, please file it in the issue tracker of the
-relevant component.
+relevant component. If you prefer to report security issues in secret, you may
```
brian-brazil: I'd prefer not spam the team list with misunderstandings of our security model.

content/docs/operating/security.md
```diff
@@ -34,10 +36,10 @@ that service discovery system.
 
 Scraped targets may be run by untrusted users. It should not by default be
 possible for a target to expose data that impersonates a different target.  The
-`honor_labels` option removes this protection, as can certain relabelling
-setups.
+`honor_labels` option removes and certain relabelling setups remove this
```
brian-brazil: This change doesn't parse.

content/docs/operating/security.md
```diff
@@ -34,10 +36,10 @@ that service discovery system.
...
+protection.
 
-As of Prometheus 2.0, the `--web.enable-admin-api` flag controls access to the
+As of Prometheus 2.x, the `--web.enable-admin-api` flag controls access to the
```
brian-brazil: This is correct as written

content/docs/operating/security.md
```diff
@@ -136,7 +139,7 @@ wherever it is stored.
 ## Denial of Service
 
 There are some mitigations in place for excess load or expensive queries.
-However, if too many or too expensive queries/metrics are provided components
+However, if too many or too expensive queries/metrics are provided, components
```
brian-brazil: This is fine as-is.

content/docs/operating/security.md
RichiH: While I agree with that sentiment, we need to have a defined way for non-public disclosure.

content/docs/operating/security.md
RichiH: time series is already an incomplete list by itself. From experience, people care most about the three I now listed. I am 50/50 on this change, though.

content/docs/operating/security.md
RichiH: Your phrasing is overly convoluted, I tried to untangle it.

RichiH: I am fine with different wording, of course.

content/docs/operating/security.md
RichiH: I agree that it's correct, but "as of" is often misunderstood as "at this time" as opposed to "from this time on". The change would make it easier on non-native speakers. We could also phrase it as "Starting from 2.0 [...]".

content/docs/operating/security.md
RichiH: How so? This was written in this file before my PR; see above.

content/docs/operating/security.md
RichiH: I disagree.

content/docs/operating/security.md
brian-brazil: We don't need one, we merely need to document how we accept security issues.

content/docs/operating/security.md
brian-brazil: memory profiles, metadata, logs (at some point) - we'll keep on adding to this

content/docs/operating/security.md
brian-brazil: Your wording still doesn't parse.

content/docs/operating/security.md
brian-brazil: It's as of 2.0, not 2.x.

content/docs/operating/security.md
brian-brazil: You cannot generate alerts this way. The current wording is fine.

content/docs/operating/security.md
brian-brazil: "However components will fall over" does not make sense.

beorn7:
The file got reworked a bit independently. I close this as obsolete.

PR #1002, opened 2018-04-14
Add Slack field configuration documentation

taharah:
Adds documentation on how to configure Slack fields, which were added to the Alertmanager as part of PR [#1135](https://github.com/prometheus/alertmanager/pull/1135).

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -454,6 +454,23 @@ channel: <tmpl_string>
 [ text: <tmpl_string> | default = '{{ template "slack.default.text" . }}' ]
 [ fallback: <tmpl_string> | default = '{{ template "slack.default.fallback" . }}' ]
 
+# An optional flag indicating whether the value(s) of all fields defined
+# are short enough to be displayed side-by-side with any other value(s).
+[ short_fields: <boolean> ]
+
+# An optional list of key/value pairs to provide further details about the incident.
+# The key/value pairs contained within it will be displayed in a table inside the
+# message attachment.
+[ fields:
+  - # Title is shown as a bold heading above the value text.
+    # It cannot contain markup and will be escaped for you.
+    title: <tmpl_string>
+    # The text value of the field. It may contain standard message
+    # markup and must be escaped as normal. May be multi-line.
+    value: <tmpl_string>
+]
+
+
 # The HTTP client's configuration.
 [ http_config: <http_config> | default = global.http_config ]
 ```


content/docs/alerting/configuration.md
```diff
@@ -454,6 +454,23 @@ channel: <tmpl_string>
 [ text: <tmpl_string> | default = '{{ template "slack.default.text" . }}' ]
 [ fallback: <tmpl_string> | default = '{{ template "slack.default.fallback" . }}' ]
 
+# An optional flag indicating whether the value(s) of all fields defined
```
brian-brazil: Keep the style consistent with what's above.

content/docs/alerting/configuration.md
```diff
@@ -454,6 +454,23 @@ channel: <tmpl_string>
...
+# are short enough to be displayed side-by-side with any other value(s).
+[ short_fields: <boolean> ]
+
+# An optional list of key/value pairs to provide further details about the incident.
+# The key/value pairs contained within it will be displayed in a table inside the
+# message attachment.
+[ fields:
```
brian-brazil: This is not the structure that is defined in the code.

taharah: Not quite sure what you mean here. The way the code reads is a little misleading since it looks like it's looping over a map at first glance [[1](https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L679-L690)], but the structure is definitely a list of key/value pairs albeit very specific, i.e., include both both the title and value for the field. Thus, do you mean that the description in the preceding comment of the structure is inaccurate and needs to be more specific?

I've tested and used the following configuration for defining two different fields:

```yaml
fields:
- title: 'severity'
  value: '{{ .GroupLabels.severity }}'
- title: 'status'
  value: '{{ .Status }}'
```

brian-brazil: The code allows for any field to be supplied, not just title and value.

taharah: Any number of fields can be defined, but it needs both the `title` [[1](https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L684)] and `value` [[2](https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L685)] defined for each field and then inherits the `short` parameter as a global setting for all fields [[3](https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L686), [4](https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L247)]. This is the required structure for fields as per the Slack documentation [[5](https://api.slack.com/docs/message-attachments#fields)].

brian-brazil: That is still not what is implemented in the alertmanager. I think you need to fix the code.

taharah:
Already resolved by #1025.

PR #1008, opened 2018-04-29
fix indentation on node_exporter example

bleything:
This is a tiny tiny change to fix an example in the introduction docs. If used verbatim in its current state, prometheus will fail to start.


RichiH:
Thanks!

PR #1009, opened 2018-05-01
added nginxplus_exporter to list of exporters

avthart:
NGINX Plus exporter. Exporter using the `ngx_http_status_module` module which provides access to various status information. This module is available as part of NGINX Plus commercial subscription.
http://nginx.org/en/docs/http/ngx_http_status_module.html

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -79,6 +79,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Apache exporter](https://github.com/Lusitaniae/apache_exporter)
    * [HAProxy exporter](https://github.com/prometheus/haproxy_exporter) (**official**)
    * [Nginx metric library](https://github.com/knyar/nginx-lua-prometheus)
+   * [Nginx Plus exporter](https://github.com/avthart/nginxplus_exporter)
    * [Nginx VTS exporter](https://github.com/hnlq715/nginx-vts-exporter)
    * [Passenger exporter](https://github.com/stuartnelson3/passenger_exporter)
    * [Tinyproxy exporter](https://github.com/igzivkov/tinyproxy_exporter)


brian-brazil:
Have you discussed adding this to the existing nginx exporter? I'd prefer that users not to have to run a separate module for every nginx module they use.

You're using direct instrumentation, you should really be using MustNewConstMetric in exporters.

avthart:
@brian-brazil 
Thank you for your feedback.

The other NGINX exporters which are listed on the Prometheus Exporters page are based on custom community nginx modules such as "nginx-module-vts" or using custom lua code. 

This exporter is using the official "ngx_http_status" module from NGINX self (only available with the commercial edition of NGINX). Therefore this exporter can only be used when you have a subscription. Besides that this exporter is exposing much more metrics than the other exporters.

I have will have a look at direct instrumentation. The current implementation is based on the official haproxy_exporter.

brian-brazil:
I don't see how that precludes a merge with others. Can you discuss it with the VTS exporter maintainer?

avthart:
I did do some research on the the VTS exporter and found that are a lot of differences with the metrics provided by the ngx_http_status_module. Some metrics and labels provided by the VTS module are not available in the NGINX Plus version and vice versa. 

I think it 's not a good idea to have one exporter which exposes different set of metrics and labels based on the module you are using. It will require some redesign work of the metrics and labelling if you want to have one exporter to support all nginx status modules (and which are we going to support, because there are many).  

I prefer to keep the implementation simple and in line with the metrics provided by the ngx_http_status_module.

brian-brazil:
Such difference are not at all unknown, both the mysqld and node exporters have that for example. Please chat with the existing exporter maintainer to see if there's a way ye can work together.

avthart:
Allright. I will try to have a chat with the maintainer of the nginx vst exporter.

avthart:
Will close this PR. Maintainer of the other nginx exporter did not answer my email. 

Besides that NGINX has created a official exporter for both NGINX and NGINX Plus. 
Maybe it is a good idea to add this one to the list of exporters, see:
See https://github.com/nginxinc/nginx-prometheus-exporter

PR #1010, opened 2018-05-02
Adding RADOSGW Usage exporter

blemmenes:
As this exporter now supports Luminous (current LTS release) I wanted to add it to the exporters page.


brian-brazil:
We try to avoid duplication in this list, so I'd suggest trying to work with the existing listed ceph exporter rather than creating a new one.

blemmenes:
@brian-brazil I can certainly get in touch with the digital ocean folks who maintain the existing exporter. However my reasoning for creating a separate exporter is that these talk to different components of the cluster.

While their exporter talks directly to the cluster monitors (a component that every Ceph cluster has) this exporter is pulling data from RADOSGW nodes. The RADOSGW component is not something that every ceph cluster operator uses and may not even be installed/running.

My feeling was as RGW was a separate optional component a separate exporter made sense.

Duplication goes the separate ceph exporter may be already be supplanted by upstream ceph adding Prometheus instrumentation directly into the base ceph-manager daemon. I have not yet had a chance to deploy that however.

Thanks,
Berant

brian-brazil:
> The RADOSGW component is not something that every ceph cluster operator uses and may not even be installed/running.

From the research I've done, RADOS is a core component in every cluster. The GW is just one API to access it.

blemmenes:
You are correct that RADOS is the core of Ceph however that is not what this exporter is reporting on. [This](http://docs.ceph.com/docs/master/architecture/) architecture page as a diagram at the top that helps show how the different components interact.

A cluster that is just doing block (RBD) say for an openstack cluster would not need any RADOSGW nodes at all, same would go for just a cluster that is providing cephfs access. A single cluster could also be providing all three.

This exporter is just for the radosgw portion (s3/swift compatable), and reports on usage stats (bytes/ops IN/OUT) for operations by bucket (GET/PUT/DEL/LIST), which allows for an operator to track and potentially bill by operation count as well as bytes IN/OUT (like AWS S3 does).

It also exports total data on disk by bucket/user for an operator that is just billing based on bytes stored

Any transactions done either directly via LIBRADOS, RBD, or cephfs would not show up in RADOSGW usage data as those interactions don't go through the RADOSGW API.

Thanks,
Berant

brian-brazil:
Thanks for the explanation.

The way you are exposing the category label doesn't follow the rules for labels, as it's not a partition across the space. I'd suggest info metrics with this information rather than double counting in the main metrics.
Taking secrets via command line is unwise, as they'll be exposed to everyone on the machine.
radosgw_usage_bucket_total is a Gauge so shouldn't end in _total, and should also have a unit.

blemmenes:
Brian,

Thanks for the feedback, I'll work to address those and let you know when it is pushed.

blemmenes:
@brian-brazil I've corrected the gauge (removed _total and added base unit). 

For the credentials, while it does accept the access/secret keys via command line it also does via environment variable. Would you prefer eliminating the option for passing it via argument or is calling out the security ramifications in the docs and recommending using an env variable sufficient? There seems to be a smattering of both among the exporters.

Regarding the category label, forgive my misunderstanding, however I'm not grasping how what I'm doing isn't a partition across the space. To me these seem the same:

radosgw_usage_ops_total{bucket="test",category="list_bucket",owner="nano"} 2.0
radosgw_usage_ops_total{bucket="test",category="put_obj",owner="nano"} 1.0

Where "category" is somewhat analogous to HTTP method.

And this from the haproxy exporter:
http_requests_total{code="200",handler="prometheus",method="get"} 2.287328e+06
haproxy_server_http_responses_total{backend="radosgws",code="2xx",server="radosgw02"} 7.2825603e+07

An info metric (as used to specify a software version) wouldn't be able to capture the timeseries for the number of operations of a given category.

Thanks again for the feedback, I appreciate it.

brian-brazil:
I can't merge this until you add the DCO.

> Where "category" is somewhat analogous to HTTP method.

My interpretation from the code is that a bucket can be in multiple categories, which is not the same as http method.

> For the credentials, while it does accept the access/secret keys via command line it also does via environment variable. Would you prefer eliminating the option for passing it via argument or is calling out the security ramifications in the docs and recommending using an env variable sufficient? There seems to be a smattering of both among the exporters.

I personally would go for a file as env vars are not designed for secrets, but getting it off the command line is the important thing.

brian-brazil:
If that's the name in the docs, then that's best.

I'd do a `git rebase -i master`, squash down to one commit, and then amend that commit.

blemmenes:
gah, now I borked it even more. yay.

blemmenes:
Yeah, I'll close this PR and resubmit.

PR #1011, opened 2018-05-04
Add bitcoind-exporter to the list of misc. exporters

lutangar:
Hi there,

Our [bitcoind-exporter](https://github.com/LePetitBloc/bitcoind-exporter) is finally ready.
`Bitcoind` is the official bitcoin daemon, this exporter rely on the **bitcoin** JsonRPC API to get chain synchronization metrics, wallet balances, address balances, etc...

Hence I'm proposing to add it to the list.

Before anything, we would be glad to get a review of the exposed metrics, here's an example:

```
# HELP bitcoind_best_block_index The block height or index
# TYPE bitcoind_best_block_index gauge
bitcoind_best_block_index 69019

# HELP bitcoind_best_block_timestamp_seconds The block time in seconds since epoch (Jan 1 1970 GMT)
# TYPE bitcoind_best_block_timestamp_seconds gauge
bitcoind_best_block_timestamp_seconds 1522746083

# HELP bitcoind_chain_difficulty The proof-of-work difficulty as a multiple of the minimum difficulty
# TYPE bitcoind_chain_difficulty gauge
bitcoind_chain_difficulty 3511060552899.72

# HELP bitcoind_wallet_version the wallet version
# TYPE bitcoind_wallet_version gauge
bitcoind_wallet_version{ticker="BTC"} 71000

# HELP bitcoind_wallet_balance_total the total balance of the wallet
# TYPE bitcoind_wallet_balance_total gauge
bitcoind_wallet_balance_total{status="unconfirmed"} 2.7345
bitcoind_wallet_balance_total{status="immature"} 0
bitcoind_wallet_balance_total{status="confirmed"} 42.73453501

# HELP bitcoind_wallet_transactions_total the total number of transactions in the wallet
# TYPE bitcoind_wallet_transactions_total gauge
bitcoind_wallet_transactions_total 77

# HELP bitcoind_wallet_key_pool_oldest_timestamp_seconds the timestamp of the oldest pre-generated key in the key pool
# TYPE bitcoind_wallet_key_pool_oldest_timestamp_seconds gauge
bitcoind_wallet_key_pool_oldest_timestamp_seconds 1516231938

# HELP bitcoind_wallet_key_pool_size_total How many new keys are pre-generated.
# TYPE bitcoind_wallet_key_pool_size_total gauge
bitcoind_wallet_key_pool_size_total 1000

# HELP bitcoind_wallet_unlocked_until_timestamp_seconds the timestamp that the wallet is unlocked for transfers, or 0 if the wallet is locked
# TYPE bitcoind_wallet_unlocked_until_timestamp_seconds gauge
bitcoind_wallet_unlocked_until_timestamp_seconds 0

# HELP bitcoind_wallet_pay_tx_fee_per_kilo_bytes the transaction fee configuration, set in Unit/kB
# TYPE bitcoind_wallet_pay_tx_fee_per_kilo_bytes gauge
bitcoind_wallet_pay_tx_fee_per_kilo_bytes 0

# HELP bitcoind_address_balance_total address balance
# TYPE bitcoind_address_balance_total gauge
bitcoind_address_balance_total{address="1FxZE15d8bt381EuDckdDdp7vw8FUiLzu6"} 41.00683469
bitcoind_address_balance_total{address="1QAm6J6jLmcm7ce87ujrSdmjPNX9fgRUYZ"} 1.72770032
```

We also reserved the port number `9439` at https://github.com/prometheus/prometheus/wiki/Default-port-allocations

Cheers.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -166,6 +166,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
 ### Miscellaneous
    * [ACT Fibernet Exporter](https://git.captnemo.in/nemo/prometheus-act-exporter)
    * [BIND exporter](https://github.com/digitalocean/bind_exporter)
+   * [Bitcoind exporter](https://github.com/LePetitBloc/bitcoind-exporter)
    * [Blackbox exporter](https://github.com/prometheus/blackbox_exporter) (**official**)
    * [BOSH exporter](https://github.com/cloudfoundry-community/bosh_exporter)
    * [cAdvisor](https://github.com/google/cadvisor)


content/docs/instrumenting/exporters.md
```diff
@@ -158,7 +158,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Transmission exporter](https://github.com/metalmatze/transmission-exporter)
    * [Unbound exporter](https://github.com/kumina/unbound_exporter)
    * [Xen exporter](https://github.com/lovoo/xenstats_exporter)
-   
+   * [Bitcoind exporter](https://github.com/LePetitBloc/bitcoind-exporter)
```

lutangar:
Thanks for the swift review.

> Counters should end with _total, gauges should not.

On `wallet_transactions_total` for example the **counter** make more sense as this value is not supposed to go down.
BUT I'm only client of the service I'm monitoring so I just can't `increment` this value, I'm just able to `set` it despite the fact that it **is** a **total**.
Could the rule have exceptions or is this a strict thing? As for the others *gauges* I get it and I'll change them.

> Don't forget the units in your metrics names,

I'm failing to see the missing units... oh wait... maybe you're talking about the *balance* metrics? 
`bitcoind_wallet_balance` for example, in this case the unit might change depending on the monitored wallet, it could be BTC as well as DASH for example.
Should I put that in the metric label?

> and you should really be prefixing everything with bitcond..

Done.

> You also seen to be using direct instrumentation rather than const metrics, you may need to switch to a different client library to enable this.

Here I'm lost. Maybe it has do with the fact I'm only client of a service. Am I supposed to *cache* these metrics or something?

Cheers.

gouthamve:
Hi @lutangar, are the updates for the above done? Could you rebase if you're happy?

brian-brazil:
The DCO and sorting were what I was waiting on, and there's just a conflict to fix now.

lutangar:
From what I can remember, @brian-brazil recommended a complete rewrite of the exporter in another language, but you probably should close this PR since my fu is limited in the suggested languages. 

Didn't check if there were any news on the *const metrics* thing for node.js, but it's probably just not doable.

...or maybe I didn't get this right at all and it wasn't a blocker?!

brian-brazil:
If it's a working sane non-duplicate exporter we'll list it, even if it's not perfect. The rest are suggestions.

lutangar:
Alright then, here you go.

brian-brazil:
Thanks!

PR #1012, opened 2018-05-04
Add Collins exporter

beorn7:
@juliusv @brian-brazil (to ask some non-SoundClouders)


beorn7:
> up and scrapeDuration should use a ConstMetric rather than direct instrumentation.

Not in this case as we possible share Collins scrapes between Prometheus scrapes.

> The state section at the bottom of the readme is obsolete, the current code looks to be doing the recommended thing.

It's not obsolete. Note that state != status in Collins.

brian-brazil:
> It's not obsolete. Note that state != status in Collins.

Ah, I missed that. What you're doing or using an info metric would be the options I see from a quick peek at the docs.

Issue #1013, opened 2018-05-05
Document alertmanager webhook receivers


SuperQ:
Like we list exporters, we should also have a list of webhook receivers to make it easier to find integrations, like IRC bots.

PR #1014, opened 2018-05-05
Add a list of AM receivers.

SuperQ:
Signed-off-by: Ben Kochie <bjk@gitlab.com>

Closes: https://github.com/prometheus/docs/issues/1013

### DIFF
--- a/content/docs/alerting/webhook_receivers.md
+++ b/content/docs/alerting/webhook_receivers.md
@@ -1,0 +1,13 @@
+---
+title: Webhook Receivers
+sort_rank: 6
+nav_icon: sliders
+---
+
+# Webhook Receivers
+
+There are a number of 3rd party systems that directly support the Alertmanager webhook receiver.
+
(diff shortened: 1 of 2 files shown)


brian-brazil:
We already have https://prometheus.io/docs/operating/integrations/, so you should add to that list.

content/docs/alerting/webhook_receivers.md
```diff
@@ -0,0 +1,13 @@
...
+sort_rank: 6
+nav_icon: sliders
+---
+
+# Webhook Receivers
+
+There are a number of 3rd party systems that directly support the Alertmanager webhook receiver.
+
+* [IRC Bot](https://github.com/metalmatze/alertmanager-bot)
+* [Audio Text-to-Speech](https://www.robustperception.io/audio-alerting-with-prometheus/)
```
brian-brazil: I don't think you should list this, it's a very silly example for a rather old version of the AM

SuperQ:
Oh, I didn't see that under integrations, it's not google ranked very well.

PR #1015, opened 2018-05-05
Add IRC bot to integrations list.

SuperQ:

Closes: https://github.com/prometheus/docs/issues/1013


brian-brazil:
As far as I can see this is not an IRC bot, it is a telegram bot of which we already have one listed.

SuperQ:
Oh, my mistake, how about https://github.com/multimfi/bot

PR #1016, opened 2018-05-06
Add nomad exporter

pcarranza:
@brian-brazil or @SuperQ I can haz PR?

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -195,6 +195,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Locust Exporter](https://github.com/ContainerSolutions/locust_exporter)
    * [Meteor JS web framework exporter](https://atmospherejs.com/sevki/prometheus-exporter)
    * [Minecraft exporter module](https://github.com/Baughn/PrometheusIntegration)
+   * [Nomad exporter](https://gitlab.com/yakshaving.art/nomad-exporter)
    * [nftables exporter](https://github.com/Intrinsec/nftables_exporter)
    * [OpenStack exporter](https://github.com/openstack-exporter/openstack-exporter)
    * [OpenStack blackbox exporter](https://github.com/infraly/openstack_client_exporter)
@@ -277,7 +278,7 @@ make use of one of the normal Prometheus client libraries under the hood. As
 for all independently maintained software, we cannot vet all of them for best
 practices.
 
-   * Clojure: [iapetos](https://github.com/clj-commons/iapetos) 
+   * Clojure: [iapetos](https://github.com/clj-commons/iapetos)
    * Go: [go-metrics instrumentation library](https://github.com/armon/go-metrics)
    * Go: [gokit](https://github.com/peterbourgon/gokit)
    * Go: [prombolt](https://github.com/mdlayher/prombolt)
@@ -287,3 +288,4 @@ practices.
    * Java/JVM: [Micrometer Prometheus Registry](https://micrometer.io/docs/registry/prometheus)
    * Python-Django: [django-prometheus](https://github.com/korfuri/django-prometheus)
    * Node.js: [swagger-stats](https://github.com/slanatech/swagger-stats)
+


SuperQ:
It looks like this is a fork of an exporter.  Why not push your changes upstream?

Also, it looks like this exporter doesn't have a [port allocation](https://github.com/prometheus/prometheus/wiki/Default-port-allocations) and is using one from another exporter.

pcarranza:
I can try pushing upstream, but it changed so much that I'm not even sure it will be accepted.

pcarranza:
https://github.com/Nomon/nomad-exporter/pull/25 maybe it gets accepted, dunno.

brian-brazil:
Thanks for sharing, in addition to Ben's comments here's some things you could improve:

raft_peers is a bit odd. This is a kube-state-metrics like exporter in which talking to a random node is okay, however raft_peers is about the cluster itself so you should have an exporter beside every node to get that in case it differs. I'd suggest dropping both that and the leader logic to avoid confusion and as the leader logic seems fragile.

Rather than having nomad_client_errors_total which would only spot errors in the previous scrape, I'd suggest either setting nomad_up to 0 if there is an error or failing the scrape. Partial data is virtually impossible to work with for users.

Info metrics should usually be constant over time, so I'd move each of status and drain to separate boolean gauge metrics.  serf_lan_member_status is an info metric, so should end in _info.

deployment_task_group_placed_allocs_total is a gauge, so should not have a _total suffix which is for counters.

allocation_memory_rss_bytes_limit limit should go before bytes, so the unit is always at the end of the metric

task_cpu_percent prefer ratios to percentages, as they're the base units.

brian-brazil:
Also, deploymentTaskGroupDesiredCanaries and a few others in collectDeploymentMetrics are not using ConstMetrics.

pcarranza:
Thanks @brian-brazil 

The reason to have partial metrics is that we have noticed that sometimes the API gets a bit flaky and it fails, I didn't want to lose a scrape just because some section of it couldn't make it.

Regarding the leader logic, I know it's a bit brittle, but I do want to avoid impacting the nodes as it seems that to get the nodes and allocations metrics nomad is effectively issuing queries to all the nodes synchronously 😒, therefore I wanted to reduce the impact of querying for metrics to a single node.

I will review the metrics names. Only one question:

> task_cpu_percent prefer ratios to percentages, as they're the base units

You mean to change the metric name or to get it differently? I was investigating nomad api and the way they declare it is by percentage, I'm not sure it's the appropriate unit though as it can go over 100, the thing is I didn't dig deep enough to understand exactly what is it.

BTW, just changed the port to the reserved one.

brian-brazil:
> The reason to have partial metrics is that we have noticed that sometimes the API gets a bit flaky and it fails, I didn't want to lose a scrape just because some section of it couldn't make it.

If you want to do that then you'll need a metric indicating which parts failed, and more importantly don't return any metrics for those parts.

> I wanted to reduce the impact of querying for metrics to a single node.

That's okay, as long as the leader and peers metrics aren't present. Otherwise you're mixing two different deployment models in one exporter.

> You mean to change the metric name or to get it differently?

Change the metric name.

> I'm not sure it's the appropriate unit though as it can go over 100

Hmm, are tasks allowed to burst above their allocation for cpu?

pcarranza:
@brian-brazil reiterating a bit on this

> raft_peers is a bit odd. This is a kube-state-metrics like exporter in which talking to a random node is okay, however raft_peers is about the cluster itself so you should have an exporter beside every node to get that in case it differs. I'd suggest dropping both that and the leader logic to avoid confusion and as the leader logic seems fragile.

I see what you mean, in fact having the peers from each node perspective makes sense, the only I think I want to extract from the server is the state metrics, so maybe it makes sense to split this -> always take the peers metric for all the hosts (this is intended to be executed next to every nomad server, which is a raft peer), then only take the stats metrics from the leader.

What I don't want is all the hosts querying all the other hosts for stats on every scrape.

pcarranza:
Changes visible here: https://github.com/pcarranza/nomad-exporter/pull/1

brian-brazil:
> always take the peers metric for all the hosts (this is intended to be executed next to every nomad server, which is a raft peer), then only take the stats metrics from the leader.

This is not the architecture for exporters though, it's usually 1:1. This would usually be a separate exporter for the metrics of nomad itself so that scheduling and service discovery remains with Prometheus. Thus I suggest dropping these metrics for simplicity.

brian-brazil:
Looks like the upstream is abandoned, an attempt was made to work upsteam, and this one has a small community already so I think it's okay to list it.

Can you add the DCO?

pcarranza:
Done, and updated the url as I moved the project to gitlab.

brian-brazil:
Thanks!

PR #1022, opened 2018-05-12
Add dingtalk to the list of integrations

timonwong:


brian-brazil:
Thanks!

You seem to be putting secrets in the --ding.profile flag, for security that should move to a configuration file so noone else on the machine can see it.

PR #1023, opened 2018-05-13
Add Minio

nitisht:
Add Minio to list of software exposing Prometheus metrics


brian-brazil:
Thank for sharing. I'm not sure this really classifies as directly instrumented, most of your metrics are coming from your own internal instrumentation system rather than Prometheus instrumentation. You can list this as something exposing Prometheus metrics, and you also need to satisfy the DCO.

minio_server_start_time_seconds is also redundant with the standard process_start_time_seconds.

nitisht:
Thanks for the info! I have updated the PR. 

We'll take a look at the redundancy you pointed out.

brian-brazil:
Thanks!

PR #1024, opened 2018-05-13
Added RavenDB exporter to the list

marcinbudny:
This PR extends the list of exporters with [one for RavenDB](https://github.com/marcinbudny/ravendb_exporter). RavenDB is a [document database](https://ravendb.net/).


marcinbudny:
Thanks for pointing out the problem with metric names. I will adjust them. 
The process collector also looks interesting, however in dockerized environment it wouldn't be easy to get it to work. Or am I missing something?

PR #1027, opened 2018-05-20
Add NVIDIA GPU exporter

mindprince:


brian-brazil:
You need to add the DCO.

Thanks for sharing, I have some suggestions to improve your exporter:

You are using direct instrumentation, you should be using MustNewConstMetric instead.

memory_total_bytes should be memory_bytes, total is generally to be avoided in metric names as most metrics are totals.

duty_cycle and fanspeed_percent should be exposed as ratio rather than a percentage, and power_usage_milliwatts would be better in watts.

mindprince:
Thanks for the review.

I added the "Signed-off-by" line in the commit message.

Regarding naming and units, I was trying to keep them similar to how they are represented in `nvidia-smi` and cAdvisor etc.

brian-brazil:
Thanks!

PR #1029, opened 2018-05-24
Replace lovoo/ipmi_exporter by soundcloud/ipmi_exporter

beorn7:
Frankly, I would prefer to list both as they have quite different
approaches. Lovoo's exporter runs on the IPMI-supervised host and uses
the `dev/ipmi0` device to server metrics. This is straight-forward but
doesn't work if the host is powered down. However, in some secnarios,
it's the whole point of IPMI to be able to work with machines while
they are powered down or stuck in the boot process. If you need
metrics in those scenarios, too, the SoundCloud exporter works
better. It uses the blackbox exporter approach and is thus (arguably)
a tad more complicated to set up.

If I _had_ to choose one and only one IPMI exporter, I would pick the
SoundCloud one as it covers all scenarios.  Since @brian-brazil has so far
insisted to only ever list one solution for a certain thing to
export, this commit _replaces_ the Lovoo exporter by the SoundCloud
exporter.

Discuss!

@brian-brazil see above.
@juliusv as another non-SC Promethean to help decide here.
@bitfehler is the main author of SC's IPMI exporter.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -54,7 +54,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [apcupsd exporter](https://github.com/mdlayher/apcupsd_exporter)
    * [Collins exporter](https://github.com/soundcloud/collins_exporter)
    * [IoT Edison exporter](https://github.com/roman-vynar/edison_exporter)
-   * [IPMI exporter](https://github.com/lovoo/ipmi_exporter)
+   * [IPMI exporter](https://github.com/soundcloud/ipmi_exporter)
    * [knxd exporter](https://github.com/RichiH/knxd_exporter)
    * [Node/system metrics exporter](https://github.com/prometheus/node_exporter) (**official**)
    * [Ubiquiti UniFi exporter](https://github.com/mdlayher/unifi_exporter)


brian-brazil:
Without looking at the exact metrics exported by either, the main question is is this exporter (largely) a superset of the existing one?

IPMI was always going to be a bit of a weird one due to both the standard exporter and blackbox-style ways in which it could reasonably be used. I personally (at least when my home server gets back from repairs) need the on-host one. Is there any way one exporter could do both?

beorn7:
Obviously, the soundcloud exporter can also monitor IPMI devices while their supervised host is running.

Not sure if there are any metrics retrievable by the on-host exporter that the soundcloud exporter doesn't export.

But fundamentally, with a certain configuration overhead, the soundcloud exporter should be usable in all use cases.

brian-brazil:
> Not sure if there are any metrics retrievable by the on-host exporter that the Soundcloud exporter doesn't export.

From a quick peek they all seem to be there. Both shell out to the same tool, so it's not hard to change anyway.

The Soundcloud one can't work only via the kernel though, and I believe it'd be a small change to make it able to work both ways. Then for me boils down to if the existing exporter plans on supporting this, as I much prefer that the community collaborate on one exporter rather than recreating exporters in isolation.

The Soundcloud one has a security issue, it's passing passwords on the command line.

beorn7:
> The Soundcloud one has a security issue, it's passing passwords on the command line.

Oops. @bitfehler, I guess that needs a fix.

About using the local IPMI device in case it exists: @bitfehler can probably judge how much work that would be. It's definitely not needed in our use case. A PR would be welcome, I guess.

Changing the Lovoo exporter to the blackbox exporter pattern would probably equal a rewrite. That was the original motivation to write our own one in the first place.

bitfehler:
I'll take care of the password issue and also look into supporting local host metrics, it should indeed not be too hard.

brian-brazil:
I don't see any need for two exporters here, especially as both as shelling
out to ipmitool rather than importing varying different protocols for ipmi
themselves. Divergence in inetrics would also be confusing to users, and
the duplication in effort seems a bit wasteful.

beorn7:
soundcloud/ipmi_exporter is now capable to collect metrics from a local IPMI device. Can we merge this now?

Hopefully nobody has written yet another IPMI exporter because the one linked so far on prometheus.io doesn't support remote collection.

PR #1032, opened 2018-05-27
Remote write to Elasticsearch

boernd:
Signed-off-by: bhois <bernd.hois@bearingpoint.com>


brian-brazil:
Thanks!

There is a HTTP header with the remote write version sent, so the user shouldn't need to configure that. Watch out also for Inf/-Inf/NaN/StaleNaN values, as JSON can't represent these floating point values.

PR #1042, opened 2018-06-02
Added EventStore exporter to the list

marcinbudny:
Adds [EventStore exporter](https://github.com/marcinbudny/eventstore_exporter) to the list. EventStore is an... [event store](https://eventstore.org/).


brian-brazil:
Thanks!

I'd suggest reading https://prometheus.io/docs/instrumenting/writing_exporters/ to see ways you can make your exporter even better, for example counters and only counters should end in _total and you should use ConstMetrics rather than the normal Gauge/Counter objects. It's also unclear what memory that `process_memory_bytes` refers to - is that resident, virtual, or something else?

PR #1044, opened 2018-06-04
Adds ACT Fibernet Exporter

captn3m0:
[ACT](https://www.actcorp.in/) is a Internet Service Provider in India. The provider exposes data usage metrics by scraping it from the ACT portal website.

Looks like this in Grafana:

image

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -133,6 +133,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [StatsD exporter](https://github.com/prometheus/statsd_exporter) (**official**)
 
 ### Miscellaneous
+   * [ACT Fibernet Exporter](https://git.captnemo.in/nemo/prometheus-act-exporter)
    * [Bamboo exporter](https://github.com/AndreyVMarkelov/bamboo-prometheus-exporter)
    * [BIG-IP exporter](https://github.com/ExpressenAB/bigip_exporter)
    * [BIND exporter](https://github.com/digitalocean/bind_exporter)


brian-brazil:
You need to add the DCO before I can merge.

As a way to improve your exporter, gauges should not have a _total suffix. It's also recommended to export bytes rather than any other unit.

RichiH:
Networks speeds are, unfortunately, normalized to using bits. In this specific case, doing calculations actually becomes more cumbersome if some data is in byets.

brian-brazil:
This exporter is exposing usage in gigabytes. The base units remain bytes - even for network speeds (in fact that's the 2nd biggest unit that needs standardisation after seconds, mixing bits and bytes does not end well).

captn3m0:
I wanted to use bytes as well, but it felt like giving a fake sense of accuracy since the portal reports usage in GB:

image

Can't do bits, since this is network transfer (and not network speed)

brian-brazil:
The standard in Prometheus is always to use base units, no matter the precision. You can always mention the precision in the help string.

captn3m0:
Will push out a new release in that case. Does this look okay?

```
# HELP act_fup_usage_bytes ACT current usage in bytes (precision is in GB)
# TYPE act_fup_usage_bytes gauge
act_fup_usage_bytes_total 41420000000

# HELP act_fup_max_bytes ACT FUP limit in bytes (precision is in GB)
# TYPE act_fup_max_bytes gauge
act_fup_max_bytes 500000000000
```

Thanks for the suggestions :100:!

brian-brazil:
act_fup_usage_bytes_total should be act_fup_usage_bytes as it's a gauge rather than a counter.

I can't merge until you add the DCO.

captn3m0:
- DCO is already passing (I force pushed with the amended commmit). Do I need to do something on my repo as well?
- Removed the `_total`
- Did a new `1.0.2` release with the change.

PR #1047, opened 2018-06-05
De-recommend checking full label consistency for custom collectors.

brian-brazil:
Signed-off-by: Brian Brazil <brian.brazil@robustperception.io>

@beorn7

### DIFF
--- a/content/docs/instrumenting/writing_clientlibs.md
+++ b/content/docs/instrumenting/writing_clientlibs.md
@@ -250,8 +250,9 @@ Client libraries MUST NOT under any circumstances allow users to have different
 label names for the same metric for Gauge/Counter/Summary/Histogram or any
 other Collector offered by the library.
 
-If your client library does validation of metrics at collect time, it MAY also
-verify this for custom Collectors.
+Metrics from custom collectors should almost always have consistent label
+names. As there are still rare but valid use cases where this is not the case,
+client libraries should not verify this.
 
 While labels are powerful, the majority of metrics will not have labels.
 Accordingly the API should allow for labels but not dominate it.


content/docs/instrumenting/writing_clientlibs.md
```diff
@@ -250,8 +250,9 @@ Client libraries MUST NOT under any circumstances allow users to have different
 label names for the same metric for Gauge/Counter/Summary/Histogram or any
 other Collector offered by the library.
 
-If your client library does validation of metrics at collect time, it MAY also
-verify this for custom Collectors.
+While metrics from custom collectors generally should also have consistent label
+names, as there are rare valid use cases where this is okay client libraries
```
RichiH: That sentence is more complex than it should be.

brian-brazil: What would you suggest?

RichiH: As a general rule, metrics from custom collectors generally should also have consistent label names. As there are still rare but valid use cases where this is not the case, client libraries should not verify this. An example of such a rare case would be XXX

Last sentence optional, but I can't think of any example off-hand, so including one seems useful to me.

brian-brazil: Adjusted

RichiH: Where? GH only shows me one commit; the initial one.

brian-brazil: Forgot to push it

RichiH:
LGTM.

@beorn7 ?

beorn7:
:+1:

Issue #149, opened 2015-08-03
Add non-goals


beorn7:
Perhaps next to the road map, add a "non-road-map" with the non-goals, i.e. things that sound like nice features but which we do not intend to implement for reasons to be stated in the new section.

Ideas so far:

- Rule templating language.
- Strings as sample values (and neither ints...).
- Support event logging / become a log db.

More?

brian-brazil:
* Supporting JSON as a scrape format
  * Long term storage in prometheus (hopefully)
  * Per-host aggregators
  * SSL and auth for our http frontends (use a reverse proxy)

brian-brazil:
* Retention policies (use remote storage for that)

juliusv:
I personally wouldn't rule out retention policies (deleting certain series completely after a configurable time, not talking about downsampling here). It shouldn't be too hard to add and could be quite useful.

brian-brazil:
* Configurable logging (use a pipe)

brian-brazil:
* A single exporter per machines/service reporting in node exporter

brian-brazil:
Thinking last night this is probably also the place to list our assumptions about the minimum state of your infrastructure (machine database, service database, configuration management), and that we won't add features solely to workaround the lack of same.

RichiH:
While I strongly disagree with the "no long-term storage" bit, a constraints section which lists design non-goals, anti-patterns, and basic requirements would make sense.

fabxc:
To clarify, the long-term storage part means we do not intend for the storage layer to become one. We do however intend to provide better integrations with existing storage solutions to serve as a LTS for Prometheus data. Compared to what we have now, this especially means read-back support.

RichiH:
Just to make sure: The plan is to put historic data into some other storage engine at a magic cut-off point to maintain performance for near-time metrics while retaining full information for old data?

Sounds good to me :)

As an aside, it might make sense to detect frequently-accessed historical data to keep that in warm storage.

beorn7:
> As an aside, it might make sense to detect frequently-accessed historical data to keep that in warm storage.

I was thinking about a simple query cache here. Queries to the LTS backend only have to be repeated if the most recent sample (per timeseries) is now beyond the rentention period of the Prometheus server (or if the query result has been evicted from the cache due to memory constraints).

But note that there a lot of more urgent features to implement before we'll tackle read-back from LTS.

brian-brazil:
* Non-HTTP scraping

[...]

PR #1054, opened 2018-06-10
docs/philosophy: Initial commit

RichiH:
This was surprisingly hard; I went through several iterations, tossing things away, re-adding some, and refactoring into different files. On the plus side, I think what I have is already relatively cleanly separated. design & patterns is a collection of half-finished thoughts and sentences with tons of duplicates. No use pushing those atm.

I would be especially interested in large conceptual comments and missing sections, less so in specific wording as that is bound to change, yet again, once there's more content.

### DIFF
--- a/content/docs/philosophy/goals.md
+++ b/content/docs/philosophy/goals.md
@@ -1,0 +1,72 @@
+---
+title: Goals and Non-Goals
+sort_rank: 3
+---
+
+# Goals
+
+## Resilience
+
+First and foremost, Prometheus must be resilient in operation.
+
+
+## Reliable alerting
+
+As a monitoring system, Prometheus is being relied upon to alert humans that
+they need to take action in order to prevent undesired system state.
+
+Thus, its most important function is to keep the pipeline of ingestion, rule
+evaluation, and alert hand-off working.
+
+The second most important function is to give humans context about these alerts
+by allowing access to the most recent data Prometheus ingested
+
+### Resulting design decisions and patterns
+
+Note that this goal might result in widely different design decisions and thus
+operational patterns for different parts of our ecosystem:
+
+For Prometheus itself, this means running every instance as an island of data
+completely detached from every other instance, except for optional federation.
+
+For Alertmanager on the other hand, it means the exact opposite: meshing all
+instances closely together, sharing knowledge about alerts and their
+notifications.
+
+## Simple operation
+
+Operation of Prometheus should be as easy and failure-tolerant as possible. We
+try to put required complexity into earlier phases, going through them less
+often and ideally still while under the control of a smaller subset of people.
+
+One example of this would be the preference of statically linked binaries over
+dynamically-built ones.
+
+## Automation
+
+Computers are good at doing the same thing over and over again, and quickly.
+Humans tend to be better at creative tasks.
+
+Prometheus will always strive to automate away all tasks whenever possible
+through various means; some specific implementations would be service discovery,
+label rewriting, and alert generation.
+
+
+# Non-Goals
+
+# Event handling
+
+Prometheus is dealing with metrics. As such, it will never process and store
+events.
+
+The only exception in our ecosystem is Alertmanager which deals with individual
+alerts and alert groups.
+
+For ways to deal with events, see TODO patterns.
+
+# Push-type system
+
+Prometheus is, and always will be, a pull-type system. We strongly believe that
+this makes operational sense in all but the very largest of scales.
+
+For ways to integrate with push-type systems, see TODO patterns.
--- a/content/docs/philosophy/philosophy.md
+++ b/content/docs/philosophy/philosophy.md
@@ -1,0 +1,69 @@
+---
+title: Philosophy
+sort_rank: 2
+---
+
+# Do one thing well
+
+We believe in the [Unix
+philosophy](https://en.wikipedia.org/wiki/Unix_philosophy), abridged from [Doug
+McIlroy's initial version from 1978](http://emulator.pdp-11.org.ru/misc/1978.07_-_Bell_System_Technical_Journal.pdf):
+
+1. Make each program do one thing well.
+  While the scope of "one thing" invariably encompasses more and more
+  elements due to increased overall system and computing complexity, we are
+  still doing one thing: ingest metric data, do computations on it, and expose
+  it to other systems.
+2. Expect the output of every program to become the input to another, as yet unknown, program.
+  Today's lingua franca is HTTP endpoints, which are used by Prometheus
+  extensively.
+  In the same vein, Prometheus relies heavily on its own libraries and strict
+  layering internally.
+3. Design and build software, even operating systems, to be tried early, ideally within weeks. Don't hesitate to throw away the clumsy parts and rebuild them.
+  Prometheus will always be available for free as in beer and as in speech.
+  We ensure that master always builds, called Continuous Integration these days,
+  and we not afraid to replace whole sections of our codebase, e.g. our storage
+  engine.
+4. Use tools in preference to unskilled help to lighten a programming task, even if you have to detour to build the tools and expect to throw some of them out after you've finished using them.
+  While this is an outdated way of stating the goal, automation where possible
+  is still one of the core characteristics any modern philosophy.
+
+## Embrace cloud-native technologies
+
+In many ways, the cloud-native approach mirrors the Unix philosphy, updating it
+for the modern world.
+
+1. Micro-services are the equivalent of doing one thing well
+2. Ubiquitous APIs enable interoperatbility in the cloud-native world
+3. Releasing early, realeasing often, and failing quickly is important when
+failure is part of expected operations
+4. Automation is key, freeing up humans to make more useful use of their time
+
+# Be pragmatic
+
+To not lose focus, we need to be honest to our users and ourselves about what we
+can do and not do.
+
+# Be open
+
+We will always put as much of our code, discussions, presentations, and other
+content into a form and place which is accessible in the long term, free of
+charge.
+
+# Play well with others
+
+Prometheus is a project of convicted and passionate individuals. As we do not
+have a profit motive, nor quarterly projections, or any other requirement to
+meet arbitrary business requirements, we can foxus on getting things right. This
+also means that we are free to suggest other implementations and projects if
+they are a better fit for a particular use-case.
+
+# Be inclusive
+
+We strongly believe that technology should be accessible to all. As such, we
+will always strive to be welcoming to everyone.
+
+As an example of this, many of us are investing their personal time helping
+individuals or communities by educating and helping them to be more productive
+in the tech sector, as well as sponsoring diversity efforts, for example paying
+for travel and accomodation at [PromCon](https://promcon.io).
--- a/content/docs/philosophy/index.md
+++ b/content/docs/philosophy/index.md
@@ -1,0 +1,5 @@
+---
+title: Philosophy
+sort_rank: 99
+nav_icon: flask
+---
--- a/content/docs/philosophy/philosophy_overview.md
+++ b/content/docs/philosophy/philosophy_overview.md
@@ -1,0 +1,16 @@
+---
+title: Philosophy overview
+sort_rank: 1
+---
+
+We have structured our considerations for developing and working with Promethues
+into four groups in decreasing order of importance, and likelihood to change.
+
+1. Philosophy
+2. Goals and Non-Goals
+3. Design decisions
+4. Patterns and Anti-Patterns
+
+While we are not implying that anything in this section is likely to change at
+all, it's still more likely for us to change a particular pattern over time than
+the underlying philosophy.


content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,72 @@
...
+First and foremost, Prometheus must be resilient in operation.
+
+
+## Reliable alerting
+
+As a monitoring system, Prometheus is being relied upon to alert humans that
+they need to take action in order to prevent undesired system state.
+
+Thus, its most important function is to keep the pipeline of ingestion, rule
+evaluation, and alert hand-off working.
```
brian-brazil: notification sending, we don't use the term "alert hand-off"

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,72 @@
...
+The second most important function is to give humans context about these alerts
+by allowing access to the most recent data Prometheus ingested
+
+### Resulting design decisions and patterns
+
+Note that this goal might result in widely different design decisions and thus
+operational patterns for different parts of our ecosystem:
+
+For Prometheus itself, this means running every instance as an island of data
+completely detached from every other instance, except for optional federation.
```
brian-brazil: This exception is weird, and incomplete. I'd go more with how any dependenceis have understandable impact.

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,72 @@
...
+
+For Alertmanager on the other hand, it means the exact opposite: meshing all
+instances closely together, sharing knowledge about alerts and their
```
brian-brazil: This is too vague. The important point is that it's AP rather than CP

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,72 @@
...
+notifications.
+
+## Simple operation
+
+Operation of Prometheus should be as easy and failure-tolerant as possible. We
```
brian-brazil: This is waaaay too open for interpretation. "easy" means I turn it on and it magically works with no configuration - which of course is far for easy in practice.

I'd talk about well defined and limited interfaces that allow integration

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,72 @@
...
+try to put required complexity into earlier phases, going through them less
+often and ideally still while under the control of a smaller subset of people.
+
+One example of this would be the preference of statically linked binaries over
```
brian-brazil: This is just an artifact of how Go works, it's not a project philosophy.

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,72 @@
...
+dynamically-built ones.
+
+## Automation
+
+Computers are good at doing the same thing over and over again, and quickly.
+Humans tend to be better at creative tasks.
+
+Prometheus will always strive to automate away all tasks whenever possible
```
brian-brazil: I don't see how this ties into our philosophy at all. Any automation like this is outside the scope of Prometheus.

What we do have is a belief that alerts going to humans should require intelligent human action.

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,72 @@
...
+through various means; some specific implementations would be service discovery,
+label rewriting, and alert generation.
+
+
+# Non-Goals
+
+# Event handling
+
+Prometheus is dealing with metrics. As such, it will never process and store
```
brian-brazil: This is incorrect, the pushgateway takes in an event when a batch job ends.

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,72 @@
...
+events.
+
+The only exception in our ecosystem is Alertmanager which deals with individual
```
brian-brazil: Alerts aren't events.

I don't see how this is helping define things.

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,72 @@
...
+alerts and alert groups.
+
+For ways to deal with events, see TODO patterns.
+
+# Push-type system
+
+Prometheus is, and always will be, a pull-type system. We strongly believe that
+this makes operational sense in all but the very largest of scales.
```
brian-brazil: This is incorrect, we believe it works at all scales.

Also both ways work at scale. The more salient point is that a lot of things we do only really work with push.

content/docs/philosophy/philosophy.md
```diff
@@ -0,0 +1,69 @@
...
+philosophy](https://en.wikipedia.org/wiki/Unix_philosophy), abridged from [Doug
+McIlroy's initial version from 1978](http://emulator.pdp-11.org.ru/misc/1978.07_-_Bell_System_Technical_Journal.pdf):
+
+1. Make each program do one thing well.
+  While the scope of "one thing" invariably encompasses more and more
+  elements due to increased overall system and computing complexity, we are
+  still doing one thing: ingest metric data, do computations on it, and expose
+  it to other systems.
+2. Expect the output of every program to become the input to another, as yet unknown, program.
+  Today's lingua franca is HTTP endpoints, which are used by Prometheus
```
brian-brazil: HTTP endpoints with JSON

But this isn't what we use everywhere. We've our own text format, and file_sd and yaml for config files.

content/docs/philosophy/philosophy.md
```diff
@@ -0,0 +1,69 @@
...
+  extensively.
+  In the same vein, Prometheus relies heavily on its own libraries and strict
+  layering internally.
+3. Design and build software, even operating systems, to be tried early, ideally within weeks. Don't hesitate to throw away the clumsy parts and rebuild them.
+  Prometheus will always be available for free as in beer and as in speech.
+  We ensure that master always builds, called Continuous Integration these days,
+  and we not afraid to replace whole sections of our codebase, e.g. our storage
+  engine.
+4. Use tools in preference to unskilled help to lighten a programming task, even if you have to detour to build the tools and expect to throw some of them out after you've finished using them.
```
brian-brazil: I don't understand what you're saying here

content/docs/philosophy/philosophy.md
```diff
@@ -0,0 +1,69 @@
...
+  While this is an outdated way of stating the goal, automation where possible
+  is still one of the core characteristics any modern philosophy.
+
+## Embrace cloud-native technologies
```
brian-brazil: This is not our philosophy, in fact I'd say this explicitly isn't our philosophy.

We're a technology that happens to work well for things that fall into the "cloud native" marketing term - but we also work just as well outside of that.

content/docs/philosophy/philosophy.md
```diff
@@ -0,0 +1,69 @@
...
+
+We will always put as much of our code, discussions, presentations, and other
+content into a form and place which is accessible in the long term, free of
+charge.
+
+# Play well with others
+
+Prometheus is a project of convicted and passionate individuals. As we do not
+have a profit motive, nor quarterly projections, or any other requirement to
+meet arbitrary business requirements, we can foxus on getting things right. This
```
brian-brazil: focus

RichiH:
I agree that it still lacks a consistent thread, but I figured I'd rather get partial feedback at this point to align.

content/docs/philosophy/goals.md
RichiH: Addressed.

content/docs/philosophy/goals.md
RichiH: I removed the half sentence completely.

content/docs/philosophy/goals.md
RichiH: Please expand those two.

brian-brazil: https://en.wikipedia.org/wiki/CAP_theorem

content/docs/philosophy/goals.md
RichiH: I would argue that Monarch-style mixed systems allow for even more scale. This is beyond this document's scope, though.

Also, I think you meant pull. It makes sense to expand on those things, though. will put in a TODO into patterns and link back to there.

content/docs/philosophy/goals.md
RichiH: s/easy/simple/ and added a section on dependencies.

content/docs/philosophy/philosophy.md
RichiH: That's why I didn't put JSON.

content/docs/philosophy/philosophy.md
RichiH: It's a literal quote. I can shorten it to stop at the first comma, though. Would that make more sense to you?

The "unskilled" part is an artefact; back then it was common to assign many tasks to what were basically filing clerks. They had the IT knowledge you would expect them to have.

content/docs/philosophy/philosophy.md
RichiH: With that structure, I wanted to say that this works well with good operations, but re-reading, I can see how that got lost over time.

Would you think it better to remove this or to expand on how cloud-native is one of the many facets of proper operations?

content/docs/philosophy/philosophy.md
RichiH: Vim stopped to `set spell` for my markdown files... I need to fix that.

RichiH:
Are there any major sections & content missing from philosophy and goals?

What I am missing, on purpose, are any cross-links, those will come last.

Anything we can reasonably push to a lower tier, we should push there.

content/docs/philosophy/goals.md
brian-brazil: I disagree, both can be scaled indefinitely. Also Monarch is more pull than push.

content/docs/philosophy/philosophy.md
brian-brazil: But we're also doing many things that aren't HTTP.

content/docs/philosophy/philosophy.md
brian-brazil: That's a bit out of the scope of the project, as we don't run an operations team. It certainly has nothing to do with the Unix philosophy.

content/docs/philosophy/philosophy.md
brian-brazil: I'd have a section saying that we work across many paradigms, but aren't going to do things just for the sake of one system that is weird/missing a feature.

content/docs/philosophy/philosophy.md
```diff
@@ -0,0 +1,69 @@
...
+4. Automation is key, freeing up humans to make more useful use of their time
+
+# Be pragmatic
+
+To not lose focus, we need to be honest to our users and ourselves about what we
+can do and not do.
+
+# Be open
+
+We will always put as much of our code, discussions, presentations, and other
```
brian-brazil: I think you're missing an "as possible" here gramatically

content/docs/philosophy/philosophy.md
```diff
@@ -0,0 +1,69 @@
...
+content into a form and place which is accessible in the long term, free of
+charge.
+
+# Play well with others
+
+Prometheus is a project of convicted and passionate individuals. As we do not
```
brian-brazil: I don't think convicted was the word you meant

content/docs/philosophy/philosophy.md
```diff
@@ -0,0 +1,69 @@
...
+have a profit motive, nor quarterly projections, or any other requirement to
+meet arbitrary business requirements, we can focus on getting things right. This
+also means that we are free to suggest other implementations and projects if
+they are a better fit for a particular use-case.
+
+# Be inclusive
```
brian-brazil: It seems weird to mix this in with the other technical stuff

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,78 @@
...
+try to put required complexity into earlier phases, going through them less
+often and ideally still while under the control of a smaller subset of people.
+
+One example of this would be the preference of statically linked binaries over
+dynamically-built ones.
+
+## Keep dependencies clear and limited
+
+Any non-trivial system needs to integrate with other systems. To keep the
+resulting complexity low, we will always try to have the fewest interfaces
```
brian-brazil: Maybe mention to make debugging/understandability easier?

content/docs/philosophy/goals.md
juliusv: I agree. Maybe RichiH was trying to say that SD (because it's mentioned as an example) dynamically discovering targets is more "automatic" than e.g. an old-style static Nagios hosts configuration? But the section also feels a bit unclear / out of place to me.

content/docs/philosophy/goals.md
juliusv: Yeah, there's no limits to scale with either approach, but there's the usual known pros+cons for both approaches, and we're quite married to the pull approach, as Prometheus is so designed around being in control of pulling and processing data when *it* wants.

content/docs/philosophy/goals.md
jamtur01: +1. The automation piece fits somewhere as advice - don't do this without CM, etc - maybe this is more about configuration and dynamism?

content/docs/philosophy/philosophy.md
juliusv: Hm yeah, I think some of these points sound a bit too forced in the context of Prometheus. I do agree about the "do one thing well" (which I would rather phrase as "keep pieces as simple as possible") and "open interfaces" points and would focus on those, while abandoning the others.. I'm not sure we need to explicitly try to tie things so ceremoniously to the Unix philosophy.

content/docs/philosophy/goals.md
jamtur01: I think you need to expand on this - perhaps even define an event from a metric? For example, my immediate thought for a new user was: "Wait, I can decorate a metric with event-like labels? Is it then not an event?"

content/docs/philosophy/goals.md
jamtur01: I'd just say the first sentence and then link to a blog post (Julius' from a while back perhaps or the FAQ entry. To the vast majority of users this is a non-issue.

content/docs/philosophy/philosophy.md
jamtur01: +1.

content/docs/philosophy/goals.md
brian-brazil: Yet many users continue to try and make it do push (and tend to be quite unhappy when told their approach won't work and/or be supported), so I think it's worth a few sentences.

content/docs/philosophy/philosophy.md
jamtur01: :lock:

juliusv:
> Are there any major sections & content missing from philosophy and goals?

Have you intentionally omitted most of the points from https://github.com/prometheus/docs/issues/149 so far?

content/docs/philosophy/philosophy.md
jamtur01: Is there another place where this is mentioned? I feel like this needs to be somewhere?

juliusv: Only indirectly in our governance: https://prometheus.io/governance/#values

Maybe if/once we have a general section for developers and getting into Prometheus contributions, it would be a better fit there?

brian-brazil: This is very vision statementy. It has little to no bearing on the design decisions we make.

content/docs/philosophy/goals.md
brian-brazil: Seems more like a place to like to My Philosophy on Alerting.

I'd also caution on putting automation as a key goal for users, elimination is better.

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,78 @@
...
+
+Computers are good at doing the same thing over and over again, and quickly.
+Humans tend to be better at creative tasks.
+
+Prometheus will always strive to automate away all tasks whenever possible
+through various means; some specific implementations would be service discovery,
+label rewriting, and alert generation.
+
+
+# Non-Goals
```
brian-brazil: I think this is where we put in stuff about assuming you already have CM/DNS/service database/machine database etc., tying back to laying/doing one thing well.

content/docs/philosophy/philosophy.md
RichiH: ...are you sure...?

content/docs/philosophy/philosophy.md
RichiH: Should we keep it in there while we don't have a better place, yet?

content/docs/philosophy/goals.md
RichiH: You mean to put in that have CMDB etc explicitly outside our scope? Isn't that obvious?

content/docs/philosophy/philosophy.md
RichiH: done.

content/docs/philosophy/goals.md
brian-brazil: We get requests in this area fairly regularly.

content/docs/philosophy/philosophy.md
RichiH: Is there something you would propose instead?

content/docs/philosophy/goals.md
RichiH: Changed.

content/docs/philosophy/philosophy.md
RichiH: Changed.

content/docs/philosophy/philosophy.md
brian-brazil: I'd remove it.

content/docs/philosophy/philosophy.md
RichiH: Rewritten completely & moved down.

content/docs/philosophy/philosophy.md
RichiH: Put in a TODO for now.

content/docs/philosophy/goals.md
RichiH: done.

content/docs/philosophy/philosophy.md
```diff
@@ -0,0 +1,69 @@
...
+To not lose focus, we need to be honest to our users and ourselves about what we
+can do and not do.
+
+# Be open
+
+We will always put as much of our code, discussions, presentations, and other
+content as possible into a form and place which is accessible in the long term,
+free of charge.
+
+# Be opionated
```
brian-brazil: I'm not sure which section it goes in here, but there should be something about not adding attractive nuisances: features that might be useful to experts for niche use cases. but would be misunderstood by and cause problems for the average user. Put another way, we don't leave chainsaws lying about for users to hurt themselves with.

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,81 @@
...
+instances closely together, sharing knowledge about alerts and their
+notifications.
+
+## Simple operation
+
+Operation of Prometheus should be as simple and failure-tolerant as possible. We
+try to put required complexity into earlier phases, going through them less
+often and ideally still while under the control of a smaller subset of people.
+
+One example of this would be the preference of statically linked binaries over
```
brian-brazil: This is still here, this is not a goal of the project.

content/docs/philosophy/goals.md
brian-brazil: I'm not seeing this change.

content/docs/philosophy/goals.md
```diff
@@ -0,0 +1,81 @@
...
+## Reliable alerting
+
+As a monitoring system, Prometheus is being relied upon to alert humans that
+they need to take action in order to prevent undesired system state.
+
+Thus, its most important function is to keep the pipeline of ingestion, rule
+evaluation, and alert notifications working.
+
+The second most important function is to give humans context about these alerts
+by allowing access to the most recent data Prometheus ingested
```
brian-brazil: fullstop

I'd add something around data that's good enough to make useful engineering decisions, rather than 100% perfect data.

content/docs/philosophy/philosophy.md
```diff
@@ -0,0 +1,69 @@
...
+# Be pragmatic
+
+To not lose focus, we need to be honest to our users and ourselves about what we
+can do and not do.
+
+# Be open
+
+We will always put as much of our code, discussions, presentations, and other
+content as possible into a form and place which is accessible in the long term,
+free of charge.
```
roidelapluie: should we add in the open, too?

content/docs/philosophy/philosophy.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Be opionated
+
+One size does never fit all, so we need to deliberately choose what we want to
+achieve and what not. We can not, and do not try to, fill every single niche.
+
+# Play well with others
+
+Prometheus is a project of convinced and passionate individuals. As we do not
```
roidelapluie: maybe we should review this about the recent discussions on tsdb code reuse etc

RichiH:
@roidelapluie I do, it's on my bucket lest, thanks!

beorn7:
@RichiH , this PR seems to get some love on the mailing list. I guess to revive it, it needs first an update from your side before it makes sense for reviewers to comment on details.

grobie:
@RichiH Any updates here? I think it would be great to get this in in some form.

jan--f:
I'll close this for now under the assumption that after 5 years this will need rework. Happy to re-open if this gets more attention.

PR #1060, opened 2018-06-13
Update Concourse URL

dtimm:
Concourse changed their official URL due to a DNS problem with .ci domains. 

Signed-off-by: David Timm <dtimm@pivotal.io>


dtimm:
@brian-brazil

brian-brazil:
Thanks!

PR #1073, opened 2018-06-20
Add OpenMetrics to roadmap

fabxc:
@RichiH @brian-brazil @jkohen


RichiH:
We should probably put this on -team for at least a hum; once we do that, I would go further and state that OM will become the preferred format.

Other than that I obviously agree :)

RichiH:
PS: You are missing the DCO sign-off.

content/docs/introduction/roadmap.md
```diff
@@ -19,10 +19,11 @@ utilized in the Prometheus server. We plan on making use of this
...
 in Prometheus and provide it via an experimental API endpoint.
 
-### Prometheus metrics format as a standard
+### Adopt OpenMetrics
 
-We intend to submit a cleaned up version of our format for standardization
-to a group such as the IETF.
+The OpenMetrics working group is developing a new standard for metric exposition.
+We plan to adopt this format in our client libraries and Prometheus itself, so it
+can consume it from clients that do not support the current Prometheus format.
```
brian-brazil: This last clause doesn't seem quite right. Any client that exports it already likely supports the Prometheus format, and is going to pick it up because it's using our client libraries.

fabxc:
This item would not go as far as "preferred format" yet – I think that's subject to how well OM gets adopted. Support should be the first goal and after my query about that on -team yesterday, it appeared relatively clear.

We can do an additional formal vote as well though.

content/docs/introduction/roadmap.md
fabxc: Isn't the whole point that OM is an open format not tied to Prometheus. Going off of the assumption that the other end definitely uses our client libraries and supports our format makes the whole idea obsolete, no?

brian-brazil: It just reads oddly. I'd suggest adopt -> support and removing everything after the comma.

RichiH:
See CfC on the team list; once we have consensus we can just write "we plan to switch over as default".

RichiH:
Re DCO: You can do it by hand on GH as well, but it's cumbersome. I would wish GH would allow you to put signatures or something like that in PRs.

brian-brazil:
:+1:

PR #1117, opened 2018-07-16
Fixed a misprint

anatolykazantsev:


RichiH:
Thanks for spotting this. Can you please add a sign-off?

````
git commit --amend --signoff
git push -f
````

Should be enough, but be careful with the latter.

anatolykazantsev:
Fixed, but github originally put "hidden" email for commit author. Should I use that "hidden" email?

RichiH:
Thanks! Ignore that error, it's not a very useful one ;)

PR #1120, opened 2018-07-16
Integrations: Add link to sshified project

hoffie:
This PR adds a reference linking to [sshified](https://github.com/hoffie/sshified) on the integrations page.

sshified is a daemon which proxies prometheus' http requests to remote hosts via ssh, therefore providing authentication, encryption and simplified network management (e.g. when crossing firewalls).
Disclaimer: Author speaking here.

Cc @brian-brazil

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -68,3 +68,4 @@ you to integrate it with your existing systems or build on top of it.
 ## Other
 
   * [PushProx](https://github.com/RobustPerception/PushProx): Proxy to transverse NAT and similar network setups
+  * [sshified](https://github.com/hoffie/sshified): Proxies scrapes over SSH for authentication, encryption and simplification of firewall rules


brian-brazil:
Thanks for sharing. This tool doesn't seem Prometheus specific, and has few users. I'm wary of listing every possible proxy/reverse-proxy in this section, so what I'd suggest doing is sharing this on the mailing lists.

hoffie:
Obviously not what I was hoping to hear, but I understand your concern. Will post something to the prometheus-users list in the next days.
I thought it was fitting well as it provides a scalable solution for the often discussed auth/encryption enquiries.

corny:
Well, I wrote my own SSH proxy for Prometheus because I could not find anything existing. Our tools are not Prometheus specific, but in specific scenarios Prometheus users could benefit a lot by using our tools.

I just started a new discussion: https://groups.google.com/d/topic/prometheus-users/W7Q8VGIBMi8/discussion
My proxy: https://github.com/digineo/http-over-ssh

I would really like to see a listing of our tools in the Prometheus documentation.

PR #1124, opened 2018-07-20
Update list with 3rd party Kibana exporter

pjhampton:
Hey, 

I would love to add the Kibana exporter to this list. In production at Citi and some other cool companies.


brian-brazil:
Why do you say this is the 3rd exporter?

Some tips to improve your exporter: Seconds are the preferred unit for durations in Prometheus, and you should mention them in your metric names. There's no need to export load average, that's the responsibility of the node exporter. If you're unsure if a metric is a gauge or a counter, use untyped. Currently some of your metrics are counter (e.g.kibana_requests_total) but exported as gauges.

pjhampton:
Hey. Sorry it was a long day. I meant to say 3rd party. 

Ok, I make those alterations. That is super helpful. Is making this change a prerequisite to getting this or merged? 

Can you help with the commit acceptance?

brian-brazil:
That;s what I thought, just checking as we try to avoid duplicates.

Improving those issues is not a requirement, however you do need to add the DCO to your commits before I can merge.

PR #1126, opened 2018-07-22
Mark ScyllaDB as having a built-in exporter

avikivity:
ScyllaDB supports a built-in exporter on port 9180 since release 1.4.

Signed-off-by: Avi Kivity <avi@scylladb.com>


brian-brazil:
Thanks!

PR #1130, opened 2018-07-28
Upgrade to Nanoc 4.0

SuperQ:
* Update Ruby gems.
* Remove nanoc-cachebuster due to incompatibility.
* Add README item about GitHub auth to improve build reliability.
* Fix various required config changes for 4.0

TODO: Find a replacement for the cachebuster.


SuperQ:
@ddfreyne and I looked into the cachebuster. We're pretty sure this is not needed anymore with the http/2 etag support between netlify and cloudflare.

SuperQ:
So, two additional comments on caching controls.
* Cloudflare is currently set to a fixed 4 hour cache header.
* Netlify suports a static asset optimizer, which basically does the same URL hashing that the old cachebuster does. It also may improve the site speed.

I would recommend we change the cache control in cloudflare to this to respect the Netlify origin cache control.

PR #1131, opened 2018-07-29
add MQTT push gateway

Svedrin:
Hi,

I'm building a push gateway for MQTT, and since I couldn't find anything that exists already, I thought I'd share.

I allocated port 9466 on the [default port allocations](https://github.com/prometheus/prometheus/wiki/Default-port-allocations) page.

I'm not sure if the section is correct, but since I didn't find an obviously better one, I thought I'd add it right next to the existing MQTT exporter.

Signed-off-by: Michael Ziegler <diese-addy@funzt-halt.net>


Svedrin:
how is MQTT user-specific :confused:

brian-brazil:
What you are doing here with MQTT is specific to you as far as I can see. An MQTT exporter would be monitoring MQTT itself, not using it as a data pipeline.

brian-brazil:
Pushing metrics is not a recommended way to use Prometheus, nor are non-HTTP data transfer mechanisms. I don't see anything to indicate that what you're working off here is some form of standard as opposed to something that you have created yourself. We want to list integrations with established systems and protocols, not facilitate the creation of brand new ones. I'm afraid we cannot list this here.

PR #1137, opened 2018-08-03
simplify inhibit_rule description

fmoor:
Most of the Prometheus docs are excellent! However, I found the `inhibit_rule` description confusing. This is my attempt to simplify the language. 50% of this pull request is me verifying that I understand how the `inhibit_rule`s work.


content/docs/alerting/configuration.md
```diff
@@ -191,9 +191,9 @@ route:
 
 ## `<inhibit_rule>`
 
-An inhibition rule is a rule that mutes an alert matching a set of matchers
-under the condition that an alert exists that matches another set of matchers.
-Both alerts must have a set of equal labels.
+An inhibition rule mutes an alert (target) matching a set of matchers
+when an alert (source) exists that matches another set of matchers.
+Both target and source must have the same labels.
```
simonpasquier: `a set of equal labels` is important here IMO otherwise I read it as the alerts must have the exact same labels (which is never going to happen).

fmoor: is it that the target and source match labels in the `inhibit_rules` need to have an intersecting set of labels or the target and source alerts at runtime need to have an intersecting set of labels?

fmoor: How does this look now? @simonpasquier @juliusv

simonpasquier:
Great! thanks a lot.

PR #1140, opened 2018-08-08
Added Traefik to list of Software exposing Prom metrics directly

vegasbrianc:
Added Traefik to list of Software exposing Prometheus metrics directly

Signed-off-by: vegasbrianc <brian.christner@gmail.com>


brian-brazil:
Thanks for your PR. I'm not sure exactly what your code is doing, but it does not appear to be directly instrumented.

vegasbrianc:
@brian-brazil Thanks for the response. OK, should I remove the *direct*? Not my project just made an assumption here as Prometheus metrics are made available

brian-brazil:
Yes, you only need to remove the direct. Direct is if you're using the Prometheus client libraries in the usual fashion.

vegasbrianc:
@brian-brazil done, thanks again for the feedback.

brian-brazil:
Thanks!

PR #1149, opened 2018-08-15
Updated link to InfluxDB Docs to use the latest

russorat:
This link was pointing to a specific version of the InfluxDB docs. This change would always point the link to the latest version of the docs.

fyi @brian-brazil


russorat:
let me try this again...

PR #1150, opened 2018-08-15
Updated link to InfluxDB Docs to use the latest

russorat:
This link was pointing to a specific version of the InfluxDB docs. This change would always point the link to the latest version of the docs.

fyi @brian-brazil


brian-brazil:
Can you add the DCO please?

You've got a stray double quote in the warning at the top of the linked doc.

russorat:
I'll fix the DCO, but what do you mean re: "...stray double quote in the warning at the top of the linked doc."?

RichiH:
https://docs.influxdata.com/influxdb/v1.6/supported_protocols/prometheus reads

Note: The Prometheus API Stability Guarantees states that remote read and remote write endpoints are features listed as experimental or subject to change, and thus considered unstable for 2.x”. Any breaking changes will be included in the InfluxDB release notes.

russorat:
ah thanks!

brian-brazil:
Thanks!

PR #1151, opened 2018-08-16
Replace erroneous “TLS” on basic auth guide

ddfreyne:
The basic auth guide mentions “TLS” instead of “basic auth“ — possibly copied from the TLS guide?


brian-brazil:
Thanks!

PR #1161, opened 2018-08-26
Use escaped #<file_sd_config> in integration links.

deniszh:
Escape the links in the same way known-working FAQ.md does.
See the link to configuration#<scrape_config>:
https://prometheus.io/docs/introduction/faq/#why-don-t-the-prometheus-server-components-support-tls-or-authentication-can-i-add-those

Thanks!


brian-brazil:
Thanks!

PR #1165, opened 2018-09-05
Add Cachet exporter to the list

diogonicoleti:
Hi!

We, at ContaAzul, created an exporter to collect some metrics from Cachet status page (number of incidents by status and number of components by status). Its use the Cachet API to scrape the data.

The metrics exposed are:
```
cachet_up
cachet_scrape_duration_seconds
cachet_incidents_total
cachet_components_total
```

And it can be used to take insights about incidents and, if necessary, make some alerts on 
forgotten open incidents.

@brian-brazil can you have a look please?


brian-brazil:
Thanks!

A few tips to improve things:

Gauges should not end in `_total`, that's for counters.

Config and metricConfig in collector.go are unused.

You should only set `up` to 1 after all requests (i.e. GetAllComponentGroups) have succeeded, otherwise you won't be able to alert on errors. Accordingly there's also no need for the ping.

The numbers for incident/component statuses are a bit oapque, using their names might be better. You're also missing incident statuses 0 and 4.

diogonicoleti:
@brian-brazil thanks for the review in the code! We're making the improvements suggested by you 😄

PR #1166, opened 2018-09-06
Add presto exporter to list

sunjiangjing:


brian-brazil:
You can add the port yourself to https://github.com/prometheus/prometheus/wiki/Default-port-allocations, there's no special process there.

Here's some things that might help you improve your exporter:

On an error, don't return no metrics. Instead send a NewInvalidMetric so the scrape fails.

What do total_input_rows, total_input_bytes, and total_cpu_time_secs represent? They seem to be for running queries, which I wouldn't presume from the name: https://github.com/prestodb/presto/blob/0d13d8e020c4a76aad31793f15da0a04539f0a74/presto-main/src/main/java/com/facebook/presto/server/ClusterStatsResource.java#L73

reserved_memory is lacking a unit, such a bytes.

I'd use queries_running rather than running_queries, so all the query metrics will be together. Similarly with other metrics.

sunjiangjing:
I'm sorry, I do not understand completely what you answered.
Would you explain more detail?

brian-brazil:
I'm suggesting that you rename running_queries to queries_running, and so on.

sunjiangjing:
As you suggested, I try to change the name like the following, is it ok.
https://github.com/yahoojapan/presto_exporter/commit/b8519f4b9b00818b2eb474ce6f8ce60a03edc550
is it necessary to change the total_input_rows, total_input_bytes, and total_cpu_time_secs ?
And if have the other corrections, please tell me 
Thanks

sunjiangjing:
> Are they really totals? Looking at the code, it seems to be only the sum over what's currently known to presto.
yes,  it is the sum. 
https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/server/ClusterStatsResource.java#L87-L89

I change the names, as the following, is it ok?  
```
running_queries --> Queries_Running

total_input_rows --> Querys_TotalInputRows
total_input_bytes --> Querys_TotalInputBytes
total_cpu_time_secs --> Querys_TotalCpuTimeSecs
```
if you have the better ideas, please give me a advice
thanks

brian-brazil:
I'm going to merge this now, as it is a working exporter. I would suggest doing running_queries --> queries_running.

PR #1167, opened 2018-09-07
Fix YAML display in node exporter's guide

splattael:
The [YAML example](https://prometheus.io/docs/guides/node-exporter/#config) for the node exporter in the guide is broken:

screenshot from 2018-09-07 10-47-40

I suppose that adding a new line will fix this display glitch.


SuperQ:
Retry on the job [looks correct](https://deploy-preview-1167--prometheus-docs.netlify.com/docs/guides/node-exporter/#config).

PR #1173, opened 2018-09-15
Fix link to instrumentation practices docs

paultcochrane:
It seems that the documentation has been rearranged at some time in the past and not all links to other documentation pages point to their intended locations.  This change fixes the link to the instrumentation practices documentation.  This pull request is submitted in the hope that it is helpful.  If you want anything changed I will gladly update it and resubmit.


paultcochrane:
Oops.  It seems like this isn't really a fix...  It fixes the link on GitHub, but wouldn't work on the prometheus.io site.  Sorry for the noise!

PR #1176, opened 2018-09-22
Fix 'make bundle' crash for nokogiri

codesome:
In my system `make bundle` crashed saying `pkg-config` not found while installing `nokogiri` gem. Though `nokogiri` doesn't mention it to be a dependency, installing `pkg-config` gem first prevents installation of `nokogiri` from crashing.

I have no idea about Ruby, so please let me know if I have missed something here.

Signed-off-by: Ganesh Vernekar <cs15btech11018@iith.ac.in>

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -68,10 +68,12 @@ GEM
     nenv (0.3.0)
     nokogiri (1.8.4)
       mini_portile2 (~> 2.3.0)
+      pkg-config (= 1.3.1)
     notiffany (0.1.1)
       nenv (~> 0.1)
       shellany (~> 0.0)
     parallel (1.12.1)
+    pkg-config (1.3.1)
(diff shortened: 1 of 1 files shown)


grobie:
The `Gemfile.lock` is an auto-generated file and should be edited by hand. It'll be overwritten next time someone installs/updates a dependency.

Looking at the installation documentation of nokogiri, it explains the explanation of pkg-config as first step using the system's dependency management tool (e.g. apt, pkg, brew, ...).

We should probably also update the Readme and document the available Docker container, which avoids having to install Ruby and its dependencies.

grobie:
Yes, it should not. Did you install `pkg-config` with apt?

codesome:
Yes, with `apt`.

codesome:
And though it is auto generated, it seems to fix the error.

PR #1181, opened 2018-10-02
Add cassandra-exporter to exporters.md

zegelin:
@brian-brazil For your review.

I removed the Cassandra reference for the _jmx\_exporter_ as I feel that _cassandra-exporter_ does a far better job -- both in exposition performance and also quality of the metrics that are exported.


zegelin:
Yes. On a fresh, empty Cassandra node (only system tables) jmx_exporter generates 2909 metrics (excluding duplicates for columnfamiles vs tables, see below) whereas cassandra-exporter generates 7105 metrics.

jmx_exporter is unable to handle a number metrics that Cassandra exports.
Cassandra's `EstimatedHistogram`s are made available over JMX as a long[], which are ignored by jmx_exporter. These are handled and exported as summaries by cassandra-exporter.

Dropwizzard timers and histograms are exported as summaries.

Additionally, Cassandra exposes a number of its metrics twice under different names (column family vs. table), and also exposes pre-aggregated values (per-keyspace/per-node). The default example jmx_exporter config for Cassandra doesn't exclude these, resulting in a number of duplicate and unnecessary metrics.

I think that cassandra-exporter should be included based on its performance and quality of output.

Cassandra-exporter bypasses JMX when run as an agent. On a Cassandra node that has 100's of tables, jmx_exporter takes _seconds_ to export. Whereas cassandra-exporter does it in _sub 100 milliseconds_.



The output has been hand-tuned to try and closely follow the Prometheus best practices -- to have nice metric names and good labels, correct metric types, and where necessary the values are scaled to Prometheus base units (e.g., Cassandra exposes timings in microseconds and cassandra-exporter converts these values to seconds).

brian-brazil:
> These are handled and exported as summaries by cassandra-exporter.

The percentiles are available via JMX, so the JMX exporter can support these. From a quick peek at your code, everything is coming from JMX - so there's nothing there that the JMX exporter can't do.

In addition it supports target labels as labels, and is not using the standard java client.

I'd suggest working on improving the example configuration for the jmx exporter.

zegelin:
> The percentiles are available via JMX, so the JMX exporter can support these. From a quick peek at your code, everything is coming from JMX - so there's nothing there that the JMX exporter can't do.

cassandra-exporter uses some JMX metadata, since the exported MBean `ObjectName`s have more info than the Dropwizzard metrics that they wrap. But I assure you _it does not use JMX MBean proxies or `getAttribute()`_ and instead directly calls the underlying getter functions, and in some cases accesses fields directly.

Do note that cassandra-exporter can run either as an agent or standalone. As standalone it _is_ using JMX, but when running as an agent it bypasses JMX when querying the values..

Please again look at the performance difference between jmx_exporter and cassandra-exporter. Performance is a real concern. Instaclustr has dealt with Cassandra clusters that have 100's of tables and scraping JMX is too expensive. Burning seconds of CPU time is never a good thing, especially on a loaded server. We have seen metrics collection via JMX actually cause increased query latency.

Just because Cassandra offers the option to collect its metrics via JMX doesn't mean you should use JMX when there are better solutions available.

> In addition it supports target labels as labels, and is not using the standard java client.

The cluster metadata/target labels can be disabled (see `--no-global-labels`).

Are you implying that cassandra-exporter is not eligible for inclusion into this list because it's not using the Prometheus Java client library?
The java client was determined to be too slow for the number of metrics a decently sized Cassandra node may have (again, see performance).

brian-brazil:
The standard set for replacing a use of the JMX exporter is that it has to offer some metrics which it would be impossible for to the JMX exporter to provide, even with additional development. It's also expected that you would engage with the maintainers of the existing exporter before trying to replace it, so that the community can work together.

What it sounds like you're doing here is putting something like the simpleclient_dropwizard in a Java agent, which has use beyond Cassandra. If there was such a thing using the normal client library I'd be happy to list it as Dropwizard is quite common and doing that is better than using JMX - however a completely independent exporter that's off the beaten path and only supports one application doesn't meet the bar for replacing the existing recommendation of the jmx exporter. 

> The cluster metadata/target labels can be disabled (see --no-global-labels).

An exporter shouldn't even offer this as a possibility, this is the responsibility of Prometheus. See https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels,-not-static-scraped-labels

> Are you implying that cassandra-exporter is not eligible for inclusion into this list because it's not using the Prometheus Java client library?

No, but it is a bit unusual. Particularly as you're reinventing things like the standard JVM exports, and won't gain OpenMetrics support when that comes along (the latter may become a blocker in future).

> The java client was determined to be too slow for the number of metrics a decently sized Cassandra node may have (again, see performance).

If you have suggestions to improve the java client let's discuss them in that repo, so they can benefit everyone.

zegelin:
What if I modify this PR so that it only adds Cassandra to the list of available database exporters, perhaps with a _see also jmx_exporter_ note, but keeps the jmx_exporter + Cassandra recommendation at the top of the page?

The exporters page lists tightly-integrated exporters for numerous applications. In this case, cassandra-exporter is tightly integrated with Cassandra and as a result produces higher quality metrics with significant performance gains.

brian-brazil:
I aim to avoid duplicates in this list, both to encourage the community to work on one option and to make it easier for end-users to choose what to use. If you want to improve the example Cassandra configuration, a PR would be appreciated.

zegelin:
One size does not fit all.

_jmx\_exporter is not suitable for monitoring any reasonably-sized Cassandra node_.

Burning _seconds_ of CPU time to collect metrics is undesirable.

I note that a number of issues have been raised against the jmx_exporter project that talk about poor performance. 

Perhaps you should allow cassandra-exporter on the list so that end-users _have options_ and can decide which exporter to use based on their use-case.

brian-brazil:
Users are free to use exporters not on this list, so they already have options. Encouraging multiple divergent projects for each applications is not ultimately something that helps users though.

kgreav:
+1 to including this in the list. I don't see the big deal with having multiple possible exporters in the list. This is open source - it's not like the exporters are in direct competition, and any exporters creation and use is good for the project. Besides, precedent has already been set for having multiple exporters for a single technology. Kafka can use the jmx_exporter, but the list also mentions a third party exporter which was added in #876 because it spouted the same benefits as this cassandra-exporter. It also removed two existing third party Kafka exporters...

RichiH:
@zegelin would you be willing to create a Dropwizard exporter, ideally using our library?

That way, we can point at Dropwizard as the default method to get at that data

As to why using our library makes sense: We can better ensure a consistent ecosystem. We will be migrating our whole ecosystem to OpenMetrics at some point; by using our library you get that for free and your users will be able to switch back and forth, have autoneg (where possible), etc.

brian-brazil:
> the list also mentions a third party exporter which was added in #876 because it spouted the same benefits as this cassandra-exporter.

The kafka exporter does something different to the jmx exporter, they're complementary.

zegelin:
@RichiH 
> @zegelin would you be willing to create a Dropwizard exporter, ideally using our library?

Dropwizzard does not have the concept of metric labels, only names, which results in an "impedance mismatch" with the Prometheus data model.

JMX ObjectNames offer richer metadata as they have attributes which can be translated into labels.

Cassandra uses a combination: It uses Dropwizzard as the underlying metrics library, i.e., all the counters, gauges, etc. are Dropwizzard implementations. Yet for JMX export, Cassandra has a custom `MetricsRegistry` sub-class implementation that allows alternate JMX ObjectNames to be specified for each registered Dropwizzard metric, which it uses to add metadata such as the keyspace and table name. 

cassandra-exporter exploits this. The agent intercepts MBean registrations, which lets it save the ObjectName, and hence the Cassandra-specific metadata, associated with every Dropwizzard metric, and the raw Dropwizzard metric object itself. To collect the actual metric data, cassandra-exporer doesn't query MBean attributes. It directly calls the getter functions of the Dropwizzard metric objects.


> As to why using our library makes sense: We can better ensure a consistent ecosystem. We will be migrating our whole ecosystem to OpenMetrics at some point; by using our library you get that for free and your users will be able to switch back and forth, have autoneg (where possible), etc.

And I intend to keep cassandra-exporter up-to-date with whatever changes to the ecosystem occur.

Re: OpenMetrics. Amusingly, I originally was targeting Protocol Buffers, but this was removed in Prometheus 2.0.

 The trade-off of using custom exposition code resulted in significant performance improvements.

---

@brian-brazil 
> The kafka exporter does something different to the jmx exporter, they're complementary.

And cassandra-exporter does _more_ than jmx_exporter.

It correctly handles the histograms exported by Cassandra as Dropwizzard `Gauge`s with `long[]` values -- these are _not_ handled  by jmx_exporter.

From the output of jmxterm, you can see that the only attribute (`Value`) is of type `Object`. It's actually a `long[]`, which contains the raw sample data for Cassandra's custom `EstimatedHistogram`.

```
#mbean = org.apache.cassandra.metrics:type=Table,keyspace=keyspace1,scope=standard1,name=EstimatedPartitionSizeHistogram
#class name = org.apache.cassandra.metrics.CassandraMetricsRegistry$JmxGauge
# attributes
  %0   - Value (java.lang.Object, r)
# operations
```

For Dropwizzard Timers and Histograms, cassandra-exporter doesn't call `getXXthPercentile` for each percentile value, as those functions are _slow_ -- they each take a lock, potentially rescale the bucket values, take a snapshot (which computes weights, etc) and then computes the percentile. Instead, there is a `getSnapshot()` function, _not available via JMX_ which does all that work once, and actually allows for arbitrary percentiles to be queried. When you have hundreds of tables, each with a number of Timers and Histograms, the performance gains are substantial.

Certain metrics are not available via JMX attributes. Either they are exposed via JMX operations, or not at all. Examples of this:
`cassandra_endpoint_downtime_total_seconds`, which over JMX is an operation (`GossiperMBean.getEndpointDowntime(String)`)
`cassandra_effective_ownership_ratio`, which over JMX is an operation (`StorageServiceMBean.effectiveOwnership(String)`) 

Others are attributes with a type that is incompatible with jmx_exporter:
`cassandra_endpoint_active`, which over JMX is available as a `Map<InetAddress, String>` 
`cassandra_ownership_ratio`, which over JMX is available as a `Map<InetAddress, Float>` 


cassandra-exporter attaches additional labels containing metadata not available via JMX.
For example, on table metrics, which represent CQL tables, views & indexes (internally, they're all tables), the labels are:

Label | ..
--- | ---
`keyspace` | The parent keyspace (JMX `keyspace`)
`table` | The table name or the parent table, in the case of an index (JMX `scope`)
`table_type` | `table`, `index` or `view`
`index` | The name of the index, if the metric represents an index
`index_type` | `keys`, `composite` or `custom`
`index_class` | The implementing class of the index, for custom indexes

Cassandra also exports certain latency metrics as two individual MBeans -- a `Latency` bean and a `TotalLatency` bean. The exporter combines these together into a single Prometheus summary.

---

And lastly, and I obviously haven't stated this enough, _it's significantly faster_. How is this not a win for everyone who runs Cassandra + Prometheus?

zegelin:
I would appreciate an update to this ticket.

ChristopheSchmitz:
> The kafka exporter does something different to the jmx exporter, they're complementary.

It appears that this Cassanrda exporter provides a 200x speed improvement. It surely must be doing something different than the standard jmx exporter :sunglasses: 

I think it will benefit the Cassandra community if this project is appropriately advertised on the Prometheus page. It doesn't necessarly need to replace the existing one on the doc, we can put this one next to the standard one. I think we should give the Cassandra community the option to use a slow and resource intensive exporter, or this 200x times faster dedicated Cassandra exporter. That's the spirit of Open Source. :+1:

jan--f:
Closing this. The exporter is archived and this hasn't seen attention in too long.

PR #1183, opened 2018-10-04
Add prosafe_exporter

dalance:
Adding an exporter for NETGEAR switches supported by ProSAFE Plus utility.


brian-brazil:
Is this data obtainable via SNMP?

This exporter seems to scrape on a fixed schedule a configured list of targets, the way to usually handle this is like the snmp exporter - take the target as a URL paramater.

SuperQ:
There are a bunch of netgear ProSAFE Plus devices that don't have SNMP, but have some stats over http.

I have one of these, but never got around to writing my own exporter. :grin:

dalance:
As @SuperQ says, NETGEAR switches categorized to "unmanaged plus" don't support SNMP.

I'll implement to take a target host through URI query string.
Thanks.

PR #1185, opened 2018-10-07
Add ccache exporter

virtualtam:
Hi @brian-brazil,

This PR adds a link to the [ccache_exporter](https://github.com/virtualtam/ccache_exporter), which provides a wrapper to the [ccache](https://ccache.samba.org/) compiler cache and exposes metrics on port `9508` (reserved on the [default port allocations](https://github.com/prometheus/prometheus/wiki/Default-port-allocations) wiki).


brian-brazil:
Thanks!

Here's some suggestions to make your exporter even better:

Hits+misses are a little annoying to work with, hits+total is easier.

max_cache_size_bytes should be cache_size_max_bytes, as naming goes from least to most specific. Similarly cache_files.

I'd suggest prefixing everything with ccache, so they're distinct from metrics from other types of cache.

virtualtam:
Hi, and thanks for the suggestions!

`ccache` is already used as a namespace for gauges and counters, I'll update the docs accordingly and rename the other metrics :)

PR #1187, opened 2018-10-09
Update user logo for PingCAP

queenypingcap:
PingCAP has updated its logo to a new version, so I filed this PR to update it.


brian-brazil:
Thanks!

PR #1188, opened 2018-10-11
Link to Net::Prometheus client library for Perl

leonerd:


leonerd:
> HELP needs to be escaped, the rules are slightly different to label values
> In the docs for histogram, buckets are upper bound - not lower bounds

Fixed in source.

> The content-type is missing the version and charset

I've added version (`version=0.0.4`) but I don't see docs/examples on charset.
https://prometheus.io/docs/instrumenting/exposition_formats/

> Is it possible to initialise a Summary/Histogram child?

Is there an example from another client library on what this would look like? I'm not quite sure I follow

brian-brazil:
> charset.

https://github.com/prometheus/client_python/blob/master/prometheus_client/exposition.py#L30

> Is there an example from another client library on what this would look like? I'm not quite sure I follow

Something like mysummary.labels("value") is the usual way, as labels both creates and returns the child.

leonerd:
> charset

Now fixed in source too.

> Something like mysummary.labels("value") is the usual way, as labels both creates and returns the child.

Ah, if I understand correctly, then it already does that:
https://metacpan.org/pod/Net::Prometheus::Metric#labels1

brian-brazil:
Ah, I'd missed that as it wasn't used in the label examples.

RichiH:
Can you cut a 0.07 for your charset change, please? Thanks for all your work :)

PR #1190, opened 2018-10-14
Adding Reference to Promregator

eaglerainbow:
Promregator provides discovery and facilitates scraping of Cloud-Foundry-based applications. Amongst another mechanism, it also uses the file-based service discovery approach to communicate the result of discovery to Prometheus. 
After some time of incubation of the project, I (as the maintainer of Promregator) would like to ask for referencing the project in Prometheus' docs.

@brian-brazil ping to you, as requested in CONTRIBUTING.md.
Signed-off-by: Nico Schmoigl <github@schmoigl-online.de>

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -76,3 +76,4 @@ you to integrate it with your existing systems or build on top of it.
 
   * [karma](https://github.com/prymitive/karma): alert dashboard
   * [PushProx](https://github.com/RobustPerception/PushProx): Proxy to transverse NAT and similar network setups
+  * [Promregator](https://github.com/promregator/promregator): discovery and scraping for Cloud Foundry applications


brian-brazil:
Is this actually a file_sd? Your docs indicated that you rejected the (recommended) approach of having on target per target.

eaglerainbow:
> Is this actually a file_sd? 

Valid question, but one which requires a little longer answer:

In fact, in the beginning, I thought that the target audience of Promregator would mainly be operators of individual apps run on Cloud Foundry. Support from operators of the Cloud Foundry Platform was explicitly out of scope for Promregator ("low touch" applications). As consequence, the number of apps/targets to scrape then would be low (per Promregator instance). Initial configuration effort then is one major concern in the entire process. That is why I initially started with an approach of having only one single endpoint in Promregator, which Prometheus would scrape (and includes the super-set of all metrics demanded).

Over the past months, however, it turned out that people want to use Promregator in environments with many more applications/targets involved. Promregator is run as central instance and may even feal like a service on the platform available to many applications (there is even [a feature request](https://github.com/promregator/promregator/issues/61) asking to separate the selection of applications from Promregator's configuration file). Thus, it's not just about the operators of the applications anymore, Moreover, operators of the underlying Cloud Foundry platform are starting to indicate interest in supporting the configuration and scraping processes. With that the "low-configuration-cost approach" does not work out anymore due to scalability issues (most likely due to the same reasons Prometheus is recommending it like it does). That is why Promregator (since 0.2.0) supports two modes in parallel (consumers are free to choose and change back and forth):

* Single Endpoint Scraping: all-in-one endpoint (initial approach)
* Single Target Scraping: Discovery happens via `file_sd` approach exposed via endpoint; Prometheus receives a set of URLs at Promregator - one URL for each potential target to scrape. Promregator then behaves like a proxy for Prometheus (similar to what you know from PushProxy) and "only" does the enrichment of labels for each CF application instance separately without aggregating metrics from different targets. 

For more details on these modes, you might be interested in having a look at https://github.com/promregator/promregator/blob/643d2c6ac0ab895ba08e11c37e18a8c85d05e090/docs/documentation.md. A discussion of the Single Target Scraping mode (with pros and cons) is also available at https://github.com/promregator/promregator/blob/643d2c6ac0ab895ba08e11c37e18a8c85d05e090/docs/singleTargetScraping.md .

Do you think that it would be helpful to mention the Single Target Scraping mode here in the line of `integrations.md`?

> Your docs indicated that you rejected the (recommended) approach of having on target per target.

As laid out above, documentation should not reject this approach altogether anymore (in the past it did for the reason and history described above). Still being the default for a quick start, Single Endpoint Scraping is only recommended in case that you have a low number of applications. If you don't like a statement in the documentation, please feel free to give me feedback - then I should have another round over that.

Alternatively, if you don't feel that Promregator fits in the `file_sd`'s category in general (e.g. as it also facilitates the scraping process), I am also fine with seeing Promregator being mentioned in a different section. The other sections in place today, however, were even less appropriate to me - and adding another category just for Promregator did not feel right from my perspective either. If you see that different (or you have some further ideas how to convey the message better), I am willing to listen to your arguments.

brian-brazil:
> "only" does the enrichment of labels for each CF application instance separately without aggregating metrics from different targets.

It is something that should be done via relabelling on the Prometheus side.

> Prometheus does not support setting the metrics path for a target directly when defining targets using the file_sd_configs approach.

`__metrics_path__` should work in file_sd_configs, though a URL parameter would be safer from an escaping standpoint.

> Do you think that it would be helpful to mention the Single Target Scraping mode here in the line of integrations.md?

It's not exactly a pure file_sd, I'm thinking that Other might be more appropriate.

eaglerainbow:
Hi Brian, I saw your reply, but due to a personal topic, I can't have a look at it appropriately. I will come back to you, once I had proper time to look at your criticism.

eaglerainbow:
With your comment, you have made a point: Single Target Scraping mode should enable users to fully take advantage of the Prometheus' way of handling labels (With Single Endpoint Scraping you can't, though). With [branch disableEnrichment](https://github.com/promregator/promregator/tree/disableEnrichment) I am currently trying to permit a way of configuration where Prometheus would be in charge of adding the labels via relabeling.

> This is something that should be done via relabelling on the Prometheus side.

I would be interested to hear your arguments on this. Besides flexibility due to configuration are there any "good reasons" for doing it like this? 
Background to my question is that metric values scraped will not be unique (and thus clashes may happen), if someone forgets to relabel them in Prometheus. I feel uneasy with that.

> `__metrics_path__` should work in file_sd_configs, though a URL parameter would be safer from an escaping standpoint.

On the escaping: Where do you see the risk exactly? 
Internally, Promregator anyhow needs to match `__metrics_path__` against a list of "known targets". Moreover, the identifiers of which `__metrics_path__` consists are all pure-ASCII characters only (i.e. it's a UUID plus some integer). 

> It's not exactly a pure file_sd, I'm thinking that Other might be more appropriate.

I will move it to the "Other" section, if you feel better with that. However, give me a little time for this - I'd like to first see that the relabeling stuff runs smoothly.

eaglerainbow:
>>> It's not exactly a pure file_sd, I'm thinking that Other might be more appropriate.
>> I will move it to the "Other" section, if you feel better with that. However, give me a little time for this - I'd like to first see that the relabeling stuff runs smoothly.

In the meantime I could finish that part. There will be an option called `promregator.scraping.labelEnrichment` which can be set to `false`. Then, relabeling is done in Prometheus using `relabel_configs`. Due to compatibility reasons the user of Promregator needs to explicitly enable the mode. 

I have updated [the documentation](https://github.com/promregator/promregator/blob/f874da4c146681f18055c4558a93038e94989116/docs/documentation.md#label-rewriting) and created [a note](https://github.com/promregator/promregator/wiki/Using-relabel_configs-with-promregator.scraping.labelEnrichment) describing when and how it can be used. Feedback is appreciated.

The feature will be available starting with 0.5.0 (the version is not released, yet - I hope that this will happen soon).

> A file_sd should ensure that the instance label ends up being unique, and thus that things work correctly without relabelling.

Well, it would be an easy thing for Promregator do take care of that. But unfortunately, `instance` may not be specified by scraping, but may only be accessed during rewriting. 

>> On the escaping: Where do you see the risk exactly?
> If there were unusual characters in there, they would not get escaped.

That cannot happen - at least as no user would be post-processing (and thus mangle) the file downloaded.
So, I think we are safe here.

brian-brazil:
Thanks!

> But unfortunately, instance may not be specified by scraping, but may only be accessed during rewriting.

You can provide an instance label in file_sd. It's not usual for a generic file sd, but it makes sense here.

PR #1191, opened 2018-10-15
Change strings to secrets in alertmanager config docs

dkess:
These fields are erroneously marked as "string" when they are really "secret."

@brian-brazil 

Signed-off-by: Daniel Kessler <daniel@dkess.me>


RichiH:
Thanks

PR #1195, opened 2018-10-17
Proper curl invocation in Testing section

ektich:
@brian-brazil for your review

The Basic Auth guide explains how to configure nginx to listen on `localhost:12321`,  but in the Testing section `curl` invocations are missing the port number.


brian-brazil:
Thanks!

PR #1197, opened 2018-10-18
docs: add reference to prometheus-kafka integration

palmerabollo:
@brian-brazil this is a minor change to add a link to [prometheus-kafka-adapter](https://github.com/Telefonica/prometheus-kafka-adapter), a write-only integration for Apache Kafka.

@brian-brazil ping as requested in CONTRIBUTING.md

Signed-off-by: Guido García <guido.garciabernardo@telefonica.com>


brian-brazil:
Thanks!

One thing to note is that JSON can't handle +Inf or NaN, so you probably want to have the value as a string like we do in our APIs.

palmerabollo:
Thanks for noticing.

PR #1203, opened 2018-10-24
Fix underscore escaping in instrumentation.md

toadjaune:
The Markdown parsing on prometheus.io/docs interprets these two underscores, but I should not. (GitHub does not)

@brian-brazil, I suppose this qualifies as a trivial change.


brian-brazil:
Thanks!

PR #1204, opened 2018-10-25
Update mailing lists

fabxc:
@brian-brazil @juliusv

### DIFF
--- a/content/community.html
+++ b/content/community.html
@@ -14,8 +14,11 @@ title: Community
       (for the easiest start, <a href="https://riot.im/app/#/room/#freenode_#prometheus:matrix.org">join via Riot</a>)
     </p>
     <p>
-      <strong>Usage mailing list:</strong>
-      <a href="https://groups.google.com/forum/#!forum/prometheus-users">prometheus-users</a> – for discussions around Prometheus usage and community support
+      <strong>User mailing lists:</strong>
+      <ul>
+        <li><a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a> – for announcements like new releases</li>
+        <li><a href="https://groups.google.com/forum/#!forum/prometheus-users">prometheus-users</a> – for discussions around Prometheus usage and community support</li>
+      </ul>
     </p>
     <p>
       <strong>Twitter:</strong>


content/community.html
```diff
@@ -15,7 +15,9 @@ <h1>Community</h1>
     </p>
     <p>
       <strong>Usage mailing list:</strong>
```
brian-brazil: This needs a tweak

content/community.html
```diff
@@ -15,7 +15,9 @@ <h1>Community</h1>
...
+      <a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a> – for announcements like new releases
       <a href="https://groups.google.com/forum/#!forum/prometheus-users">prometheus-users</a> – for discussions around Prometheus usage and community support
+      <a href="https://groups.google.com/forum/#!forum/prometheus-developers">prometheus-developers</a> – for discussions around Prometheus development
```
grobie: We already list this developers list below in the contributions section.

juliusv:
:+1:

juliusv:
Should we clarify who is allowed to post to the announcements list? Is it everyone or just -team?

juliusv:
(at the moment it seems I even need approval to even join the list as a subscriber)

RichiH:
We should clarify it so it's documented.

Only -team should be allowed to send. We can gatekeep for any non-team
maintainers etc if there's ever a need.

grobie:
So what about announcements about third-party releases? I've seen people
following our [ANN] style messages.

What about letting everyone post to that list, until we see it gets abused?

juliusv:
We could allow everyone to post, but with moderator approval required for
non-owners.

RichiH:
I might be old and grumpy here, but I would argue that a project-specific list should be at least curated, better filled, by people trusted within that project.

fabxc:
I'm open to either. We just need to decide something regarding the "read-only" part in the docs.

grobie:
I like @juliusv idea, so change it to "moderated"?

RichiH:
No hard objection from me, just keep in mind that we need to actively moderate that list, then.

juliusv:
That should be fine. We also sometimes get moderation requests for our
other mailing lists when Google thinks something might be spam, and we
usually handle them quickly.

RichiH:
Can someone from -team please revert and have a pre-Callum team member
merge this? I don't think new member should merge themselves in; looks
fishy.

Need to run onto my roof so can't do it now and don't want to leave a
revert dangle; it should all be done in less than a minute :)

RichiH:
I personally don't see any problem at all with this given that it happened
after a vote and all. And the main commit didn't come from Callum.

juliusv:
Oh wow, that last comment was actually by me, not by @RichiH, but GitHub is displaying it as @RichiH (probably because I replied to an email address in his mail thread that was only meant to be used by him).

juliusv:
Seems like what I (Julius) write here shows up as being written by Richard
Hartmann now.

beorn7:
Github, you are drunk! (This should all be in #1487.)

I also don't have a problem with how the PR got merged.

RichiH:
Actually Richard: if everyone else is fine, I won't block on it. For
future, it wou be better to not mix that one

RichiH:
I did it that way just to verify that his permissions were all set.

I don't see any problem with this either, especially for such a trivial
change to something that is not mission critical code.

mdlayher:
Same thing happened here, that was my reply above.

RichiH:
On Tue, Nov 26, 2019 at 3:04 PM Richard Hartmann <notifications@github.com>
wrote:

> I did it that way just to verify that his permissions were all set.
>
> I don't see any problem with this either, especially for such a trivial
> change to something that is not mission critical code.
>

NB: If you look at this thread in GH, it's hilarious.

To Matt, who appears as me: I know the docs are not "critical", but from
the legal point of view, the governance is about the most important thing
we have. That's why I would rather do things by the book as details like
this might become relevant if shit ever hits the fan.

This is more a note for the future though. For this round, let's leave it.

>

juliusv:
> That's why I would rather do things by the book

I don't see the governance text saying anything about who may merge this kind of PR (after decisions have been made). If you feel strongly about it, you may want to propose an amendment to make it explicit. I guess Callum was only able to merge it anyway because Matt added him to the GitHub owners in the step before the PR.

PR #1206, opened 2018-10-29
Add micrometer (jvm) to list third party utilities

fitzoh:
Micrometer is the default metrics library for spring boot (as of 2.0).

http://micrometer.io/
http://micrometer.io/docs/registry/prometheus
https://spring.io/blog/2018/03/16/micrometer-spring-boot-2-s-new-application-metrics-collector
https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html


brian-brazil:
Thanks for your PR. To keep things sane for users, we aim to avoid duplication in this list and there's already an official Java client listed.

fitzoh:
No worries!  For what it's worth, there are some differentiators: micrometer is an "instrument once, report to many backends" library, much like go-metrics, which is also on the list.

PR #1207, opened 2018-10-29
Add Couchbase exporter to Databases exporters list

blakelead:
I'm writing an exporter for Couchbase community for versions 4.5.1 and 5.1.1 (other versions may work but I did not test them).

I believe it is mature enough to be submitted to the official list (I'm using it for work), but I also know that I still have some work to do, especially when it comes to testing, performance and best-practices.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -33,6 +33,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Aerospike exporter](https://github.com/alicebob/asprom)
    * [ClickHouse exporter](https://github.com/f1yegor/clickhouse_exporter)
    * [Consul exporter](https://github.com/prometheus/consul_exporter) (**official**)
+   * [Couchbase exporter](https://github.com/blakelead/couchbase_exporter)
    * [CouchDB exporter](https://github.com/gesellix/couchdb-exporter)
    * [ElasticSearch exporter](https://github.com/justwatchcom/elasticsearch_exporter)
    * [EventStore exporter](https://github.com/marcinbudny/eventstore_exporter)


brian-brazil:
Thanks for your PR. As I mentioned on the list this accepts passwords on the command line which is not a safe security practice, and you will need to fix that before I can merge this PR.


Here's some other suggestions to improve your exporter, I've not included everything but there's at least one example of every type of thing I spotted.

>	// The two following lines are used to get rid of go metrics. Should be removed after wip.

These collectors should be used, as they help monitor the health of the exporter itself.

You are using normal counters/gauges, as this is an exporter you should use MustNewConstMetric instead and then you can also remove the mutexes.

You don't seem to be checking is the return value of `Fetch` is nil (i.e. there was an error), you should fail the scrape when this happens.

> cb_cluster_fts_ram_quota_bytes

I'd go with cb_cluster_quota_ram_fts_bytes so that all the quota metrics share a prefix.

> cb_cluster_max_bucket_count	

This sounds like a Gauge, so should not end in `_count` which is used for Summary/Histogram metrics. I'd go with cb_cluster_buckets_max.

> cb_cluster_rebalance_success_count	

This sounds like a counter, so should end with _total.

Some of the node metrics sound like machine-level metrics, any of those should be removed as that's the Node Exporter's responsibility.

> cb_node_stats_couch_docs_actual_disk_size

A unit should be included here.

> cb_node_stats_ops

This and a few others sounds like counters.

> cb_bucket_ram_quota_percent_used

Ratios are preferred, and even better is raw total and used metrics.

> cb_bucket_hit_ratio

Similarly, total and hits is preferred.

> cb_bucket_ops_per_second

Could you have this as a raw counter instead?

> cb_bucket_ep_resident_items_rate

Why the '_rate'?

> cb_bucket_vb_avg_active_queue_age

What are the units for this?

> cb_bucket_vb_active_resident_items_ratio

Why the `_ratio`?

> cb_bucket_bytes_read

The unit should go at the end of the metric, and this sounds like a counter.

> cb_bucket_curr_connections

Implicitly all metrics are current information, so you don't need that in the metric name.

> cb_bucket_ep_dcp_*_*

I think you might want a label here.

blakelead:
Hi @brian-brazil ,

Thanks a lot for your feedback!

I'll work the next few days on everything you said, and I have a question for you about password and security issue: 
Is it accepted to pass password with environment variables? If not, can you tell me what is considered best practice in this matter?
I did not find satisfactory answers on the web apart maybe storing credentials in a config file and loading it at the start of the exporter.

brian-brazil:
> Is it accepted to pass password with environment variables? If not, can you tell me what is considered best practice in this matter?

It's not as good as using a file, but it wouldn't block you here.

blakelead:
I just removed the possibility to pass credentials via command-line and added a configuration file.

brian-brazil:
Thanks!

Issue #1193, opened 2018-10-15
Add proxy exporter pattern explanation


vonneudeck:
The blackbox, SNMP and IPMI exporter make use of the proxy exporter pattern, which is in itself special enough to be explained to clarify their usage.

The proxy pattern is that the exporter is getting the target and the module to apply on it via an HTTP GET request from Prometheus and then starts its own request over network and returns the answer with the HTTP request response. In practice this often coupled with using relabeling in the Prometheus config.yml that is not trivial, especially for people new to Prometheus.

The blackbox exporter will be used as an example as it is:
- introducing no second protocol
- the most popular of those three according to docker hub statistics
- often used with relabeling

The purpose is neither a complete step-by-step guide nor a complete reference. It should enable people who have not heard of the proxy pattern before to deploy and configure such exporters and Prometheus to do something useful for them.

This explanation will be linked to by the SNMP exporter explanation #1192 .

As proposed on the [users mailing list](https://groups.google.com/d/msg/prometheus-users/jFsTY8g3pk8/nG5Ua0pcCgAJ) and announced on the [devs mailing list](https://groups.google.com/d/msg/prometheus-developers/edSzZuPI7Mc/VoWpl61_BAAJ) I want to write these explanations, but help is very welcome.

brian-brazil:
> proxy exporter pattern

There's no such term used in the project, and all exporters are fundamentally proxies so the name would be confusing.

brian-brazil:
I also find proxy exporter to be acceptable for exporters that are 1:1 with their targets, and commonly explain exporters as being proxies.

SuperQ:
I usually call them sidecars, or sidecar exporters. This is a more common term used by Kubernetes users.

brian-brazil:
I suggest we find a different term, sidecar is a bit k8 specific. Maybe multi-target or provided target?

RichiH:
The fundamental difference between node_exporter, snmp_exporter, and ipmi_exporter is that they are 1:1, 1:n, and 1:1/n, respectively. That's the reason why I dislike ipmi_exporter's design decision.

I would also call it multi-target, multiplexer, 1:n, or something along those terms.

As we are mangling data, I would propose `1:n` as the official term; this is well-defined in the DB and data structure world.

vonneudeck:
Sorry, I pressed the button accidentally.

matthiasr:
[...]

PR #1209, opened 2018-11-01
Added guide about proxy exporters/blackbox exporter

vonneudeck:
Solves #1193 

This guide should answer a lot of common questions on the mailing list
and help people understand how this type of exporters work while also
giving them practical experience.

It handles a few common pitfalls and is written to be as accessible as
possible. Hopefully it will help many people to have a less frustrating
experience using Prometheus and free the mailing list from some load of
often repeated questions.

It is a prerequirement for an SNMP exporter guide, that is useful for beginners.

Signed-off-by: Nicolai von Neudeck <nicolai.von.neudeck@sap.com>

### DIFF
--- a/content/docs/guides/multi-target-exporter.md
+++ b/content/docs/guides/multi-target-exporter.md
@@ -220,7 +220,7 @@ sudo docker \
 With this command, you told `docker` to:
 
 1. `run` a container with the port `9115` outside the container mapped to the port `9115` inside of the container.
-1. `mount` from your current directory (`$(pwd)` stands for *p*rint *w*orking *d*irectory) the file `blackbox.yml` into `/blackbox.yml` in `readonly` mode.
+1. `mount` from your current directory (`$(pwd)` stands for print working directory) the file `blackbox.yml` into `/blackbox.yml` in `readonly` mode.
 1. use the image `prom/blackbox-exporter` from [Docker hub](https://hub.docker.com/r/prom/blackbox-exporter/).
 1. run the blackbox-exporter with the flag `--config.file` telling it to use `/blackbox.yml` as config file.
 
--- a/content/docs/guides/multi-target-exporter.md
+++ b/content/docs/guides/multi-target-exporter.md
@@ -35,7 +35,7 @@ Now let’s try it out for yourself!
 Use [Docker](https://www.docker.com/) to start a blackbox exporter container running by running this in a terminal:
 
 ```bash
-docker run -p 9115:9115 prom/blackbox-exporter
+sudo docker run -p 9115:9115 prom/blackbox-exporter
 ```
 
 You should see a few log lines and if everything went well the last one should report `msg="Listening on address"` as seen here:
@@ -228,8 +228,8 @@ probe_ip_protocol 6
 probe_success 0
 ```
 
-Unfortunately almost all with the value `0`. The last one a very disheartening `probe_success 0`. This means the prober could not sucessfully reach `prometheus.io`.  
-The reason is hidden in the metric `probe_ip_protocol` with the value `6`. By default the prober uses [IPv6](https://en.wikipedia.org/wiki/IPv6) until told otherwise. Unfortunately the Docker daemon blocks IPv6 until told otherwise. Hence our blackbox exporter running in a Docker container can’t connect via IPv6.
+Notice that almost all metrics have a value of `0`. The last one reads `probe_success 0`. This means the prober could not sucessfully reach `prometheus.io`.  
+The reason is hidden in the metric `probe_ip_protocol` with the value `6`. By default the prober uses [IPv6](https://en.wikipedia.org/wiki/IPv6) until told otherwise. But the Docker daemon blocks IPv6 until told otherwise. Hence our blackbox exporter running in a Docker container can’t connect via IPv6.
 
 We could now either tell Docker to allow IPv6 or the blackbox exporter to use IPv4. In the real world both can make sense and as so often the answer to the question "what is to be done?" is "it depends". Because this is an exporter guide we will change the exporter and take the oportunity to configure a custom module.
 
@@ -245,13 +245,12 @@ First download the file using curl or your browser:
 curl -o blackbox.yml https://raw.githubusercontent.com/prometheus/blackbox_exporter/master/blackbox.yml
 ```
 
-Opened in an editor the first few lines look like this:
+Open it in an editor. The first few lines look like this:
 
 ```yaml
 modules:
   http_2xx:
     prober: http
-    http:
   http_post_2xx:
     prober: http
     http:
@@ -259,7 +258,7 @@ modules:
 ```
 
 [Yaml](https://en.wikipedia.org/wiki/YAML) uses whitespace indentation to express hierarchy, so you can recognise that two `modules` named `http_2xx` and `http_post_2xx` are defined, and that they both have a prober `http` and for one the method value is specifically set to `POST`.  
-We will now change the module `http_2xx` by setting the `preferred_ip_protocol` of the prober `http` explicitly to the string `ip4`.
+You will now change the module `http_2xx` by setting the `preferred_ip_protocol` of the prober `http` explicitly to the string `ip4`.
 
 ```yaml
 modules:
@@ -275,7 +274,7 @@ modules:
 
 If you want to know more about the available probers and options check out the [documentation](https://github.com/prometheus/blackbox_exporter/blob/master/CONFIGURATION.md).
 
-Now we need to tell the blackbox exporter to use our freshly changed file. We can do that with the flag `--config.file="blackbox.yml"`. But because we are using Docker, we first must make this file [available](https://docs.docker.com/storage/bind-mounts/) inside the container using the `--mount` command.  
+Now we need to tell the blackbox exporter to use our freshly changed file. You can do that with the flag `--config.file="blackbox.yml"`. But because we are using Docker, we first must make this file [available](https://docs.docker.com/storage/bind-mounts/) inside the container using the `--mount` command.  
 
 NOTE: If you are using macOS you first need to allow the Docker daemon to access the directory in which your `blackbox.yml` is. You can do that by clicking on the little Docker whale in menu bar and then on `Preferences`->`File Sharing`->`+`. Afterwards press `Apply & Restart`.
 
@@ -286,7 +285,7 @@ Then you run this command. It is long, but we will explain it:
 <a name="run-exporter"></a>
 
 ```bash
-docker \
+sudo docker \
   run -p 9115:9115 \
   --mount type=bind,source="$(pwd)"/blackbox.yml,target=/blackbox.yml,readonly \
   prom/blackbox-exporter \
@@ -308,10 +307,10 @@ level=info ts=2018-10-19T12:40:51.653357722Z caller=main.go:220 msg="Loaded conf
 level=info ts=2018-10-19T12:40:51.65349635Z caller=main.go:324 msg="Listening on address" address=:9115
 ```
 
-Now we can try our new IPv4-using module `http_2xx` in a terminal:
+Now you can try our new IPv4-using module `http_2xx` in a terminal:
 
 ```bash
-curl "localhost:9115/probe?target=prometheus.io&module=http_2xx"
+curl 'localhost:9115/probe?target=prometheus.io&module=http_2xx'
 ```
 
 Which should return Prometheus metrics like this:
@@ -319,10 +318,10 @@ Which should return Prometheus metrics like this:
 ```
 # HELP probe_dns_lookup_time_seconds Returns the time taken for probe dns lookup in seconds
 # TYPE probe_dns_lookup_time_seconds gauge
-probe_dns_lookup_time_seconds 0.060584416
+probe_dns_lookup_time_seconds 0.02679421
 # HELP probe_duration_seconds Returns how long the probe took to complete in seconds
 # TYPE probe_duration_seconds gauge
-probe_duration_seconds 0.37071774
+probe_duration_seconds 0.461619124
 # HELP probe_failed_due_to_regex Indicates if probe failed due to regex
 # TYPE probe_failed_due_to_regex gauge
 probe_failed_due_to_regex 0
@@ -331,11 +330,11 @@ probe_failed_due_to_regex 0
 probe_http_content_length -1
 # HELP probe_http_duration_seconds Duration of http request by phase, summed over all redirects
 # TYPE probe_http_duration_seconds gauge
-probe_http_duration_seconds{phase="connect"} 0.041410336000000006
-probe_http_duration_seconds{phase="processing"} 0.19369339
-probe_http_duration_seconds{phase="resolve"} 0.062867198
-probe_http_duration_seconds{phase="tls"} 0.092468257
-probe_http_duration_seconds{phase="transfer"} 0.000145375
+probe_http_duration_seconds{phase="connect"} 0.062076202999999996
+probe_http_duration_seconds{phase="processing"} 0.23481845699999998
+probe_http_duration_seconds{phase="resolve"} 0.029594103
+probe_http_duration_seconds{phase="tls"} 0.163420078
+probe_http_duration_seconds{phase="transfer"} 0.002243199
 # HELP probe_http_redirects The number of redirects
 # TYPE probe_http_redirects gauge
 probe_http_redirects 1
@@ -345,6 +344,9 @@ probe_http_ssl 1
 # HELP probe_http_status_code Response HTTP status code
 # TYPE probe_http_status_code gauge
 probe_http_status_code 200
+# HELP probe_http_uncompressed_body_length Length of uncompressed response body
+# TYPE probe_http_uncompressed_body_length gauge
+probe_http_uncompressed_body_length 14516
 # HELP probe_http_version Returns the version of HTTP of the probe response
 # TYPE probe_http_version gauge
 probe_http_version 1.1
@@ -353,20 +355,23 @@ probe_http_version 1.1
 probe_ip_protocol 4
 # HELP probe_ssl_earliest_cert_expiry Returns earliest SSL cert expiry in unixtime
 # TYPE probe_ssl_earliest_cert_expiry gauge
-probe_ssl_earliest_cert_expiry 1.552175999e+09
+probe_ssl_earliest_cert_expiry 1.581897599e+09
 # HELP probe_success Displays whether or not the probe was a success
 # TYPE probe_success gauge
 probe_success 1
+# HELP probe_tls_version_info Contains the TLS version used
+# TYPE probe_tls_version_info gauge
+probe_tls_version_info{version="TLS 1.3"} 1
 ```
 
-We can see that the probe was successful and get many useful metrics, like latency by phase, status code, ssl status or cert expiry in [epoch](https://en.wikipedia.org/wiki/Unix_time).  
+You can see that the probe was successful and get many useful metrics, like latency by phase, status code, ssl status or cert expiry in [epoch](https://en.wikipedia.org/wiki/Unix_time).  
 The blackbox exporter also offers a tiny web interface at [localhost:9115](http://localhost:9115) for you to check out the last few probes, the loaded config and debug information. It even offers a direct link to probe `prometheus.io`. Handy if you are wondering why something does not work.
 
 ## Querying multi-target exporters with Prometheus
 
-So far, so good. Congratulate yourself. The blackbox exporter works and you can manually tell it to query a remote target. We are almost there. We now need to tell Prometheus to do the queries for us.  
+So far, so good. Congratulate yourself. The blackbox exporter works and you can manually tell it to query a remote target. You are almost there. Now you need to tell Prometheus to do the queries for us.  
 
-Below you find a minimal prometheus config. It is telling Prometheus to scrape the exporter itself as we did [before](#query-exporter) using `curl "localhost:9115/metrics"`:
+Below you find a minimal prometheus config. It is telling Prometheus to scrape the exporter itself as we did [before](#query-exporter) using `curl 'localhost:9115/metrics'`:
 
 NOTE: If you use Docker for Mac or Docker for Windows, you can’t use `localhost:9115` in the last line, but must use `host.docker.internal:9115`. This has to do with the virtual machines used to implement Docker on those operating systems. You should not use this in production.
 
@@ -396,12 +401,22 @@ scrape_configs:
       - host.docker.internal:9115
 ```
 
-Now run a Prometheus container and tell it to mount our config file from above:
+Now run a Prometheus container and tell it to mount our config file from above. Because of the way networking on the host is addressable from the container you need to use a slightly different command on Linux than on MacOS and Windows.:
 
 <a name=run-prometheus></a>
 
+Run Prometheus on Linux (don’t use `--network="host"` in production):
 ```bash
-docker \
+sudo docker \
+  run --network="host"\
+  --mount type=bind,source="$(pwd)"/prometheus.yml,target=/prometheus.yml,readonly \
+  prom/prometheus \
+  --config.file="/prometheus.yml"
+```
+
+Run Prometheus on MacOS and Windows:
+```bash
+sudo docker \
   run -p 9090:9090 \
   --mount type=bind,source="$(pwd)"/prometheus.yml,target=/prometheus.yml,readonly \
   prom/prometheus \
@@ -412,13 +427,20 @@ This command works similarly to [running the blackbox exporter using a config fi
 
 If everything worked, you should be able to go to [localhost:9090/targets](http://localhost:9090/targets) and see under `blackbox-exporter-metrics` an endpoint with the state `UP` in green. If you get a red `DOWN` make sure that the blackbox exporter you started [above](run-exporter) is still running. If you see nothing or a yellow `UNKNOWN` you are really fast and need to give it a few more seconds before reloading your browser’s tab.
 
-To tell Prometheus to query `"localhost:9115/probe?target=prometheus.io&module=http_2xx"` we set the `metrics_path` to `/probe` and the parameters under `params:` in the Prometheus config file.
+To tell Prometheus to query `"localhost:9115/probe?target=prometheus.io&module=http_2xx"` you add another scrape job where you set the `metrics_path` to `/probe` and the parameters under `params:` in the Prometheus config file `prometheus.yml`:
 
 <a name="prometheus-config"></a>
 
 ```yaml
 scrape_configs:
 
+- job_name: blackbox-exporter-metrics
+  scrape_interval: 5s
+  metrics_path: /metrics
+  static_configs:
+    - targets:
+      - localhost:9115   # for Windows and macOS replace with - host.docker.internal:9115
+      
 - job_name: 'blackbox'
   scrape_interval: 5s
   metrics_path: /probe
@@ -448,10 +470,17 @@ The details are worthy of a [talk](https://www.youtube.com/watch?v=b5-SvvZ7AwI) 
 * You can set internal labels that are called `__param_<name>`. Those set URL parameter with the key `<name>` for the scrape request.
 * There is an internal label `__address__` which is set by the `targets` under `static_configs` and whose value is the hostname for the scrape request. By default it is later used to set the value for the label `instance`, which is attached to each metric and tells you were the metrics came from.
 
-Don’t worry if you are confused now, we will go through our new config step by step’
+Here is the config you will use to do that. Don’t worry if this is a bit much at once, we will go through it step by step:
 
 ```yaml
 scrape_configs:
+
+- job_name: blackbox-exporter-metrics
+  scrape_interval: 5s
+  metrics_path: /metrics
+  static_configs:
+    - targets:
+      - localhost:9115   # for Windows and macOS replace with - host.docker.internal:9115
 
 - job_name: 'blackbox'
   scrape_interval: 5s
@@ -485,7 +514,7 @@ This way we can have the actual targets there, get them as `instance` label valu
 
 Often people combine these with a specific service discovery. Check out the [configuration documentation](/docs/prometheus/latest/configuration/configuration) for more information. Using them is no problem, as these write into the `__address__` label just like `targets` defined under `static_configs`.
 
-That is it. Restart the Prometheus docker container and look at your [metrics](http://localhost:9090/graph).
+That is it. Restart the Prometheus docker container and look at your [metrics](http://localhost:9090/graph?g0.range_input=5m&g0.stacked=0&g0.expr=probe_http_duration_seconds&g0.tab=0).
 
 # Summary
 
--- a/content/docs/guides/multi-target-exporter.md
+++ b/content/docs/guides/multi-target-exporter.md
@@ -1,33 +1,34 @@
 ---
-title: Understanding and using the proxy exporter pattern
+title: Understanding and using the multi-target exporter pattern
 ---
 
-# Understanding and using the proxy exporter pattern
+# Understanding and using the multi-target exporter pattern
 
-This guide will introduce you to the proxy exporter pattern. To achieve this we will:
+This guide will introduce you to the multi-target exporter pattern. To achieve this we will:
 
-* describe the proxy exporter pattern and why it is used,
+* describe the multi-target exporter pattern and why it is used,
 * run the [blackbox](https://github.com/prometheus/blackbox_exporter) exporter as an example of the pattern,
 * configure a custom query module for the blackbox exporter,
 * let the blackbox exporter run basic metric queries against the Prometheus [website](https://prometheus.io),
 * examine a popular pattern of configuring Prometheus to scrape exporters using relabeling.
 
-## The proxy exporter pattern?
+## The multi-target exporter pattern?
 
-Any exporter receives requests from Prometheus and answers them with data forwarded from other systems. So technically every [exporter](/docs/instrumenting/exporters/) is a [proxy](https://en.wikipedia.org/wiki/Proxy_server). But by proxy exporter pattern we refer to a specific design, in which:
+By multi-target [exporter](/docs/instrumenting/exporters/) pattern we refer to a specific design, in which:
 
 * the exporter will get the target’s metrics via a network protocol.
 * the exporter does not have to run on the machine the metrics are taken from.
 * the exporter gets the targets and a query config string as parameters of Prometheus’ GET request.
 * the exporter subsequently starts the scrape after getting Prometheus’ GET requests and once it is done with scraping.
+* the exporter can query multiple targets.
 
 This pattern is only used for certain exporters, namely the [blackbox](https://github.com/prometheus/blackbox_exporter) and the [SNMP exporter](https://github.com/prometheus/snmp_exporter).
 
 The reason is that we either can’t run the exporter on the targets, e.g. network gear speaking SNMP, or that we are explicitly interested in the distance, e.g. latency and reachability of a website from a specific point outside of our network.
 
-## Running proxy exporters
+## Running multi-target exporters
 
-Proxy exporters are flexible regarding their environment and can be run in many ways. As regular programs, in containers, as background services, on baremetal, virtual machines. Because they are queried and do query over network they do need appropriate open ports. Otherwise they are frugal.
+Multi-target exporters are flexible regarding their environment and can be run in many ways. As regular programs, in containers, as background services, on baremetal, virtual machines. Because they are queried and do query over network they do need appropriate open ports. Otherwise they are frugal.
 
 Now let’s try it out for yourself!
 
@@ -43,7 +44,7 @@ You should see a few log lines and if everything went well the last one should r
 level=info ts=2018-10-17T15:41:35.4997596Z caller=main.go:324 msg="Listening on address" address=:9115
 ```
 
-## Basic querying of proxy exporters
+## Basic querying of multi-target exporters
 
 There are two ways of querying:
 
@@ -280,7 +281,7 @@ NOTE: If you are using macOS you first need to allow the Docker daemon to access
 
 First you stop the old container by changing into its terminal and press `ctrl+c`.
 Make sure you are in the directory containing your `blackbox.yml`.
-Then you run this command. Don’t be intimidated, it is long, but we explain it:
+Then you run this command. It is long, but we will explain it:
 
 <a name="run-exporter"></a>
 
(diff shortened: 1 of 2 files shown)


content/docs/guides/proxy-exporter.md
```diff
@@ -0,0 +1,375 @@
+---
+title: Understanding and using the proxy exporter pattern
+---
+
+# Understanding and using the proxy exporter pattern
```
brian-brazil: Proxy exporter is still not a term that we use blackbox-like is the closest.

content/docs/guides/proxy-exporter.md
```diff
@@ -0,0 +1,375 @@
...
+## The proxy exporter pattern?
+
+Any exporter receives requests from Prometheus and answers them with data forwarded from other systems. So technically every [exporter](/docs/instrumenting/exporters/) is a [proxy](https://en.wikipedia.org/wiki/Proxy_server). But by proxy exporter pattern we refer to a specific design, in which:
+
+* the exporter will get the target’s metrics via a network protocol.
+* the exporter does not have to run on the machine the metrics are taken from.
+* the exporter gets the targets and a query config string as parameters of Prometheus’ GET request.
+* the exporter subsequently starts the scrape after getting Prometheus’ GET requests and once it is done with scraping.
+
+This pattern is only used for certain exporters, namely the [blackbox](https://github.com/prometheus/blackbox_exporter), the [SNMP](https://github.com/prometheus/snmp_exporter) and the [IPMI exporter](https://github.com/soundcloud/ipmi_exporter).
```
brian-brazil: IPMI is mixed now, best not to mention it

content/docs/guides/proxy-exporter.md
```diff
@@ -0,0 +1,375 @@
...
+
+The response should be a list of metrics of mostly starting with `go_` and `process_` in the Prometheus [format](/docs/instrumenting/exposition_formats/#text-format-example).
+
+For the second type of querying we need to provide a target and module as parameters in the HTTP GET Request. The target is a URI or IP and the module must defined in the exporter’s configuration. The blackbox exporter container comes with a meaningful default configuration.  
+We will use the target `prometheus.io` and the predefined module `http_2xx`. It tells the exporter to make a GET request like a browser would if you go to `prometheus.io` and to expect a [200 OK](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success) response.
+
+You can now tell your blackbox exporter to query `prometheus.io` in the terminal with curl:
+
+```bash
+curl "localhost:9115/probe?target=prometheus.io&module=http_2xx"
```
brian-brazil: I'd add an indication of why you have the quotes here, this trips people up

content/docs/guides/proxy-exporter.md
```diff
@@ -0,0 +1,375 @@
...
+# HELP probe_ip_protocol Specifies whether probe ip protocol is IP4 or IP6
+# TYPE probe_ip_protocol gauge
+probe_ip_protocol 6
+# HELP probe_success Displays whether or not the probe was a success
+# TYPE probe_success gauge
+probe_success 0
+```
+
+Unfortunately almost all with the value `0`. The last one a very disheartening `probe_success 0`. This means the prober could not sucessfully reach `prometheus.io`.  
+The reason is hidden in the metric `probe_ip_protocol` with the value `6`. By default the prober uses [IPv6](https://en.wikipedia.org/wiki/IPv6) until told otherwise. Unfortunately the Docker daemon blocks IPv6 until told otherwise. Hence our blackbox exporter running in a Docker container can’t connect via IPv6.
```
brian-brazil: This is getting into the blackbox exporter, rather than blackbox-like exporters.

vonneudeck: Yes. This is a feature. This is a good opportunity to gently introduce how to modify/define modules while also avoiding frustration for many beginners with the most popular proxy exporter.

content/docs/guides/proxy-exporter.md
vonneudeck: Oh, I did not know that. Thanks for the hint.

content/docs/guides/proxy-exporter.md
vonneudeck: Before I made the PR I tested this guide thoroughly with someone completely new to the command line, containers, Prometheus and exporters. They did not stumble over this, hence I do not worry too much.

content/docs/guides/proxy-exporter.md
RichiH: As per https://github.com/prometheus/docs/issues/1193#issuecomment-435324102 I would suggest `1:n exporter` - that way you can also directly explain why they need target info from their caller in the text below.

content/docs/guides/proxy-exporter.md
RichiH: If IPMI is so confusing that it should not be part of the docs, that says a lot about IPMI exporter.

content/docs/guides/proxy-exporter.md
```diff
@@ -0,0 +1,375 @@
...
+
+```
+level=info ts=2018-10-17T15:41:35.4997596Z caller=main.go:324 msg="Listening on address" address=:9115
+```
+
+## Basic querying of proxy exporters
+
+There are two ways of querying:
+
+1. Querying the exporter itself. It has its own metrics, usually available at `/metrics`.
```
RichiH: I think this is non-obvious to new users, but very important. It would be a good place to quickly explain why and link to instrumentalization.

content/docs/guides/proxy-exporter.md
RichiH: I am torn; intermediate users might get confused where new ones just type things without questioning it. OTOH, this is not a shell guide. A quick note that quoting protects the URL's special chars can't hurt, though. Also, you should probably use `'`, not `"`.

content/docs/guides/proxy-exporter.md
RichiH: Maybe add an intro sentence what exporters you will use in the guide, why, and what you can expect to learn in it. That way, explaining two things alongside is expected.

content/docs/guides/proxy-exporter.md
vonneudeck: I think we should have that discussion where everyone else is having it. In the issue.
Also please be careful with the "we".  I am sure you don’t want to imply that all the people who expressed different proposals are not part of the project.

content/docs/guides/proxy-exporter.md
vonneudeck: I tested the whole guide on someone who is very new to the shell, containers, Prometheus and exporters and this did not came up as a tripping hazard. I am not overly worried about this.

content/docs/guides/proxy-exporter.md
vonneudeck: I did mention this it in the intro, but maybe this is not clear for readers as I hoped it would be. What would you change there?

content/docs/guides/proxy-exporter.md
vonneudeck: Good point. I added something.

content/docs/guides/proxy-exporter.md
brian-brazil: It's not confusing just a bit unusual, but would cause confusion in an introductory guide like this as it is both types of exporter.

content/docs/guides/proxy-exporter.md
vonneudeck: I changed the `"` to `'`. If I put a note in for the `'` I should likely remove them further up at the first invocation of curl with no special signs. And I think that make things more confusing, and there is already enough content people need to learn in this guide.

vonneudeck:
I am glad you like the gentle style. I spent much effort to make it friendly, low barrier and encouraging. Do you think the "unfortunately" is to negative?

content/docs/guides/proxy-exporter.md
```diff
@@ -0,0 +1,491 @@
...
+probe_http_version 0
+# HELP probe_ip_protocol Specifies whether probe ip protocol is IP4 or IP6
+# TYPE probe_ip_protocol gauge
+probe_ip_protocol 6
+# HELP probe_success Displays whether or not the probe was a success
+# TYPE probe_success gauge
+probe_success 0
+```
+
+Unfortunately almost all with the value `0`. The last one a very disheartening `probe_success 0`. This means the prober could not sucessfully reach `prometheus.io`.
```
SuperQ: How about `Notice the metircs that have a value of 0`, this brings attention, rather than judgment of the results.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,521 @@
...
+
+* the exporter will get the target’s metrics via a network protocol.
+* the exporter does not have to run on the machine the metrics are taken from.
+* the exporter gets the targets and a query config string as parameters of Prometheus’ GET request.
+* the exporter subsequently starts the scrape after getting Prometheus’ GET requests and once it is done with scraping.
+* the exporter can query multiple targets.
+
+This pattern is only used for certain exporters, namely the [blackbox](https://github.com/prometheus/blackbox_exporter) and the [SNMP exporter](https://github.com/prometheus/snmp_exporter).
+
+The reason is that we either can’t run the exporter on the targets, e.g. network gear speaking SNMP, or that we are explicitly interested in the distance, e.g. latency and reachability of a website from a specific point outside of our network.
```
brian-brazil: an exporter

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,521 @@
...
+## Running multi-target exporters
+
+Multi-target exporters are flexible regarding their environment and can be run in many ways. As regular programs, in containers, as background services, on baremetal, virtual machines. Because they are queried and do query over network they do need appropriate open ports. Otherwise they are frugal.
+
+Now let’s try it out for yourself!
+
+Use [Docker](https://www.docker.com/) to start a blackbox exporter container running by running this in a terminal:
+
+```bash
+sudo docker run -p 9115:9115 prom/blackbox-exporter
```
brian-brazil: Why the sudo?

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,521 @@
...
+## Basic querying of multi-target exporters
+
+There are two ways of querying:
+
+1. Querying the exporter itself. It has its own metrics, usually available at `/metrics`.
+1. Querying the exporter to scrape another target. Usually available at a "descriptive" endpoint, e.g. `/probe`.
+
+You can manually try the first query type with curl in another terminal. If you don’t have curl, you can also copy the part between the `'` into a browser’s address bar:
+
+<a name="query-exporter"></a>
```
brian-brazil: Is this meant to be here?

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,521 @@
...
+```
+
+## Basic querying of multi-target exporters
+
+There are two ways of querying:
+
+1. Querying the exporter itself. It has its own metrics, usually available at `/metrics`.
+1. Querying the exporter to scrape another target. Usually available at a "descriptive" endpoint, e.g. `/probe`.
+
+You can manually try the first query type with curl in another terminal. If you don’t have curl, you can also copy the part between the `'` into a browser’s address bar:
```
brian-brazil: I think it's reasonably to presume that everyone has curl, otherwise use an `<a>` to link there directly.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,521 @@
...
+<a name="query-exporter"></a>
+
+```bash
+curl 'localhost:9115/metrics'
+```
+
+The response should be something like this:
+
+```
+# HELP blackbox_exporter_build_info A metric with a constant '1' value labeled by version, revision, branch, and goversion from which blackbox_exporter was built.
```
brian-brazil: I'd trim this down, it's largely not relevant to the goal of the page

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,521 @@
...
+probe_ssl_earliest_cert_expiry 1.581897599e+09
+# HELP probe_success Displays whether or not the probe was a success
+# TYPE probe_success gauge
+probe_success 1
+# HELP probe_tls_version_info Contains the TLS version used
+# TYPE probe_tls_version_info gauge
+probe_tls_version_info{version="TLS 1.3"} 1
+```
+
+You can see that the probe was successful and get many useful metrics, like latency by phase, status code, ssl status or cert expiry in [epoch](https://en.wikipedia.org/wiki/Unix_time).
```
brian-brazil: I wouldn't refer to this as just "epoch", which epoch? What unit?
"Unix time" would be the usual term.

content/docs/guides/multi-target-exporter.md
vonneudeck: Because this docker command does not work without using sudo.

content/docs/guides/multi-target-exporter.md
vonneudeck: Yes, I link to that anchor later in the guide to make it more comfortable to compare two queries for interested users.

content/docs/guides/multi-target-exporter.md
vonneudeck: You are right, most people do have curl, but it is not preinstalled on many popular distributions.
A link is a very good idea, I will change that.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+  run -p 9115:9115 \
+  --mount type=bind,source="$(pwd)"/blackbox.yml,target=/blackbox.yml,readonly \
+  prom/blackbox-exporter \
+  --config.file="/blackbox.yml"
+```
+
+With this command, you told `docker` to:
+
+1. `run` a container with the port `9115` outside the container mapped to the port `9115` inside of the container.
+1. `mount` from your current directory (`$(pwd)` stands for *p*rint *w*orking *d*irectory) the file `blackbox.yml` into `/blackbox.yml` in `readonly` mode.
```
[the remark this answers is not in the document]
vonneudeck: Does it? It looks fine in the GitHub file preview.
But my heart is not attached to the emphasis of the initials.
If you think it is better, I can take them away.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+probe_ip_protocol 4
+# HELP probe_ssl_earliest_cert_expiry Returns earliest SSL cert expiry in unixtime
+# TYPE probe_ssl_earliest_cert_expiry gauge
+probe_ssl_earliest_cert_expiry 1.552175999e+09
+# HELP probe_success Displays whether or not the probe was a success
+# TYPE probe_success gauge
+probe_success 1
+```
+
+We can see that the probe was successful and get many useful metrics, like latency by phase, status code, ssl status or cert expiry in [epoch](https://en.wikipedia.org/wiki/Unix_time).
```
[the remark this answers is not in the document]
vonneudeck: Sure, I will change it as you said.

content/docs/guides/multi-target-exporter.md
vonneudeck: Sure.

Like this:
```
TEXT text text text
TEXT text text text
some_metric 0

…

HELP text text text
TEXT text text text
another_metric 1
```

or this:

```
TEXT text text text
TEXT text text text
some_metric 0

[…]

HELP text text text
TEXT text text text
another_metric 1
```

content/docs/guides/multi-target-exporter.md
brian-brazil: I don't understand why this should require sudo, this sounds like something odd with your setup rather than a general thing about docker.

content/docs/guides/multi-target-exporter.md
brian-brazil: Whatever you think is best

content/docs/guides/multi-target-exporter.md
vonneudeck: I tested this on a very vanilla Ubuntu Linux like many potential readers of the guide would use and it needs sudo. What more can I tell you about it?

brian-brazil: The command works fine for me without sudo, and I'm also on vanilla Ubuntu. I'm guessing your user isn't a member of the docker group?

In generally let's not recommend use of sudo when it's not 100% required.

matthiasr: I haven't tried it yet but docker on Mac / Windows is strongly tied to the _user_ so I expect `sudo docker` tp either blow up spectacularly or misbehave subtly. let's give the command without it, and mention in a separate sentence that in some setups you may need sudo?

content/docs/guides/multi-target-exporter.md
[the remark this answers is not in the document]
matthiasr: In the diff view everything afterwards is italic. To me that shows that it's at least ambiguous, Markdown being the underspecified mess that it is. I'd say remove it to avoid issues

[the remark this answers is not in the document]
vonneudeck: Well, sad, but okay. I’ll remove the *s.

content/docs/guides/multi-target-exporter.md
vonneudeck: Okay, that sounds like a reasonable approach between best practices and raising the bar to entry. I’ll add it.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,448 @@
...
+```
+level=info ts=2018-10-17T15:41:35.4997596Z caller=main.go:324 msg="Listening on address" address=:9115
+```
+
+## Basic querying of multi-target exporters
+
+There are two ways of querying:
+
+1. Querying the exporter itself. It has its own metrics, usually available at `/metrics`.
+1. Querying the exporter to scrape another target. Usually available at a "descriptive" endpoint, e.g. `/probe`.
```
RichiH: Maybe state that this is what people usually want but that the former is better for demonstration?

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,448 @@
...
+
+* the exporter will get the target’s metrics via a network protocol.
+* the exporter does not have to run on the machine the metrics are taken from.
+* the exporter gets the targets and a query config string as parameters of Prometheus’ GET request.
+* the exporter subsequently starts the scrape after getting Prometheus’ GET requests and once it is done with scraping.
+* the exporter can query multiple targets.
+
+This pattern is only used for certain exporters, namely the [blackbox](https://github.com/prometheus/blackbox_exporter) and the [SNMP exporter](https://github.com/prometheus/snmp_exporter).
+
+The reason is that we either can’t run an exporter on the targets, e.g. network gear speaking SNMP, or that we are explicitly interested in the distance, e.g. latency and reachability of a website from a specific point outside of our network.
```
RichiH: Mention blackbox for the second type? You're using it for examples below so that gives some mental reference.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,448 @@
...
+
+After saving the config file switch to the terminal with your Prometheus docker container and stop it by pressing `ctrl+C` and start it again to reload the configuration by using the existing [command](#run-prometheus).
+
+The terminal should return the message `"Server is ready to receive web requests."` and after a few seconds you should start to see colourful graphs in [your Prometheus](http://localhost:9090/graph?g0.range_input=5m&g0.stacked=0&g0.expr=probe_http_duration_seconds&g0.tab=0).
+
+This works, but it has a few disadvantages:
+
+1. The actual targets are up in the param config, which is very unusual and hard to understand later.
+1. The `instance` label has the value of the blackbox exporter’s address which is technically true, but not what we are interested in.
+1. We can’t see which URL we probed. This is unpractical and will also mix up different metrics into one if we probe several URLs.
```
RichiH: Maybe "target URL"?

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,448 @@
...
+
+To fix this, we will use [relabeling](/docs/prometheus/latest/configuration/configuration/#<relabel_config>).
+Relabeling is very powerful for advanced usecases because behind the scenes many things in Prometheus are configured with internal labels.
```
RichiH: This sentence is somewhat complicated and doesn't really tell the reader much.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,448 @@
...
+The details are worthy of a [talk](https://www.youtube.com/watch?v=b5-SvvZ7AwI) on its own. For now is suffices if you understand this:
```
RichiH: Maybe rephrase that the the details are complicated, outside of scope, you will limit yourself to what's needed here and then refer to the talk as further info?

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,448 @@
...
+ 1. `relabel_configs` contains the new relabeling rules.
+ 1. We take the values from the label `__address__` (which contain the values from `targets`) and write them to a new label `__param_target` which will add a parameter `target` to the Prometheus scrape requests.
+ 1. We take the values from the label `__param_target` and create a label instance with the values.
+ 1. We write the value `localhost:9115` to the the label `__address__`. This will be used as the hostname and port for the Prometheus scrape requests.
+
+This way we can have the actual targets there, get them as `instance` label values while letting Prometheus make a request against the blackbox exporter.
+
+Often people combine these with a specific service discovery. Check out the [configuration documentation](/docs/prometheus/latest/configuration/configuration) for more information. Using them is no problem, as these write into the `__address__` label just like `targets` defined under `static_configs`.
+
+That is it. Restart the Prometheus docker container and look at your [metrics](http://localhost:9090/graph?g0.range_input=5m&g0.stacked=0&g0.expr=probe_http_duration_seconds&g0.tab=0).
```
RichiH: Maybe a closing sentence about what differences to look for, or how to extend the time range of the graph in case they took longer to do it?

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,448 @@
...
+  relabel_configs:
+    - source_labels: [__address__]
+      target_label: __param_target
+    - source_labels: [__param_target]
+      target_label: instance
+    - target_label: __address__
+      replacement: localhost:9115  # The blackbox exporter’s real hostname:port. For Windows and macOS replace with - host.docker.internal:9115
+```
+
+So what is new compared to the [last config](#prometheus-config)?
```
RichiH: I like the below, but maybe split it in "results" and "what are the steps to get there ", this again subsplit for each of the three sections under `relabel_configs`?

content/docs/guides/multi-target-exporter.md
vonneudeck:
```suggestion
1. Querying the exporter to scrape another target. Usually available at a "descriptive" endpoint, e.g. `/probe`. This is likely what you are primarily interested in, when using multi-target exporters.
```

content/docs/guides/multi-target-exporter.md
vonneudeck:
```suggestion
The reason is that we either can’t run an exporter on the targets, e.g. network gear speaking SNMP, or that we are explicitly interested in the distance, e.g. latency and reachability of a website from a specific point outside of our network, a common use case for the [blackbox](https://github.com/prometheus/blackbox_exporter) exporter.
```

content/docs/guides/multi-target-exporter.md
vonneudeck:
```suggestion
That is it. Restart the Prometheus docker container and look at your [metrics](http://localhost:9090/graph?g0.range_input=30m&g0.stacked=0&g0.expr=probe_http_duration_seconds&g0.tab=0). Pay attention that you selected the period of time when the metrics were actually collected.
```

content/docs/guides/multi-target-exporter.md
vonneudeck: Good point.
I tried to change this using the suggestion feature, but github is confused by nested markdown code blocks within suggestions. I will change it directly. Sorry for the inconvenience.

RichiH: Looking good.

RichiH:
Thanks!

vonneudeck:
Thank you und Bitteschön!

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+
+By multi-target [exporter](/docs/instrumenting/exporters/) pattern we refer to a specific design, in which:
+
+* the exporter will get the target’s metrics via a network protocol.
+* the exporter does not have to run on the machine the metrics are taken from.
+* the exporter gets the targets and a query config string as parameters of Prometheus’ GET request.
+* the exporter subsequently starts the scrape after getting Prometheus’ GET requests and once it is done with scraping.
+* the exporter can query multiple targets.
+
+This pattern is only used for certain exporters, namely the [blackbox](https://github.com/prometheus/blackbox_exporter) and the [SNMP exporter](https://github.com/prometheus/snmp_exporter).
```
brian-brazil: s/namely/such as/

These aren't the only ones.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+* configure a custom query module for the blackbox exporter,
+* let the blackbox exporter run basic metric queries against the Prometheus [website](https://prometheus.io),
+* examine a popular pattern of configuring Prometheus to scrape exporters using relabeling.
+
+## The multi-target exporter pattern?
+
+By multi-target [exporter](/docs/instrumenting/exporters/) pattern we refer to a specific design, in which:
+
+* the exporter will get the target’s metrics via a network protocol.
+* the exporter does not have to run on the machine the metrics are taken from.
```
brian-brazil: (almost) no exporter has to run on the same machine, so this is a bit misleading.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+* the exporter subsequently starts the scrape after getting Prometheus’ GET requests and once it is done with scraping.
+* the exporter can query multiple targets.
+
+This pattern is only used for certain exporters, namely the [blackbox](https://github.com/prometheus/blackbox_exporter) and the [SNMP exporter](https://github.com/prometheus/snmp_exporter).
+
+The reason is that we either can’t run an exporter on the targets, e.g. network gear speaking SNMP, or that we are explicitly interested in the distance, e.g. latency and reachability of a website from a specific point outside of our network, a common use case for the [blackbox](https://github.com/prometheus/blackbox_exporter) exporter.
+
+## Running multi-target exporters
+
+Multi-target exporters are flexible regarding their environment and can be run in many ways. As regular programs, in containers, as background services, on baremetal, virtual machines. Because they are queried and do query over network they do need appropriate open ports. Otherwise they are frugal.
```
brian-brazil: on virtual machines.

I'm not sure this sentence structure will work for non-native speakers.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+Below you find a minimal prometheus config. It is telling Prometheus to scrape the exporter itself as we did [before](#query-exporter) using `curl 'localhost:9115/metrics'`:
+
+NOTE: If you use Docker for Mac or Docker for Windows, you can’t use `localhost:9115` in the last line, but must use `host.docker.internal:9115`. This has to do with the virtual machines used to implement Docker on those operating systems. You should not use this in production.
+
+`prometheus.yml` for Linux:
+
+```yaml
+scrape_configs:
+
+- job_name: blackbox-exporter-metrics
```
brian-brazil: This is a bit verbose, and metrics is redundant here. Just blackbox will do.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+Now run a Prometheus container and tell it to mount our config file from above. Because of the way networking on the host is addressable from the container you need to use a slightly different command on Linux than on MacOS and Windows.:
+
+<a name=run-prometheus></a>
+
+Run Prometheus on Linux (don’t use `--network="host"` in production):
+
+```bash
+docker \
+  run --network="host"\
+  --mount type=bind,source="$(pwd)"/prometheus.yml,target=/prometheus.yml,readonly \
```
brian-brazil: Did you mean $PWD?

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+scrape_configs:
+
+- job_name: blackbox-exporter-metrics
+  scrape_interval: 5s
+  metrics_path: /metrics
+  static_configs:
+    - targets:
+      - localhost:9115   # for Windows and macOS replace with - host.docker.internal:9115
+
+- job_name: 'blackbox'
```
brian-brazil: This isn't monitoring blackbox so this is misleading. blackbox_http or similar would be clearer.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+
+NOTE: If you use Docker for Mac or Docker for Windows, you can’t use `localhost:9115` in the last line, but must use `host.docker.internal:9115`. This has to do with the virtual machines used to implement Docker on those operating systems. You should not use this in production.
+
+`prometheus.yml` for Linux:
+
+```yaml
+scrape_configs:
+
+- job_name: blackbox-exporter-metrics
+  scrape_interval: 5s
```
brian-brazil: Per scrape-config intervals are generally an anti-pattern, do this up in global.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+    - source_labels: [__address__]
+      target_label: __param_target
+    - source_labels: [__param_target]
+      target_label: instance
+    - target_label: __address__
+      replacement: localhost:9115  # The blackbox exporter’s real hostname:port. For Windows and macOS replace with - host.docker.internal:9115
+```
+
+Before applying the relabeling rules, the URI of a request Prometheus would make would look like this:
+`"http://prometheus.io/probe?module=http_2xx"`. After relabeling it will look like this `"localhost:9115/probe?target=http://prometheus.io&module=http_2xx"`.
```
brian-brazil: The leading http should still be there.

content/docs/guides/multi-target-exporter.md
```diff
@@ -0,0 +1,492 @@
...
+
+Our request is now `"localhost:9115/probe?target=http://prometheus.io&module=http_2xx"`. This way we can have the actual targets there, get them as `instance` label values while letting Prometheus make a request against the blackbox exporter.
+
+Often people combine these with a specific service discovery. Check out the [configuration documentation](/docs/prometheus/latest/configuration/configuration) for more information. Using them is no problem, as these write into the `__address__` label just like `targets` defined under `static_configs`.
+
+That is it. Restart the Prometheus docker container and look at your [metrics](http://localhost:9090/graph?g0.range_input=30m&g0.stacked=0&g0.expr=probe_http_duration_seconds&g0.tab=0). Pay attention that you selected the period of time when the metrics were actually collected.
+
+# Summary
+
+In this guide, you learned how the multi-target exporter pattern works and how to run a blackbox exporter with a customised module and to configure Prometheus using relabeling to scrape metrics with prober labels.
```
brian-brazil: This is a bit run on, replace the first and with a comma.

content/docs/guides/multi-target-exporter.md
vonneudeck: No, I didn’t.
Please try to avoid passive-aggressive language. It is off-putting and not good for fostering collaboration.

content/docs/guides/multi-target-exporter.md
vonneudeck: I get why you say that, but this is to clarify that it is about the metrics of the exporter itself.
It is aimed at people who want to learn about the concept and named like that for pedagogical reasons.

content/docs/guides/multi-target-exporter.md
vonneudeck: Good catch. Namely is a false friend between German and English I was not aware of. Thank you for pointing it out.

content/docs/guides/multi-target-exporter.md
vonneudeck: I tested the wording and the guide as a whole with non-native speakers, hence I am optimistic.

content/docs/guides/multi-target-exporter.md
vonneudeck: Good Point.

content/docs/guides/multi-target-exporter.md
brian-brazil: It's redundant though, and I consider it an anti-pattern. This will be blindly copy&pasted, so we should follow best practices.

content/docs/guides/multi-target-exporter.md
brian-brazil: It's still important in terms of label taxonomy, it means something to humans and that's importnat.

content/docs/guides/multi-target-exporter.md
brian-brazil: Hmm, I didn't realise that this works.

It was an honest question.

content/docs/guides/multi-target-exporter.md
vonneudeck: True. Thanks for pointing that out.

content/docs/guides/multi-target-exporter.md
vonneudeck: Will do.

content/docs/guides/multi-target-exporter.md
vonneudeck: Will do.

content/docs/guides/multi-target-exporter.md
vonneudeck: How about `blackbox-exporter-itself`?

brian-brazil: That's still a bit redundant, I'd personally use blackbox, but could live with blackbox_exporter

content/docs/guides/multi-target-exporter.md
vonneudeck: Some do. Especially the node-exporter, which is among the first most people come in contact with. This is a guide aimed at people relatively new to Prometheus and I want to make it easy to grasp the concepts while saying the truth.

brian-brazil: This point is misleading though. That you can run e.g. the haproxy exporter on a different machine doesn't make it multi-target.

vonneudeck: All of the points have to be true. One is not sufficient. To me that was clear by "in which". But we can change that. Any ideas?

content/docs/guides/multi-target-exporter.md
vonneudeck: Redundancy is a feature in this context.

content/docs/guides/multi-target-exporter.md
vonneudeck: I am sorry Brian. I read it like a smug "please use this instead". I am very glad you did not mean it like that.
All of this works, I tested it under both, Linux and Mac OS myself and let a new to prometheus engineer try the whole guide to see how it does from a pedagogical perspective as well.

content/docs/guides/multi-target-exporter.md
vonneudeck: Yes. Exactly because it does mean something for humans I left it aside.

content/docs/guides/multi-target-exporter.md
brian-brazil: It's a bit ambiguous as written. "does not usually" maybe?

content/docs/guides/multi-target-exporter.md
brian-brazil: It's a bug, as it's extra typing/reading. We don't need every job name ending with prometheus-telemetry-metrics-job (I've seen much worse in practice). Job names should be succinct.

content/docs/guides/multi-target-exporter.md
brian-brazil: Ah no, I'd actually mis-read it as {} rather than () and thus thought that this was a error.

content/docs/guides/multi-target-exporter.md
brian-brazil: But it's not sufficiently distinguishing, you wouldn't be mixing http and tcp probes under the same job. This name to me could also realistically mean the exporter itself.

vonneudeck: I propose we discuss this further up at the other job name. Those two issues are connected anyway.

content/docs/guides/multi-target-exporter.md
vonneudeck: Yes, in a normal context it is absolutely a bug. I don’t disagree with you. And I share your frustration about redundant job names in production systems.
But this is an exercise to help people form their mind. We are trying to solve a different problem than usually. Everything is only true within a context and when it changes things can become tricky.
Maybe we can find a solution that satisfies both requirements at once? Something pedagogical and not to verbose.

We have two jobs:
* One about the metrics of the exporter itself. That must be clear to people who are new.
* One about the targets we tell the exporter to test for us. That must also be clear for new people and not implying magic. 

At the same time we want people not to form the bad habit of really long redundant job names.

How about we use a comment? It is a bit worse for understanding and less elegant, but the risk that people adopt bad habits is much lower.

```yaml
 - job_name: blackbox-exporter # to get metrics about the exporter itself 
[…]
 - job_name: blackbox # to get metrics from targets we want the exporter to query for us
```

content/docs/guides/multi-target-exporter.md
vonneudeck: I am a bit confused. Where exactly would you put the "does not usually"?

content/docs/guides/multi-target-exporter.md
vonneudeck: Resolving that just to make the view clearer. The issue will be discussed above.

content/docs/guides/multi-target-exporter.md
roidelapluie: I think you underestimate how we just all copy and past examples from the internet. I am on Brian's side on this one.

vonneudeck: Good point. I’ll go with the comments solution.

content/docs/guides/multi-target-exporter.md
brian-brazil: the exporter does not usually run on the machine

content/docs/guides/multi-target-exporter.md
brian-brazil: We can use comments reasonably liberally, the type of probe should be in the job name though as you can't put all your different probe types in one job.

PR #1210, opened 2018-11-01
Add VictoriaMetrics to remote storage integrations

valyala:
Mention that VictoriaMetrics supports direct PromQL querying.

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -47,7 +47,7 @@ data volumes.
   * [OpenTSDB](https://github.com/prometheus/prometheus/tree/master/documentation/examples/remote_storage/remote_storage_adapter): write
   * [PostgreSQL/TimescaleDB](https://github.com/timescale/prometheus-postgresql-adapter): read and write
   * [SignalFx](https://github.com/signalfx/metricproxy#prometheus): write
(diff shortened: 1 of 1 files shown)


content/docs/operating/integrations.md
```diff
@@ -47,6 +47,7 @@ data volumes.
   * [OpenTSDB](https://github.com/prometheus/prometheus/tree/master/documentation/examples/remote_storage/remote_storage_adapter): write
   * [PostgreSQL/TimescaleDB](https://github.com/timescale/prometheus-postgresql-adapter): read and write
   * [SignalFx](https://github.com/signalfx/metricproxy#prometheus): write
+  * [VictoriaMetrics](https://github.com/VictoriaMetrics/VictoriaMetrics): write and direct read via [Extended PromQL](https://github.com/VictoriaMetrics/VictoriaMetrics/wiki/ExtendedPromQL).
```
brian-brazil: Are you sure this supports remote read?

This looks like a write-only system per the docs, and what is offered is not PromQL and calling it that is confusing - and we don't list that sort of thing here anyway.

valyala: It doesn't support remote read API. Removed the part about read support to be consistent with other integrations.

brian-brazil:
Thanks!

PR #1219, opened 2018-11-13
Added CardRates.png to index.html

db-roberto:
Signed-off-by: Roberto Sanchez <roberto@digitalbrands.com>


brian-brazil:
Thanks for your PR. I believe this company was already added in #1111

db-roberto:
This PR is for CardRates.com

brian-brazil:
Both PRs appear to be from the same company, and we try to list each company only once. I'd suggest changing the existing one to Digital Brands to show that the company is using Prometheus rather than just a website.

db-roberto:
Oh, I see what you're saying. Thanks!

PR #1220, opened 2018-11-15
Add release 2.5 of Prometheus

simonpasquier:
cc @brian-brazil @brancz


brancz:
Oops, sorry I missed to do this. We have forgotten this a number of times, maybe it's time for some automation?

PR #1222, opened 2018-11-19
Add openmetrics-php/exposition-text as PHP client library

hollodotme:
## Proposed changes

This PR adds [openmetrics-php/exposition-text](https://github.com/openmetrics-php/exposition-text) as a second client library for PHP.

The library implements the text exposition format described by prometheus / openmetrics.io and ships with a [PSR-7](https://www.php-fig.org/psr/psr-7/) compatible HTTP response implementation.

The library supports PHP >= 7.1.

### DIFF
--- a/content/docs/instrumenting/clientlibs.md
+++ b/content/docs/instrumenting/clientlibs.md
@@ -31,7 +31,7 @@ Unofficial third-party client libraries:
 * [.NET / C#](https://github.com/andrasm/prometheus-net)
 * [Node.js](https://github.com/siimon/prom-client)
 * [Perl](https://metacpan.org/pod/Net::Prometheus)
-* [PHP](https://github.com/Jimdo/prometheus_client_php)
+* PHP - [Jimdo](https://github.com/Jimdo/prometheus_client_php) / [openmetrics-php](https://github.com/openmetrics-php/exposition-text)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thank you for your PR. We try to avoid duplication in this list, so I'd suggest working with the existing library.

Also, I'm not sure you're actually producing valid OpenMetrics output. You should check with the Python parser.

hollodotme:
Hi @brian-brazil, thanks for the quick feedback.

I'm not very familiar with python, can you point me to some kind of how-to use the python parser, assuming you were talking about [this parser](https://github.com/prometheus/client_python#parser), so I can check the output.

What is the reason to limit users to one library per language?

brian-brazil:
> assuming you were talking about this parser, so I can check the output.

You want the one in the openmetrics package, otherwise it's the same.

> What is the reason to limit users to one library per language?

Users can use whatever they like, however leaving it up to new users to select which of the various libraries to use would not be a good experience as they lack the understanding to know what's good and what's not. Yours for example is not a client library, as it only does exposition. I also want to encourage everyone to work together, and not create brand new clients/exporters every few months.

hollodotme:
Thanks for the pointer and clarification.
Obviously I totally misread the scope of a client lib. 

I try to work smarter next time.

PR #1227, opened 2018-11-26
group software rather than million bajillion links, added Portworx Storage

mh720:
group software rather than million bajillion links, added Portworx Storage


brian-brazil:
Could you break this into two PRs please so I can review each change indivudally?

mh720:
Hi Brian, let me know how you’d need to see the PRs. This documentation page has 2 sections which got alphabetized, a similar treatment was done on the no-exporter-needed list, and the only software addition was the Portworx line for storage.

PR #1228, opened 2018-11-26
add libvirt exporter

sunnoy:
add libvirt exporter


brian-brazil:
Thanks for your PR, this looks like a fork of the main repo. I'd prefer to list the main one, and that you work with the maintainers of it to include the changes you need so that everyone can benefit.

PR #1230, opened 2018-12-02
Add FreeRADIUS exporter

gen2brain:
This adds a `FreeRADIUS exporter` to `Miscellaneous` list.
I didn't found exporter anywhere so I wrote one, it needs to have `Status-Server` enabled.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -154,6 +154,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Dovecot exporter](https://github.com/kumina/dovecot_exporter)
    * [eBPF exporter](https://github.com/cloudflare/ebpf_exporter)
    * [Ethereum Client exporter](https://github.com/31z4/ethereum-prometheus-exporter)
+   * [FreeRADIUS exporter](https://github.com/bvantagelimited/freeradius_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thank you for sharing. This exporter takes a password on the command line which is a security issue as anyone on the machine could see it, could you please remove that?

You should grab the next port number off https://github.com/prometheus/prometheus/wiki/Default-port-allocations

Some other things that may help you improve your exporter would be adding _total to the names of all the counter metrics, removing the existing "total_" in the names as it's redundant, I sense that there may be a case for labels in these metrics, you should fail the query with NewInvalidMetric if there's an error, and I'm not sure you need that mutex.

gen2brain:
Hi, thank you for feedback.

Sure, I will add some simple json config file for now, and password can configured also via env. But this is actually useful as it allows one to have default password defined, that is `testing123`. By default freeradius installation in config have defined localhost client with secret `testing123`, that is used usually for testing. 

I did check that ports age, and saw there are non standard ports. I picked 9812 because it is similar to radius ports (i.e. 1812, 18120, 18121) but I can change that, doesn't really matter.

Not sure if I understand what you wrote about total, will need to check, but I will fix some things there and reply here when done.

Thanks,
Milan

gen2brain:
I didn't have time to work on this, but I have a few more ideas on how to improve it, to work with both version 3 and 4, and with `radmin` OR `Status-Server`. Will send another PR when I find the time.

PR #1231, opened 2018-12-03
Add acurite_exporter and smartthings_exporter

kadaan:
Addition of links to acurite_exporter and smartthings_exporter


brian-brazil:
Thanks for sharing, I believe both of these fall under hardware.

For the acurite exporter, I'd suggest fewer labels on the invalid value metrics and relying on logs to narrow things down. Don't forget units on your metrics.

For the smartthings_exporter it's taking a secret on the command line, which is not secure. It should preferably come from a file. "Energy usage in kilowatts" - power is measured in kilowatts, not energy. Base units such as ratios, Celsius and joules/watts are preferred.

kadaan:
Thanks for the feedback.

For the acurite_exporter, I will make those changes  

For smartthings_exporter, a couple questions. 

1. The secret is only needed for the first call. Once it is done, you don’t need to do it again because the oauth token is cached in a file. Would it be better add a command “register” to the cli, which performs the oauth step, gets the token, and prints it to a user specified file. Then require you to provide the oauth token to the normal run as a file path.

2. I’ll fix up the units, but when they aren’t provided by the underlying system in base units should the exporter concert them to the base unit?  For instance the kWh one is provided from smartthings at that granularity.

kadaan:
Also, the energy/power units are defined in the smartthings spec backwards. Should the exporter correct them?

See “Energy Meter” and “Power Meter” on
https://docs.smartthings.com/en/latest/capabilities-reference.html

brian-brazil:
> The secret is only needed for the first call. Once it is done, you don’t need to do it again because the oauth token is cached in a file. 

The key thing is that secrets shouldn't be on the command line. Beyond that how you handle it is largely up to you.

> I’ll fix up the units, but when they aren’t provided by the underlying system in base units should the exporter concert them to the base unit? 

Yes, where possible.

> Also, the energy/power units are defined in the smartthings spec backwards. Should the exporter correct them?

Those look right to me in that spec.

kadaan:
Regarding the spec, I see now.  I just left the hour off the unit.  Should it be watt_hour or watthour?

kadaan:
What suffix would your recommend for metrics that simply indicate state.  For example, motion and presence are 1 when they are detected and 0 otherwise.

brian-brazil:
There's no specific suffix. Here I'd use something like _detected.

kadaan:
@brian-brazil I've seen that you don't recommend the constant metrics with timestamp approach, except in certain circumstances.  It feels like with acurite, that is the correct circumstance.  The acurite  access or smarthub device will post metrics to atlasapi.myacurite.com every 5 minutes.  On the local network atlasapi.myacurite.com resolves to the server running acurite_exporter.  So, seems like the acurite_exporter needs to keep the samples around for eventual scraping from prometheus.  Does this seem right?

kadaan:
Seem good now?
```
...
# HELP smartthings_alarm_cleared 0 if the alarm is clear.
# TYPE smartthings_alarm_cleared gauge
smartthings_alarm_cleared{id="",name="Workshop Smoke Detector"} 0
# HELP smartthings_battery_percentage Percentage of battery remaining.
# TYPE smartthings_battery_percentage gauge
smartthings_battery_percentage{id="",name="Workshop Smoke Detector"} 94
# HELP smartthings_contact_closed 1 if the contact is closed.
# TYPE smartthings_contact_closed gauge
smartthings_contact_closed{id="",name="Workshop Smoke Detector"} 0
# HELP smartthings_energy_usage_joules Energy usage in joules.
# TYPE smartthings_energy_usage_joules gauge
smartthings_energy_usage_joules{id="",name="Heater, Workshop"} 3.532301728438278e+08
# HELP smartthings_exporter_build_info A metric with a constant '1' value labeled by version, revision, branch, and goversion from which smartthings_exporter was built.
# TYPE smartthings_exporter_build_info gauge
smartthings_exporter_build_info{branch="",goversion="go1.11.2",revision="",version=""} 1
# HELP smartthings_invalid_metric Total number of metrics that were invalid.
# TYPE smartthings_invalid_metric counter
smartthings_invalid_metric 0
# HELP smartthings_power_usage_watts Current power usage in watts.
# TYPE smartthings_power_usage_watts gauge
smartthings_power_usage_watts{id="",name="Heater, Workshop"} 1350.7
# HELP smartthings_presence_detected 1 if presence is detected.
# TYPE smartthings_presence_detected gauge
smartthings_presence_detected{id="",name="iPhone"} 0
# HELP smartthings_smoke_detected 1 if smoke is detected.
# TYPE smartthings_smoke_detected gauge
smartthings_smoke_detected{id="",name="Workshop Smoke Detector"} 0
# HELP smartthings_switch_enabled 1 if the switch is on.
# TYPE smartthings_switch_enabled gauge
smartthings_switch_enabled{id="",name="Heater, Workshop"} 1
```

kadaan:
Made all the changes you suggested and released v0.0.1 of both.

brian-brazil:
> So, seems like the acurite_exporter needs to keep the samples around for eventual scraping from prometheus. Does this seem right?

This is a case where timestamps may make sense, however as the data is more than 5 minutes old it'd cause staleness issues for most users. I'd not use timestamps in this case.

You're still taking a password on the command line.

kadaan:
Yes, that password is part of a registration step needed to complete OAuth.  I guess we could prompt for the password (maybe using something like (Terminal.ReadPassword)[https://godoc.org/golang.org/x/crypto/ssh/terminal#Terminal.ReadPassword]), but it would need to be cross platform.  During the normal execution of the exporter, there is no password

Acurite SmartHub/Access only posts data every 5 minutes.  Can you explain how the data will cause staleness.  Also, what are the issues?

SuperQ:
For timestamps data, I tend to expose those directly as a metric. For example if you have the timestamp of of the last alarm event, you could expose this:

```
smarthings_last_alarm_seconds X
```

This way you can do things like `time() - smarthings_last_alarm_seconds` for the time since last alarm. Or `changes(smarthings_last_alarm_seconds[28d])` to find out how many alarms have happened in the last 4 weeks.

One other suggestion is to make the alarm state a "positive" bool. `smartthings_alarm_cleared` -> `smartthings_alarm_present`.

As for the password, the recommended method is to read it from a file.

brian-brazil:
> Acurite SmartHub/Access only posts data every 5 minutes. Can you explain how the data will cause staleness. Also, what are the issues?

If you exposed the true timestamp, once the data was 5 minutes old then it'd no longer be returned by instant queries.

kadaan:
Thanks for the feedback. Unfortunately I thjnk I only have whether it is alarmed or not, not when it happened. 

I like the idea of using _present. How would you do it if there are 3+ states (smoke detector can be clear, detected, or tested).

My thinking is that at exporter config time the password would be prompted for. The config is interactive. The resultant token needs to be provided to the exporter during its standard run as a file.

kadaan:
@brian-brazil. Got it. So with the acurite posted time stamp there may be an edge case where there are no instant metrics. In stat case, it will probably be fine to have the exported remove the time stamp from the metric, but keep it for ageist off the sample.

SuperQ:
> I like the idea of using _present. How would you do it if there are 3+ states (smoke detector can be clear, detected, or tested).

Depends on those states, can it be both "clear" and "tested" simultaneously?
If it comes back as one of 3, I would probabl make an ENUM.

```
smarthings_smoke_detector_status{state="clear"} 1
smarthings_smoke_detector_status{state="detected"} 0
smarthings_smoke_detector_status{state="tested"} 0
```

brian-brazil:
If we're following openmetrics, that'd be:

```
smarthings_smoke_detector_status{smarthings_smoke_detector_status="clear"} 1
smarthings_smoke_detector_status{smarthings_smoke_detector_status="detected"} 0
smarthings_smoke_detector_status{smarthings_smoke_detector_status="tested"} 0
```

kadaan:
That’s great!

—
Joel Baranick
jbaranick@gmail.com

________________________________

kadaan:
@brian-brazil Does the exporter need to know about all the status values for the smarthings_smoke_detector_status metrics  and when it pulls the value `clear` from Smartthings then explicitly set the `smarthings_smoke_detector_status{status="clear"}==1` and the others to 0?

What if you don't know all the status values up front, or don't want to require them to be coded into the exporter?

brian-brazil:
> Does the exporter need to know about all the status values

It should.

> What if you don't know all the status values up front, or don't want to require them to be coded into the exporter?

Then you're best leaving it as an integer. I imagine this list won't change though.

gouthamve:
@brian-brazil/@superQ as far as I can see the token is being read from a file now. 

There is a step to generate the secrets file which requires a different secret to be passed in the CLI. As it is just the generation step and not involved in the actual running of the exporter, can we merge this in?

brian-brazil:
Passing secrets on the command line is passing secrets on the command line. The code shouldn't allow it, and the docs shouldn't mention it.

The entries also need to be sorted.

jan--f:
Hello from the bug scrub!
I'll close this for now. Seems like the proposed exporters aren't very active anymore.

PR #1232, opened 2018-12-05
add link to twemproxy_exporter

stuartnelson3:
twemproxy (https://github.com/twitter/twemproxy) proxies requests for memcached or redis pools.


brian-brazil:
Can you add the DCO?

That JSON unmarshalling is a bit odd. server_connections_active would be clearer than active_server_connections.

stuartnelson3:
@brian-brazil do you want to open issues against the repo? it's still being worked on and I would welcome the input. There's also the issue of their README being incomplete and describing every metric as a counter, even though some of them are definitely gauges.

stuartnelson3:
>That JSON unmarshalling is a bit odd.

Yes, definitely. I would very much like it to not be this weird unmarshal-remarshal-unmarshal dance.

PR #1233, opened 2018-12-07
Add Wavefront as a remote write

puckpuck:
@brian-brazil can you please take a look?  This adapter is already deployed at multiple locations in production today.


brian-brazil:
Can you add the DCO?

Looking at your code, note that there's no requirement that timeseries from Prometheus have an instance label which would result in an empty source - which I believe Wavefront will reject. Also, underscore is a valid character in Wavefront metrics names, so there's no need to sanitize it.

puckpuck:
DCO done.

Empty source does not get rejected by Wavefront, it will get set to "unknown" if missing.
Wavefront's UI creates hierarchies from metrics based on the `.`. Converting a `_` to a `.` of the metric name to take advantage of the UI benefits is this is the typical use case for Wavefront.

brian-brazil:
Thanks!

PR #1238, opened 2018-12-13
Add exporter for App Connect Enterprise

MartinRoss-IBM:
App Connect Enterprise is an integration product (formally known as IBM Integration Bus or WebSphere Message Broker)

The exporter is available at https://github.com/ot4i/ace-docker and further work is planned to de-couple from this repository so it can be more readily included in projects

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -66,6 +66,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Ubiquiti UniFi exporter](https://github.com/mdlayher/unifi_exporter)
 
 ### Messaging systems
+   * [App Connect Enterprise exporter](https://github.com/ot4i/ace-docker)
    * [Beanstalkd exporter](https://github.com/messagebird/beanstalkd_exporter)
    * [Gearman exporter](https://github.com/bakins/gearman-exporter)
    * [Kafka exporter](https://github.com/danielqsj/kafka_exporter)
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -66,7 +66,6 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Ubiquiti UniFi exporter](https://github.com/mdlayher/unifi_exporter)
 
 ### Messaging systems
-   * [App Connect Enterprise exporter](https://github.com/ot4i/ace-docker)
    * [Beanstalkd exporter](https://github.com/messagebird/beanstalkd_exporter)
    * [Gearman exporter](https://github.com/bakins/gearman-exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Could you point me at the source code please? That's asking for a login.

MartinRoss-IBM:
Apologies @brian-brazil - I have updated the link in the description:
https://github.com/ot4i/ace-docker/tree/master/internal/metrics

brian-brazil:
This sounds like more a piece of software that exposes Prometheus metrics, rather than a typical exporter. Those are listed further down the page.

You need to add a DCO.

I'd suggest using ConstMetrics rather than normal gauges/counters, you'll find them a lot easier to work with and you should be able to remove all the channel stuff.

MartinRoss-IBM:
The software gets built into the linked Docker image and can run in isolation - it exposes metrics over a websocket... There is a separate exporter included in the project linked that listens on the web socket and make the metrics available to be scraped. We should be separating it out into a separate repository at some point, but it is separate... The intention is that users can use the example ace-docker repo which bundles the software and exporter (and other things) into the image, or they can build their own Docker image from the software (ie. https://github.com/ot4i/ace-docker#build-an-image-with-app-connect-enterprise-only) and then should be able to use the exporter as well... I hope that makes sense... Happy to move it down if you feel that is where you believe it should be listed - but it's intention is to be separate...

MartinRoss-IBM:
I will sort out the DCO over the weekend and look into the ConstMetrics - thank you for the pointer

MartinRoss-IBM:
I was going to move the new entry down to the `Software exposing Prometheus metrics` section and then once we have separated it into a new repo send another PR to move it back up, but the description in the lower section is a little misleading for our case:
"Some third-party software exposes metrics in the Prometheus format, so no
separate exporters are needed"

The software is "App Connect Enterprise" (http://www-01.ibm.com/common/ssi/ShowDoc.wss?docURL=/common/ssi/rep_ca/3/877/ENUSZP18-0043/index.html&request_locale=en) and this provides metrics via a webhook. There is an exporter in the https://github.com/ot4i/ace-docker repository that listens on that webhook and exposes metrics in the Prometheus format. It could be said that the code in "ace-docker" is software that we have open sourced, but that is not the software the exporter is for - which could be misleading, just a sample to show how to package that software (App Connect Enterprise) into a container (that users may use as-is or modify for their requirements).

What would you suggest @brian-brazil ?

brian-brazil:
It's a bit odd to have an exporter that's just an aside of another process, and reading those docs it's not clear to me how I'd just run it as an exporter. I'd say put it in the other section for now, to avoid confusion.

MartinRoss-IBM:
Have moved the link down to the `Software exposing Prometheus metrics` section @brian-brazil

brian-brazil:
Thanks!

PR #1241, opened 2018-12-17
Cleanup old releases

SuperQ:
Cleanup docs for Prometheus 2.0 -> 2.2.

### DIFF
--- a/nanoc.yaml
+++ b/nanoc.yaml
@@ -113,36 +113,6 @@ data_sources:
   -
     type: repo_docs
     identifier_type: legacy
-    items_root: /docs/prometheus/2.2/
-    config:
-      name: '2.2'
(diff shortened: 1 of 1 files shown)


simonpasquier:
What is the gain from removing those outdated versions? Do we have an established policy of removing outdated versions (eg keep documentation for the x last releases)?

brian-brazil:
We don't have any such policy yet, this is all still (relatively) new for us.

RichiH:
Unless we are running into actual issues, I would like us to keep docs around longer rather than shorter.

SuperQ:
I was just looking to clean up clutter. I went for this list because 2.0-2.2 are known to be very buggy and we don't support them.

beorn7:
I'd also like to keep old documentation for reference.

PR #1250, opened 2018-12-31
Add JavaMelody in Software exposing Prometheus metrics

evernat:
JavaMelody has native integration of Prometheus for some time now. Link: https://github.com/javamelody/javamelody/wiki/UserGuideAdvanced#exposing-metrics-to-prometheus

A sample of output is available at http://javamelody.org/demo/monitoring?format=prometheus

Signed-off-by: Emeric Vernat <evernat@free.fr>


brian-brazil:
Can you add the DCO?

It's recommended to use the Java client, rather than marshalling the format by hand. For example all valid UTF-8 is allowed in label values, so you don't need sanitizeName (but you do need some escaping). I'd also suggest reading https://prometheus.io/docs/instrumenting/writing_exporters/, as there's a few ways you can improve these for your users such as using base units.

evernat:
Sorry for DCO. I misunderstood its syntax.
Created another PR #1251

PR #1251, opened 2018-12-31
Add JavaMelody in Software exposing Prometheus metrics

evernat:
JavaMelody has native integration of Prometheus for some time now. Link: https://github.com/javamelody/javamelody/wiki/UserGuideAdvanced#exposing-metrics-to-prometheus

A sample of output is available at http://javamelody.org/demo/monitoring?format=prometheus

Signed-off-by: evernat <evernat@free.fr>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -202,6 +202,7 @@ separate exporters are needed:
    * [Flink](https://github.com/apache/flink)
    * [FreeBSD Kernel](https://www.freebsd.org/cgi/man.cgi?query=prometheus_sysctl_exporter&apropos=0&sektion=8&manpath=FreeBSD+12-current&arch=default&format=html)
    * [Grafana](http://docs.grafana.org/administration/metrics/)
+   * [JavaMelody](https://github.com/javamelody/javamelody/wiki/UserGuideAdvanced#exposing-metrics-to-prometheus)
    * [Kubernetes](https://github.com/kubernetes/kubernetes) (**direct**)
    * [Linkerd](https://github.com/BuoyantIO/linkerd)
    * [mgmt](https://github.com/purpleidea/mgmt/blob/master/docs/prometheus.md)


evernat:
in previous [PR](https://github.com/prometheus/docs/pull/1250#issuecomment-450647922), @brian-brazil said:

> Can you add the DCO?
>
>It's recommended to use the Java client, rather than marshalling the format by hand. For example all >valid UTF-8 is allowed in label values, so you don't need sanitizeName (but you do need some escaping). >I'd also suggest reading https://prometheus.io/docs/instrumenting/writing_exporters/, as there's a few >ways you can improve these for your users such as using base units.

evernat:
> It's recommended to use the Java client, rather than marshalling the format by hand. For example all valid UTF-8 is allowed in label values, so you don't need sanitizeName (but you do need some escaping).

Given that javamelody is a monitoring tool used by a lot of users which are often not using Promotheus, I prefer to not add Promotheus java client as a dependency. But the output result is certainly the one wanted now (after fixing use of `java.io.PrintWriter.println()` which was printing `\r\n` on Windows and which is not accepted by Prometheus).

> I'd also suggest reading https://prometheus.io/docs/instrumenting/writing_exporters/, as there's a few ways you can improve these for your users such as using base units.

I did not write the original code for that output, but I may not change that, because javamelody is a monitoring tool with its own reports and it may be better to keep javamelody units for better understanding and comparison between reports in Prometheus and reports in javamelody, not to mention breaking usages of javamelody integration with Prometheus since 15 months.

evernat:
> The dependency is quite light (by design), and you can use the bits that just do the marshalling and JVM stats.

Being a non-intrusive library in webapps, javamelody is very strict on the no-required dependencies rule (except one for RRD graphs). For example, there are no required dependencies such as commons-io, commons-lang, spring, jpa, ejb, jsf, struts, quartz 1, quartz 2, ehcache 1.2, ehcache 1.6, ehcache 2.7, statsd/aws/datadog/graphite/influxdb etc, which would be too many and which would cause version conflicts with the users' dependencies some time or another.

> how is javamelody_log_duration_millis negative?

Good question. It should be excluded from the Prometheus data, because it is not a real value for "log". I will fix that asap. Thanks.

PR #1252, opened 2019-01-05
Add SNMP notifier as Alertmanager Webhook.

maxwo:
@brian-brazil 
Hi,
I had to use a similar webhook to send Prometheus' Alertmanager alerts to a legacy SNMP notification system. It runs well on a production system, and here is an open-source version.
It may be useful to someone else.
Cheers.


brian-brazil:
Thanks for sharing. This is taking a secret (the community) on the command line which isn't secure, could you fix that?

I'd also suggest reconsidering the default grouping, the alerts already come grouped from the Alertmanager so I'd rely on that rather than having two layers of grouping.

maxwo:
Indeed clever suggestion about using the Alertmanager grouped labels instead of implementing another.
Same for using an environment variable for the community.
I made the changes, with a new release.

brian-brazil:
Thanks!

PR #1253, opened 2019-01-06
Fix typo in operating/security

pharaujo:
Changed `aribtrary` to `arbitrary`.


RichiH:
Thanks! Can you please add a DCO?

pharaujo:
@RichiH yeah, just did that. I did the change from the web UI as it was such a minor change. You were faster than the time it took to clone and amend the repo! 😄

RichiH:
Well, we aim to please :)

PR #1265, opened 2019-01-21
[docs/instrumenting] Add the Openstack Exporter.

niedbalski:
This commits adds the Linaro OpenStack exporter [0]
to the list of misc. exporters.

[0] https://github.com/Linaro/openstack-exporter

Signed-off-by: Jorge Niedbalski <jorge.niedbalski@linaro.org>


brian-brazil:
Thanks!

A few suggestions that may help you improve the exporter:
I'd split openstack_neutron_agent_state into two boolean metrics, one for up/down and another for admin status. Similarly openstack_nova_service_state.

openstack_nova_local_gb. Is this local storage? Bytes are also the standard unit to use when exporting metrics.

I don't think AddMetrics/GetMetrics is buying you much, I'd suggest using only local variables within each file.

PR #1268, opened 2019-01-24
Replace "goroutines" with "concurrent requests" in gauge example

Daniel15:
Not everyone reading this paragraph will know what a "goroutine" is. Since the previous paragraph mentions "you can use a counter to represent the number of requests served", I updated this one to also use requests as an example.


Daniel15:
Sorry about that! I usually do my minor edits through the GitHub web UI, which doesn't have an option to automatically 'sign off' on the commit.

brian-brazil:
Thanks!

No problem, it's a little annoying that the DCO gets in the way of quick fixes like this.

PR #1270, opened 2019-01-29
container_cpu_usage_seconds_total is collected as total cpu usage, not per-cpu

bysnupy:
* Description:
  Eliminate `(split up by core)`, because `container_cpu_usage_seconds_total` are collecting the cpu usage as total from the `cadvisor` >=`v0.29` unlike before `per-cpu`.

* Refer here: [Add Prometheus total CPU metric and allow per-cpu stats to be disabled](https://github.com/google/cadvisor/pull/1852#issuecomment-356361780)

### DIFF
--- a/content/docs/guides/cadvisor.md
+++ b/content/docs/guides/cadvisor.md
@@ -115,7 +115,7 @@ The table below lists some other example expressions
 
 Expression | Description | For
 :----------|:------------|:---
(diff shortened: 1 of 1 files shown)


brian-brazil:
That appears to be configurable, and the way it works is also quite odd. Maybe switch to using a different metric?

bboreham:
It's configurable in cAdvisor but many people will be using cAdvisor inside kubelet where you can't configure this setting.

bysnupy:
Thank you for your review @brian-brazil

@bboreham I can understand because of your comment.

My point is `per-cpu` is confusing in typical use cases, It does not mean not to be configurable as eliminating `(split up by core)`.

brian-brazil:
But what you're proposing is also confusing, and a label="total" is not something we should encourage or expose to users. Thus I suggest avoiding this and using a different example.

bysnupy:
@brian-brazil No problem, my purpose is enhancement :)

brian-brazil:
Thanks!

PR #1272, opened 2019-01-31
[docs/instrumenting] Add EMQ exporter

HagaiBarel:
This commit add [EMQ exporter](https://github.com/nuvo/emq_exporter) to the list of messaging systems exporters

Signed-off-by: Hagai Barel <hagai.barel@nuvo-group.com>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -68,6 +68,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
 
 ### Messaging systems
    * [Beanstalkd exporter](https://github.com/messagebird/beanstalkd_exporter)
+   * [EMQ exporter](https://github.com/nuvo/emq_exporter)
    * [Gearman exporter](https://github.com/bakins/gearman-exporter)
    * [Kafka exporter](https://github.com/danielqsj/kafka_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for sharing. This exporter is accepting passwords on the command line, which is not secure. Can you fix that?

I also noticed a few races. You should have scrape going directly to MustNewConstMetric, and remove e.metrics to fix those.

brian-brazil:
Whatever is not command line arguments. Files are best, but envvars are an option.

In Collect up is shared between scrapes, so might get things wrong with concurrent scrapes. Similarly with e.metrics.

HagaiBarel:
Alright, I'll work on both your points. 
Thanks!

brian-brazil:
Thanks!

HagaiBarel:
@brian-brazil Thanks for the merge!

Quick follow up to the points you raised - 
* I've deprecated the ability to pass credentials via command line, only via a file or `env var`
* I've introduced a `mutex` to handle concurrency better, refactored a lot of the code and added unit tests to detect potential data races

Thanks for the pointers, It really helped me to better rewrite things

brian-brazil:
> I've introduced a mutex to handle concurrency better, refactored a lot of the code and added unit tests to detect potential data races

When implemented in the normal way with ConstMetrics, you shouldn't need a mutex.

PR #1275, opened 2019-02-02
Add Netgear consumer router exporter

DRuggeri:
Signed-off-by: Daniel Ruggeri <DRuggeri@primary.net>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -65,6 +65,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [NVIDIA GPU exporter](https://github.com/mindprince/nvidia_gpu_prometheus_exporter)
    * [ProSAFE exporter](https://github.com/dalance/prosafe_exporter)
    * [Ubiquiti UniFi exporter](https://github.com/mdlayher/unifi_exporter)
+   * [Netgear router exporter](https://github.com/DRuggeri/netgear_exporter)
 
 ### Messaging systems
    * [Beanstalkd exporter](https://github.com/messagebird/beanstalkd_exporter)
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -61,11 +61,11 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [IPMI exporter](https://github.com/soundcloud/ipmi_exporter)
    * [knxd exporter](https://github.com/RichiH/knxd_exporter)
    * [Netgear Cable Modem Exporter](https://github.com/ickymettle/netgear_cm_exporter)
+   * [Netgear Router exporter](https://github.com/DRuggeri/netgear_exporter)
    * [Node/system metrics exporter](https://github.com/prometheus/node_exporter) (**official**)
    * [NVIDIA GPU exporter](https://github.com/mindprince/nvidia_gpu_prometheus_exporter)
    * [ProSAFE exporter](https://github.com/dalance/prosafe_exporter)
    * [Ubiquiti UniFi exporter](https://github.com/mdlayher/unifi_exporter)
-   * [Netgear router exporter](https://github.com/DRuggeri/netgear_exporter)
 
 ### Messaging systems
    * [Beanstalkd exporter](https://github.com/messagebird/beanstalkd_exporter)


content/docs/instrumenting/exporters.md
```diff
@@ -65,6 +65,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [NVIDIA GPU exporter](https://github.com/mindprince/nvidia_gpu_prometheus_exporter)
    * [ProSAFE exporter](https://github.com/dalance/prosafe_exporter)
    * [Ubiquiti UniFi exporter](https://github.com/mdlayher/unifi_exporter)
+   * [Netgear router exporter](https://github.com/DRuggeri/netgear_exporter)
```

DRuggeri:
Thanks, @brian-brazil for such a thorough review of my hobbyist code! Here are a few comments in response:

 > How does this compare to the cable modem exporter already listed? Could they be combined?
They gather data from very different sources and export very different metrics. The cable modem exporter scrapes HTML data from the administrative UI of a select set of models and exports data specific to the functions of a cable modem only.

This exporter interfaces with a go client implementation of the standard SOAP API that Netgear consumer routers offer. The SOAP API is supported by all models of gear supported by Netgear's Genie app.


> The exporter is taking a password on the command line, which is insecure. Can you remove that?

I opted to follow the same pattern used by the  [node_exporter](https://github.com/prometheus/node_exporter) and [cf_exporter](https://github.com/bosh-prometheus/cf_exporter) projects: Use [kingpin](https://github.com/alecthomas/kingpin) to parse command line options with the option to provide the same data via environment variables. This gives the user the ability to choose how to provide the credential. I hope that helps.

> In Collect you're maintaining gauges in a way that can't handle concurrent scrapes. It's easier and more correct to expose whatever the device returns back (adjusting units if needed) via ConstMetrics.

Totally agreed. The challenge lies with the way the SOAP API exposes the metrics. I've [updated the documentation around this metric](https://github.com/DRuggeri/netgear_exporter#TrafficDelta) with an explanation making it clear what the shortcomings are of this collector. I've also implemented a new collector as you suggest to publish "raw data" from the API [here](https://github.com/DRuggeri/netgear_exporter#Traffic).

content/docs/instrumenting/exporters.md
DRuggeri: Roger that - fixed in 48ec9debdce68db10c7f297a410f0c4eb84b8e89

brian-brazil:
> The SOAP API is supported by all models of gear supported by Netgear's Genie app.

So is this exporter a superset of the other one?

> This gives the user the ability to choose how to provide the credential. I hope that helps.

Taking a credential on the command line should not be offered as an option, 

> If you want to know just the number of bytes in/out, you have to calculate it yourself since the stats are incrementing counters that reset daily.

Prometheus will have no problem calculating that.

DRuggeri:
> So is this exporter a superset of the other one?

Negative - they share nothing in common except for the brand on the front of the device. Think: Samsung refrigerator versus Samsung television.

> Taking a credential on the command line should not be offered as an option,

I disagree with taking user choice away, but have removed this citing the mandate from the Prometheus project: [130dec583c4e4a2c3a93b15a982ebe1637bc7129](https://github.com/DRuggeri/netgear_exporter/commit/130dec583c4e4a2c3a93b15a982ebe1637bc7129)

> Prometheus will have no problem calculating that.

This is great to hear, but I couldn't figure out how to do so with the time series that increments until it resets at midnight... what did you have in mind? Something like `clamp_min(idelta(netgear_traffic_todaydownload), 0)`?

SuperQ:
> I disagree with taking user choice away

While I agree that it's bad to take choices away, we do like to follow the principal of "one obvious way to do things", as well as be secure by default. In addition, credential files and ENV vars are the least-unsafe, while command line flag is obviously unsafe.

SuperQ:
Looking over the metrics you're exposing, I see a lot of redundant data.

In reality, I think you could get away with dropping a all of the Today, Yesterday, Week, and Last month data. You really only need to collect these things:

* MonthConnectionTime
* MonthDownload
* MonthUpload

Everything can be calculated in PromQL from there.

Examples:
* TodayUpload = `increase(netgear_traffic_monthupload[1d])`
* TodayDownloadAverage = `rate(netgear_traffic_monthupload[1d])`

Of course, the above examples are for "the last day" and not "since midnight", but you can adjust the range vector in whatever UI you use. Grafana can easily do "Today" with `$__range`.

brian-brazil:
Thanks!

> we do like to follow the principal of "one obvious way to do things", as well as be secure by default.

I'll add that this is our internal policy, so it's not required for exporters to follow it. Security issues are one of the few things that will block addition to this list though.

PR #1276, opened 2019-02-04
operating/security: Remove mention of public security vul reports

mxinden:
Instead giving people the choice of either reporting security
vulnerabilities publicly via the issue trackers, or privately to the
maintainers of the relevant project, this patch removes the former,
giving us more time to mitigate the problem.

What are your thoughts?

### DIFF
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -13,9 +13,8 @@ This page describes the general security assumptions of Prometheus and the
 attack vectors that some configurations may enable.
 
 As with any complex systems it is not possible to guarantee that there are no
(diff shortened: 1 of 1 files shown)


mxinden:
This came up, looking into https://github.com/prometheus/alertmanager/issues/1734.

brian-brazil:
:+1:

simonpasquier:
Definitely yes!

SuperQ:
:+1:

RichiH:
Makes sense.

I do wonder if we should ask them to CC -team in case maintainer[s] are on holiday, etc.

brancz:
Yes I was thinking the same, it ends up being forwarded to team anyways.

PR #1279, opened 2019-02-04
Update security.md

RichiH:


beorn7:
But you need the DCO, of course.

PR #1280, opened 2019-02-05
added micrometer prometheus registry

gysel:


gysel:
@brian-brazil can you please merge this PR? Thanks!

brian-brazil:
Thanks for your PR. We only list one client library per language/runtime, and there's already a Java client which follows our best practices.

gysel:
Micrometer is the default metrics library for all Spring Boot projects, I would suggest it makes sense to mention it in the documentation.

Furthermore there are currently already 3 JVM libs listed...?

brian-brazil:
Those are collectors, not client libraries.

gysel:
I quite the introduction to that section:

> They are not Prometheus client libraries themselves but make use of 
> one of the normal Prometheus client libraries under the hood. 

That is exactly what Micrometer is. A generic metrics library for the JVM that collects data and offers a Prometheus endpoint.

Please instruct me if you want me to change the wording in my change or proceed and close the PR.

Thanks for considering it.

brian-brazil:
If it goes anywhere, it'd be in the section above.

gysel:
> it'd be in the section above.

You're referring to "Software exposing Prometheus metrics"? That's where I as a reader would certainly not expect it.

Micrometer is a library, not a software.

You could argue that Spring Boot could be mentioned in that section as it can export Prometheus metrics using Micrometer. But Micrometer is also an independent library that has nothing to do with Spring. I've just migrated a large enterprise application to export Prometheus metrics using Micrometer and we're not using Spring.

Again, thanks for maintaining this documentation. I find it very helpful and would like to contribute where I see a need. But you'll need to give me clearer advice.

brian-brazil:
> You're referring to "Software exposing Prometheus metrics"?

Yes.

fitzoh:
If Micrometer belongs in the `Software exposing Prometheus metrics`, doesn't that imply that `go-metrics` belongs in the `Software exposing Prometheus metrics` section as well?

Micrometer and go-metrics perform the same function, they are libraries that you can use to instrument your code and export metrics to different metrics backends.

SuperQ:
I think this makes sense where it is, it's in the same class as go-metrics.

PR #1281, opened 2019-02-07
Add Oracle Database Exporter

ricardona:


brian-brazil:
Thanks for your PR. In order to make things easy for users, we try to only list one exporter for a given thing and we already have an Oracle DB exporter listed. I'd suggest that if there's something missing from the current one, to work with the author(s) to improve it.

ricardona:
Thanks for the clarification and understand it. But, Oracle Database Exporter is a better option to monitor the Oracle database.

PR #1288, opened 2019-02-11
Update exporters.md: mention performance co-pilot

fche:


content/docs/instrumenting/exporters.md
```diff
@@ -214,6 +214,7 @@ separate exporters are needed:
    * [midonet-kubernetes](https://github.com/midonet/midonet-kubernetes) (**direct**)
    * [Minio](https://github.com/minio/minio)
    * [Netdata](https://github.com/firehol/netdata)
+   * [PCP Performance Co-Pilot](https://pcp.io/man/man3/pmwebapi.3.html) (also includes [prometheus scraping](https://pcp.io/man/man1/pmdaprometheus.1.html))
```
brian-brazil: This should only mention the exposition.

fche:
The name mapping is that way (using colons) in order to make the name transformation reversible.  Underscores are already used.

brian-brazil:
Usually the goal of an exporter is to get the data into a good format for Prometheus, not to be suitable for another monitoring system.

That build failure looks spurious, and I can't get it to re-run. Can you force push a null change?

brian-brazil:
Hmm, this change itself is fine however the linked docs say that this supports OpenMetrics - which it doesn't. The Prometheus text format and OpenMetrics are slightly different things.

fche:
> Hmm, this change itself is fine however the linked docs say that this supports OpenMetrics - which it doesn't. The Prometheus text format and OpenMetrics are slightly different things.

You're right - I don't know why the linked docs would say that.  It didn't say that back when I wrote the code.  Perhaps @goodwinos can advise.

gouthamve:
Is there any update on this? I think this is a really good integration but we should remove the Openmetrics reference. Is there a PR I could submit to remove it?

fche:
commits like the following suggest a deliberate terminology switch that is unlikely to be reconsidered

https://github.com/performancecopilot/pcp/commit/48e4cf3ae79b732ca39134cecefadc0f0f6cf1fd

brian-brazil:
I wouldn't worry about internals naming like that, long term that could will (hopefully) be more openmetrics than prometheus. The confusion caused by the current user-facing docs is the concern here.

jan--f:
Hello from the bug scrub!
Closing this for now as the proposed link is a 404 now.

PR #1289, opened 2019-02-12
Add JMeter exporter to list

johrstrom:
@brian-brazil for your review. I notice you look at the repositories of the exporters, so I'll add this info that you may be concerned with.

* I publish a shaded (fat) jar for users with a lot of the io.prometheus java_client libs within it for convenience. I build the jar with your LICENSE and NOTICE, but [it's not actually copyrighted](https://github.com/prometheus/client_java/blob/master/LICENSE#L189) to anyone.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -161,7 +161,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Ethereum Client exporter](https://github.com/31z4/ethereum-prometheus-exporter)
    * [Jenkins exporter](https://github.com/lovoo/jenkins_exporter)
    * [JIRA exporter](https://github.com/AndreyVMarkelov/jira-prometheus-exporter)
-   * [JMeter exporter](https://github.com/johrstrom/jmeter-prometheus-plugin)
+   * [JMeter plugin](https://github.com/johrstrom/jmeter-prometheus-plugin)
    * [Kannel exporter](https://github.com/apostvav/kannel_exporter)
    * [Kemp LoadBalancer exporter](https://github.com/giantswarm/prometheus-kemp-exporter)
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/exporters.md
```diff
@@ -161,6 +161,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Ethereum Client exporter](https://github.com/31z4/ethereum-prometheus-exporter)
    * [Jenkins exporter](https://github.com/lovoo/jenkins_exporter)
    * [JIRA exporter](https://github.com/AndreyVMarkelov/jira-prometheus-exporter)
+   * [JMeter exporter](https://github.com/johrstrom/jmeter-prometheus-plugin)
```
brian-brazil: JMeter Plugin sounds more accurate to me

brian-brazil:
Thanks for sharing. 

In BaseCollectorConfig why do you have random metric names?

In ThreadCollector I'd recommend using a GaugeMetricFamily, and those metrics would be better separate rather than as labels.

In the README, the convention is not to expose metric names with colons so it'd be good to remove that from examples. You've also a typo "destoryed"

johrstrom:
Thanks for the feedback.

> In BaseCollectorConfig why do you have random metric names?
This is a historic thing where I wanted to be sure there where no registration errors when the user is mucking about in the UI.  I've since re-factored where now the actual collectors are created and registered when the test starts. I think I could use sane defaults like `jmeter_counter` now that multiple listeners can have reference to a single collector. I can't really get more specific than that (like http_response_time_ms) because of JMeter's nature, users typically collect HTTP results, but they could also collect JDBC, or JMS or any number of things. 

As to the other notes, I'll take care of those directly.

brian-brazil:
Thanks!

PR #1290, opened 2019-02-14
Link to Clojure client

carlduevel:
I'd like to add a Clojure client I wrote a while ago.
All feedback is welcome!

### DIFF
--- a/content/docs/instrumenting/clientlibs.md
+++ b/content/docs/instrumenting/clientlibs.md
@@ -22,6 +22,7 @@ Unofficial third-party client libraries:
 
 * [Bash](https://github.com/aecolley/client_bash)
 * [C++](https://github.com/jupp0r/prometheus-cpp)
+* [Clojure](https://github.com/hackbert/clometheus)
 * [Common Lisp](https://github.com/deadtrickster/prometheus.cl)
 * [Elixir](https://github.com/deadtrickster/prometheus.ex)
 * [Erlang](https://github.com/deadtrickster/prometheus.erl)


SuperQ:
Interesting, your README talks about not having to register metrics. This is a bit odd for a typical Prometheus implementation. In Prometheus, we expect things like counters to exist from startup time, and stay existing until the process exits. We try to [avoid metrics that appear and disappear](https://www.robustperception.io/existential-issues-with-metrics).

The classic problem here is having an error counter that starts at 1, because the counter gets created on the first increment. So if you want to alert on `increase(errors_total[5m])`, a single increment won't fire because it didn't start out at zero.

brian-brazil:
Thanks for sharing. To make things easier for users, we try to list only one client per language/runtime and there's already a Java client.

SuperQ:
While I agree that it's good to have one option per language/runtime. Clojure is different enough that it probably deserves a dedicated client.

brian-brazil:
It's still on the JVM, and can use the Java client which has better usage in this case at least. Otherwise we'll have every JVM language looking for a link.

hackbert:
Thanks for your quick feedback @SuperQ and @brian-brazil!

To your point about the avoidance of metrics that appear and disappear
@SuperQ, I will update the documentation of the project to advocate
defining metrics upfront (aka registering them).

About the question of listing clients for different languages:
When I initially wanted to instrument some Clojure and Scala
applications I was surprised to not find any clients in those
languages on the list, because at least in my experience as soon
as a Java library gets popular it immediately gets at least wrapped
by somebody to accommodate the language better than Java does.
And of course that is what people have done in many JVM languages
which is an indicator of the popularity of Prometheus in those
communities. So in a way this client list does not do justice to
that widespread use.

There is an Elixir and Erlang client on the list and I think that
is a good choice you made because people writing Elixir will prefer
a client with an interface in Elixir to one in Erlang though they are running
their code on the same VM.

brian-brazil:
Wrappers are fine (there's a section down the bottom for them), brand new client libraries on a given runtime are not in terms of duplication. If duplicates have slipped in, we'll clean them up at some point.

PR #1297, opened 2019-02-26
Update self-inhibition prevention docs for AM v0.17

beorn7:
As promised…

Please only merge once v0.17 is actually released.


beorn7:
0.17 is being released right now. Merging this.

beorn7:
Of course, would be great to have versioned docs for AM, too.

PR #1298, opened 2019-02-27
Add docs for alertmanager group_by statement

smelchior:
This adds the docs for the group_by statement which are currently only present in the alertmanager README. I think the info, especially for the magic value '...' should also be present here

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -128,6 +128,16 @@ current node.
 
 ```yaml
 [ receiver: <string> ]
+# The labels by which incoming alerts are grouped together. For example,
+# multiple alerts coming in for cluster=A and alertname=LatencyHigh would
+# be batched into a single group.
+#
+# To aggregate by all possible labels use the special value '...' as the sole label name, for example:
+# group_by: ['...'] 
+# This effectively disables aggregation entirely, passing through all 
+# alerts as-is. This is unlikely to be what you want, unless you have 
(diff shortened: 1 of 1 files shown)


content/docs/alerting/configuration.md
```diff
@@ -128,6 +128,15 @@ current node.
 
 ```yaml
 [ receiver: <string> ]
+# The labels by which incoming alerts are grouped together. For example,
+# multiple alerts coming in for cluster=A and alertname=LatencyHigh would
+# be batched into a single group.
+#
+# To aggregate by all possible labels use '...' as the sole label name.
```
brian-brazil: We also use ... to mean "more labels" so it'd be good to clarify the syntax is literal here

smelchior: thanks for the quick feedback, i rephrased it a bit

content/docs/alerting/configuration.md
```diff
@@ -128,6 +128,17 @@ current node.
 
 ```yaml
 [ receiver: <string> ]
+# The labels by which incoming alerts are grouped together. For example,
+# multiple alerts coming in for cluster=A and alertname=LatencyHigh would
+# be batched into a single group.
+#
+# To aggregate by all possible labels use the special value '...' as the sole label name:
```
brian-brazil: You can combine this and the next line,  

...name for example:

smelchior: done

brian-brazil:
Thanks!

PR #1307, opened 2019-03-15
Add VerneMQ to software exposing Prometheus metrics

larshesel:
Hi @brian-brazil 

VerneMQ exports metrics in the Prometheus format on port 8888 by default, so thought I'd add it to the list.

Signed-off-by: Lars Hesel Christensen <lars@larshesel.dk>


brian-brazil:
Thanks!

I'd suggest using an existing Erlang client library, rather than doing it yourself.

PR #1309, opened 2019-03-22
Adds Zipkin as a software that exposes prometheus metrics directly

devinsba:


content/docs/instrumenting/exporters.md
```diff
@@ -226,6 +226,7 @@ separate exporters are needed:
    * [Traefik](https://github.com/containous/traefik)
    * [VerneMQ](https://github.com/vernemq/vernemq)
    * [Weave Flux](https://github.com/weaveworks/flux)
+   * [Apache Zipkin (incubating)](https://zipkin.incubator.apache.org/)
```
devinsba: I put this at the end because Zipkin would be last

SuperQ:
Is this for Zipkin library use? Maybe this would fit better under "Other third-party utilities"?

devinsba:
No, the Zipkin server exposes prometheus metrics. https://github.com/openzipkin/zipkin/tree/master/zipkin-server#metrics

The default distributions include the required library

content/docs/instrumenting/exporters.md
brian-brazil: There's no need to mention that it's incubating, and Apache would go first. I always hear it referred to as just Zipkin, and that's what's on its webpage.

I gather this isn't on /metrics, can you link to the docs saying where it is?

gouthamve:
@devinsba Could you rebase and drop the "incubating"? And if you really want to keep Apache, could you move it to the top in line with the normal ordering?

devinsba:
Looks like I deleted my fork at some point, recreated as https://github.com/prometheus/docs/pull/1530

PR #1312, opened 2019-03-26
Add Promitor, an Azure Monitor exporter for Prometheus

tomkerkhove:
Add Promitor, an Azure Monitor exporter for Prometheus

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -123,6 +123,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Akamai Cloudmonitor exporter](https://github.com/ExpressenAB/cloudmonitor_exporter)
    * [Alibaba Cloudmonitor exporter](https://github.com/aylei/aliyun-exporter)
    * [AWS CloudWatch exporter](https://github.com/prometheus/cloudwatch_exporter) (**official**)
+   * [Azure Monitor exporter / Promitor](https://github.com/tomkerkhove/promitor)
    * [Cloud Foundry Firehose exporter](https://github.com/cloudfoundry-community/firehose_exporter)
    * [Collectd exporter](https://github.com/prometheus/collectd_exporter) (**official**)
    * [Google Stackdriver exporter](https://github.com/frodenas/stackdriver_exporter)


brian-brazil:
Thanks for sharing. We generally list the most canonical and guideline compliant exporter. As far as I can see https://github.com/RobustPerception/azure_metrics_exporter is more featureful, more used, and more compliant. Do you want to add whatever features are missing from yours to that?

tomkerkhove:
We are adding more features as we speak but you mention that it is more "compliant" which is good, but compliant with what? Would love to read more about it.

brian-brazil:
https://prometheus.io/docs/instrumenting/writing_exporters/

This exporter appears to be fetching on a schedule, rather than on a scrape.

> The most used argument is one I don't really agree with given if it cannot be listed it cannot get adopted more

An Azure metrics exporter should be listed, the question is which one.

tomkerkhove:
> This exporter appears to be fetching on a schedule, rather than on a scrape.

That is correct given we handle the scraping intervals for the consumer, Prometheus just has to scrape and base itself on the timestamp.

Push support is on the way.

> An Azure metrics exporter should be listed, the question is which one.

Why not both? But I get what you are saying.

brian-brazil:
> That is correct given we handle the scraping intervals for the consumer, Prometheus just has to scrape and base itself on the timestamp.

That's not in line with our guidelines, and will miss data more frequent for example.

> Why not both?

To avoid ending up with N exporters for a given thing, for which users are not equipped to evaluate and which would also not encourage everyone to work together.

tomkerkhove:
Will investigate as part of https://github.com/tomkerkhove/promitor/issues/467

Makes sense!

tomkerkhove:
You mention that https://github.com/RobustPerception/azure_metrics_exporter is more featureful, can you please elaborate on what features it has that Promitor doesn't have?

From the looks of it it's pretty much similar at a first glance.

brian-brazil:
The RP one has far more options for how to select what's scraped from what I can see.

tomkerkhove:
Feel free to have a closer look and let me know what, maybe docs are lacking

brian-brazil:
I've dug around a bit more. Both repos have similar stats in terms of stars etc., though the other is older. The other has more Docker pulls (and has had a dockerfile for about half the time).

Looking at yours I note that it has no label support, a user has to list every single resource they want scraped and give it a unique metric name. The other allows beyond that specifying entire resource groups, with multiple aggregations, filtering, and handles labels automatically.

Accordingly I think #1332 is the appropriate PR to accept, as it's easier for users and more in line with general Prometheus philosophy. Thank you for sharing yours.

tomkerkhove:
For what it's worth, here are my remarks.

> I've dug around a bit more. Both repos have similar stats in terms of stars etc., though the other is older. The other has more Docker pulls (and has had a dockerfile for about half the time).

That is correct, although we've only switched from [tomkerkhove/promitor-scraper](https://hub.docker.com/r/tomkerkhove/promitor-scraper) to [tomkerkhove/promitor-agent-scraper](https://hub.docker.com/r/tomkerkhove/promitor-agent-scraper) but I don't believe pulls are that good of measurement but it's your call)

> Looking at yours I note that it has no label support, a user has to list every single resource they want scraped and give it a unique metric name

That is correct, and also on purpose.

> The other allows beyond that specifying entire resource groups, with multiple aggregations, filtering, and handles labels automatically.

- Scraping entire RGs is bad
- Filtering is WIP

That said, it's ok for me to merge yours as long one is in the list. People can still use mine if they want to, however, I'm not sure why we cannot add both of them but that's ok.

PR #1315, opened 2019-03-30
add glusterfs exporter

kocoten1992:
add official gluster exporter


SuperQ:
Thanks for the contribution.

It looks like this exporter has some issues. For one, it's using Gauge metrics for counters. It should probably be refactored to use "Const" metrics.

See: https://prometheus.io/docs/instrumenting/writing_exporters/#collectors

SuperQ:
It also looks like the exporter has it's own timer loop, this is a violation of Prometheus best practices. Exporters should gather metrics at scrape time, not implement their own internal timers.

See: https://prometheus.io/docs/instrumenting/writing_exporters/#scheduling

brian-brazil:
How does this differ from the existing gluster exporter that's listed?

kocoten1992:
Hi, I found this exporter while searching for glusterfs exporter, don't know how it differ with current one (good part it was supported by official team).

But if it not meet the standard, think we could close this ticket.

brian-brazil:
From a quick peek, the existing one isn't perfect either however it is generally in line. If the official one could be made to follow the guidelines better we could potentially switch, as the other looks a little unloved.

PR #1318, opened 2019-04-01
Add nginx-prometheus-exporter

Dean-Coakley:
Add Nginx exporter maintained by NGINX Inc.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -90,6 +90,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
 ### HTTP
    * [Apache exporter](https://github.com/Lusitaniae/apache_exporter)
    * [HAProxy exporter](https://github.com/prometheus/haproxy_exporter) (**official**)
+   * [Nginx exporter](https://github.com/nginxinc/nginx-prometheus-exporter)
    * [Nginx metric library](https://github.com/knyar/nginx-lua-prometheus)
    * [Nginx VTS exporter](https://github.com/hnlq715/nginx-vts-exporter)
    * [Passenger exporter](https://github.com/stuartnelson3/passenger_exporter)


brian-brazil:
Thanks for sharing. How does this differ from the nginx integrations already listed on this page?

Dean-Coakley:
As far as I understand:
* The `nginx metric library` requires `libnginx-mod-http-lua` to be installed with additional configuration.
* The `nginx vts exporter` requires you to recompile nginx including the [vts module](https://github.com/vozlt/nginx-module-vts#installation).

This exporter scrapes metrics directly from `stub_status` and does not have any dependencies on any third party nginx modules. Therefore It can be run on any existing nginx deployment without any additional configuration as long as `/stub_status` or `/api` is enabled and accessible.

Hope that makes sense

brian-brazil:
So you're taking that this is a superset of the vts exporter's functionality?

Dean-Coakley:
Yeah, I would say this gives this exporter more flexibility in it's use cases over the others. 

It's also the only exporter to support nginx plus metrics for active nginx plus subscribers.

Finally it has the support of NGINX maintainers directly so it's the closest thing to an official exporter.

brian-brazil:
> Yeah, I would say this gives this exporter more flexibility in it's use cases over the others.

We try to only list one exporter for a given thing so that the community works together on one exporter rather than everyone writing their own. To replace an existing exporter you'd need to provide a superset of features, and be at least as compliant with the exporter guidelines.
Looking at it you'd need the commercial version of Nginx in order to get the rough functionality of the current vts-exporter.

What I'd suggest is extending the community version of Nginx to have metric parity with the commercial version, as the 7 metrics it currently exposes are already covered by the vts exporter plus many more.

On an implementation front I only see a few small issues, upMetric should also use MustNewConstMetric and then you can remove the locking. Counters should end in _total, and metrics should contain their units (seconds, bytes). You may also want to consider an enum instead of a gauge for upstreamServerStates as it's easier to work with - but does use more time series.

Dean-Coakley:
> We try to only list one exporter for a given thing so that the community works together on one exporter rather than everyone writing their own.

I agree. However, the goal of this exporter is to support the built-in NGINX metrics, not to provide an advanced alternative for the listed exporters.

If a user gets NGINX from DockerHub (https://hub.docker.com/_/nginx) or the NGINX software repositories (http://nginx.org/en/linux_packages.html), the only available metrics are the stub_status metrics, which this exporter support. In contrast, the listed exporters  (1) do not support the stub_status metrics and (2) require the user to extend NGINX with third-party modules.  Thus, this exporter is not a replacement for the listed exporters, but the solution for users who want to collect metrics from the vanilla NGINX.

I believe this exporter is useful for NGINX users. This is based on the activity in the exporter GitHub repo and the downloads of the exporter from the DockerHub (https://hub.docker.com/u/nginx). 

Additionally, this exporter supports NGINX Plus metrics, but this is not a requirement for users.

Thanks for the comments about the implementation.

SuperQ:
The problem is that stub_status doesn't provide a useful set of metrics, hence the need for extra 3rd-party code. Unfortunately this makes this exporter insufficient to be something we would recommend using.

Dean-Coakley:
That makes sense. 

Thanks for taking the time to review the exporter. 

> I would recommend NGINX Inc add a native Prometheus/OpenMetrics API endpoint in the code to replace stub_status. This would elevate NGINX to "native" level support in our recommendations.

Hopefully someday!

Dean-Coakley:
@brian-brazil Can you clarify what you meant by 

> You may also want to consider an enum instead of a gauge for upstreamServerStates as it's easier to work with - but does use more time series.

Perhaps it's related to this issue?: https://github.com/nginxinc/nginx-prometheus-exporter/issues/35

brian-brazil:
> Perhaps it's related to this issue?:

That's the one. This is one convention of a way to do things, it's one of several ways to approach this.

As an aside: https://www.robustperception.io/why-can-countx-5-not-return-0

PR #1319, opened 2019-04-03
add ping exporter

anarcat:
The "ping" exporter is really an "fping" exporter but that's how it is named so I picked that.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -177,6 +177,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [SABnzbd exporter](https://github.com/msroest/sabnzbd_exporter)
    * [Script exporter](https://github.com/adhocteam/script_exporter)
    * [Shield exporter](https://github.com/cloudfoundry-community/shield_exporter)
+   * [Smokeping prober](https://github.com/SuperQ/smokeping_prober)
    * [SMTP/Maildir MDA blackbox prober](https://github.com/cherti/mailexporter)
    * [SoftEther exporter](https://github.com/dalance/softether_exporter)
    * [Transmission exporter](https://github.com/metalmatze/transmission-exporter)


SuperQ:
How does this compare to this one: https://github.com/SuperQ/smokeping_prober

anarcat:
differences:

 * it's written in Python, not Golang
 * it uses fping instead of a fork of [go-ping](https://github.com/sparrc/go-ping)

similarities:

 * both are not in the current list (and there's no ICMP prober that i can find, except maybe the blackbox exporter, which is not an immediately obvious replacement for new users)

Maybe both should be added?

anarcat:
darn it, here's another one:

https://github.com/localleon/pingpong

brian-brazil:
If we're going to add one, it's be https://github.com/SuperQ/smokeping_prober as it's the most in line with the Prometheus way of doing things. We try to list only one exporter for a given purpose, and there's already overlap with the blackbox exporter.

anarcat:
okay, i rebuild the patch to use the smokeping thing even though i think it's way overkill. :)

i do wonder about the overlap with the blackbox exporter as well.. maybe we should add a note on the blackbox exporter saying it can also do ping instead?

SuperQ:
Yea, there are a number of Prometheus-compatible probers. The smokeping one I made I call it a prober, rather than exporter, as it's a stand-alone metrics generator rather than something like the blackbox exporter.

You're right, I should add some more description to the smokeping_prober's readme about how it works and why it's similar and different from the blackbox_exporter.

content/docs/instrumenting/exporters.md
```diff
@@ -138,6 +138,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [NRPE exporter](https://github.com/robustperception/nrpe_exporter)
    * [Osquery exporter](https://github.com/zwopir/osquery_exporter)
    * [OTC CloudEye exporter](https://github.com/tiagoReichert/otc-cloudeye-prometheus-exporter)
+   * [Smokeping prober](https://github.com/SuperQ/smokeping_prober)
```
brian-brazil: This is really misc, as you're not talking to an existing smokeping setup

RichiH: Agreed; else LGTM.

anarcat: so the reason i put it in "other monitoring system" in the first place is that the original solution i proposed was using "fping" (the "other") instead of a home-grown implementation. but I guess that is, indeed, not a valid argument anymore.

i'm concerned about adding yet another entry to "misc". there's already way too much disparate stuff in there, it's hard to make head or tails in there...

brian-brazil: That's what tends to happen with any Misc category in any system.

anarcat: heh. i guess what i'm saying is we might be able to create a new category from the misc one... maybe we could regroup the following in a "Network service" section of some sort?

* BIND
* Blackbox
* Dovecot
* PowerDNS
* rTorrent
* SMTP...
* Transmission
* Unbound

Then the ICMP/Ping probers could fit right in.

brian-brazil: Why would Dovecot be a network service and BIG-IP not? Almost all of the misc list are for services provided over the network.

I'd suggest getting this in, and then if you've a better idea for taxonomy exploring that in another issue/PR.

anarcat: because i don't know what BIG-IP is. :) My argument is that there *are* a bunch of things in misc that are not *primarily* network services. cAdvisor, JIRA, Jenkins, kube-state... those are of course available over the network, but their primary purpose is not to offer some standard network protocol (like DNS or IMAP) but to provide a service (like bugtracking or CI).

but you are right: maybe that peculiar bikeshed should be painted somewhere else.

SuperQ: BIG-IP is a hardware http load-balancer appliance product/company.

brian-brazil: Actually BIG-IP should be under hardware.

brian-brazil:
Thanks!

PR #1322, opened 2019-04-04
Clarify usage of 'instance' and 'target'.

jameshochadel:
The term 'target' is used later on this page, but is never explicitly defined.

Signed-off-by: James Hochadel <james.hochadel@gmail.com>

### DIFF
--- a/content/docs/concepts/jobs_instances.md
+++ b/content/docs/concepts/jobs_instances.md
@@ -6,7 +6,7 @@ sort_rank: 3
 # Jobs and instances
 
 In Prometheus terms, an endpoint you can scrape is called an _instance_,
-usually corresponding to a single process. A collection of instances with the same purpose, a process replicated for scalability or reliability for example, is called a _job_.
+usually corresponding to a single process. An instance is also sometimes referred to as a _target_. A collection of instances with the same purpose, a process replicated for scalability or reliability for example, is called a _job_.
 
 For example, an API server job with four replicated instances:
 


jameshochadel:
@brian-brazil addressing you per `CONTRIBUTING.md`.

 I have never used a DCO check before, and I'm not sure why it's asking for a different email; the one I included is the one associated with my Github account. Any advice?

SuperQ:
I'm not sure why, but the DCO is expecting `james@carbonrelay.com`.

content/docs/concepts/jobs_instances.md
```diff
@@ -6,7 +6,7 @@ sort_rank: 3
 # Jobs and instances
 
 In Prometheus terms, an endpoint you can scrape is called an _instance_,
-usually corresponding to a single process. A collection of instances with the same purpose, a process replicated for scalability or reliability for example, is called a _job_.
+usually corresponding to a single process. An instance is also sometimes referred to as a _target_. A collection of instances with the same purpose, a process replicated for scalability or reliability for example, is called a _job_.
```
brian-brazil: Targets are not the same as instances.

jameshochadel: Thank you for your response! Could you explain what they each are, or point me towards canonical definitions? 

I think the source of my confusion is that in this doc, instance is defined as:
> In Prometheus terms, an endpoint you can scrape is called an _instance_

...but from what I'm reading in _Prometheus Up & Running_ + the Prometheus configuration docs and from using Prometheus itself, that definition more closely matches target. Unless instance refers to the endpoint, and target refers to the host?

Thank you again— getting started, this is confusing, which is why I wanted to update the docs upstream when I found an answer. I appreciate the help.

jameshochadel: Just found https://prometheus.io/docs/introduction/glossary/. I'll close this PR; might submit a separate one that references the glossary. Thanks again.

Issue #1320, opened 2019-04-04
split the ever-growing misc exporters section


anarcat:
The misc section in the [list of exporters](https://prometheus.io/docs/instrumenting/exporters/) is getting out of control, with now 38 entries of various unrelated things mixed together.

Here are the item counts at the time of writing:

 * Databases: 22
 * Hardware: 12
 * Messaging: 10
 * Storage: 6
 * HTTP: 9
 * APIs: 14
 * Logging: 3
 * Other monitoring systems: 23
 * Miscellaneous: 38
 * Other software: 31
 * Third-party: 10

It's the largest section, and because there's so much stuff in there, there are a lot of hidden gems new users will miss. I totally missed and underestimated the importance of the blackbox exporter for example. In #1319, I suggested at least one group might be usefully split out of the miscellaneous section, but I'm not sure yet if I could come up with a good definition.

But somehow I feel those belong together:

## Network services

   * [BIND exporter](https://github.com/digitalocean/bind_exporter)
   * [Blackbox exporter](https://github.com/prometheus/blackbox_exporter) (**official**)
   * [Dovecot exporter](https://github.com/kumina/dovecot_exporter)
   * [Kannel exporter](https://github.com/apostvav/kannel_exporter)
   * [PowerDNS exporter](https://github.com/ledgr/powerdns_exporter)
   * [SABnzbd exporter](https://github.com/msroest/sabnzbd_exporter)
   * [Smokeping prober](https://github.com/SuperQ/smokeping_prober)
   * [SMTP/Maildir MDA blackbox prober](https://github.com/cherti/mailexporter)
   * [SoftEther exporter](https://github.com/dalance/softether_exporter)
   * [Unbound exporter](https://github.com/kumina/unbound_exporter)

I would possibly even include those in the list:

   * [rTorrent exporter](https://github.com/mdlayher/rtorrent_exporter)
   * [Transmission exporter](https://github.com/metalmatze/transmission-exporter)

We could also have a section on:

## Issue trackers and CI

   * [Bitbucket exporter](https://github.com/AndreyVMarkelov/prom-bitbucket-exporter)
   * [Confluence exporter](https://github.com/AndreyVMarkelov/prom-confluence-exporter)
   * [JIRA exporter](https://github.com/AndreyVMarkelov/jira-prometheus-exporter)
   * [Jenkins exporter](https://github.com/lovoo/jenkins_exporter)
[...]

PR #1323, opened 2019-04-04
create issue trackers and CI section, fix Presto database

anarcat:
This is a partial fix for #1320. It creates a issue trackers and CI section, along with moving Presto in the right place.

I would also create a "load balancers" and "networking" sections, along with moving Cachet, cAdvisor, Kibana and kube-stat to "Other monitoring" but that seemed to have been less popular. ;)


anarcat:
another one i think might belong to "hardware" is the [ACT Fibernet Exporter](https://git.captnemo.in/nemo/prometheus-act-exporter) which seem to talk to DSL modems to get bandwidth data... should i move it as well?

brian-brazil:
Thanks!

PR #1325, opened 2019-04-06
Another Kafka metrics exporter

nipuntalukdar:
Another Kafka metrics exporter for Prometheus. Grafana dashboard generation tool is included.


brian-brazil:
Thanks for sharing. We try to avoid duplication in this list, and it appears that this exporter is offering a only a subset of the functionality of the existing one that we list. If there's features the existing one is missing, I'd suggest talking to them.

PR #1326, opened 2019-04-06
fix missed download page

chy168:
Hi,
I found the download page in the document is 404 on github.
How about point to the release page of prometheus directly?


simonpasquier:
This works as expected from https://prometheus.io/docs/introduction/first_steps/ so there's no need to change anything.

chy168:
i see, thanks for your comment :)

PR #1334, opened 2019-04-15
Clarify that prometheus-announce is _not_ mirrored to prometheus-users

beorn7:
I hope that Google groups doesn't prevent mirroring like this in some
way. So far we have done the mirroring manually, but whenever I asked
around, the consensus seemed to be that we just want to subscribe
prometheus-users@ to prometheus-announce@. If there are no objections
now, please approve this PR, and then I'll try to subscribe as
described.

The reviewers selected here are those that have recently posted to prometheus-announce@. Of course, everybody is encouraged to chime in.

Different thought: There are only 69 subscribers of prometheus-announce@. Perhaps we should, well, announce prometheus-announce@ on prometheus-users@? After we have clarified the mirroring, of course, so that we can explain that you don't need to subscribe to -announce if you are on -users anyway (but you can subscribe to -announce if you want to avoid the traffic from -users but not miss the announcements).

### DIFF
--- a/content/community.html
+++ b/content/community.html
@@ -16,8 +16,8 @@ title: Community
     <p>
       <strong>User mailing lists:</strong>
       <ul>
-        <li><a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a> – for announcements like new releases</li>
-        <li><a href="https://groups.google.com/forum/#!forum/prometheus-users">prometheus-users</a> – for discussions around Prometheus usage and community support</li>
+        <li><a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a> – low-traffic list for announcements like new releases.</li>
+        <li><a href="https://groups.google.com/forum/#!forum/prometheus-users">prometheus-users</a> – for discussions around Prometheus usage and community support. Announcements are <em>not</em> generally mirrored from <a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a>.</li>
       </ul>
     </p>
     <p>


beorn7:
Hmmgrmbl… guess what, subscribing one group to another is an enterprise feature. The work around would be to create a gmail address that simply forwards all its mails to prometheus-users@. I'm not sure if I would like such a work around in our setup. We could also keep the current practice of mirroring manually. (In which case I would adjust the explanation in this PR accordingly.)

What does everybody think?

matthiasr:
hmm in that case I would rather mirror manually. separate google account means one more set of credentials to maintain

brian-brazil:
This was a to-do for me from the last monthly meeting, and I discovered
what you did. I was going to bring this up at the next one.

My thought was to properly announce -announce, and then only send
announcements there. Longer term I expect we would have unsubscribed -users
from -announce anyway.

beorn7:
Oops, yeah, I should have read the meeting notes.

It looks we all agree that we should not do the contraption with a fake Gmail account.

Which leaves us with manually mirroring (@matthiasr) or not mirroring at all (@brian-brazil). I guess relying on humans to mail both lists is doomed to fail (at least now and then). Which makes me gravitate towards "not mirroring at all". Which is still something we should clarify in the docs.

If there are no objections, I'll update this PR for the latter. And once this is merged, I'll announce -announce on -users.

beorn7:
Note to self: We also need to update prometheus/prometheus/RELEASE.md.

Suggested announcement mail:

"""
Subject: [ANN] The prometheus-announce mailing list

Dear Prometheus users,

A while ago, we created the [prometheus-announce](https://groups.google.com/forum/#!forum/prometheus-announce) mailing list, mostly for users who want to get important announcements (usually about releases) but are not interested in the many discussions going on on prometheus-users. It has been common practice to send those announcements to both, prometheus-users and prometheus-announce. However, for better separation of concerns, we are planning to stop this manual procedure and send announcements only to prometheus-announce.

Thus, if you are interested in announcements of releases etc., please subscribe to [prometheus-announce](https://groups.google.com/forum/#!forum/prometheus-announce) at your next convenience.

_Anticipated FAQ: “Why don't you just subscribe prometheus-users to prometheus-announce?” Unfortunately, that's not possible in a clean way with the free version of Google groups, which contributed to our decision to keep things clearly separated._
"""

beorn7:
Updated the PR to clarify the non-mirroring.

brian-brazil:
:+1:

beorn7:
Two :+1:s and no objections. Merging now…

PR #1338, opened 2019-04-17
add Dnsmasq exporter into exporters list

mpursley:
Signed-off-by: Matt Pursley <mpursley@gmail.com>


mpursley:
@brian-brazil , this PR just adds dnsmasq_exporter into exporters.md...
exporters.md -> Miscellaneous -> Dnsmasq exporter](https://github.com/google/dnsmasq_exporter)

In case you haven't come across it before, "Dnsmasq is a lightweight, easy to configure DNS forwarder, designed to provide DNS (and optionally DHCP and TFTP) services to a small-scale network."...
  -- https://en.wikipedia.org/wiki/Dnsmasq

It's small, easy and included in a bunch of Linux distros.  So, it probably gets enough use to worth having the exporter mentioned here too...

brian-brazil:
Thanks!

All of these metrics sound like Counters rather than Gauges.

PR #1339, opened 2019-04-22
add Lastmile exporter into exporters list

bsingr:
https://github.com/coliquio/lastmile

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -217,6 +217,7 @@ separate exporters are needed:
    * [Grafana](http://docs.grafana.org/administration/metrics/)
    * [JavaMelody](https://github.com/javamelody/javamelody/wiki/UserGuideAdvanced#exposing-metrics-to-prometheus)
    * [Kubernetes](https://github.com/kubernetes/kubernetes) (**direct**)
+   * [Lastmile](https://github.com/coliquio/lastmile)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. From a quick look this seems to overlap with the blackbox exporter, and we try to avoid duplication in this list.

bsingr:
I use the blackbox exporter heavily, but the following use-case it does not work for (afaik):

Our team spreads across multiple office and remote working locations. We want to be notified if any of the locations is not be able to access our shared cloud infrastructure. With some locations using VPN and some not we have different routes traffic can hit our shared cloud infrastructure (e.g. different public/private IP space, different public/private DNS resolution). This is why we monitor DNS + HTTP.

In its core this is achievable with blackbox exporter.

What makes it hard is:

- Edge locations don't have fixed IPs / DNS themselves, so how does a central prometheus reach those locations? (<=> this is why we thought central pushgateway makes sense)
- Edge locations are not always behind VPN, so it would require some secure https+auth reverse proxy (<=> again with central pushgateway behind https+auth quite simple)
- Re-deploying edge locations is for the same reason not very easy to do; at the same time the probe config changes quite often (<=> here we let the lastmile instances pull the config from time-to-time)

Might be very specific use-case not useful for many, so again, close if you like :)

brian-brazil:
That feels more like something in the pushprox space, as this is more a networking thing than a metrics thing. This particular way of combining these sort of things sounds a bit use-case specific.

PR #1340, opened 2019-04-22
add http-json-query exporter into exporters list

bsingr:
https://github.com/coliquio/http-json-query-exporter


brian-brazil:
Hmm, there's also https://github.com/kawamuray/prometheus-json-exporter out there which is more popular and we do try to avoid duplication in this list.

SuperQ:
@bsingr Thanks for the contribution, but I think more people are using the existing https://github.com/kawamuray/prometheus-json-exporter right now. However it doesn't seem to be currently maintained.

Plus, the prometheus-json-exporter is written in Go, so it should be a bit more efficient than a javascript service.

bsingr:
No problem @brian-brazil  @SuperQ. Did not know about the https://github.com/kawamuray/prometheus-json-exporter.

Compares to me sth. like so:

Pro https://github.com/coliquio/http-json-query-exporter

- The transformation with http://jsonata.org/ are far more flexible that just jsonpath
- It supports basically any kind of http request (not just GET) including headers etc.
- Allows querying different http endpoints from a single process

Pro https://github.com/kawamuray/prometheus-json-exporter

- I assume you are right with the superior effiency of using go
- More people using it seems also like big plus

brian-brazil:
Looks like prometheus-json-exporter is still by far the one the community is using, so that's what we would list.

PR #1341, opened 2019-04-22
add Tyk API Gateway exporter into exporters list

bsingr:
https://github.com/coliquio/tyk-gateway-prometheus-exporter

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -105,6 +105,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Passenger exporter](https://github.com/stuartnelson3/passenger_exporter)
    * [Squid exporter](https://github.com/boynux/squid-exporter)
    * [Tinyproxy exporter](https://github.com/igzivkov/tinyproxy_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Hmm, where's the actual source code for this?

The metrics seem to be info metrics, so should end in `_info` and be of type gauge. Your example in the README has a secret on the command line, which is not secure.

bsingr:
There is no source code, it's actually using https://github.com/coliquio/http-json-query-exporter + generic config for tyk. See https://github.com/coliquio/tyk-gateway-prometheus-exporter/blob/master/Dockerfile#L1

Copy that `_info`. Will change that.

Agree that secrets is not best-practice, any suggestions how you deal with secrets in doc?

Besides, I now understand you want to keep the exporters list to be very focused. So just close if you like.

brian-brazil:
> There is no source code, it's actually using https://github.com/coliquio/http-json-query-exporter + generic config for tyk.

Hmm, for similarish things what usually happens is that the main exporter has a contrib directory (or similar) with such configs.

> Agree that secrets is not best-practice, any suggestions how you deal with secrets in doc?

Reading from a file is generally good.

> Besides, I now understand you want to keep the exporters list to be very focused.

The notability of what the exporter is for generally isn't relevant, unless it's something that e.g. only the creator of the exporter uses.

PR #1342, opened 2019-04-22
add Kong API Gateway exporter into exporters list

bsingr:
https://github.com/Kong/kong-plugin-prometheus

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -216,6 +216,7 @@ separate exporters are needed:
    * [FreeBSD Kernel](https://www.freebsd.org/cgi/man.cgi?query=prometheus_sysctl_exporter&apropos=0&sektion=8&manpath=FreeBSD+12-current&arch=default&format=html)
    * [Grafana](http://docs.grafana.org/administration/metrics/)
    * [JavaMelody](https://github.com/javamelody/javamelody/wiki/UserGuideAdvanced#exposing-metrics-to-prometheus)
+   * [Kong](https://github.com/Kong/kong-plugin-prometheus)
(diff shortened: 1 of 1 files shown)


brian-brazil:
I think this might better belong in the Software exposing Prometheus metrics list given it's a plugin rather than an exporter.

Metrics should indicate their unit, what are the units for bandwidth and latency?

bsingr:
Bandwith is bytes, latencies is milliseconds.

bsingr:
Agree + moved to `Software exposing Prometheus metrics`

content/docs/instrumenting/exporters.md
```diff
@@ -216,6 +216,7 @@ separate exporters are needed:
    * [FreeBSD Kernel](https://www.freebsd.org/cgi/man.cgi?query=prometheus_sysctl_exporter&apropos=0&sektion=8&manpath=FreeBSD+12-current&arch=default&format=html)
    * [Grafana](http://docs.grafana.org/administration/metrics/)
    * [JavaMelody](https://github.com/javamelody/javamelody/wiki/UserGuideAdvanced#exposing-metrics-to-prometheus)
+   * [Kong API Gateway exporter](https://github.com/Kong/kong-plugin-prometheus)
```
brian-brazil: The link text should just be Kong

brian-brazil:
Seconds are the standard unit for time in Prometheus, so those metrics could be improved.

bsingr:
> Seconds are the standard unit for time in Prometheus, so those metrics could be improved.

Best practice is to add suffix `_milliseconds` or so?

brian-brazil:
Best practice would be to divide it by 1000, and append `_seconds`.

brian-brazil:
Thanks!

PR #1344, opened 2019-04-23
Adding reference to Rocket.Chat AlertManager webhook

louisfelix:
Hi,
This PR proposes the addition of a Rocket.Chat AlertManager webhook integration to the list.
Thanks!

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -64,6 +64,7 @@ For notification mechanisms not natively supported by the Alertmanager, the
   * [JIRAlert](https://github.com/free/jiralert)
   * [Phabricator / Maniphest](https://github.com/knyar/phalerts)
   * [prom2teams](https://github.com/idealista/prom2teams): forwards notifications to Microsoft Teams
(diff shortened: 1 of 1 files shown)


brian-brazil:
Hmm, the official Rocket docs point to https://github.com/pavel-kazhavets/AlertmanagerRocketChat. We try to avoid duplicates, and that one seems more popular.

louisfelix:
Hi @brian-brazil,
I completely agree with the duplicate rules, but I think here we have two different types of components that are used in very different scenarios.

The first project is a script you copy/paste in Rocket.Chat custom incoming webhook section. It is completely tied to Rocket.Chat installation, and is more static.

The other I propose is a standalone receiver that can be configured, deployed and updated independently from Rocket.Chat (it uses the Rocket.Chat API and will evolve using the full power of the API). As such, it shares the same modularity spirit of all the other Alertmanager Webhook Receiver listed here and this is why I think it fits here.
(I also proposed this addition to the Rocket.Chat official doc also).

Thanks!

PR #1348, opened 2019-04-30
content/docs/alerting: Add reference to Alertmanager API v2

mxinden:
Alertmanager v0.16.0 introduces a new API, v2, replacing API v1 in the
future. This patch adds a reference to the new API in the alerting docs.

Reference: https://github.com/prometheus/prometheus/pull/5482#issuecomment-487057373


brian-brazil:
:+1:

PR #1350, opened 2019-05-13
Add perf exporter

hodgesds:
Signed-off-by: Daniel Hodges <hodges.daniel.scott@gmail.com>


brian-brazil:
Hmm, does this cover things that the ebpf exporter doesn't? We try to avoid overlap on this page.

hodgesds:
It uses kprobes directly, eBPF programs (generally) need to be compiled with kernel headers unless another eBPF frontend is used like in [this project](https://github.com/weaveworks/tcptracer-bpf). The main benefit that the `perf_exporter` has is that by using kprobes directly you don't need to compile any eBPF programs. So much in the same way a helicopter and airplane are both methods of flying the means of doing so is rather different.

brian-brazil:
From https://github.com/cloudflare/ebpf_exporter#examples there seems to be overlap, and your exporter is quite new.

brian-brazil:
I haven't heard back, so I'll presume that this is already covered by the ebpf exporter. It also looks like nothing has happened on this project since.

PR #1354, opened 2019-05-15
Update AlertManager configuration page for AM v0.18

simonpasquier:
Follow-up of https://github.com/prometheus/alertmanager/pull/1863. It should be merged only after AlertManager v0.18 is released. cc @bwplotka

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -445,9 +445,9 @@ links:
 The fields are documented in the [PagerDuty API documentation](https://v2.developer.pagerduty.com/v2/docs/send-an-event-events-api-v2#section-the-images-property).
 
 ```yaml
+href: <tmpl_string>
 source: <tmpl_string>
 alt: <tmpl_string>
-text: <tmpl_string>
 ```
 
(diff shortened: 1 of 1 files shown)


content/docs/alerting/configuration.md
```diff
@@ -603,6 +604,18 @@ OpsGenie notifications are sent via the [OpsGenie API](https://docs.opsgenie.com
 [ http_config: <http_config> | default = global.http_config ]
 ```
 
+### `<responder>`
+
+```yaml
+# One of these fields should be defined.
```
brian-brazil: Exactly or at least?

simonpasquier: I think it's "exactly" but @bwplotka could confirm.

bwplotka: They say nothing about this stricly so I would assume `exactly.
```
 Either id or name of each responder should be provided.You can refer below for example values.
```

simonpasquier:
@brian-brazil ok to merge?

simonpasquier:
Thanks!

PR #1357, opened 2019-05-20
Added WireGuard exporter

MindFlavor:


content/docs/instrumenting/exporters.md
```diff
@@ -190,6 +190,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Transmission exporter](https://github.com/metalmatze/transmission-exporter)
    * [Unbound exporter](https://github.com/kumina/unbound_exporter)
    * [Xen exporter](https://github.com/lovoo/xenstats_exporter)
+   * [WireGuard exporter](https://github.com/MindFlavor/prometheus_wireguard_exporter)
```

MindFlavor:
> Counters should end in _total

Thank you, I will correct that ASAP ([https://github.com/MindFlavor/prometheus_wireguard_exporter/issues/2](https://github.com/MindFlavor/prometheus_wireguard_exporter/issues/2)).

> Export timestamps of events, not time since

Thank you, I will do that too. I hope the numbers given by WireGuard are precise otherwise the timestamp will keep changing anyway. Ref: [https://github.com/MindFlavor/prometheus_wireguard_exporter/issues/3](https://github.com/MindFlavor/prometheus_wireguard_exporter/issues/3).

> Please note your port number on the wiki page, and unless there's some reason you chose this port in particular consider using something from our standard range.

I've just picked the next available one (and it seems another Golang implementation picked the same port). Is it a problem? Should I/we change it?

> I don't think you've handling escaping correctly in label values.

I'll check but AFAIK the labels cannot contain invalid values by construction.

brian-brazil:
It needs to be sorted before I can merge.

> Is it a problem? Should I/we change it?

It's an optional thing, unless it's highly likely to conflict what the other service is.

brian-brazil:
Thanks!

PR #1358, opened 2019-05-25
Adding Thanos to integration of remote write.

bwplotka:
Thanos mainly uses remote read to integrate with Prometheus. Recently Thanos receiver was added so we can put it as `write` integration as well. I added `read as client` as well as to me this is important detail. Thanos integrates with Prometheus as remote read client, so technically it interates with read? Thanos does not expose remote read on it's own though.

Thoughts? (:

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -49,6 +49,7 @@ data volumes.
   * [SignalFx](https://github.com/signalfx/metricproxy#prometheus): write
   * [Splunk](https://github.com/lukemonahan/splunk_modinput_prometheus#prometheus-remote-write): write
   * [TiKV](https://github.com/bragfoo/TiPrometheus): read and write
+  * [Thanos](http://github.com/improbable-eng/thanos): write
   * [VictoriaMetrics](https://github.com/VictoriaMetrics/VictoriaMetrics): write
   * [Wavefront](https://github.com/wavefrontHQ/prometheus-storage-adapter): write
 


content/docs/operating/integrations.md
```diff
@@ -49,6 +49,7 @@ data volumes.
   * [SignalFx](https://github.com/signalfx/metricproxy#prometheus): write
   * [Splunk](https://github.com/lukemonahan/splunk_modinput_prometheus#prometheus-remote-write): write
   * [TiKV](https://github.com/bragfoo/TiPrometheus): read and write
+  * [Thanos](http://github.com/improbable-eng/thanos): write and read client
```
brian-brazil: I thought you only supported write? Can I read from Thanos via remote read?

bwplotka: I explained that in my PR description (:

bwplotka: Happy to move it to `write` only, depending how you define integration here, as we hook into remote read (:

brian-brazil: The question is whether Prometheus can read/write, not whether something uses read/write itself.

brancz: Agreed, for now (and potentially ever) thanos is going to be wrote only.

bwplotka: agree, changed.

content/docs/operating/integrations.md
```diff
@@ -49,6 +49,7 @@ data volumes.
   * [SignalFx](https://github.com/signalfx/metricproxy#prometheus): write
   * [Splunk](https://github.com/lukemonahan/splunk_modinput_prometheus#prometheus-remote-write): write
   * [TiKV](https://github.com/bragfoo/TiPrometheus): read and write
+  * [Thanos](http://github.com/improbable-eng/thanos): write\*
```
brian-brazil: The * is to makes a note below, it's not of any use on its own

brancz: agreed

bwplotka: Any preferred info there? (:  Th

Something like:
```
* Thanos has experimental support for sending metrics via remote write. It uses remote read as the main way to get metrics from Prometheus. 
```

brancz: sgtm

brian-brazil: The experimentalness of things isn't relevant on this page. I'd mention that it doesn't have remote read, but exposes a Prometheus-compatible HTTP API.

bwplotka: > but exposes a Prometheus-compatible HTTP API.

I think this is can confuse users, as most of those solution has this. Also remote-read and `federate` are Prometheus HTTP API as well and Thanos does not support them (: I would stick to simple version:

`Thanos does not expose remote read API, but uses remote read as the main way to get metrics from Prometheus.`

brancz: Talking about the remote read API will just confuse people who don't know Thanos I think.

bwplotka: ok, so bit confused now (: Do we need any `*` then.

bwplotka:
@brian-brazil @brancz 

I removed `*` as we don't have consensus what that could mean (: Clearly there is no good place for Thanos in those integration docs, as it uses Prometheus via remote read to expose valid Prometheus Query API, however this section talks about things that implements remote read and remote write backends, so any extra info seems weird. 

Proposing just `write` as it's better than nothing, and with remote receiver it's true, even though it's very optional way of using Thanos (:

PR #1360, opened 2019-05-28
Add unofficial Go library

valyala:
Add https://github.com/VictoriaMetrics/metrics , which is less powerful than the official Go library, but is easier to use.


brian-brazil:
Thanks for your contribution. We aim to avoid duplicates, so have only one library listed per language/runtime. If there's features you feel the official library is lacking, I'd suggest talking to the maintainer.

valyala:
Got it! Then closing the PR.

PR #1361, opened 2019-05-29
Add Dell Hardware OMSA exporter

galexrt:
Signed-off-by: Alexander Trost <galexrt@googlemail.com>

This adds my [galexrt/dellhw_exporter](https://github.com/galexrt/dellhw_exporter) exporter to the hardware exporter section.
The exporter can export metrics from Dell Hardware using [Dell's OpenManage Server Administrator](https://www.dell.com/support/article/de/de/dedhs1/sln312492/openmanage-server-administrator-omsa?lang=en) service.

If needed, I can provide the `/metrics` output in a few days from a Dell server.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -58,6 +58,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [apcupsd exporter](https://github.com/mdlayher/apcupsd_exporter)
    * [BIG-IP exporter](https://github.com/ExpressenAB/bigip_exporter)
    * [Collins exporter](https://github.com/soundcloud/collins_exporter)
+   * [Dell Hardware OMSA exporter](https://github.com/galexrt/dellhw_exporter)
    * [IBM Z HMC exporter](https://github.com/zhmcclient/zhmc-prometheus-exporter)
    * [IoT Edison exporter](https://github.com/roman-vynar/edison_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

I can't see the exporter metrics, but I'm wondering if things like volts and temperature have appropriate units. Also, if things like storage controllers should have a label for the controller in question.

galexrt:
@brian-brazil Thank you very much for the feedback and the merge!

Volts and temps are exported as floats (e.g., 
https://github.com/galexrt/dellhw_exporter/blob/master/collector/volts.go#L29-L38, https://github.com/galexrt/dellhw_exporter/blob/master/collector/temps.go#L29-L38).
Looking at the other collectors right now, the metrics are mostly exposed as floats right now. "appropriate unit" for the metrics is definitely a point, which I will look into again soon, e.g., that the simpler three way state (`0`, `1`, `2`) metrics are using a better type than float.

The storage controllers and alike, have their id added as a label, e.g., https://github.com/galexrt/dellhw_exporter/blob/master/pkg/omreport/omreport.go#L169-L177 the label as added to the created Prometheus metrics.

SuperQ:
@galexrt By "units", Brian meant "Celsius" or "Fahrenheit", not float. :smile:

galexrt:
Haha. I see, the temperatures are in Celsius. I'll add that to the readme of the temperature collector on the weekend.

PR #1362, opened 2019-06-10
add RocketMQ expoter link

hdchen:
RocketMQ  graduated from the community in 2017 to become Apache's top project，
    RocketMQ covers more than 40% of the news domain scenarios in China alone.
    I found that the RocketMQ project does not have a very good open source solution for monitoring, so I personally developed a prometheus RocketMQ exporter, which can very well collect RocketMQ monitoring indicators and leverage the powerful ProQL capabilities and completeness of prometheus。
     The repository address is as follows: https://github.com/apache/rocketmq-exporter. Now, many users in the Apache RocketMQ community are using the prometheus RocketMQ exporter as their monitoring solution for the operation and maintenance of RocketMQ. This project has also received strong support from the founder of Apache RocketMQ, vongosling

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -90,6 +90,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [MQTT blackbox exporter](https://github.com/inovex/mqtt_blackbox_exporter)
    * [RabbitMQ exporter](https://github.com/kbudde/rabbitmq_exporter)
    * [RabbitMQ Management Plugin exporter](https://github.com/deadtrickster/prometheus_rabbitmq_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

Some suggestions to improve your exporter per https://prometheus.io/docs/instrumenting/writing_exporters/:

If possible use counters rather than gauges. From a quick peek RocketMQ has them internally, but I'm not sure if it's exposing them.

rocketmq_producer_offset sounds like a gauge. Also, are the units bytes there? 

rocketmq_producer_put_size is missing units, and as a counter should end in _total.

rocketmq_group_get_latency is missing a unit, and should be a Summary without quantiles if possible.

cluster and broker sound like target labels, which should come from Prometheus rather than the exporter.

Your port number isn't in the usual range from the wiki page.

PR #1363, opened 2019-06-11
Add Tankerkönig API Exporter

lukasmalkmus:
Updated https://github.com/lukasmalkmus/tankerkoenig_exporter recently and thought why not add it to the docs.

@brian-brazil

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -126,6 +126,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Pagespeed exporter](https://github.com/foomo/pagespeed_exporter)
    * [Rancher exporter](https://github.com/infinityworksltd/prometheus-rancher-exporter)
    * [Speedtest exporter](https://github.com/nlamirault/speedtest_exporter)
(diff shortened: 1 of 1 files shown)


lukasmalkmus:
Ups, did the edit online, will fix.

lukasmalkmus:
Fixed.

brian-brazil:
Thanks for sharing. You're taking a secret on the command line, which isn't secure. Could you move that elsewhere?

The exporter will be simpler and more correct if you use ConstMetrics rather than direct instrumentation.

lukasmalkmus:
> Thanks for sharing. You're taking a secret on the command line, which isn't secure. Could you move that elsewhere?

What is recommended here? Environment?

> The exporter will be simpler and more correct if you use ConstMetrics rather than direct instrumentation.

Alright, will address.

brian-brazil:
> What is recommended instead? Environment variables for secrets?

Files are best, but the important thing is that it's not on the command line.

lukasmalkmus:
Finally had a few mins to fix. Transitioned to direct instrumentation and read Token from environment instead of command line.

Cheers.

brian-brazil:
Your docker example still has the token on the command line.

lukasmalkmus:
Ups, resolved.

brian-brazil:
Thanks!

PR #1364, opened 2019-06-13
Add a link to a new graphite integration

InformatiQ:
A project to integrate with graphite in read and write.


brian-brazil:
We try to avoid duplicates in this list to reduce confusion for users. Is this strictly better than what we currently list? 

Also as what's currently listed is kinda official and historical, can this one insert data into Graphite in the same format?

brian-brazil:
Also, you've a password on the command line in the URL in the README. That's insecure, and should be fixed.

brian-brazil:
We never heard back from you, if you're still interested in replacing what we've got currently listed please let us know.

PR #1365, opened 2019-06-17
Add Splunk(read and write) to list of remote storages

kebe7jun:
Ropee is a prometheus remote storage adapter for splunk designed by @DaoCloud.

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -47,7 +47,6 @@ data volumes.
   * [OpenTSDB](https://github.com/prometheus/prometheus/tree/master/documentation/examples/remote_storage/remote_storage_adapter): write
   * [PostgreSQL/TimescaleDB](https://github.com/timescale/prometheus-postgresql-adapter): read and write
   * [SignalFx](https://github.com/signalfx/metricproxy#prometheus): write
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for sharing. It looks like this is a superset of the currently listed one, so you can replace it.

Your README has a password on the commandline, can you move it somewhere more secure?

There's also no need for read_recent to be true.

kebe7jun:
Thanks for your comment, I have updated my [repo's README](https://github.com/kebe7jun/ropee) and removed old Splunk adaptor.

brian-brazil:
You're still taking a password on the command line, both in the docker example and the adapter itself.

kebe7jun:
> You're still taking a password on the command line, both in the docker example and the adapter itself.

@brian-brazil 
I have removed the auth from command line for secure. 
https://github.com/kebe7jun/ropee#configuring-prometheus

brian-brazil:
I still see it in the code: https://github.com/kebe7jun/ropee/blob/master/main.go#L67

We ask that there's no obvious security issues before listing something here, which basically means not exposing passwords on the command line.

kebe7jun:
Sorry about that, I have removed the password option from command.
commit: https://github.com/kebe7jun/ropee/commit/5ae55ed39a4d977e67c4e073426318e487ded09e

brian-brazil:
Thanks!

PR #1367, opened 2019-06-18
 name error. It is "TencentCloud"

Himer:
name error. It is "TencentCloud"..


brian-brazil:
Can you squash/add the DCO?

PR #1370, opened 2019-06-18
remove ref to HipChat

eightnoneone:
HipChat is no more. Replaced specific product references with generic descriptions.


brian-brazil:
Thanks!

PR #1372, opened 2019-06-18
edit links to be in context. 'click here' must die.

eightnoneone:
Signed-off-by: Matt Warren <matt@tramat.com>


RichiH:
Thanks!

PR #1373, opened 2019-06-21
Monitoring Kubernetes Ingress

nbkrause:
Hi,

I wanted to spin up Prometheus in Kubernetes so I could monitor ingress to my cluster. I spent some time hacking away at getting a working config and I felt it would be useful to write up my process for documentation.

I used the [coreos/prometheus-operator](https://github.com/coreos/prometheus-operator) to get Prometheus running in cluster and am controlling Ingress with Envoy via [Ambassador](https://github.com/datawire/ambassador). 

I have documented the steps you need to take to get:

1. An example application running in Kubernetes 
2. Ambassador running and exposing the application to the internet
3. Prometheus running in cluster and scraping ingress statistics from Envoy

Let me know if you have any feedback on ways to improve this document. Just thought this would be a useful addition to the guides.


brancz:
I feel like this would probably make more sense to be in the Prometheus Operator docs rather than these.

PR #1375, opened 2019-06-25
Add criteo cassandra exporter

guilhem:


brian-brazil:
Thanks for sharing. We try to avoid duplication in this list to avoid confusion for users. Cassandra is already covered by the JMX exporter, does this do something that it can't?

BarthV:
https://github.com/criteo/cassandra_exporter#design-explanation
```
JMX is an heayweight RPC mechanism and some cassandra metrics calls are expensive to scrap (i.e: snapshots size) as they trigger some heavy operations for cassandra. Not caching results mean that you can bring down your nodes by just requesting the metrics page
```

brian-brazil:
Caching is not recommended for the exporter guidelines, and you can always ignore an mBean for the jmx exporter. I'm trying not to list things that are only minor deviations from what's already there.

erebe:
> Thanks for sharing. We try to avoid duplication in this list to avoid confusion for users. Cassandra is already covered by the JMX exporter, does this do something that it can't?

Hello,
I am the author of Cassandra exporter, not that I care much if the name appears or not in the list, except being proud, but regarding your question.
Initially we were using jmx_exporter but decided to fork it to make cassandra_exporter for the following reasons.

- Cassandra expose EstimatedHistogram that jmx_exporter is not able to export properly
- Being able to filter on attribute of MBeans. Cassandra expose a lot of unhelpful metrics like 999 percentile which put too much memory pressure on our prometheus. Even we can blacklist things in prometheus config, in order to get a single point of config, this exporter allows to filter attributes of MBeans
- Caching results, this is not the prometheus philosophy, but some jmx calls trigger heavy operations (like scanning all the sstables on disk) which can bring down the node if scrapped too often. (i.e: a bad deployment of prometheus)
- Simple configuration file, jmx exporter is very flexible but the final config file is a nightmare to maintain

That was the main reasons for forking jmx_exporter, note that I haven't followed the developpement of jmx_exporter since them.

Cheers

brian-brazil:
I'm not seeing anything there that can't be done by the jmx exporter. I don't think it makes sense to have an entire exporter in preference to a config file for a more generic exporter.

PR #1376, opened 2019-06-27
docs/instrumentation/exporters: Add modbus exporter

mxinden:
Prometheus exporter which retrieves stats from a modbus tcp system and
exports them via HTTP for Prometheus consumption.

Very much looking forward to any feedback!


brian-brazil:
From looking at the code I think this exporter is broken as-is, as it's an snmp-style exporter however not using const metrics and sharing metrics across requests. The Counter handling is also accordingly broken, and there's no locking. If labelnames varied across modules for a metric I'd expect this to panic.

In addition it's putting in the module as a label on all results, which is not in line with our guidelines. The KeepAlive setting appears unused. I'd suggest getting everything on a single http port, and using a port from the wiki.

brian-brazil:
Looks like this particular one is abandoned. @RichiH do you want to send a PR to add your one?

PR #1377, opened 2019-06-27
Use Yaml indentation hints to avoid otherwise needless parentheses

beorn7:
As inspired by discussions of the alert mixin.

### DIFF
--- a/content/docs/practices/rules.md
+++ b/content/docs/practices/rules.md
@@ -47,6 +47,11 @@ conflicts and give you more useful metrics and alerts.
 
 ## Examples
 
+_Note the indentation style with outdented operators on their own line between
+two vectors. To make this style possible in Yaml, [block quotes with an
+indentation indicator](https://yaml.org/spec/1.2/spec.html#style/block/scalar)
+(e.g. `|2`) are used._
+
 Aggregating up requests per second that has a `path` label:
 
 ```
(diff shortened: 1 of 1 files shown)


brian-brazil:
I'd considered this previously, however `|2` is not something I'd expect users to be aware of (I certainly wasn't until I dug through the YAML docs). Thus what we have is likely better in terms of understanding and maintainability.

beorn7:
On the contrary, I think `|2` is really important for more readable rules, and thus we should spread the goodness.

beorn7:
Speaking with the experience of shepherding 120 engineers using 50k lines of Prometheus rules files.

beorn7:
I added an explicit hint so that the usage of `|2` doesn't go unnoticed.

brian-brazil:
That could work. @juliusv what do you think?

beorn7:
I have now linked the relevant section of the Yaml spec.

juliusv:
I'm really unhappy with this whole issue with YAML as well :( I'm not sure which variant I find less bad. Maybe the one with the obscure indentation feature, but subsequent lack of paren requirements.

beorn7:
While initially weird, it's not that hard to get used to `|2` (or whatever leading indentation you need). There was very little friction at SC once it was introduced.

juliusv:
Ship it!

brian-brazil:
:+1:

beorn7:
Thanks.

PR #1378, opened 2019-07-01
announce ss exporter

cherusk:


brian-brazil:
Thanks for sharing. This looks like it'd have extremely high cardinality, and thus I'm not sure it'd be appropriate to list it.

cherusk:
Nigh, I've specified the circumstances and mitigation steps to reduce cardinality systematically to a viable level. It highly depends on the actual scenario. No reason to debar it.

brian-brazil:
Any safe uses of this would be very niche, and it'd likely cause more problems than it solves. You're free to announce it on the -users list.

PR #1379, opened 2019-07-02
naming: add mention of version pseudo-metrics and _info suffix

cdanis:
The _info suffix is used in the Prometheus server itself for this
purpose, so this seems to be something of an informal standard;
let's make it formal.


cdanis:
It's been over a week -- anything else needed here?

PR #1381, opened 2019-07-03
Update routing-tree to add tooltip that shows receiver configs on text mouseOver

jfchevrette:
This is a rather small change (read: hack) I am using on a local branch which I find useful when working with complex receiver configs. It shows a tooltip when hovering on top of the receiver name in the tree. This currently shows the raw yaml of the receiver config.

I'm far from a javascript expert, so I'll be the first to admit the code is probably not that great but possibly 'good enough' to be considered, hence why I'm opening this PR.

Additional ideas:
- Add a checkbox to toggle showing receiver configs or not (disabled by defaul?)
- Improve format and readability of the tooltip content since a complex recipient config can span many lines.

### DIFF
--- a/static/routing-tree.js
+++ b/static/routing-tree.js
@@ -291,6 +291,7 @@ function aggregateMatchers(node) {
 }
 
 function getReceiverConfigs(name, receivers) {
+  if (!receivers) return;
   return receivers.find(function(e){
     return e.name == name;
   });
--- a/static/routing-tree.js
+++ b/static/routing-tree.js
@@ -232,7 +232,11 @@ function update(root) {
       .on("mouseover", function(d) {
         d3.select(this).style("fill", color);
 
-        text = jsyaml.dump(d.receiver_configs).replace(/ /g, '\u00a0').split("\n");
+        text = ["<receiver config missing>"];
+        if (typeof(d.receiverConfig) !== 'undefined') {
+          text = jsyaml.dump(d.receiverConfig).replace(/ /g, '\u00a0').split("\n");
+        }
+
         text.forEach(function(t) {
           tooltip.append("div").text(t);
         });
@@ -290,7 +294,7 @@ function aggregateMatchers(node) {
   return matchers
 }
 
-function getReceiverConfigs(name, receivers) {
+function getReceiverConfig(name, receivers) {
   if (!receivers) return;
   return receivers.find(function(e){
     return e.name == name;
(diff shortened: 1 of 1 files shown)


static/routing-tree.js
```diff
@@ -167,11 +168,13 @@ function massage(root) {
 
   root.matchers = matchers;
 
+  root.receiver_configs = getReceiverConfigs(root.receiver, receivers);
```
juliusv, line 171:
```suggestion
  root.receiverConfig = getReceiverConfig(root.receiver, receivers);
```

(singular since it's just one receiver config per node, and camelCase for style consistency)

static/routing-tree.js
```diff
@@ -225,7 +228,25 @@ function update(root) {
       .attr("dy", ".31em")
       .attr("text-anchor", function(d) { return d.x < 180 ? "start" : "end"; })
       .attr("transform", function(d) { return d.x < 180 ? "translate(8)" : "rotate(180)translate(-8)"; })
-      .text(function(d) { return d.receiver; });
+      .text(function(d) { return d.receiver; })
+      .on("mouseover", function(d) {
+        d3.select(this).style("fill", color);
+
+        text = jsyaml.dump(d.receiver_configs).replace(/ /g, '\u00a0').split("\n");
```
juliusv, line 235: Would be nice to detect `undefined` receiver configs here and not try to render them as YAML, but as `<receiver config missing>` or something like that.

static/routing-tree.js
```diff
@@ -167,11 +168,13 @@ function massage(root) {
 
   root.matchers = matchers;
 
+  root.receiverConfig = getReceiverConfig(root.receiver, receivers);
```
juliusv, line 171: You updated this line, but you'll also need to update all the other references to those names.

jfchevrette: Indeed, sorry but this oversight. I've updated the code accordingly and also rebased on top of master. I've tested a few different scenarios locally and everything seem fine

static/routing-tree.js
jfchevrette: I've added your suggestion in the latest commit

jfchevrette:
Implemented suggestions & rebased on top of master. I've tested locally with a few different configs and things look good.

I've sent this PR over a year ago so it took me some time to get back into it. I personally don't really need this change anymore, but I still think it's a nice addition.

juliusv:
Awesome, thank you! Ah sorry, I didn't even see that you first submitted this PR a year ago. I only noticed it when @roidelapluie assigned me a couple of days ago :)

juliusv:
:+1:

Issue #1304, opened 2019-03-09
Links in prometheus config page are bad


roidelapluie:
1. Go to https://prometheus.io/docs/prometheus/latest/configuration/configuration/
2. Go to `<scrape_config>`
3. Click on  `<azure_sd_config>` in scrape config example
4. expectation: arrive on azure_sd_config config section
5. what appens instead: arrive on azure_sd_config element in TOC

PR #1382, opened 2019-07-04
Fixed anchor links in Configuration pages

pracucci:
The generation of "Configuration" pages runs the `config_linker.rb` filter to add anchors to `<placeholder>` in code blocks. The generation of these anchors doesn't "slugify" the placeholder text, thus leading to anchors like `<tls_config>`. This is conflicting with the default anchors generated by the markdown compiler for the HTML headers, as well as the extra anchors injected by `add_anchors.rb`.

In this PR I'm uniforming the way anchors are generated, to always "slugify" the text, and have links within the configuration code blocks working correctly. After this change, the `id` in the TOC also matches with the one of non-"Configuration" pages, which is currently different due to side effects introduced by the current logic. 

Fixes https://github.com/prometheus/docs/issues/1304 and https://github.com/prometheus/docs/issues/833.

@brian-brazil 

### Example

**Before the change**:

```
[ <a href="#&lt;tls_config&gt;">&lt;tls_config&gt;</a> ]
```

**After the change**:

```
[ <a href="#tls_config">&lt;tls_config&gt;</a> ]
```

### DIFF
--- a/lib/filters/config_linker.rb
+++ b/lib/filters/config_linker.rb
@@ -11,8 +11,8 @@ class ConfigLinker < ::Nanoc::Filter
     configs = doc.xpath('//pre//code')
 
     configs.each do |config|
-      definitions.each do |text, html|
-        config.inner_html = config.inner_html.gsub(html, %(<a href="##{text}">#{html}</a>))
+      definitions.each do |anchor, html|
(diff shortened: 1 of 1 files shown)


brian-brazil:
That seems to do it, @grobie do you want to take a look?

PR #1383, opened 2019-07-04
Enforce utf-8 encoding on data sources

pracucci:
[Nanoc defaults to the current environment encoding](https://nanoc.ws/doc/troubleshooting/#wrong-input-encoding). If - for any reason - the default encoding is not `utf-8`, the compilation breaks with the following error:

```
Nanoc::DataSources::Filesystem::Errors::InvalidEncoding: Could not read repositories/prometheus.git/latest (2.10)/docs//storage.md because the file is not valid US-ASCII.
```

In this PR, I'm enforcing `utf-8` encoding on all data sources.

@brian-brazil

### DIFF
--- a/nanoc.yaml
+++ b/nanoc.yaml
@@ -71,6 +71,7 @@ data_sources:
       name: 'latest (2.10)'
       repository: https://github.com/prometheus/prometheus.git
       refspec: release-2.10
+    encoding: utf-8
   -
     type: repo_docs
     identifier_type: legacy
@@ -80,6 +81,7 @@ data_sources:
       repository: https://github.com/prometheus/prometheus.git
       refspec: release-2.10
       canonical: /docs/prometheus/latest/
+    encoding: utf-8
   -
     type: repo_docs
     identifier_type: legacy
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can we set this in one place?

pracucci:
> Can we set this in one place?

I'm not familiar with `nanoc`. I looked for it in the [`nanoc` doc](https://nanoc.ws/doc/data-sources/#configuration) and I haven't found any reference about a global setting (which I agree would be ideal).

brian-brazil:
@grobie Do you know if there's one?

pracucci:
I checked the source code:

- `nanoc` doesn't offer a way to globally set the `encoding` on the `Filesystem` data source (`RepoDocsDataSource` extends it)
- Ruby's `File.read()` infers the default encoding from the `LANG` environment variable

We may enforce it overriding the `LANG` env variable in the `Makefile`, but it's value may change on different systems (ie. `UTF-8` vs `C.UTF-8`). All in all, I would personally suggest to enforce it for each data source in the `nanoc.yaml` (this PR), but I'm open to suggestions and further discussion.

grobie:
I'm unhappy about the current repo docs config structure anyway, it's too convoluted when tracking multiple versions. I'd hope to simplify that in a future version. For now, it's just one line more to copy&paste, I'd go forward with that.

pracucci:
@brian-brazil Do you think there's interest into merging this PR?

brian-brazil:
Thanks!

PR #1387, opened 2019-07-08
Update PagerDuty/image config

asherf:
see https://github.com/prometheus/alertmanager/pull/1931 and https://github.com/prometheus/alertmanager/issues/1784
https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L228


simonpasquier:
Thanks for the PR but it is already covered by https://github.com/prometheus/docs/pull/1354 which includes other configuration changes for v0.18.0.

PR #1390, opened 2019-07-15
Mention Packet file_sd

t0mk:
This PR adds link to the Packet host service discovery repo in the integrations listing


brian-brazil:
Thanks! In your example prometheus.yml you want instance rather than hostname, see https://www.robustperception.io/target-labels-are-for-life-not-just-for-christmas

PR #1393, opened 2019-07-18
Added an additional link to a new ibm mq java based exporter

sberdyshev:
Signed-off-by: Sergey Berdyshev <sberdyshev@cinimex.ru>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -83,6 +83,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [EMQ exporter](https://github.com/nuvo/emq_exporter)
    * [Gearman exporter](https://github.com/bakins/gearman-exporter)
    * [IBM MQ exporter](https://github.com/ibm-messaging/mq-metric-samples/tree/master/cmd/mq_prometheus)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. To avoid confusion for users and to encourage cooperation and discourage NIH we try to avoid duplicates in this list. How does this differ from the existing one we list, which still appears to be maintained?

sberdyshev:
Thnks for the reviewing pr.
The main point - the exporter is written in java that allows to easily run it in AIX environment. 
I stated it here: https://groups.google.com/forum/?fromgroups#!topic/prometheus-developers/NbE7zt34aAw
As extra we tried to meet prometheus standards on metrics naming (default IBM MQ metrics names are confusing) and to provide the best docs we could offer.

brian-brazil:
Have you talked to the maintainers of the existing exporter to take on your improvements?

sberdyshev:
No, I haven't.
As I stated before, we wanted exporter to be java-based in order to make it possible to run it in AIX environment with no effort.

brian-brazil:
Considering that AIX is from IBM and the existing exporter is also related to them, I'd suggest talking to them. Having an uncommon deployment scenario is not enough reason to have duplication in this list.

brian-brazil:
Per above, we aim not to have duplicates and there's an existing maintained official exporter in this space.

PR #1397, opened 2019-07-19
Add link to openstack_client_exporter

ctrlaltdel:


brian-brazil:
How does this differ from the openstack exporter we already list, and have you talked to the listed one's maintainers about changes you'd like to make?

ctrlaltdel:
They have both different functionalities and a different level of access to the OpenStack APIs.openstack_client_exporter works with an unprivileged user within a single project to ensure that user facing functionalities are working. It does so by creating an instance and sshing into it, uploading and retrieving files from the object store, ...The existing openstack_exporter needs admin credentials and is useful for retrieving global cloud metrics but create or delete an resources by itself.I'm not sure it would make much sense to try to combine both of them in a single exporter.On Jul 19, 2019 21:51, Brian Brazil <notifications@github.com> wrote:How does this differ from the openstack exporter we already list, and have you talked to the listed one's maintainers about changes you'd like to make?

brian-brazil:
Okay, so this is more of a blackbox exporter. Can you put Blackbox in the entry name?

openstack_client_spawn_timing and friends should include their unit in the metric name, openstack_client_spawn_success and openstack_client_object_store_success should not have the error as a label - that's a matter for logs.

ctrlaltdel:
> Okay, so this is more of a blackbox exporter. Can you put Blackbox in the entry name?

Yep, done.
 
> openstack_client_spawn_timing and friends should include their unit in the metric name

Good point, I'll look into releasing a new version with s/_timing/_seconds/.

> openstack_client_spawn_success and openstack_client_object_store_success should not have the error as a label - that's a matter for logs.

Conceptually, I agree. However, it is a pretty convenient way to create annotations in a grafana dashboard with the reason why something failed.

brian-brazil:
> Conceptually, I agree. However, it is a pretty convenient way to create annotations in a grafana dashboard with the reason why something failed.

Doing logging in a metrics system can easily lead to both cardinality and semantic issues, it should be avoided.

PR #1401, opened 2019-07-23
Improve base units table

beorn7:
- Fixed numerous typographical and puncuation nits.
- Fixed weight → mass.
- Added note about deliberately using grams rather than kilogram
  and aligned existing similar note about celsius with it.
- Call out that `bytes` for bits is not a typo.
- Reworded and -formatted the note for `ratio` as the asterisk seemed
  needless and disrupting the flow.

@janvanoort

### DIFF
--- a/content/docs/practices/naming.md
+++ b/content/docs/practices/naming.md
@@ -75,13 +75,13 @@ The list is not exhaustive.
 | Family | Base unit | Remark | 
 | -------| --------- | ------ |
 | Time   | seconds   |        |
-| Temperature | celsius | Celsius is the most common one encountered in practice |
+| Temperature | celsius | _celsius_ is preferred over _kelvin_ for practical reasons. |
 | Length | meters | |
 | Bytes  | bytes | | 
-| Bits   | bytes | |
-| Percent | ratio(*) | Values are 0-1. <br/> *) Usually 'ratio' is not used as suffix, but rather A\_per\_B. Exceptions are e.g. disk\_usage\_ratio  |
+| Bits   | bytes | To avoid confusion combining different metrics, always use _bytes_, even where _bits_ appear more common. |
+| Percent | ratio | Values are 0–1 (rather than 0–100). `ratio` is only used as a suffix for names like `disk_usage_ratio`. The usual metric name follows the pattern `A_per_B`. |
 | Voltage | volts | |
-| Electric current | amperes | | 
+| Electric current | amperes | |
 | Energy | joules | |
-| Weight | grams | |
+| Mass   | grams | _grams_ is preferred over _kilograms_ to avoid issues with the _kilo_ prefix. |
  


content/docs/practices/naming.md
```diff
@@ -75,13 +75,13 @@ The list is not exhaustive.
 | Family | Base unit | Remark | 
 | -------| --------- | ------ |
 | Time   | seconds   |        |
-| Temperature | celsius | Celsius is the most common one encountered in practice |
+| Temperature | celsius | _celsius_ is preferred over _kelvin_ for practical reasons. |
```
brian-brazil: I'd say more than "practical reasons"

I'd leave it as Celsius rather than _celsius_ as we're talking about the unit names rather than the string in the metric name.

content/docs/practices/naming.md
```diff
@@ -75,13 +75,13 @@ The list is not exhaustive.
...
-| Bits   | bytes | |
-| Percent | ratio(*) | Values are 0-1. <br/> *) Usually 'ratio' is not used as suffix, but rather A\_per\_B. Exceptions are e.g. disk\_usage\_ratio  |
+| Bits   | bytes | _Sic!_ |
+| Percent | ratio | Values are 0–1 (rather than 0–100). `ratio` is only used as a suffix for names like `disk_usage_ratio`. The usual suffix follows the pattern `A_per_B`. |
 | Voltage | volts | |
-| Electric current | amperes | | 
+| Electric current | amperes | |
 | Energy | joules | |
-| Weight | grams | |
+| Mass   | grams | _grams_ is preferred over _kilograms_ for practical reasons. |
```
brian-brazil: Ditto, it's due to the prefix.

content/docs/practices/naming.md
```diff
@@ -75,13 +75,13 @@ The list is not exhaustive.
...
 | -------| --------- | ------ |
 | Time   | seconds   |        |
-| Temperature | celsius | Celsius is the most common one encountered in practice |
+| Temperature | celsius | _celsius_ is preferred over _kelvin_ for practical reasons. |
 | Length | meters | |
 | Bytes  | bytes | | 
-| Bits   | bytes | |
-| Percent | ratio(*) | Values are 0-1. <br/> *) Usually 'ratio' is not used as suffix, but rather A\_per\_B. Exceptions are e.g. disk\_usage\_ratio  |
+| Bits   | bytes | _Sic!_ |
+| Percent | ratio | Values are 0–1 (rather than 0–100). `ratio` is only used as a suffix for names like `disk_usage_ratio`. The usual suffix follows the pattern `A_per_B`. |
```
brian-brazil: The usual metric name

content/docs/practices/naming.md
```diff
@@ -75,13 +75,13 @@ The list is not exhaustive.
 | Family | Base unit | Remark | 
 | -------| --------- | ------ |
 | Time   | seconds   |        |
-| Temperature | celsius | Celsius is the most common one encountered in practice |
+| Temperature | celsius | _celsius_ is preferred over _kelvin_ for practical reasons. |
 | Length | meters | |
 | Bytes  | bytes | | 
-| Bits   | bytes | |
-| Percent | ratio(*) | Values are 0-1. <br/> *) Usually 'ratio' is not used as suffix, but rather A\_per\_B. Exceptions are e.g. disk\_usage\_ratio  |
+| Bits   | bytes | _Sic!_ |
```
brian-brazil: I'd explain a little here.

content/docs/practices/naming.md
beorn7: The string in the metric name would be `celsius`. I used italics here because it's a verbatim reproduction of the entry in the "Base unit" column. I'm happy to capitalize, but then we should do in in the base unit column, too.

beorn7: What other reasons would you mention here?

brian-brazil: The reason is related to the concepts of Celsius and Kelvin, not the strings `celsius` and `kelvin` though.

The existing one is what I'd use, Celsius is much more common than Kelvin

content/docs/practices/naming.md
beorn7: I'm not sure if it is worth it to discuss the detailed reason in a short note like this. I'll add something about the prefix to satisfy your remark.

janvanoort:
Just to be pedantic: the SI unit for temperature is the kelvin ( all
lowercase ), abbreviated with an uppercase K.

If and when you speak about Celsius, the actual unit name is the *degree
Celsius*, not just "Celsius".

Jan van Oort

Head of Research

KIVU Technologies GmbH

www.kivu.tech
<https://www.google.com/url?q=http://www.kivu.tech/&sa=D&ust=1510066916265000&usg=AFQjCNG-LM2XmW9HyYfBBtKpazy1m_XcNg>

content/docs/practices/naming.md
beorn7: Again, I'm not using `celsius` and `kelvin`. I'm using _celsius_ and _kelvin_ exactly because I'm referring to units, _not_ to strings. With capitalization, that would be kind of clear as well, but then we need to capitalize in the _base unit_ column as well for consistency. Do you want me to do that?

I would not needlessly go down the rabbithole of discussing what unit is more common. I wrote a master thesis and a PhD thesis and read gazillions of papers that all used _K_ and never _C_. 300 million US Americans will happily swear that _F_ is most common.

I think the true reason is that most uses of Prometheus metric for temperature is things like CPU temperature or even temperature of cat beds, where the numbers get really high if we used _K_. Also, _C_ is probably what most sensors give you readily. But again, I wouldn't go down the rabbithole of discussing it scientifically. That's why staying vague here is most productive. We essentially only want to signal that we thought about _K_, to avoid the next bug report telling us that _K_ is the SI base unit.

content/docs/practices/naming.md
beorn7: Done (although it's again a bit long for a short remark).

beorn7:
> Just to be pedantic: the SI unit for temperature is the kelvin ( all
lowercase ), abbreviated with an uppercase K.
> 
> If and when you speak about Celsius, the actual unit name is the *degree
Celsius*, not just "Celsius".

Both even more reasons to use _italics_ as I suggested because that just refers to the content of the _base unit_ column, where we will probably be forgiven to not follow typographical standards.

content/docs/practices/naming.md
brian-brazil: > I'm using celsius and kelvin exactly because I'm referring to units, not to strings. 

Oh sorry, my mistake. I had mis-parsed the markdown.

content/docs/practices/naming.md
```diff
@@ -75,13 +75,13 @@ The list is not exhaustive.
 | Family | Base unit | Remark | 
 | -------| --------- | ------ |
 | Time   | seconds   |        |
-| Temperature | celsius | Celsius is the most common one encountered in practice |
+| Temperature | celsius | _celsius_ is preferred over _kelvin_ for practical reasons. |
 | Length | meters | |
 | Bytes  | bytes | | 
-| Bits   | bytes | |
-| Percent | ratio(*) | Values are 0-1. <br/> *) Usually 'ratio' is not used as suffix, but rather A\_per\_B. Exceptions are e.g. disk\_usage\_ratio  |
+| Bits   | bytes | To avoid confusion combining different metrics, always use _bytes_, even where _bits_ appear more commen. |
```
brian-brazil: common

beorn7: Just saw it. Fixed.

beorn7:
OK, at least it's less bad now. :o)

PR #1403, opened 2019-07-25
Pushgateway doc — added link to node.js client

razonrus:


content/docs/instrumenting/pushing.md
```diff
@@ -24,3 +24,5 @@ class.
  * For use from Python see [Exporting to a Pushgateway](https://github.com/prometheus/client_python#exporting-to-a-pushgateway).
 
  * For use from Ruby see the [Pushgateway documentation](https://github.com/prometheus/client_ruby#pushgateway).
+ 
+ * For use from Node.js see the [Pushgateway documentation](https://github.com/siimon/prom-client#pushgateway).
```
brian-brazil: We only link the official clients here, other clients are on the general client library list.

beorn7: Perhaps we can put an explanation here, along the lines: “The links above are only for official client libraries. Unofficial client libraries may also offer means to push to the Pushgateway.”

brian-brazil: We could, this is the first time this has come up tbh.

beorn7:
I have created #1437 for the doc clarification.

Closing this one.

PR #1404, opened 2019-07-25
Adds link to a collection of Prometheus alerting rules

samber:


brian-brazil:
Last I checked we encourage such things to live with the given exporter. Alerts also tend to be very org-specific, as the symptoms are tied to that org's SLOs.

PR #1406, opened 2019-07-26
Fix grammar.

shields-fn:


brian-brazil:
This seems grammatically fine to me.

shields-fn:
How about "There are various client libraries"?

PR #1414, opened 2019-08-09
Add RabbitMQ to software that exposes Prometheus metrics natively

gerhard:
RabbitMQ - as of v3.8.0 - exposes Prometheus metrics directly.

The link points to the next version of RabbitMQ. We will update it to https://www.rabbitmq.com/prometheus.html as soon as v3.8.0 becomes generally available.

cc @michaelklishin

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -234,6 +234,7 @@ separate exporters are needed:
    * [Netdata](https://github.com/firehol/netdata)
    * [Pretix](https://pretix.eu/)
    * [Quobyte](https://www.quobyte.com/) (**direct**)
+   * [RabbitMQ](https://rabbitmq.com/prometheus.html)
    * [RobustIRC](http://robustirc.net/)
    * [ScyllaDB](http://github.com/scylladb/scylla)
    * [Skipper](https://github.com/zalando/skipper)


brian-brazil:
This doesn't sound like it's directly instrumented, or even rabbitmq exposing Prometheus metrics considering it requires a 3rd party plugin. The existence of `collect_statistics_interval` also indicates that this is not direct instrumentation. Can you expand?

gerhard:
RabbitMQ ships with built-in plugins, such as `rabbitmq_management`, `rabbitmq_federation` etc. One of the plugins that RabbitMQ 3.8.0 ships with is `rabbitmq_prometheus`. When this plugin is enabled, it exposes Prometheus metrics on a dedicated TCP port, 15692 by default.

In other words, to have RabbitMQ 3.8.0 expose metrics in Prometheus format, the only requirement is to enable this built-in functionality. No plugin needs to be installed & no exporter needs to be used.

Demo:

```bash
$ docker run --name rmq_prom --rm --publish 15692:15692 rabbitmq:3.8-rc
...
2019-08-10 12:56:16.623 [info] <0.8.0> Server startup complete; 0 plugins started.

$ docker exec rmq_prom rabbitmq-plugins enable rabbitmq_prometheus
Enabling plugins on node rabbit@cff0baef1f5e:
rabbitmq_prometheus
The following plugins have been configured:
  rabbitmq_management_agent
  rabbitmq_prometheus
  rabbitmq_web_dispatch
Applying plugin configuration to rabbit@cff0baef1f5e...
The following plugins have been enabled:
  rabbitmq_management_agent
  rabbitmq_prometheus
  rabbitmq_web_dispatch

started 3 plugins.

$ curl -s localhost:15692/metrics | head -n 3
# TYPE erlang_mnesia_held_locks gauge
# HELP erlang_mnesia_held_locks Number of held locks.
erlang_mnesia_held_locks{node="rabbit@cff0baef1f5e",cluster="rabbit@cff0baef1f5e"} 0
```

brian-brazil:
I've had a poke around, and it seems this is a very new exporter not really based on what's already out there. Having a piece of software exposing metrics out of the box is preferred, so I'm willing to consider having this replace the ones we already list (we try to avoid dupllicates) if you can indicate if it's largely a superset of the other options and compliant with exporter guidelines.

From a quick peek the global_labels look off, counters lack _total, and many units aren't right.

michaelklishin:
@brian-brazil can you elaborate on what looks "off" specifically?

This exporter does not expose a lot more metrics as the metrics available haven't meaningfully changed (there were some additions, though). However, as soon as RabbitMQ 3.8.0 goes. GA, I predict that the existing Prometheus exported will very soon become unmaintained. Max shelf life it has is that of RabbitMQ 3.7.x, which is 6 months after 3.8.0 final. Why? Because for better or worse, a 3rd party plugin stands no chance against one that is is built-in and [documented](http://next.rabbitmq.com/prometheus.html) on rabbitmq.com. I don't mean to be belittling or arrogant but please don't assume that the existing plugin has much of a future long term.

gerhard:
I am going to review the [exporter guidelines](https://prometheus.io/docs/instrumenting/writing_exporters/) and see which areas need improving.

> From a quick peek the global_labels look off

Can you expand please?

From my perspective, I insisted that all Prometheus metrics that RabbitMQ emits are tagged with specific labels (a.k.a. _global labels_) so that the concern of identity is not spread across multiple layers.

Since every metric is linked to a particular node that is part of a cluster and runs a specific version, this is part of the metric identity that should be generated at origin. While we could add this information via the scraping config (static or relabel), we would end up with two separate systems that would need to be kept in sync - so what keeps them in sync?

This gets even more complicated when we take into account scenarios such as multiple nodes on the same host, rolling upgrades where versions change **over time** (minutes / hours / days), mixed versions, mistakes (e.g. _Scraping config tells me that this is node A from cluster 1, but it's actually node A from cluster 2_) etc. By letting nodes determine their identity, we have a self-contained metrics system that always exposes the perspective that matters: the origin's.

I would be happy to chat with you about this, GitHub comments are not the most efficient way of having these types of discussions.

> ... counters lack _total

This is legit, happy to address this point.

> ... and many units aren't right

Which ones? In what way?

brian-brazil:
> From my perspective, I insisted that all Prometheus metrics that RabbitMQ emits are tagged with specific labels (a.k.a. global labels) so that the concern of identity is not spread across multiple layers.

In Prometheus that layer is Prometheus SD. See https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels

https://www.robustperception.io/exposing-the-software-version-to-prometheus is a better way to handle this.

> Which ones? In what way?

I noticed times not in seconds, missing units, and in same cases the unit wasn't a suffix.

Some of the types also seem incorrect, io_sync_time_microseconds sounds like a counter and all the raft metrics sound like gauges.

> This new exporter does not expose a lot more metrics as the metrics available haven't meaningfully changed (there were some additions, though). 

The question is more is something important missing.

> Why? Because for better or worse, a 3rd party plugin stands no chance against one that is is built-in and documented on rabbitmq.com. I don't mean to be belittling or arrogant but please don't assume that the existing plugin has much of a future long term.

I try to encourage those looking to write new exporters to work with existing exporters, rather than everyone doing their own thing and users having to choose from a selection of exporters in various stages of maintenance scattered across github. There is a preference for things being first party, but that's not the only consideration when replacing an existing established exporter.

Have you chatted with @deadtrickster and @kbudde about their exporters?

gerhard:
> In Prometheus that layer is Prometheus SD. See https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels
> 
> https://www.robustperception.io/exposing-the-software-version-to-prometheus is a better way to handle this.

I can see how a single label for versions would work. I can also see how a single label for node / cluster identity would work as well. Thank you!

> I noticed times not in seconds, missing units, and in same cases the unit wasn't a suffix.

All time metrics are in microseconds, a second is too large as a unit.

I will go over all metrics and double-check for missing units & units in the wrong place.

> Some of the types also seem incorrect, io_sync_time_microseconds sounds like a counter

You are right, `io_sync_time_microseconds` should be a counter. Will check again the types of all other metrics.

> and all the raft metrics sound like gauges.

All `raft_` metric types are correct. Raft metrics that increase monotonically are counters. The rest are gauges.

> The question is more is something important missing.

Not as far as I am aware. `rabbitmq_prometheus` depends on `prometheus.erl`, same as `prometheus_rabbitmq_exporter`. `prometheus.erl` exposes certain Erlang VM metrics by default. These metrics will be available via either exporter.

All the core metrics that RabbitMQ exposes via the HTTP API, which both existing exporters convert into Prometheus metrics, are exposed via `rabbitmq_prometheus`. The primary difference is that metrics exposed via HTTP API are aggregated, which means that they put pressure on the entire cluster, and are likely to timeout when the Erlang distribution is under load - meaning no metrics when they are needed the most.

`rabbitmq_prometheus` addresses all these problems since it only exposes node-local metrics, in the most efficient way possible, with no unnecessary conversations. Metrics are read from internal, in-memory data structures called ETS tables and converted into Prometheus format just-in-time.

A long-term goal that I am 100% behind is to stop exposing metrics via the HTTP API and stop doing any metric aggregation & cluster-wide operations in RabbitMQ. This new exporter is the first significant step towards that goal. From a certain perspective, it's a re-design of RabbitMQ's metrics system.

> I try to encourage those looking to write new exporters to work with existing exporters, rather than everyone doing their own thing and users having to choose from a selection of exporters in various stages of maintenance scattered across github. There is a preference for things being first party, but that's not the only consideration when replacing an existing established exporter.
 
We have been working with @deadtrickster on prometheus_rabbitmq_exporter since 2018:

* https://github.com/deadtrickster/prometheus_rabbitmq_exporter/pull/47
* https://github.com/deadtrickster/prometheus_rabbitmq_exporter/pull/54
* https://github.com/deadtrickster/prometheus_rabbitmq_exporter/pull/55
* https://github.com/deadtrickster/prometheus_rabbitmq_exporter/pull/57

After this https://github.com/deadtrickster/prometheus_rabbitmq_exporter/pull/73, we realised that we need to invest more in the whole metrics. It's not enough to expose metrics, we also need to invest time in visualising them. The [RabbitMQ-Overview Grafana dashboard](http://next.rabbitmq.com/prometheus.html#rabbitmq-overview-dashboard) was build for exactly this reason. Same for [RabbitMQ-Raft](https://grafana.gcp.rabbitmq.com/d/f1Mee9nZz/rabbitmq-raft?orgId=1&refresh=5s&from=now-1h&to=now&var-cluster=rmq-gcp-38-qq&var-node=All).

The features that didn't fit in prometheus_rabbitmq_exporter we added to prometheus.erl:

* https://github.com/deadtrickster/prometheus.erl/pull/75
* https://github.com/deadtrickster/prometheus.erl/pull/92

And also BEAM Dashboards: https://github.com/deadtrickster/beam-dashboards/pull/9

> Have you chatted with @deadtrickster and @kbudde about their exporters?

I have not chatted to @kbudde about his exporter. I hope that my previous point addresses the problems inherent to exporting cluster-wide metrics in JSON format. While we are not deprecating that functionality, large-scale enterprise deployments keep hitting various problems specific to this approach.

This was a great conversation, I am glad that I have submitted the tiny PR and learned so much in the process. I plan on focusing on the rabbitmq_prometheus shortcomings that you have identified, will be back once I've made progress on that front.

FWIW, [I will be talking about RabbitMQ's Observability journey over the next couple of months](https://gerhard.io/talk/index.html). This is an [alpha preview](https://gerhard.io/slides/observable-systems-alpha/#/) of what to expect in the final talk. The topics that I have touched on here are at the core of this journey. Thank you for helping out @brian-brazil 😉

brian-brazil:
> All time metrics are in microseconds, a second is too large of a unit.

We use floats, so that's not a problem.

> All raft_ metric types are correct. Raft metrics that increase monotonically are counters. The rest are gauges.

"Raft log snapshot index" sounds like a gauge to me. If you care about the absolute value rather than only how fast it's increasing, that's a gauge.

> You are right, io_sync_time_microseconds should be a counter. 

You could also do it as a Summary. That'll make it more obvious to users how to calculate latency from it.

> This new exporter is the first significant step towards that goal.

If you haven't heard, we're working on standardising the Prometheus format into the OpenMetrics format so at some point the Erlang client library will (presumably) provide you with that.

> I hope that my previous point addresses the problems inherent to exporting cluster-wide metrics in JSON format. 

I don't need any convincing on that point. We usually have to actively argue against such designs :)

Would you mind sharing with me a dump of a full /metrics? I spotted some stuff that looked odd, and my Erlang knowledge isn't exactly the best.

What I suggest we do here generally is once this is in an official RabbitMQ release pointing to the new docs, and remove the existing two exporters. This should not be listed as direct (ETS is your instrumentation here, not Prometheus).

gerhard:
> Would you mind sharing with me a dump of a full /metrics? I spotted some stuff that looked odd, and my Erlang knowledge isn't exactly the best.

[rabbitmq_prometheus-3.8.0-beta.802-sample-20190813.txt](https://github.com/prometheus/docs/files/3496278/rabbitmq_prometheus-3.8.0-beta.802-sample-20190813.txt)

In case you want to DIY:

```
git clone https://github.com/rabbitmq/rabbitmq-prometheus.git
cd rabbitmq-prometheus
make overview
curl -s localhost:15693/metrics
```
 
> What I suggest we do here generally is once this is in an official RabbitMQ release pointing to the new docs, and remove the existing two exporters. This should not be listed as direct (ETS is your instrumentation here, not Prometheus).

I've just removed `direct`. The last few official RabbitMQ 3.8 **pre-releases** already ship the new exporter. [rabbitmq:3.8-rc](https://hub.docker.com/_/rabbitmq) Docker image has it too. Will update when 3.8 goes GA.

brian-brazil:
Thanks. From a quick peek:

rabbitmq_uptime_milliseconds - It's preferred to export timestamps rather than time since. The client library should in principle be providing process_start_time_seconds for you
rabbitmq_memory_used_limit_bytes - this is a peak, not a limit
erlang_vm_memory_processes_bytes_total - this is a gauge, it shouldn't end in _total
erlang_vm_allocators - I think the usage label should be in the metric name instead

gerhard:
> rabbitmq_uptime_milliseconds - It's preferred to export timestamps rather than time since. The client library should in principle be providing process_start_time_seconds for you

There are multiple uptimes, all measured in milliseconds:

* `erlang_vm_statistics_wallclock_time_milliseconds`
* `erlang_vm_statistics_runtime_milliseconds `
* `rabbitmq_uptime_milliseconds`

It's important to stick to the runtime standard - `erlang:statistics(wall_clock)` - so that we can compare uptime differences across various layers.

> rabbitmq_memory_used_limit_bytes - this is a peak, not a limit

This is actually the memory limit at which point publishing connections get blocked (a.k.a. [memory alarm](https://www.rabbitmq.com/memory.html)). While `rabbitmq_memory_limit_bytes` would have been more descriptive, the idea was to link it to `rabbitmq_memory_used_bytes`.

> erlang_vm_memory_processes_bytes_total - this is a gauge, it shouldn't end in _total
> erlang_vm_allocators - I think the usage label should be in the metric name instead

Both good points. This is a change that we will need to do in [prometheus.erl](https://github.com/deadtrickster/prometheus.erl).

RichiH:
> It's important to stick to the runtime standard - erlang:statistics(wall_clock) - so that we can compare uptime differences across various layers.

If all of them are exposed as seconds, you can still compare them. It's less than ideal to have to switch orders of magnitudes around, agreed; but it's even worse to have to do it in Prometheus between different data sources.

What if you want to compare to the uptime of, say, a container? You need to convert anyway, and then it's best to do it before ingestion.

gerhard:
> If all of them are exposed as seconds, you can still compare them. It's less than ideal to have to switch orders of magnitudes around, agreed; but it's even worse to have to do it in Prometheus between different data sources.

Having read [Prometheus: Up & Running](https://www.oreilly.com/library/view/prometheus-up/9781492034131/) (thank you @brian-brazil for writing it), this makes perfect sense, will convert all time metrics to `_seconds`.

gerhard:
@brian-brazil time to merge this?

PR #1415, opened 2019-08-12
add link to aws emr yarn exporter

jeremietharaud:
@brian-brazil 

Hi,
This is my first contribution to the project, please find a link to my exporter: [AWS EMR YARN Exporter](https://github.com/jeremietharaud/emr-yarn-exporter)

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -127,6 +127,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Rancher exporter](https://github.com/infinityworksltd/prometheus-rancher-exporter)
    * [Speedtest exporter](https://github.com/nlamirault/speedtest_exporter)
    * [Tankerkönig API Exporter](https://github.com/lukasmalkmus/tankerkoenig_exporter)
+   * [AWS EMR YARN Exporter](https://github.com/jeremietharaud/emr-yarn-exporter)
 
 ### Logging
    * [Fluentd exporter](https://github.com/V3ckt0r/fluentd_exporter)


brian-brazil:
Hmm, does this do anything that the Cloudwatch exporter doesn't? We try to avoid duplication in this list.

jeremietharaud:
Hi Brian,
Thank you for your comment.
My exporter is using YARN API directly to get the metrics. it provides more up-to-date metrics compared to Cloudwatch. Indeed, EMR Cloudwatch metrics are [only updated every 5 minutes](https://docs.aws.amazon.com/emr/latest/ManagementGuide/UsingEMR_ViewingMetrics.html#UsingEMR_ViewingMetrics_HowDoI), which is a bit annoying where your spark jobs take less than one minute.
Thank you.

brian-brazil:
In that case, is this anything the JMX exporter can't expose?

jeremietharaud:
Hi,
I don't think so. My exporter is not a Java agent, it queries the YARN master cluster API that exposes metrics (for cluster, node, queues and scheduler), so no need to get JMX beans. Plus it runs outside of the EMR clusters, like the SQS or ECS exporter.

brian-brazil:
The question is more, is this available via JMX? My vague recollection is that it is for YARN.

jeremietharaud:
Probably, but it is likely to be harder to configure (plus EMR clusters are managed by AWS so we don't have access to the whole configuration).
Using the API is much more straightforward in that case.

brian-brazil:
That counts as duplicating an existing exporter then I'm afraid, if we had an exporter for every possible deployment variation that would lead more to confusion than anything.

jeremietharaud:
I don't think it's a duplication of an existing exporter. My exporter is specific to AWS EMR service insofar as there is an auto-discover of EMR clusters, it also adds the tags of the clusters as label values on the metrics as well as cluster id, name and status, which cloudwatch and jmx exporter cannot do.
I think it would be too bad to limit to a "generic" exporter which configuration is much more complex (conf of the jmx, look for the beans, etc.) and less accurate than my exporter.

brian-brazil:
> as there is an auto-discover of EMR clusters, it also adds the tags of the clusters as label values on the metrics as well as cluster id, name and status

That's the responsibility of Prometheus, not an exporter.

jeremietharaud:
> 
> 
> > as there is an auto-discover of EMR clusters, it also adds the tags of the clusters as label values on the metrics as well as cluster id, name and status
> 
> That's the responsibility of Prometheus, not an exporter.

OK thank you. I will review that.

PR #1417, opened 2019-08-13
Add a best practice page for tuning remote write

csmarchbanks:
Documentation for what each remote write parameter does and when to adjust them. Any feedback is appreciated!

Closes: https://github.com/prometheus/prometheus/issues/5803

cc: @beorn7 @cstyan

### DIFF
--- a/content/docs/practices/remote_write.md
+++ b/content/docs/practices/remote_write.md
@@ -40,8 +40,13 @@ of the data. For each series in the WAL, the remote write code caches a mapping
 of series ID to label values, causing large amounts of series churn to
 significantly increase memory usage.
 
-Memory usage will also increase based on the number of shards used, which can be
-capped by setting `max_shards`.
+In addition to the series cache, each shard and its queue increases memory
+usage. Shard memory is proportional to the `number of shards * (capacity +
+max_samples_per_send)`. When tuning, consider reducing `max_shards` alongside
+increases to `capacity` and `max_samples_per_send` to avoid inadvertantly
+running out of memory. The default values for `capacity` and
+`max_samples_per_send` will constrain shard memory usage to less than 100 kB per
+shard.
 
 ## Parameters
 
@@ -56,7 +61,7 @@ appended to any shards and all throughput will cease.
 
 Capacity should be high enough to avoid blocking other shards in most
 cases, but too much capacity can cause excess memory consumption and longer
-times to clear queues during resharding. It is recommended to set capacity 
+times to clear queues during resharding. It is recommended to set capacity
 to 3-10 times `max_samples_per_send`.
 
 ### `max_shards`
@@ -64,7 +69,11 @@ to 3-10 times `max_samples_per_send`.
 Max shards configures the maximum number of shards, or parallelism, Prometheus
 will use for each remote write queue. Prometheus will try not to use too many
 shards, but if the queue falls behind the remote write component will increase
-the number of shards up to max shards to increase thoughput.
+the number of shards up to max shards to increase thoughput. Unless remote
+writing to a very slow endpoint, it is unlikely that `max_shards` should be
+increased beyond the default. However, it may be necessary to reduce max shards
+if there is potential to overwhelm the remote endpoint, or to reduce memory
+usage when data is backed up.
 
 ### `min_shards`
 
--- a/content/docs/practices/remote_write.md
+++ b/content/docs/practices/remote_write.md
@@ -1,0 +1,103 @@
+---
+title: Remote write tuning
+sort_rank: 8
+---
+
+# Remote write tuning
+
+Prometheus implements sane defaults for remote write, but many users have
+different requirements and would like to optimize their remote settings.
+
+This page describes the tuning parameters available via the [remote write
+configuration.](/docs/prometheus/latest/configuration/configuration/#remote_write)
+
+## Remote write characteristics
+
+Each remote write destination starts a queue which reads from the write-ahead
+log (WAL), writes the samples into an in memory queue owned by a shard, which
+then sends a request to the configured endpoint. The flow of data looks like:
+
+```
+      |-->  queue (shard_1)   --> remote endpoint
+WAL --|-->  queue (shard_...) --> remote endpoint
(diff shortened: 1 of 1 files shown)


beorn7:
Many thanks. @juliusv could you have a look as well, as an authoritative -team reviewer? You are, after all, the maintainer of the remote storage parts.

content/docs/practices/index.md
```diff
@@ -1,5 +1,5 @@
 ---
 title: Best practices
-sort_rank: 8
+sort_rank: 9
```
juliusv: I think this needs to stay the same, as this is about the sort order of the "Best practices" menu item among the top-level items.

content/docs/practices/remote_write.md
```diff
@@ -0,0 +1,103 @@
...
+Prometheus implements sane defaults for remote write, but many users have
+different requirements and would like to optimize their remote settings.
+
+This page describes the tuning parameters available via the [remote write
+configuration.](/docs/prometheus/latest/configuration/configuration/#remote_write)
+
+## Remote write characteristics
+
+Each remote write destination starts a queue which reads from the write-ahead
+log (WAL), writes the samples into a queue owned by a shard, which then sends a
```
juliusv: Might be good to clarify that these queues are in-memory only ("writes the samples into a queue" sounds like it's on-disk queues). At least that's how I assume it still works.

content/docs/practices/remote_write.md
```diff
@@ -0,0 +1,103 @@
...
+WAL --|-->  queue (shard_...) --> remote endpoint
+      |-->  queue (shard_n)   --> remote endpoint
+```
+
+When one shard backs up and fills its queue, Prometheus will block reading from
+the WAL into any shards. Failures will be retried without loss of data unless
+the remote endpoint remains down for more than 2 hours. After 2 hours, the WAL
+will be compacted and data that has not been sent will be lost.
+
+During operation, Prometheus will continuously calculate the the optimal number
```
juliusv: "the the" -> "the"

content/docs/practices/remote_write.md
```diff
@@ -0,0 +1,103 @@
...
+of shards to use based on the incoming sample rate, number of outstanding
+samples not sent, and time taken to send each sample.
+
+### Memory usage
+
+Using remote write increases the memory footprint of Prometheus. Most users
+report ~25% increased memory usage, but that number is dependent on the shape
+of the data. A mapping of series id to label values is stored for each series
+in the WAL, so large amounts of series churn significantly increases memory
+usage.
```
juliusv: Might be good to clarify how and which portions of the on-disk WAL are mapped / copied into memory, to give more context here.

content/docs/practices/remote_write.md
```diff
@@ -0,0 +1,103 @@
...
+will use for each remote write queue. Prometheus will try not to use too many
+shards, but if the queue falls behind the remote write component will increase
+the number of shards up to max shards to increase thoughput.
+
+### `min_shards`
+
+Min shards configures the minimum number of shards used by Prometheus, and is
+the number of shards used when remote write starts. If remote write falls
+behind, Prometheus will automatically scale up the number of shards so most
+users do not have to adjust this parameter. Setting to a higher number will
```
juliusv: Setting -> Setting this

content/docs/practices/remote_write.md
```diff
@@ -0,0 +1,103 @@
+---
+title: Remote Write Tuning
```
juliusv: We've standardized on not using title case for menu items, so:

"Remote Write Tuning" -> "Remote write tuning"

juliusv:
:+1: besides comments.

I haven't even followed any of the recent remote write developments closely, so either @tomwilkie or @csmarchbanks should probably take over maintainership.

beorn7:
@cstyan might also be a candidate. Let's ask around who wants the job. :o)

content/docs/practices/remote_write.md
csmarchbanks: This paragraph tries to specify that it is a mapping of series id to label values, do you have any suggestions for how to clarify that?

juliusv: Ahh, I parenthesized sentence fragments together incorrectly in my mind: that for each series, there is a mapping cached in the WAL, not that for each series in the WAL, there is a mapping cached (in memory).

How about (to make the structure extra clear): "For each series in the WAL, the remote write code caches a mapping of series ID to label values."

csmarchbanks: :+1: That sounds good. I have made the change.

csmarchbanks:
I think I have addressed all the comments at this point, let me know if I missed anything.

As far as maintainership goes, I think all of the people mentioned would be good candidates. I would be happy to take over remote write at some point; it is an area that I am familiar with and use heavily. But I would support @cstyan or @tomwilkie as well.

juliusv:
:+1: from my side

content/docs/practices/remote_write.md
```diff
@@ -0,0 +1,103 @@
...
+cases, but too much capacity can cause excess memory consumption and longer
+times to clear queues during resharding. It is recommended to set capacity 
+to 3-10 times `max_samples_per_send`.
+
+### `max_shards`
+
+Max shards configures the maximum number of shards, or parallelism, Prometheus
+will use for each remote write queue. Prometheus will try not to use too many
+shards, but if the queue falls behind the remote write component will increase
+the number of shards up to max shards to increase thoughput.
```
beorn7: Perhaps add something about the memory impact (as the number of shardes multiplies with the capacity to determine how many samples are queued at max). Also, [there were reports of lock contention with too many shards](https://github.com/prometheus/prometheus/issues/5803#issuecomment-520973662). Perhaps give some sane ballpark numbers here so that people don't set this to thousands of shards?

csmarchbanks: I added more information about the memory impact of shards. I did not add anything specific about lock contention, but did say that it is unlikely to need to increase the default value for `max_shards`.

beorn7:
Thank you very much. Let's merge this now. We can still add more if that appears necessary in the future.

PR #1421, opened 2019-08-17
Add AWS EC2 security group exporter

cytopia:
# Add AWS EC2 security group exporter

### PR Scope
This PR adds [aws-ec2-sg-exporter](https://github.com/cytopia/aws-ec2-sg-exporter) to the list of available exporters.

### What is it
A dockerized<strong><sup>[1]</sup></strong> Prometheus exporter that compares desired/wanted IPv4/IPv6 CIDR against currently applied inbound CIDR rules by protocol and port number in your AWS security group(s) per region.

> <strong><sup>[1]</sup></strong>: If you want to use this exporter without Docker jump here: [Usage without Docker](https://github.com/cytopia/aws-ec2-sg-exporter#usage-without-docker)

### Motivation

Some IP addresses ranges such as Cloudfront edge nodes or SaaS hosts might change frequently and you possibly want to ensure that those are always in sync with what you have currently defined in
your security group. This exporter does exactly this and can easily be hooked up with Alertmanager to trigger alerts in case you get out of sync.

### Meta
Signed-off-by: cytopia <cytopia@everythingcli.org>

@brian-brazil

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -112,6 +112,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [WebDriver exporter](https://github.com/mattbostock/webdriver_exporter)
 
 ### APIs
+   * [AWS EC2 security group exporter](https://github.com/cytopia/aws-ec2-sg-exporter)
    * [AWS ECS exporter](https://github.com/slok/ecs-exporter)
    * [AWS Health exporter](https://github.com/Jimdo/aws-health-exporter)
    * [AWS SQS exporter](https://github.com/jmal98/sqs_exporter)


brian-brazil:
Thank you for your PR. This doesn't seem like a metrics exporter, more of a set of nagios-style checks that's doing configuration validation. In addition this sounds like cause-based alerting, and closing the loop and having automation change the security groups would seems better than having to call in a human.

cytopia:
> Thank you for your PR. This doesn't seem like a metrics exporter, more of a set of nagios-style checks that's doing configuration validation.

The metric in this case is: available security group rules per security group and if they match anything specified in the input. You might also just monitor the number of security group rules and visualize this in Grafana. Additionally this let's you use Alertmanager provided by Prometheus instead of using something like Nagios (in case you only have that in place).


> and closing the loop and having automation change the security groups would seems better than having to call in a human.

Definitely, and this should be the first thing of having security groups in sync. Secondly this automation can always break and having something which monitors it, wouldn't be too bad or?

brian-brazil:
> Secondly this automation can always break and having something which monitors it, wouldn't be too bad or?

It's still cause based monitoring. There should be a symptom such as request failures that's caused by the firewall.

cytopia:
> There should be a symptom such as request failures that's caused by the firewall.

I think this is something to gathered by a different exporter. I was trying to keep the current one as small and stable as possible by following KISS.

> It's still cause based monitoring.

Fair enough, I leave it up to you to close this PR, if this exporter does not bring the right value to it.

brian-brazil:
Yeah, this is more a configuration management/validation tool than a metric exporter.

PR #1424, opened 2019-08-21
Add mysqlrouter_exporter

rluisr:


brian-brazil:
Thanks for sharing. Your exporter it taking a password on the command line, which is not secure. Can you fix that?

Some other tips to improve your exporter:
Several label names should be separate metrics as they are numeric, including process_id, time_started, time_refresh_in_ms, group_replication_id, time_last_refresh_succeeded, refresh_failed. router_status should be router_info.
Metrics should be gathered at scrape time and use MustNewConstMetric, not on a 60s cycle independent of scrapes.

rluisr:
OK I'll change reading credentials from command line to an environment variable.

> Several label names should be separate metrics as they are numeric, including process_id, time_started, time_refresh_in_ms, group_replication_id, time_last_refresh_succeeded, refresh_failed. router_status should be router_info.

It's means change label like below?
```
router_info{process_id=x, time_started=x, time_refresh_in_ms=x, group_replication_id=x, time_last_refresh_succeeded=x, refresh_failed=x}
```

brian-brazil:
> It's means change label like below?

Yes, and remove the numeric labels. Instead have them as separate gauge metrics. You can't do math on label values.

rluisr:
Thanks quick reply!
I know labels can't math. but for example `process_id` there is no problem even if wen can't math currently.

I updated code of getting credentials.
please check again :-)
https://github.com/rluisr/mysqlrouter_exporter

brian-brazil:
R sorts before S

rluisr:
Ah sorry. fixed.

brian-brazil:
y is after S :)

rluisr:
oh my god.... sorry

brian-brazil:
Thanks!

PR #1431, opened 2019-08-28
Update doc link

counter2015:
Update link of influxDB to latest version.


counter2015:
This is a trivial fix, one thing makes me confused that the diff show two lines have been modified since I only change the first line.

Is that ok? @brian-brazil

brian-brazil:
There's some noop change on the very last line too, odd but not a problem. Thanks!

PR #1432, opened 2019-08-29
Add Elasticsearch/Metricbeat as a reader

jamiesmith:
The Metricbeat prometheus module supports reading from prometheus and prometheus exporters.  Linked to the docs as they are more informative than the [module itself](https://github.com/elastic/beats/tree/master/metricbeat/module/prometheus)


brian-brazil:
We don't have a section for things that support the Prometheus text format, which'd be most monitoring systems out there by now.

PR #1434, opened 2019-09-04
Change "type" to "operation" for label name

ChrisSimmons:
"type" goes against the guidance in "Writing exporters" to avoid using that word.  "operation" seems a good enough replacement.


brian-brazil:
Thanks!

PR #1435, opened 2019-09-05
Add Locust Exporter

pperzyna:
+ signed
+ sorted


brian-brazil:
Thanks!

Some tips: You should use MustNewConstMetric, rather than using direct instrumentation and a mutex. Add units to time metrics where possible. common/log is deprecated, we use common/promlog these days.

PR #1436, opened 2019-09-06
Remove reference to --web.enable-remote-shutdown

snebel29:


beorn7:
IIRC this was true for Prometheus 1.x. The sentence you remove here is referring to 1.x only.

beorn7:
If we wanted to remove anything, we should remove the whole reference to 1.x. What does everybody think about that?

snebel29:
Ah right, I wasn't aware of this existing in 1.X and since the sentence start with `As of Prometheus 2.0`... I keep it open for whatever the fix finally is

brian-brazil:
> If we wanted to remove anything, we should remove the whole reference to 1.x. What does everybody think about that?

Hmm, there's stragglers out there running 1.x. It's out almost 2 years now.

snebel29:
Ok, that was my fault reading too fast, it's in fact stated that this only apply to 1.x

PR #1440, opened 2019-09-18
Update ShowMax -> Showmax (new name) and a current logo

zloo:
@brian-brazil 
Update the logo on prometheus.io website to be the current one, and also change the filename to be consistent with the company name (lowercase m).


brian-brazil:
Thanks!

PR #1446, opened 2019-09-25
add a Nim client library to the list

stefantalpalaru:
Signed-off-by: Ștefan Talpalaru <stefantalpalaru@yahoo.com>

### DIFF
--- a/content/docs/instrumenting/clientlibs.md
+++ b/content/docs/instrumenting/clientlibs.md
@@ -29,6 +29,7 @@ Unofficial third-party client libraries:
 * [Lua](https://github.com/knyar/nginx-lua-prometheus) for Nginx
 * [Lua](https://github.com/tarantool/prometheus) for Tarantool
 * [.NET / C#](https://github.com/andrasm/prometheus-net)
+* [Nim](https://github.com/status-im/nim-metrics)
 * [Node.js](https://github.com/siimon/prom-client)
 * [Perl](https://metacpan.org/pod/Net::Prometheus)
 * [PHP](https://github.com/Jimdo/prometheus_client_php)


brian-brazil:
Thanks for sharing, that's a nice readme.

Some comments from looking over it:

> counter("one_off_counter").inc()

This is not in line with the client guidelines, see https://www.robustperception.io/why-predeclare-metrics

> mySummary_count 2.0 1569332171696

Timestamps should not be exported by default, they should only ever come from custom collectors (or if you're pushing to systems that require the client to provide a timestamp - in which case it's the exposition's job to worry about it).

> mySummary_created 1569332171696.0

Looks like you're folowing the openmetrics data model, this value should be in seconds. In which case using milliseconds internally may not be the best choice.

Also the main counter time series name should end with _total per Prometheus convention, and openmetrics requirements.

> lCounter.inc(labelValues = ["2", "b"])

You should have a `labels()` call for this, see https://www.robustperception.io/label-lookups-and-the-child

> declareCounter cCounter, "counter with colons in name", name = "foo:bar:baz"

No direct instrumentation should be using metric names with colons, so this isn't something to worry about.

> Start an HTTP server listening on 127.0.0.1:9093

I'd suggest not using the Alertmanager's port in examples.

> nim_gc_total_mem_bytes nim_gc_occupied_mem_bytes

I'd go for nim_gc_mem_bytes and nim_gc_mem_occupied_bytes so the memory metrics are together.

Watch out for /proc/stat, as the command can contain spaces. So anything after the command you need to count from the end.

nim_runtime_info isn't mentioned in the readme.

Do the statsd/graphite pushers correctly handle metrics which have been created but not yet used? What I'd suggest is having them send everything every (say) 10s, rather than hooking into instrumentation directly.

stefantalpalaru:
Thanks for the review!

>> counter("one_off_counter").inc()

> This is not in line with the client guidelines

I know. It's extra and repeatedly discouraged in the docs, but some team members wanted something quick and dirty. I guess there's value in an API with no barriers to entry - barriers that might discourage casual uses.

> Timestamps should not be exported by default

Not a good idea if you want precision, specially in the pull model where you have externally set delays between metric updates and collection.

> they should only ever come from custom collectors

On the contrary, those are the ones updated upon collection request, so with the smallest time difference.

Are you worried about Prometheus' extra parsing load with default timestamps? Worth benchmarking, when we get enough metrics to care about that.

>> mySummary_created 1569332171696.0

> Looks like you're folowing the openmetrics data model, this value should be in seconds.

Changed to seconds.

> Also the main counter time series name should end with _total per Prometheus convention, and openmetrics requirements.

Done.

>> lCounter.inc(labelValues = ["2", "b"])

> You should have a labels() call for this

Caching the result of hash table lookups is not worth the complexity increase for a feature that I'm only supporting to comply with the standard. I'll keep it in mind, if it turns out to be a bottleneck in real-world scenarios.

> No direct instrumentation should be using metric names with colons, so this isn't something to worry about.

It still felt weird not allowing the full use of the allowed regex.

> I'd suggest not using the Alertmanager's port in examples.

Changed to 8000.

> I'd go for nim_gc_mem_bytes and nim_gc_mem_occupied_bytes so the memory metrics are together.

Done.

> Watch out for /proc/stat, as the command can contain spaces.

Right. Parsing fixed.

> nim_runtime_info isn't mentioned in the readme.

Because that name is not used for an actual metric. It's just a placeholder name for the collector that spits out all Nim-related metrics (only 2 right now).

> Do the statsd/graphite pushers correctly handle metrics which have been created but not yet used? What I'd suggest is having them send everything every (say) 10s, rather than hooking into instrumentation directly.

It's much easier to have instrumentation updates trigger the pushing. They all end up "pushed" to the same channel buffer, waiting to be read by the same networking thread that does the actual pushing to the StatsD/Carbon backend.

An attempt to flood this system and see the maximum metric update rate (with data dropping kicking in, of course) shows acceptable values: https://github.com/status-im/nim-metrics/pull/13#issuecomment-533761947

I'd rather not increase the implementation complexity until I have to.

brian-brazil:
> I know. It's extra and repeatedly discouraged in the docs, but some team members wanted something quick and dirty.

Unfortunately it leads to quite a lot of problems for downstream users as appearing and disappearing time series are difficult to deal with, and should be removed.

> Not a good idea if you want precision, specially in the pull model where you have externally set delays between metric updates and collection.

Timestamps break things like staleness handling, they must not be present for direct instrumentation. A client library that does otherwise is not usable, and thus cannot be recommended on this page.

> Changed to seconds.

You can have subsecond precision on those as the value is a float, not that it'll matter much in practice.

> It still felt weird not allowing the full use of the allowed regex.

It's fine to allow it, but there's no need to go out of your way to document how to do it as it should never come up.

> Because that name is not used for an actual metric.

Ah, I thought it was a metric. It'd be a typical metric name to put things like runtime version number in.

> waiting to be read by the same networking thread that does the actual pushing to the StatsD/Carbon backend.

While that's how StatsD works, it's not how Graphite works. You should push everything on a regular basis to Graphite, and this shouldn't be much code for a compliant client library as you can see in https://github.com/prometheus/client_python/blob/7ced8d36d6cd0f8fc9a405bd1a9840f7dc13c5a3/prometheus_client/bridge/graphite.py#L63-L73

stefantalpalaru:
> should be removed

Not your call. You can impose minimum client library features for Prometheus compliance, but you cannot veto extra features.

> Timestamps break things like staleness handling

You mean this? https://www.robustperception.io/staleness-and-promql

The only way for a metric to no longer appear on nim-metrics' HTTP endpoint is for the corresponding collector to be unregistered. Timestamps have nothing to do with it.

> A client library that does otherwise is not usable, and thus cannot be recommended on this page.

That settles it, then.

brian-brazil:
> Not your call. You can impose minimum client library features for Prometheus compliance, but you cannot veto extra features.

I can recommend against misfeatures which experience has shown are unwise, and go against our data model. Extra features in general are encouraged as makes sense for a given language/runtime.

> The only way for a metric to no longer appear on nim-metrics' HTTP endpoint is for the corresponding collector to be unregistered. Timestamps have nothing to do with it.

Timestamps disable the staleness handling, and your example output (inconsistently) includes timestamps. If you don't actually output timestamps and the examples are out of date then that's okay.

stefantalpalaru:
> Timestamps disable the staleness handling

Good. I don't need any.

> your example output (inconsistently) includes timestamps

Don't confuse string serialisation with the HTTP output. The latter consistently includes timestamps and looks like this:

```text
# HELP process_info CPU and memory usage
# TYPE process_info gauge
process_virtual_memory_bytes 1032871936.0 1569508593100
process_resident_memory_bytes 34828288.0 1569508593100
process_start_time_seconds 1569508575.69 1569508593100
process_cpu_seconds_total 2.8 1569508593100
process_max_fds 1024.0 1569508593100
process_open_fds 46.0 1569508593100

# HELP nim_runtime_info Nim runtime info
# TYPE nim_runtime_info gauge
nim_gc_mem_bytes 2101248.0 1569508593100
nim_gc_mem_occupied_bytes 1121800.0 1569508593100

# HELP connected_peers number of peers in the pool
# TYPE connected_peers gauge
connected_peers 3.0 1569508591330
connected_peers_created 1569508576.0
```

brian-brazil:
> Good. I don't need any.

It's something that users require though, and was one of the big features of Prometheus 2.0.

There are valid use cases for exposing timestamps, related to exporters from other monitoring systems. Not for direct instrumentation though. The OpenMetrics spec will also strongly discourage timestamps outside of these use cases.

> connected_peers_created 1569508576.0

This line has no timestamp.

stefantalpalaru:
> It's something that users require though, and was one of the big features of Prometheus 2.0.

I still don't see what client-set timestamps have to do with you deleting metrics after a certain interval of not seeing them on the HTTP endpoint.

> This line has no timestamp.

Since the timestamp is already in the value, it makes no sense to duplicate it in another field.

brian-brazil:
> I still don't see what client-set timestamps have to do with you deleting metrics after a certain interval of not seeing them on the HTTP endpoint.

Staleness doesn't work when there's exposed timestamps, as I couldn't figure out sane semantics. Stale metrics aren't deleted, they're not returned for evaluations after a given time.

> Since the timestamp is already in the value, it makes no sense to duplicate it in another field.

To Prometheus this is a normal metric, and in general you should have timestamps on everything or nothing. This would be explicitly invalid in OpenMetrics.

PR #1448, opened 2019-09-27
add logstash exporter

Wing924:
Signed-off-by: Wing924 <weihe924stephen@gmail.com>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -133,6 +133,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Fluentd exporter](https://github.com/V3ckt0r/fluentd_exporter)
    * [Google's mtail log data extractor](https://github.com/google/mtail)
    * [Grok exporter](https://github.com/fstab/grok_exporter)
+   * [Logstash exporter](https://github.com/Wing924/logstash-exporter)
 
 ### Other monitoring systems
    * [Akamai Cloudmonitor exporter](https://github.com/ExpressenAB/cloudmonitor_exporter)
@@ -256,7 +257,7 @@ make use of one of the normal Prometheus client libraries under the hood. As
 for all independently maintained software, we cannot vet all of them for best
 practices.
 
-   * Clojure: [iapetos](https://github.com/clj-commons/iapetos) 
+   * Clojure: [iapetos](https://github.com/clj-commons/iapetos)
    * Go: [go-metrics instrumentation library](https://github.com/armon/go-metrics)
    * Go: [gokit](https://github.com/peterbourgon/gokit)
    * Go: [prombolt](https://github.com/mdlayher/prombolt)


brian-brazil:
I see that there are a few logstash exporters around. How does this compare to the others?

Wing924:
I’ve tried other logstash exporters but only this one works with latest logstash.

brian-brazil:
Have you tried working with the existing exporters to fix that, rather than creating a brand new one?

Wing924:
https://github.com/BonnierNews/logstash_exporter this one is not be maintained any more since 2017. 
https://github.com/alxrem/prometheus-logstash-exporter this one don’t use correct gauge, counter and it’s hard to fix because it just convert json to metrics without any logic.

brian-brazil:
> it’s hard to fix because it just convert json to metrics without any logic.

That's an acceptable design decision for an exporter (see https://prometheus.io/docs/instrumenting/writing_exporters/#maintainability-and-purity), though it should be using Untyped rather than Gauge @alxrem. 

Does it have other issues? I'd much prefer that the community work together, rather than continuously writing exporters for scratch and losing the knowledge that existing code contains.

Wing924:
For the Maintainability and purity, I think logstash has only a handful of metrics that rarely change. You can check my code, which don't have too many metrics. So I think getting everything perfect is better than just converting.

> I'd much prefer that the community work together

Yes, working together is better. It's about time to deciding which one should be in the list on the docs and make it as standard.

brian-brazil:
From what I can see the community is largely around https://github.com/sequra/logstash_exporter

Wing924:
Yes, I also checked that. It’s designed to work with logstash 5, which is EoL.  It don’t works with logstash 7, the latest one.

alxrem:
>  it should be using Untyped rather than Gauge @alxrem.

I released version 0.6.0 with all metrics marked as untyped.

brian-brazil:
@alexrem, thanks. It's a minor thing. If you've any input on this topic feel free to chime in.

A quick look at logstash does indicates to me that the nodestats API hasn't changed between 5 and 7. So if there's something missing in the @sequra one I'd suggest working to fix it.

Wing924:
@brian-brazil I'll send PR to @sequra 's repo. Since that repo don't update or open issue for about 7 months, I'm not sure if it is active.

brian-brazil:
It is not always a good sign, but it doesn't necessarily mean its inactive. We'll find out one way or the other.

gouthamve:
@Wing924 Is there any update? While I don't see any activity there, I think it's stable and might have more users. Did you find any issues with that exporter?

brian-brazil:
Looks like there was just a bugfix PR sent to the sequra one, so we'll know if it's actually dead in a while.

Wing924:
https://github.com/sequra/logstash_exporter looks active.

brian-brazil:
Yip, so looks like sequra is the one to list.

jan--f:
Hello from the bug scrub!
I'll close this for now, none of the proposed exporters seem active anymore.

PR #1450, opened 2019-10-05
Add Diffusion to third-party software list.

philipa:
@brian-brazil 

Diffusion is a commercial software product from Push Technology.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -219,6 +219,7 @@ separate exporters are needed:
    * [Collectd](https://collectd.org/wiki/index.php/Plugin:Write_Prometheus)
    * [Concourse](https://concourse-ci.org/)
    * [CRG Roller Derby Scoreboard](https://github.com/rollerderby/scoreboard) (**direct**)
+   * [Diffusion](https://https://developers.pushtechnology.com)
    * [Docker Daemon](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-metrics)
    * [Doorman](https://github.com/youtube/doorman) (**direct**)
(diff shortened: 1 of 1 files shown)


brian-brazil:
That URL isn't right, and I think https://developers.pushtechnology.com/hc/en-us/articles/360034215494-How-To-Use-Metric-Collectors would be more useful.

You don't appear to actually document how to get the metrics. I presume "gateway" doesn't mean the pushgateway, I can imagine that confusing some of your users so would suggest clarifying.

philipa:
Thanks Brian; yes I agree. The blog you found is about general configuration of Diffusion's metric collection and not specific to Prometheus.

I've updated the URL to point to the relevant page in our manual.

brian-brazil:
Thanks!

diffusion_sessions_peak sounds like a gauge rather than a counter. Counters should also generally end in _total (except if part of a Summary/Histogram), and _count should be for that series in a Summary/Histogram.

philipa:
`diffusion_sessions_peak` can only go up. It is the highest number of sessions the server has ever hosted (it is the maximum value of `diffusion_sessions_open`). There is a separate `diffusion_sessions_total` metric

brian-brazil:
That's still a Gauge. You care about the absolute value, not how fast it is going up.

philipa:
Since `_peak` is monotonically increasing, isn't a counter appropriate? What benefit would using a gauge bring?

brian-brazil:
Counters always go up, but that doesn't mean that everything that always goes up is a counter. Counters are not meant to be usable without using rate() first, whereas diffusion_sessions_peak is usable as-is.

SuperQ:
@philipa Even tho it's monotonically increasing, it's not counting anything.

philipa:
OK, schooled! I'll get this fixed. Thanks for your patience.

PR #1453, opened 2019-10-07
Add an introductory definition of Label

avivace:
Signed-off-by: Antonio Vivace <avivace4@gmail.com>

### DIFF
--- a/content/docs/introduction/glossary.md
+++ b/content/docs/introduction/glossary.md
@@ -53,11 +53,9 @@ An instance is a label that uniquely identifies a target in a job.
 
 ### Label
 
-A label is a key-value pair you can associate to your metric, allowing to provide more details by appending those additional fields.
+A label is a key-value pair associated with a metric. Labels allow for identification of different instances of the same metric.
 
-For example, describing the CPU rate you could append the core number as a label.
-
-You can then aggregate and filter metrics based on labels.
+For example, different CPUs on a node, or different nodes in a cluster.
 
 ### Job
 
--- a/content/docs/introduction/glossary.md
+++ b/content/docs/introduction/glossary.md
@@ -51,6 +51,14 @@ metrics that are exposed in a non-Prometheus format into a format Prometheus sup
 
 An instance is a label that uniquely identifies a target in a job.
 
+### Label
+
+A label is a key-value pair you can associate to your metric, allowing to provide more details by appending those additional fields.
+
+For example, describing the CPU rate you could append the core number as a label.
+
+You can then aggregate and filter metrics based on labels.
+
 ### Job
 
 A collection of targets with the same purpose, for example monitoring a group of like processes replicated for scalability or reliability, is called a job.


content/docs/introduction/glossary.md
```diff
@@ -51,6 +51,14 @@ metrics that are exposed in a non-Prometheus format into a format Prometheus sup
 
 An instance is a label that uniquely identifies a target in a job.
 
+### Label
+
+A label is a key-value pair you can associate to your metric, allowing to provide more details by appending those additional fields.
```
SuperQ: Good idea for an addition, but this doesn't fit with the writing style.

How about this:
```
A label is a key-value pair associated with a metric. They allow for identification of different instances of the same metric.

For example, different CPUs on a node, or different nodes in a cluster.
```

avivace: Sounds good!

SuperQ:
It looks like you need to sign off the second commit with `git commit -s --amend`.

avivace:
Whoops, should be good now

content/docs/introduction/glossary.md
```diff
@@ -51,6 +51,12 @@ metrics that are exposed in a non-Prometheus format into a format Prometheus sup
 
 An instance is a label that uniquely identifies a target in a job.
 
+### Label
+
+A label is a key-value pair associated with a metric. Labels allow for identification of different instances of the same metric.
```
brian-brazil: instance has a meaning in Prometheus, so I'd go for another word.

avivace: Occurence?

brian-brazil: Occurrence could be misunderstood to have a time aspect. Variation maybe?

Hmm, this definition makes it seem like there's only one label for a metric. And in Prometheus terms we should be talking about time series rather than metrics here.

SuperQ: We typically define "metric" as a unique time-series. This would also be good to add to the glossary term list.

brian-brazil: That's not what a metric is in Prometheus, a metric can contain any number of time series.

beorn7: That would be a metric family… =:-o
We've always been lacking complete consistency here.

brian-brazil: Yeah, which is why it's best to avoid "metric".

beorn7: Indeed. (But that's a different reason for not using metric than what you said above.)

(In different news, perhaps we should one day officially define those terms for Prometheus instead of avoiding them. IIRC, we deferred the discussion at some point in the past, and then forgot about it.)

brian-brazil: Metric is ambiguous here, though it does apply to all uses of it.

(I'm not sure we'll get to fix this, I put a bit in my book about all the meanings of metric).

avivace: One reason I want this to be clarified is the confusion I had when I was starting with Prometheus: time series/metric concept and the fact a time series is identified by its "metric name" but a metric can be a _set_ of time series.

beorn7: As @brian-brazil said, for the purpose of this PR, I would avoid using "metric" as a meaningful term.
Establishing a canonical meaning for "metric" is a different issue and might not be feasible at all.

brian-brazil:
If you'd like to continue working on this please let us know, but I'm going to close it off for now.

PR #1467, opened 2019-10-16
fix typo in /content/docs/practices/naming.md

iameli:
Seconds aren't easily confused with bytes. Bits are, though.


brian-brazil:
I think the example makes sense, as time and data are completely different units.

SuperQ:
I'm not even sure what `must have a single unit` means in the context of a single metric name. I think some better counter-examples would help explain it.

brian-brazil:
I think it's pretty clear, and the existing examples cover it.

beorn7:
I don't understand the meaning of this line, either. That's too senior project members that don't find it "clear". @brian-brazil could you enlighten us what that line is trying us to tell? Once we got it, we can suggest a wording that works for us mere mortals.

brian-brazil:
For example don't put http request size and http request duration in the same metric.

grobie:
@SuperQ @beorn7 with the added information, any suggestions how to improve this line?

beorn7:
> For example don't put http request size and http request duration in the same metric.

Sorry for missing @brian-brazil's response here. I understand now what this sentence is referring to, but _unit_ is only the right word for mixing seconds and milliseconds. The problem with seconds and bytes is that it mixes different _quantities_ (which usually goes along with mixing different units, but that's more an effect than a cause).

I'll suggest something in the comments.

content/docs/practices/naming.md
```diff
@@ -26,7 +26,7 @@ A metric name...
    (exported by many client libraries)
  * <code><b>http</b>\_request\_duration\_seconds</code>
    (for all HTTP requests)
-* ...must have a single unit (i.e. do not mix seconds with milliseconds, or seconds with bytes).
+* ...must have a single unit (i.e. do not mix seconds with milliseconds, or bits with bytes).
```
beorn7, line 29: "have" is perhaps the wrong word here anyway. A metric has a unit, but a metric name contains a unit or refers to it…

To also address the quantity aspect:

```suggestion
* ...must refer to a single unit (e.g. do not mix seconds with milliseconds) and to a single quantity (e.g. do not mix request size with request duration).
```

This is kind of related or even redundant with L30 and L40, but perhaps being overly explicit is not a bad thing here.

jan--f:
Closing this in favor of #2818

PR #1468, opened 2019-10-17
Add huawei cloudeye exporter

zhongjun2:
Signed-off-by: zhongjun2 <jun.zhongjun2@gmail.com>


brian-brazil:
Thanks!

The metric name prefix should not be configurable, and I'd also advise against munging of label names.

zhongjun2:
Thanks brian-brazil!  " I'd also advise against munging of label names. " ----- Where I can find what label I can use, what label I can not use? Is there a link or something else?

brian-brazil:
I'm looking at https://github.com/huaweicloud/cloudeye-exporter/blob/master/collector/collector.go#L30

https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels are the docs, if there are invalid characters you'd usually change them to underscores.

zhongjun2:
Got it, thanks brian-brazil!

PR #1470, opened 2019-10-23
Update clientlibs.md

miroswan:
Adding a new client library

@brian-brazil

### DIFF
--- a/content/docs/instrumenting/clientlibs.md
+++ b/content/docs/instrumenting/clientlibs.md
@@ -21,6 +21,7 @@ HTTP endpoint on your application’s instance:
 Unofficial third-party client libraries:
 
 * [Bash](https://github.com/aecolley/client_bash)
+* [C](https://github.com/digitalocean/prometheus-client-c)
 * [C++](https://github.com/jupp0r/prometheus-cpp)
 * [Common Lisp](https://github.com/deadtrickster/prometheus.cl)
 * [Elixir](https://github.com/deadtrickster/prometheus.ex)


brian-brazil:
Some comments from a quick peek:

Is rendering the output over http threadsafe?

I'd allow users to not have to pass labels if they're not using them. Unlike Go, you can do function overloading in C. Similarly there's no need for inc vs add here due to overloading.

`prom_metric_sample_histogram_from_labels` - it'd be good if the user could cache this, rather than it having to be computed on every observation. Similarly for the other types.

A Summary without quantiles should be easy to add.

`promhttp_set_active_collector_registry` - I think an optional argument to `promhttp_start_daemon` would be better than a global.

miroswan:
Hello @brian-brazil. Thanks for the good feedback. 

* Good idea regarding promhttp_start_daemon. I'll make an adjustment. 

* Do you have a proposal for the approach to overloading in C? AFAIK know it's not supported without using odd hacks. If there is a way to natively support function overloading where multiple declarations of the same function name are supported (like Java), definitely let me know. Interested. 

* Technically, users can cache metric samples in variables at compile time if they are aware of all of their labels ahead of time. We have an existing demonstration of this use-case in a staging environment. I'll update the API/Documentation to make this capability more obvious. If you're looking for more capability than what I've described, please let me know and I'll see if I can address it as well. 

* I'm interested in adding Summaries. I'll see about getting to this. 

* I _believe_ metric exposition over HTTP should be thread-safe. I haven't had any issues in my smoke tests or staging environments but I'll definitely take a closer look to make sure there aren't bugs there.

brian-brazil:
> Do you have a proposal for the approach to overloading in C?

I must have been thinking of a different language, it's been a while since I've used pure-C.

https://stackoverflow.com/a/25026358/83591 indicates that it's possible with C11, which you already require.

miroswan:
> https://stackoverflow.com/a/25026358/83591 indicates that it's possible with C11, which you already require.

Neat. I'll look into this more.

brian-brazil:
> I believe metric exposition over HTTP should be thread-safe.

I didn't notice any locking when glancing over the code.

PR #1471, opened 2019-10-30
Update Loodse company logo

xrstf:


beorn7:
Thanks @xrstf . Are you affiliated in some way with Loodse? It would be good to get some kind of approval from a Loodse representative that this is indeed their new logo and they are fine with our usage here.

xrstf:
I am indeed employed at Loodse and just offered to create the PR. Just made my membership in our Github org public: https://github.com/xrstf

PR #1473, opened 2019-10-31
Add reference to the ServiceNow Webhook

louisfelix:
Hi!

As discussed with @simonpasquier in [this issue](https://github.com/prometheus/alertmanager/issues/931), this PR add a reference to the ServiceNow Webhook.

It supports alerts grouping in one incident, updating an existing incident based on the group key, dynamic fields configurations and GO templating on fields values (from alert's annotations and labels).

Thanks!


brian-brazil:
Thanks!

PR #1475, opened 2019-11-03
Fix method name

fiunchinho:
I believe this is meant to say `Collect()`, as it's the method in the `Collector` interface.


brian-brazil:
Thanks!

PR #1481, opened 2019-11-11
Added exporter for Mosquitto broker

daviddetorres:
Added link to the exporter of the Mowsquitto MQTT message broker.

Signed-off-by: David de Torres <detorres.david@gmail.com>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -92,6 +92,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [RabbitMQ exporter](https://github.com/kbudde/rabbitmq_exporter)
    * [RabbitMQ Management Plugin exporter](https://github.com/deadtrickster/prometheus_rabbitmq_exporter)
    * [RocketMQ exporter](https://github.com/apache/rocketmq-exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for sharing. You're taking a password on the command line which is insecure, can you change that?

$SYS/broker/uptime and $SYS/broker/clients/maximum are gauges, not counters. I'd include $SYS/broker/timestamp and $SYS/broker/version. Don't forget your _total on counters.

I'd suggest including a metric for when you last received an update.

daviddetorres:
Thanks a lot for the comments. There is an open PR to add metrics for the time of the last update and status of the connection. 

About I'll open issues in the repo for the password on the command line and for the change of the names in the metrics to comply with the metrics naming convention. Hopefully they will be also added in the open pull request.

brian-brazil:
We can add it here once the password is fixed, as we aim not to list things with security issues.

daviddetorres:
@RichiH Now that the exporter support passing authentication credentials from env variables, the previous requirement to include it in the official doc is met.

jan--f:
Hello from the bug scrub!
Apologies from the lack of replies here. It looks like the exporter is no longer very active. Hence I'll close this for now. Please feel free to re-open at any time.

PR #1482, opened 2019-11-13
edit passive sentence

cuchalin:
Change the sentence: 'New members may be proposed by any existing member by email to Prometheus-team.
Into: 'Any existing member may propose new members by email to Prometheus-team'
The new sentence is in an active voice. It makes the sentence easier to read for most people.

### DIFF
--- a/content/governance.md
+++ b/content/governance.md
@@ -30,7 +30,7 @@ Each project must have a `MAINTAINERS.md` file with at least one maintainer. Whe
 
 Team member status may be given to those who have made ongoing contributions to the Prometheus project for at least 3 months. This is usually in the form of code improvements and/or notable work on documentation, but organizing events or user support could also be taken into account. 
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. The existing wording seems consistent with the style of the rest of the text.

beorn7:
Yeah, agree with @brian-brazil . Passive voice is used a lot here anyway.
In different news, a change to the governance requires a qualified vote. So if we ever did an "improve the language" change, it should probably batch up a bunch of things.

brian-brazil:
This would qualify as an editorial change, so I don't think we'd need a full vote here.

beorn7:
Yes, you are right about the editorial change.

Still, since both of us seem to be happy with the general usage of passive voice in the governance document, and nobody else chimed in, I'm closing this.

Issue #1490, opened 2019-11-28
resolve_timeout documentation is outdated


Shadi:
From the documentation at alerting configuration [page](https://prometheus.io/docs/alerting/configuration/):
```
  # ResolveTimeout is the time after which an alert is declared resolved
  # if it has not been updated.
  [ resolve_timeout: <duration> | default = 5m ]
```

Looking at prometheus implementation, and this [comment]( https://github.com/prometheus/prometheus/pull/4538#discussion_r212914500)

it seems that this is not the case anymore.

brian-brazil:
Would you like to send a PR indicating this is merely a default if the client doesn't send it?

Given the amount of confusion this particular setting hurts, it's likely worth moving it from the top and mentioning that this has no impact on Prometheus.

Shadi:
Yes I will prepare a PR that will clear this up.

Shadi:
Hi Brian, I created #1492, please let me know if you have any notes on the wording(English is not my first language)

Shadi:
@stuartnelson3

Shadi:
PR merged.

PR #1492, opened 2019-11-29
Update resolve_timeout docs to indicate it is a fallback value

Shadi:
See #1490

Signed-off-by: Shadi Abdelfatah <shadi.abdelfatah@gmail.com>

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -96,8 +96,7 @@ global:
 
   # ResolveTimeout is the default value after which alertmanager
   # can declare the alert as resolved if it has not been updated.
-  # resolve_timeout configuration value is a fallback value and only used if
-  # Prometheus does not send the value with the alert(Prometheus pre 2.4.0 versions)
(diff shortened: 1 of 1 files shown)


content/docs/alerting/configuration.md
```diff
@@ -98,6 +94,12 @@ global:
   # The default HTTP client configuration
   [ http_config: <http_config> ]
 
+  # ResolveTimeout is the default value after which alertmanager
```
brian-brazil: If a client does not include EndsAt in an alert, default it to this many seconds from now.
This has no impact on alerts from Prometheus, as they always include EndsAt.



2.4.0 was a while back, and I'd prefer to keep this short. I'm only proposing to go beyond pure reference docs here as it's a common confusion.

Shadi: I changed it a bit more to reflect that it has no affect on alerts from Prometheus.

I left `default it to this many seconds from now.` out since this is a duration and can be set in minutes, and also the default value is in minutes too.

brian-brazil:
Thanks!

PR #1495, opened 2019-12-01
Added CockroachDB (no exporter needed, built-in endpoint)

lgueye:


brian-brazil:
Can you add the DCO?


I'd suggest looking at using the Prometheus client directly, it'll be simpler and more efficient.

lgueye:
@brian-brazil I don't know what a DCO is but I'll take a look

lgueye:
@brian-brazil figured out what it is. I might ask cockroachdb developers to do it since I'm not an expert on their licensing strategy.
Thank you for your time

brian-brazil:
`git commit --amend -s` should do it, then force push.

PR #1496, opened 2019-12-02
Add QuasarDB to list of remote storage integrations

kontrarian:
[QuasarDB](https://www.quasardb.net/) is a high performance timeseries database. We've recently added support for Prometheus to our REST API tool.

I've updated the list of remote storage integrations to include QuasarDB and a link to our documentation for Prometheus.


brian-brazil:
Thanks!

> Authentication is currently not supported as Prometheus’s configuration file does not allow for environmental variables or secrets. You can expect authentication support in a future release.

This is incorrect, and irrelevant. We support several forms of client auth, see https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write

kontrarian:
Thanks, I'll update our docs.

PR #1499, opened 2019-12-06
add empty label value  introduction.

johncming:
Fixes https://github.com/prometheus/alertmanager/issues/2102

### DIFF
--- a/content/docs/concepts/data_model.md
+++ b/content/docs/concepts/data_model.md
@@ -37,6 +37,8 @@ are reserved for internal use.
 
 Label values may contain any Unicode characters.
 
+A label with an empty label value is considered equivalent to a label that does not exist.
+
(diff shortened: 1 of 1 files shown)


content/docs/concepts/data_model.md
```diff
@@ -37,6 +37,24 @@ are reserved for internal use.
 
 Label values may contain any Unicode characters.
 
+A label with an empty label value is considered equivalent to a label that does not exist. For example:
```
brian-brazil: We don't have extended examples on this page, and this example features PromQL which isn't part of our data model. Just this sentence should do.

johncming: ok

brian-brazil:
Thanks!

PR #1500, opened 2019-12-09
Add CockroachDB to exporters documentation: no external process need,…

lgueye:
… CockroachDB natively exposes a prometheus compatible endpoint

Signed-off-by: louis gueye <louis.gueye@gmail.com>


brian-brazil:
Thanks!

PR #1501, opened 2019-12-10
Update community.html

denshirenji:
We are adding MetricFire.com to this list of commercial support because we offer a Hosted Prometheus Service. You can find us here, at www.metricfire.com, thank you! Signed-off-by: Lindsey Rogerson <lindsey@metricfire.com>

### DIFF
--- a/content/community.html
+++ b/content/community.html
@@ -107,7 +107,7 @@ title: Community
       <li><a href="https://grafana.com">Grafana Labs</a></li>
       <li><a href="https://juliusv.com">Julius Volz</a> (independent contractor)</li>
       <li><a href="https://tina.tincho.org">Martina Ferrari</a> (independent contractor)</li>
-      <li><a href="https://metricfire.com">MetricFire</a></li>
+      <li><a href="https://www.metricfire.com/hosted-prometheus">MetricFire</a></li>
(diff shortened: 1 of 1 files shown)


content/community.html
```diff
@@ -107,6 +107,7 @@ <h1>Commercial Support</h1>
       <li><a href="https://grafana.com">Grafana Labs</a></li>
       <li><a href="https://juliusv.com">Julius Volz</a> (independent contractor)</li>
       <li><a href="https://tina.tincho.org">Martina Ferrari</a> (independent contractor)</li>
+      <li><a href="https://metricfire.com">MetricFire</a> (SaaS)</li>
```
brian-brazil: SaaS shouldn't be mentioned, we don't mention what any of the others do.

denshirenji:
Hi Brian,

Thank you! I removed "(Saas)" and added DCO, how's it looking? Does the link not work for you? It works for me but it's slow, we probably need to improve that page. Thank you!

Best,
Lindsey

brian-brazil:
You still need to add the DCO.

Linking to https://www.metricfire.com/hosted-prometheus would be better I think.

denshirenji:
looking good now? Thank you!

brian-brazil:
DCO still isn't happy, try squashing.

PR #1504, opened 2019-12-11
Update "Software exposing Prometheus metrics" section in exporters.md

iyangsj:
Signed-off-by: Sijie Yang <yangsijie@baidu.com>


brian-brazil:
Thanks!

You appear to be hand-producing the format from your own client library that doesn't support labels. I'd suggest looking into using client_golang.

PR #1505, opened 2019-12-11
mention PushProx in alternative strategies

belm0:
Outcome of https://github.com/prometheus/pushgateway/pull/312


brian-brazil:
:+1: 

It is.

As a general note I'm not fully sure on recommending 3rd party tools from official docs, though this one obviously makes sense (even if I'm a little biased).

beorn7:
I think it makes sense, too. And so does @belm0 . So you have two more or less unbiased witnesses to back your case. :o)

PR #1508, opened 2019-12-14
Update prometheus-net URL

sandersaares:
The prometheus-net library was migrated from a user repository to a GitHub organization some time ago. There's an automatic redirect in place so the link still works but just to be on the safe side, here's the updated link.


brian-brazil:
Thanks!

PR #1509, opened 2019-12-15
Add Xandikos to list of software providing metrics without a separate exporter.

jelmer:
Add Xandikos to list of software providing metrics without a separate exporter.


brian-brazil:
Thanks!

requests_in_progress_total is a gauge, so shouldn't end in _total.

jelmer:
Wow, that was not the degree of review I was expecting from this pull request. Thank you!

brian-brazil:
I was mainly checking if there was actually direct instrumentation, and noticed that in passing.

PR #1515, opened 2019-12-23
Adding checkbot to the list of exporters

tobiasdenzler:
@brian-brazil: Checkbot is able to run custom bash script in a container running on OpenShift. These scripts can check functionality and compliance settings in your cluster and will expose the result as Prometheus metrics.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -171,6 +171,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [cAdvisor](https://github.com/google/cadvisor)
    * [Cachet exporter](https://github.com/ContaAzul/cachet_exporter)
    * [ccache exporter](https://github.com/virtualtam/ccache_exporter)
+   * [checkbot](https://github.com/tobiasdenzler/checkbot)
    * [Dovecot exporter](https://github.com/kumina/dovecot_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. This seems pretty niche, there's other exporters out there that run scripts and both the pushgateway and node exporter can both take care of metrics from batch jobs.

tobiasdenzler:
Might look like niche, but can be used for a variety of use cases. We apply this approach for compliance/security metrics on AWS services and Kafka management for example.

brian-brazil:
We try to avoid duplication in this list, and this doesn't seem to be doing anything new plus it's restricted to openshift.

tobiasdenzler:
Got it.. seems reasonable. Can you point me to the tools which provide similar functionality? Support for Kubernetes is on my list 😉

brian-brazil:
In addition to the pushgateway and node exporter, there's the NRPE exporter and the various script exporters.

tobiasdenzler:
I added support for Kubernetes and improved the documentation. I hope it is now clear which additional use cases checkbot can provide for monitoring. Otherwise feel free to close this PR.

brian-brazil:
I'm afraid I'm still not seeing how this is usefully different compared to what's already listed, it can only work on some platforms and is inventing a new metrics format. You're free to announce it on the users list, but it doesn't seem to belong on this list.

PR #1516, opened 2019-12-23
Add webdevops exporters

mblaschke:
- Add apprelease-exporter
- Add azure-devops-exporter
- Add azure-keyvault-exporter
- Add azure-metrics-exporter
- Add azure-resourcemanager-exporter
- Add azure-scheduledevents-exporter
- Add pagerduty-exporter
- Add public-holiday-exporter


brian-brazil:
Thanks for your PR. This is quite a lot of exporters to review at once, and at first glance at least three of them seem to duplicate other exporters either already on this list or other integrations known within the ecosystem. They all also seem to be collecting information independently of scrapes.

Can you explain what each of these are? For example I don't believe there's any piece of software called "apprelease". Why are the various Azure exporters split out this way?

mblaschke:
apprelease-exporter
Exporter for GitHub releases/tags and Docker repositories. The goal is to export new releases of applications and also attach CVE/CVSS metrics for each release.
More systems (eg. GitLab) are planed.

azure-devops-exporter
Exporter for CI/CD/Issue tracking system Azure DevOps (previously known as VSTS/TFS). Used to fetch builds,releases,agentpool,... metrics and alert if eg. releases are failing or the agentpool is overloaded.

azure-keyvault-exporter
Exporter for Azure KeyVaults and their content (eg. created, expired dates). Used eg. for alerting for expiry of secerts, keys and certificates.
The Azure KeyVault API is a different from the other APIs so it's an own exporter, also for security reasons.

azure-metrics-exporter
Exporter for Azure Insights metrics and LogAnalytics which uses the official Azure SDK and is only configured using Prometheus scrape config (except service principal connection).
This exporter can scape one target or using Azure resource filter to scape multiple targets at once (eg. all Redis instances in one resourcegroup or in the whole subscription).

azure-resourcemanager-exporter
Exporter for Azure inventory (uses same labels as azure-keyvault-exporter and azure-metrics-exporter, can be coupled for alerting), uses official Azure SDK.
Can be used for metrics exporter to get more informations about the resource or to get an overview about the inventory.

azure-scheduledevents-exporter
Exporter for maintenance annoucements from Azure Metadata API (forced VM freezes, restarts and redeployments). Need to be deployed on each VM or in each AvailabilitySet to get alerts about upcoming Azure ScheduledEvents. Talking directly to the MetaData API without the need of a service principal.
Alerting for upcoming restarts/redeployments (downtime of a VM).

pagerduty-exporter
Exporter for PagerDuty incidents, schedules and teams. Used for eg. showing the current oncall person in Grafana and overview about active incidents. Uses official PagerDuty SDK.

public-holiday-exporter
Exporter for public holiday for services which should not generate alerts when there is a public holiday (eg. telematics and logistics). Uses API at https://date.nager.at/.

Especially the Azure exporter are only covering one job as some jobs are already very huge (eg. azure-metrics-exporter or azure-resourcemanager-exporter). Each exporter is one API.

Some exporters are using own scraping in the background as it can take very long to fetch all information and to optimize given ratelimiting or to avoid overloading of the remote system.

brian-brazil:
Starting with the ones that are easier for me to review:

> public-holiday-exporter

I don't think this should be listed as this does not provide useful metrics, there's no sane way to use a dump of holidays with the dates in labels and historical changes in label values with PromQL. There's also other solutions in this space like https://github.com/roidelapluie/prometheus-timezone-holidays by @roidelapluie 

If you exposed timestamps when holidays started/ended, or a bool indicating if there was a holiday currently in progress, for a country/region that'd be more useful and something that could be listed. The number of labels would also need reducing, things like the launchYear are only going to get in the way when this is used to filter alerts. A separate set of info metrics would be a better way to cover this.

> azure-metrics-exporter

We aim to avoid duplicates in this list, to encourage the community to work on one good exporter rather than everyone creating a new one from scratch every few months. This duplicates the azure exporter already listed. If there's features that one is missing please work with the maintainers of that exporter (which happens to be me in this case). Replacing an existing listing is generally only done when the existing one is abandoned, the community has rallied around a different integration and/or the new one is substantially better in terms of following best practices.

I'm also wondering where you're keeping the secrets here, and if you're indeed using azurerm_resource_metric for all metrics. The Azure metric name would be a better choice.

> Some exporters are using own scraping in the background as it can take very long to fetch all information and to optimize given ratelimiting or to avoid overloading of the remote system.

The primary way to control that is to adjust the scrape interval. Only when you're getting up to minutes for the majority of users should caching/independent scrape intervals be considered, it's a last resort rather than something you do on day one.

brian-brazil:
> apprelease-exporter

In the readme releax -> reduce

apprelease_project_docker_release should have the unit as a suffix on the metric name.

What sort of alerts are you writing on all of these? Generally an exporter is to provide data for graphing/alerting, not to dump arbitrary databases into Prometheus.

apprelease_project_docker_release_cve should have the unit in the name, which is _cvss. The metric is also nothing to do with docker, as it's general CVE information. Consider moving the non-identifying labels to an _info metric.

The name of this exporter doesn't make it obvious what it does as "apprelease" is not a application. This will need a brief comment so someone browsing the list would have an idea of how it'd be useful.

You're silently ignoring API errors, which can result in partial and/or old data. If you don't have full consistent data, then you should fail the scrape. The writing of the cache files is not atomic, write to a temporary file and then rename. Data collection and scraping are not thread safe.

> azure-devops-exporter

This takes a secret on the command line, which is not safe security practice.

With a 30m scrape interval, I wonder how useful this can be in terms of metrics. Metrics should be realtime, otherwise they're not of much use and you may aswell look at the service itself for up to date data rather than complicating and delaying things by passing the data through Prometheus. Everything should use the same scrape interval, and given that's 30s then the collection should be all synchronous.

Are you sure it's both wise and necessary to do concurrent API requests?

Having to use label_replace in your query examples indicates that you labelnames aren't quite right. label_replace and friends are a stopgap while you get the instrumentation fixed to use consistent labels.

This also does not appear to be threadsafe, which synchronous collection and ConstMetrics will resolve for you.

maxParallelism is presumably a number, so should be a metric rather than a labelname. Prometheus can't do maths on labelnames. provisioningState, enabled, status, and hasAssignedRequest also sound like they should be metrics as I presume they change over time for a given agentpool.

id is not a great label name on azure_devops_workitem_data, there's a high chance it'll clash and it's too generic to tell me what it is. Something like workitem would be better. All those dates should be metrics.

Similarly on azure_devops_repository_stats type is a label name to avoid.

azure_devops_repository_commits - this is a gauge, not a counter. Commits can go backwards, and I would presume you care about the absolute number rather than the rate of increase. Similarly with others.

https://github.com/webdevops/azure-devops-exporter/blob/66ae6991de051332a63f3ed8c6a8e0edf44f3c6b/collector_structs.go#L75 - I am confused by these 3 functions. Why would you need to bulk-set/increment childs by the same value?

azure_devops_stats_agentpool_builds_wait - don't forget your units.

jan--f:
Hello from the bug scrub!
Apologies for the long silence. @mblaschke are you interested at all in updating this PR?

jan--f:
Hello from the bug scrub!
Closing this for now, please feel free to reopen.

PR #1518, opened 2019-12-29
Fix link to best practices for writing exporters

spaghetti-:
The old link to https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/writing_exporters.md 404's and this fixes it.


brian-brazil:
The link works fine on the website.

SuperQ:
These files are processed by nanoc before publishing. If you look at the netlify test output, you will see that this breaks the link.

PR #1526, opened 2020-01-10
Added prometheus docker swarm discovery (python)

llacroix:
Added a link to my docker swarm implementation

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -22,7 +22,8 @@ these categories.
 For service discovery mechanisms not natively supported by Prometheus,
 [file-based service discovery](/docs/operating/configuration/#%3Cfile_sd_config%3E) provides an interface for integrating.
 
- * [Docker Swarm](https://github.com/ContainerSolutions/prometheus-swarm-discovery)
+ * [Docker Swarm (Go)](https://github.com/ContainerSolutions/prometheus-swarm-discovery)
+ * [Docker Swarm (Python)](https://github.com/llacroix/prometheus-swarm-discovery)
  * [Lightsail](https://github.com/n888/prometheus-lightsail-sd)
  * [Packet](https://github.com/packethost/prometheus-packet-sd)
  * [Scaleway](https://github.com/scaleway/prometheus-scw-sd)


brian-brazil:
To encourage the community to rally around one solution, and not continuously create new ones from scratch,  we aim to avoid duplicates in these lists. Which of these is the most popular and/or in line with best practices?

llacroix:
My project is hardly going to win a popularity contest in a week, that's sure. But I will take my odds vs a project that is advertised as a POC that has issues that are still unresolved, if memory leaks and file descriptor depletion is best practice, my odds shouldn't be so bad.

brian-brazil:
It'd be better to work to improve what's already there, rather than creating a brand new thing from scratch. Otherwise we end up with many okay integrations, rather than one really good one.

llacroix:
> It'd be better to work to improve what's already there, rather than creating a brand new thing from scratch. Otherwise we end up with many okay integrations, rather than one really good one.

I respectfully disagree. Having many integrations doesn't mean you'll have many okay integration instead of a good one. With the way things are going it's closer to one poor integration than many okay ones. Sometimes it's better to rewrite from 0 than to carry old code with its bugs around. Having multiple integration is not inherently bad. As some developer from one community might not be in a position to contribute in a language they don't know enough to contribute. 

If there was an active maintainer and the project was being updated at least headed somewhere, things would be a bit different. This project has been dead a week after its creation and it did get a few forks and receives changes in forks that arguably "improvements" but more personal requirements.

My integration drastically changes the way it works and how it can be implemented so even if I wanted to contribute, I'd still probably end up with a complete rewrite... I'd want to get it merged I'd receive the answer: "It's too different from the original project." if I received any answer at all. So instead of wasting my time, I implemented something that fulfils most of my needs and gives similar feature as the original (as a compatibility feature) including improved functionality and stability. The developers of the original project had 3 years to do that. If nobody help improve the project in 3 years, may be there's nothing worth improving. 

On the other side, my project, while pretty stable and functional does still have ways to be improved. It has been written in roughly 3 days, so there's a lot of things to clean up and simplify without breaking existing functionality. Eventually support other backend than docker swarm.

Then it wouldn't fall into the Docker Swarm integration wouldn't it?

brian-brazil:
I was stating the general goal behind the curation rules applied here, specifics vary of course. It is not a good strategy to list every new integration here of a given type that has been created as that ultimately does not serve the long terms needs of the community. For example we'd be listing at least 5 different Swarm integrations here, and be leaving it up to the users to figure out which ones actually work. Google can provide that service, without us having to maintain an explicit list.

Keep in mind that this is only one list, and it does not aim to be definitive or exhaustive. Nothing stops you from developing and promoting your integration outside of this (e.g. posting to prometheus-users) and building a good community around it. It'd be premature to list it here though.

llacroix:
There, just for the information, I added tests and coverage. I'm at 95% at the moment and tests not only cover usual tests but also service discovery in an actual docker swarm node. Pretty much all functionality of the project are being tested. The remaining 5% of the missing coverage are related to some specific branching and setup.py.

Docker images generated with version numbers.

Building a community around the projects sounds like a difficult task. It doesn't look like a lot of people use prometheus in a docker swarm.

brian-brazil:
> It doesn't look like a lot of people use prometheus in a docker swarm.

A fair few do. The number of users of a given piece of software has not been a factor as to whether integrations for it get listed here, and there's stuff listed on the exporter page that's far rarer.

The thing I'm looking for here is signs that this is the one the community is moving to, such as others organically filing issues, PRs, github stars etc. None of that will likely happen overnight, so maybe come back in a few months and we'll see how it's going.

I'd also note that while for these lists I don't care as long as a sane language is used, for this particular list and integration I'd give some preference to things built on top of https://github.com/prometheus/prometheus/tree/master/documentation/examples/custom-sd

PR #1529, opened 2020-01-13
Add solace exporter

GreenRover:
Add exporter for [Solace PubSub+](http://www.solace.com)

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -92,6 +92,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [RabbitMQ exporter](https://github.com/kbudde/rabbitmq_exporter)
    * [RabbitMQ Management Plugin exporter](https://github.com/deadtrickster/prometheus_rabbitmq_exporter)
    * [RocketMQ exporter](https://github.com/apache/rocketmq-exporter)
+   * [Solace exporter](https://github.com/dabgmx/solace_exporter)
 
 ### Storage
    * [Ceph exporter](https://github.com/digitalocean/ceph_exporter)


brian-brazil:
You're taking a password on the command line, which needs to be removed as that's insecure.


A few suggestions to improve things:

Why do you have a mutex in Collect? You're correctly using MustNewConstMetric, so you shouldn't need that.

Why are you resetting metrics on scrape? That'll break if there's more than one scraper.

You should probably drop all the _avg metrics, I'd guess that they're since process start and thus of little use. Similarly the _rate metrics can be calculated from the _total metrics in PromQL.

GreenRover:
Hi Brian,

the password via command line is for docker the way to go i gues. Would you prever ENV?

The reset is for debug case and is not used in production.

_avg and _rate will be removed if there is a total.

Can you please explain the the mutex issue?

brian-brazil:
> the password via command line is for docker the way to go i gues. Would you prever ENV?

I don't mind how you approach it, as long as there isn't a password being put on the command line by users.

> The reset is for debug case and is not used in production.

I'd suggest a disclaimer for that option then, users have a tendency to set such things and not understand why everything then breaks.

> Can you please explain the the mutex issue?

You've a mutex for no apparent reason.

juliusv:
@GreenRover Is the mutex perhaps only needed for the global reset, so you don't do it twice? The other read-only data gathering calls to Solace that happen in `Collect()` should be safe to do concurrently/in parallel, right?

GreenRover:
@brian-brazil 
For the password  i added a security hint that the command line option only should be use for hte config option of kingpin

All _avg _rate metrics was removed if a counter exists.

The reset option was refactored and documentation was improved. A warning was added to only usw with care.

@juliusv @brian-brazil 
The mutex is to protect the solace broker to not bean overloaded. The target is not perform parallel scrapes.

GreenRover:
@brian-brazil pleas not jet merge. @dabgmx needs a last confirmation that we are permitted to publish.

brian-brazil:
> For the password i added a security hint that the command line option only should be use for hte config option of kingpin

The binary still takes a password on the command line, and in the docker example.

> The mutex is to protect the solace broker to not bean overloaded. The target is not perform parallel scrapes.

That's not the duty of an exporter generally.

GreenRover:
@brian-brazil This password options is still there. yes. This is beacuse kingpin use the same flags for commandline options and config file. What solution whould you advice? That support config file + ENV support.

brian-brazil:
If Kingpin forces you to take command line arguments for this, then I'd not use kingpin for this setting.

GreenRover:
@brian-brazil Now we fixed all issues and got the last go for the release.

brian-brazil:
You still have passwords on the commandline in the docker example.

GreenRover:
@brian-brazil that is the way how to do this with docker. There when using open shift this ca be secured. But for normal docker this is not better then write it to a file

brian-brazil:
I very much doubt that a known insecure practice like this is the standard way to do this with Docker. Please change the example to not include a password on the command line.

dabgmx:
"docker run" has a an argument "--env-file" to pass variables. I have modified the example to consume a file with the variables. Like this they are not passed on the command line.

brian-brazil:
Thanks!

PR #1531, opened 2020-01-13
fixed layout of the routing-tree-editor to avoid stripping bottom of the canvas

FUSAKLA:
Fixes https://github.com/prometheus/alertmanager/issues/2159

The routing tree editor has stripped bottom of the canvas when having odd number of routes and one going straight down. This should fix the stripping and also make sure the footer is below the tree.


juliusv:
:+1: Thanks!

PR #1532, opened 2020-01-14
add Kuma (Service Mesh) to the list of file_sd based integrations

yskopets:


brian-brazil:
Thanks!

You shouldn't be suggesting that users change refresh_interval. That's only a fallback in case inotify breaks, the default is fine.

PR #1536, opened 2020-01-28
Document 'mrkdwn_in' field in Alertmanager's Slack configuration

simonpasquier:
From https://github.com/prometheus/alertmanager/pull/1967#issuecomment-578686833


content/docs/alerting/configuration.md
```diff
@@ -531,6 +531,7 @@ actions:
 fields:
   [ <field_config> ... ]
 [ footer: <tmpl_string> | default = '{{ template "slack.default.footer" . }}' ]
+[ mrkdwn_in: [ <string> ] | default = ["fallback", "pretext", "text"] ]
```
brian-brazil: This isn't the syntax we use, look up at equal.

PR #1538, opened 2020-02-01
Remove link to Unifi exporter from exporters.md

mhansen:
The Unifi exporter repo has been archived, no more development, and no nominated successor.


brian-brazil:
As long as the link works, and the exporter isn't completely broken, I'd keep it here. When a suitable successor comes along it can replace it.

mhansen:
OK, I'll drop this.

PR #1539, opened 2020-02-01
Add prometheus-msteams integrations

bzon:
I'm adding the project https://github.com/bzon/prometheus-msteams to the alert manager integrations.


roidelapluie:
We generally only list one integration per third party.

bzon:
@roidelapluie may I ask why? Our intention is to provide an alternative to https://github.com/idealista/prom2teams. I understand that it is bad to have a large number list of alternatives but I think having two considered stable projects is not that bad? Our project is also contributor friendly for Go developers. The alternate is written in Python.

roidelapluie:
I will quote @brian-brazil : 
> To encourage the community to rally around one solution, and not continuously create new ones from scratch, we aim to avoid duplicates in these lists. Which of these is the most popular and/or in line with best practices?

bzon:
> To encourage the community to rally around one solution, and not continuously create new ones from scratch, we aim to avoid duplicates in these lists.

I totally agree on this but we also have to be open-minded and consider **thriving** and **proven** projects out there.

> Which of these is the most popular and/or in line with best practices?

prometheus-msteams has 2 million docker pulls with 111 Github stars as of this writing.

With regards to best practices, I will not say that the project is better than the prom2teams. This is subjective from the reviewer perspective and standard. But we do have:

* 70% test coverage
* Code quality is validated with golangci lint standards.
* Changelogs.
* Well documented installation and features.
* Also, the templating library that the project uses is from Alertmanager Go code.

Lastly, we are in the work of moving the main project from **my** account to its own organization https://github.com/prometheus-msteams including the docker image to https://quay.io/organization/prometheusmsteams so the image can be scanned for vulnerabilities.

That is all that I have to say. If the maintainers of Prometheus is still against this, then I, the main maintainer of this project will not object furthermore with peace.  🍻

brian-brazil:
For best practices we mean Prometheus best practices, which don't really apply to notifiers - it's more an exporter thing.

I will only list one project, but it appears we have a case here where there are two projects started within 6 months of each other about 2 years ago, and both have a non-trivial community around them.

I try to follow the community in relation to these, and it's not immediately clear to me which is the more popular. What do the engagement metrics look like for both? Are there features that one has but not the other? Has any attempt ever been made to work together?

I also notice that you appear to have secrets (webhook urls) on the command line in your readme, which is a security issue which needs to be fixed.

bzon:
@brian-brazil 

> I will only list one project, but it appears we have a case here where there are two projects started within 6 months of each other about 2 years ago, and both have a non-trivial community around them.

I don't want their project to be removed in the integrations list and I don't want this to be a competition between two projects. If it really has to be only one, then I don't want this pull request to proceed anymore.

> I also notice that you appear to have secrets (webhook urls) on the command line in your readme, which is a security issue which needs to be fixed.

Do you mean the example webhook urls like `https://outlook.office.com/webhook/xxxx/aaa/bbb`? Those are examples and not valid urls but I can remove them.

brian-brazil:
This is a messier situation one than we usually deal with. Usually it's a brand new project with no community yet wanting to be additionally listed.

> Those are examples and not valid urls but I can remove them.

Yes, the problem is that you're encouraging users to put secrets on the command line. That command line flag should probably be removed.

jan--f:
Hello from the bug scrub!
Nowadays alertmanager can send [MS Teams notifications](https://prometheus.io/docs/alerting/latest/configuration/#msteamsv2_config). Hence I will close this for now. Please feel free to reopen at any time if this addition is still needed.

PR #1543, opened 2020-02-08
Query log user guide

roidelapluie:

### DIFF
--- a/content/docs/guides/query-log.md
+++ b/content/docs/guides/query-log.md
@@ -1,0 +1,181 @@
+---
+title: Query Log
+sort_rank: 1
+---
+
+# Using the Prometheus query log
+
+Prometheus has the ability to log all the queries run by the engine to a log
+file. This guide demonstrates how to use that log file, which fields it does
+contain and provides advanced tips about how to operate the log file.
+
+The query log has been introduced in v2.16.0 and is disabled by default.
+
+## Enable the query log
+
+The query log can be enabled and disabled at runtime. It can therefore be
+activated when you want to investigate slownesses or high load on your
+Prometheus instance.
+
+To enable or disable the query log, two steps are needed:
+1. Adapt the configuration to add or remove the query log configuration.
+1. Reload the Prometheus server configuration.
+
+### Logging all the queries to a file.
+
+This example demonstrates how to log all the queries to
+a file called `/var/log/prometheus/query.log`.
+
+First, let's ensure that Prometheus has write access to a directory
+`/var/log/prometheus`. We will assume that Prometheus is running under a user
+called Prometheus:
+
+```shell
+# mkdir /var/log/prometheus
+# chown prometheus: /var/log/prometheus
+```
+
+Then, adapt the `prometheus.yml` configuration file:
+
+```yaml
+global:
+  scrape_interval:     15s
+  evaluation_interval: 15s
+  query_log_file: /var/log/prometheus/query.log
+scrape_configs:
+  - job_name: 'prometheus'
+    - targets: ['localhost:9090']
+```
+
+And, reload the Prometheus configuration:
+
+```shell
+$ curl -X POST http://127.0.0.1:9090/-/reload
+```
+
+Or, if Prometheus is not launched with `--web.enable-lifecycle`:
+
+```shell
+$ killall -HUP prometheus
+```
+
+The file `/var/log/prometheus/query.log` should now exist and all the queries
+will be logged to that file.
+
+To disable the query log, repeat the operation but remove `query_log_file` from
+the configuration.
+
+## Verifying if the query log is enabled
+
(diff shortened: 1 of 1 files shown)


content/docs/guides/query-log.md
```diff
@@ -0,0 +1,182 @@
...
+sort_rank: 1
+---
+
+# Using the Prometheus query log
+
+Prometheus has the ability to log all the queries run by the engine to a log
+file. This guide demonstrates how to use that log file, which fields it does
+contain and provides advanced tips about how to operate the log file.
+
+The query log has been introduced in v2.16.0 and is disabled by default.
```
brian-brazil: This is very awkward. The version should only be mentioned in passing, and the enablement is what the rest is about.

content/docs/guides/query-log.md
```diff
@@ -0,0 +1,182 @@
+---
+title: Query Log
+sort_rank: 1
+---
+
+# Using the Prometheus query log
+
+Prometheus has the ability to log all the queries run by the engine to a log
+file. This guide demonstrates how to use that log file, which fields it does
```
brian-brazil: it contains

content/docs/guides/query-log.md
```diff
@@ -0,0 +1,182 @@
...
+contain and provides advanced tips about how to operate the log file.
```
brian-brazil: aren't we doing Oxford commas?

content/docs/guides/query-log.md
```diff
@@ -0,0 +1,182 @@
...
+
+This example demonstrates how to log all the queries to
+a file called `/var/log/prometheus/query.log`.
+
+First, let's ensure that Prometheus has write access to a directory
+`/var/log/prometheus`. We will assume that Prometheus is running under a user
+called Prometheus:
+
+```shell
+# mkdir /var/log/prometheus
```
brian-brazil: This fails to mention the need for root. Let's keep to examples that don't require root unless it's strictly required.

content/docs/guides/query-log.md
```diff
@@ -0,0 +1,182 @@
...
+1. Reload the Prometheus server configuration.
+
+### Logging all the queries to a file
+
+This example demonstrates how to log all the queries to
+a file called `/var/log/prometheus/query.log`.
+
+First, let's ensure that Prometheus has write access to a directory
+`/var/log/prometheus`. We will assume that Prometheus is running under a user
+called Prometheus:
```
brian-brazil: prometheus is not the same as Prometheus.

content/docs/guides/query-log.md
```diff
@@ -0,0 +1,182 @@
...
+    - targets: ['localhost:9090']
+```
+
+And, reload the Prometheus configuration:
+
+```shell
+$ curl -X POST http://127.0.0.1:9090/-/reload
+```
+
+Or, if Prometheus is not launched with `--web.enable-lifecycle`:
```
brian-brazil: I'd do this the other way around, mention the flag first.

content/docs/guides/query-log.md
```diff
@@ -0,0 +1,182 @@
...
+
+```shell
+$ killall -HUP prometheus
```
brian-brazil: You're making an assumption of OS here, the purpose of killall varies quite a bit.

content/docs/guides/query-log.md
```diff
@@ -0,0 +1,182 @@
...
+
+The first metric, `prometheus_engine_query_log_enabled` is set to 1 of the
+query log is enabled, and 0 otherwise.
+The second one, `prometheus_engine_query_log_failures_total`, indicates the
+number of queries that could not be logged.
+
+## Format of the query log
+
+The query log is a JSON-formatted log. Here is an overview of the fields
+present on any query:
```
brian-brazil: for a query

content/docs/guides/query-log.md
```diff
@@ -0,0 +1,182 @@
...
+        }
+    },
+    "ts": "2020-02-08T14:59:50.387Z"
+}
+```
+
+- params: the query itself. The start and end timestamp, the step and the actual
+  query.
+- stats: statistics. Currently, it contains internal engine timers.
+- ts: timestamp of the log line.
```
brian-brazil: Capitalisation is inconsistent here.

Surely it's the timestamp of the query, not the log line. Is that when the query starts or ends?

content/docs/guides/query-log.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+
+## Rotating the query log
+
+Prometheus will not rotate the query log itself. Instead, you can use external
+tools to do so.
+
+One of those tools is logrotate. Please refer to the documentation of your
```
brian-brazil: Link to those docs.

roidelapluie:
I will followup on your comments in another pr

PR #1550, opened 2020-02-13
When new team members are added, notify CNCF

matthiasr:
As team members, we are CNCF maintainers. Note that we should notify
them on changes.

There is currently no role-email for this. To avoid publishing
indivduals' email addresses, link to the private prometheus-team mailing
list topic that has the contacts and context.

I believe this is an editorial change?

### DIFF
--- a/content/governance.md
+++ b/content/governance.md
@@ -40,6 +40,7 @@ If they choose to accept, the following steps are taken:
 * Team members are added to the [team mailing list][team].
 * Team members are added to the list of team members in this document.
 * New team members are announced on the [developers mailing list][devs] by an existing team member.
+* New team members are announced to the [CNCF team][cncf]. Refer to [this private topic][cncf-maintainer-contact] for contact details.
 
 Team members may retire at any time by emailing [the team][team].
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
:+1: 

Yeah, I'd view this as editorial as we're not changing any of the governance rules. Let's leave merging until say tomorrow just in case anyone disagrees with that.

matthiasr:
at what point should we break those out into a different document?

beorn7:
> at what point should we break those out into a different document?

That's another thing I had been thinking about: We could keep "procedural" things in a separate doc to minimize editorial changes to the governance so that we are less often in the situation where we have to make the call if a change is just editorial or not.

beorn7:
But of course, all of that is just brainstorming and out of scope of this particular PR.

RichiH:
As this changes a process, I would argue that it's not an editorial change and would thus disagree with this PR.

As I don't want to be the one blocking without offering a solution, I volunteer to rewrite this, split out on-/offboarding, create a small process for that, and then put it all up for a vote.

matthiasr:
> I volunteer to rewrite this, split out on-/offboarding, create a small process for that, and then put it all up for a vote.

perfect, thank you very much!

RichiH:
https://github.com/prometheus/docs/pull/1552

PR #1551, opened 2020-02-13
Adding azure-health-exporter to the exporter list

louisfelix:
Hi,
Adding `azure-health-exporter`, an exporter exposing Azure Health API status as metrics.
Thanks!


brian-brazil:
Thanks!

Your metrics should start with azure_health or similar, as that what they represent.

I see some code that's very similar to what's in the azure metrics exporter here. Someone might want to start a shared library for that at some point.

louisfelix:
Thanks @brian-brazil for merge and comments!

> Your metrics should start with azure_health or similar, as that's what they represent.

I would change it to start with `azure_resource_health...`, as there is 2 other types of health not yet supported by the exporter (Azure Service health and Azure status page). 
What do you think?

> I see some code that's very similar to what's in the azure metrics exporter here. Someone might want to start a shared library for that at some point.

I would be happy to work on that (as soon as find time for it), I didn't know if it was something you would be OK with. Doing so, we may want to take the opportunity to completely switch azure_metrics_exporter to Go modules.

brian-brazil:
> What do you think?

That sounds reasonable.

PR #1552, opened 2020-02-14
Update governance, split out on-/offboarding

RichiH:

### DIFF
--- a/content/governance.md
+++ b/content/governance.md
@@ -46,6 +46,9 @@ Upon death of a member, they leave the team automatically.
 
 In case a member leaves, the [offboarding](#offboarding) procedure is applied.
 
+In case of certified mental incapacitation of a member, the [offboarding](#offboarding) procedure can be applied partially or in full to ensure account and access security and voting rights and quorum count will be suspended.
+This is applicable until the certified mental incapacitation is over.
+
 The current team members are:
 
 * Bartłomiej Płotka ([Red Hat](https://www.redhat.com/))
--- a/content/governance.md
+++ b/content/governance.md
@@ -38,7 +38,10 @@ If they choose to accept, the [onboarding](#onboarding) procedure is followed.
 
 Team members may retire at any time by emailing [the team][team].
 
-Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any changes to this and related documents any more. In case more than one removal vote runs at the same time, these limitations apply only to to respective removal votes, not globally.
+Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team].
+For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any changes to this and related documents any more.
+Multiple members can be removed at the same time, but every vote can only cover one single person.
+For parallel, distinct removal votes every team member retains full voting rights and quorum relevance except their own.
 
 Upon death of a member, their team membership ends automatically.
 
@@ -166,7 +169,7 @@ The new member is
 
 The ex-member is
 * removed from the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR. In case of forced removal, no approval is needed.
-* removed from the [GitHub organization][gh] and related organizations and repositories.
+* removed from the [GitHub organization][gh] and related organizations and repositories. Optionally, they can retain maintainership of one or more repositories if the [team](#team-members) agrees.
 * removed from the [team mailing list][team].
 * announced as removed to CNCF. We will explicitly request reconfirmation of removal from CNCF.
 * removed from shared password storage. All passwords, API tokens, etc are cycled.
--- a/content/governance.md
+++ b/content/governance.md
@@ -160,7 +160,7 @@ The new member is
 * added to the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR.
 * announced on the [developers mailing list][devs] by an existing team member. Ideally, the new member replies in this thread, acknowledging team membership.
 * added to the [GitHub organization][gh] as _Owner_.
-* optionally added to the community, junkyard, and related repositories.
+* optionally added to the community, junkyard, and related organizations and repositories.
 * added to the [team mailing list][team].
 * announced to CNCF.
 * given access to the shared password storage.
--- a/content/governance.md
+++ b/content/governance.md
@@ -34,18 +34,15 @@ New members may be proposed by any existing member by email to [prometheus-team]
 
 If the new member proposal is accepted, the proposed team member should be contacted privately via email to confirm or deny their acceptance of team membership. This email will also be CC'd to [prometheus-team][team] for record-keeping purposes.
 
-If they choose to accept, the following steps are taken:
-
-* Team members are added to the [GitHub organization][gh] as _Owner_.
-* Team members are added to the [team mailing list][team].
-* Team members are added to the list of team members in this document.
-* New team members are announced on the [developers mailing list][devs] by an existing team member.
+If they choose to accept, the [onboarding](#onboarding) procedure is followed.
 
 Team members may retire at any time by emailing [the team][team].
 
-Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote and does not count towards the quorum.
+Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any [editorial](#editorial-changes) changes to this and related documents any more.
 
 Upon death of a member, their team membership ends automatically.
+
+In case the member leaves, the [offboarding](#offboarding) procedure is applied.
 
 The current team members are:
 
@@ -94,6 +91,11 @@ Decisions are usually made by [lazy consensus](#consensus). If no consensus can 
 ### Governance changes
 
 Material changes to this document are discussed publicly on the [developer mailing mailing list][devs]. Any change requires a [supermajority](#supermajority-vote) in favor. Editorial changes may be made by [lazy consensus](#consensus) unless challenged.
+
+### Editorial changes
+
+Editorial changes are changes which fix spelling or grammar.
+They must be made via PR, be left unmerged for at least 48 hours, and need the approval of at least one other [team member](#team-members).
 
 ### Other matters
 
@@ -144,6 +146,32 @@ Votes may take the form of a single proposal, with the option to vote yes or no,
 A vote on a single proposal is considered successful if at least two thirds of those eligible to vote vote in favor.
 
 If there are multiple alternatives, members may vote for one or more alternatives, or vote “no” to object to all alternatives. A vote on multiple alternatives is considered decided in favor of one alternative if it has received the most votes in favor, and a vote from at least two thirds of those eligible to vote. Should no alternative reach this quorum, another vote on a reduced number of options may be called separately.
+
+## On- / Offboarding
+
+Changes in the On- /Offboarding section can be taken by [lazy consensus](#consensus) unless challenged.  If no consensus can be reached, the matter may be resolved by [majority vote](#majority-vote).
+
+### Onboarding
+
+The new member is
+* added to the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR.
+* announced on the [developers mailing list][devs] by an existing team member. Ideally, the new member replies in this thread.
+* added to the [GitHub organization][gh] as _Owner_.
+* added to the [team mailing list][team].
+* announced to CNCF via Amye Scavarda Perrin and Chris Aniszczyk, or via a role account once existing.
+* given access to the shared password storage.
+
+### Offboarding
+
+The ex-member is
+* removed from the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR. In case of forced removal, no approval is needed.
+* removed from the [GitHub organization][gh].
+* removed from the [team mailing list][team].
+* announced as removed to CNCF via Amye Scavarda Perrin and Chris Aniszczyk, or via a role account once existing. We will explicitly request reconfirmation of removal from CNCF.
+* removed from shared password storage. All passwords, API tokens, etc are cycled.
+* not allowed to call themselves an active team member any more, nor allowed to imply this to be the case.
+
+If needed, we reserve the right to publicly announce removal.
 
 ## FAQ
 
--- a/content/governance.md
+++ b/content/governance.md
@@ -46,9 +46,6 @@ Upon death of a member, they leave the team automatically.
 
 In case a member leaves, the [offboarding](#offboarding) procedure is applied.
 
-In case of certified mental incapacitation of a member, the [offboarding](#offboarding) procedure can be applied partially or in full to ensure account and access security and voting rights and quorum count will be suspended.
-This is applicable until the certified mental incapacitation is over.
-
 The current team members are:
 
 * Bartłomiej Płotka ([Red Hat](https://www.redhat.com/))
--- a/content/governance.md
+++ b/content/governance.md
@@ -38,7 +38,7 @@ If they choose to accept, the [onboarding](#onboarding) procedure is followed.
 
 Team members may retire at any time by emailing [the team][team].
 
-Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any changes to this and related documents any more.
+Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any changes to this and related documents any more. In case more than one removal vote runs at the same time, these limitations apply only to to respective removal votes, not globally.
 
 Upon death of a member, their team membership ends automatically.
 
@@ -149,15 +149,15 @@ If there are multiple alternatives, members may vote for one or more alternative
 
 ## On- / Offboarding
 
-Changes in the On- /Offboarding section can be taken by [lazy consensus](#consensus) unless challenged.  If no consensus can be reached, the matter may be resolved by [majority vote](#majority-vote).
+The On- / Offboarding section is informational and can be changed by [lazy consensus](#consensus) unless challenged.  If no consensus can be reached, the matter may be resolved by [majority vote](#majority-vote).
 
 ### Onboarding
 
 The new member is
 * added to the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR.
-* announced on the [developers mailing list][devs] by an existing team member. Ideally, the new member replies in this thread, acknowleding team membership.
+* announced on the [developers mailing list][devs] by an existing team member. Ideally, the new member replies in this thread, acknowledging team membership.
 * added to the [GitHub organization][gh] as _Owner_.
-* optinally added to the community, junkyard, and related repositories.
+* optionally added to the community, junkyard, and related repositories.
 * added to the [team mailing list][team].
 * announced to CNCF.
 * given access to the shared password storage.
@@ -171,13 +171,13 @@ The ex-member is
 * announced as removed to CNCF. We will explicitly request reconfirmation of removal from CNCF.
 * removed from shared password storage. All passwords, API tokens, etc are cycled.
 * not allowed to call themselves an active team member any more, nor allowed to imply this to be the case.
-* added to an alumni list if they agree to it.
+* added to a list of alumnae if they agree to it.
 
 If needed, we reserve the right to publicly announce removal.
 
 ## FAQ
 
-This section is informational. In case of disagreement, the rules above overrule any FAQ.
+The FAQ section is informational and can be changed by [lazy consensus](#consensus) unless challenged.  If no consensus can be reached, the matter may be resolved by [majority vote](#majority-vote).
 
 ### How do I propose a decision?
 
--- a/content/governance.md
+++ b/content/governance.md
@@ -39,9 +39,8 @@ If they choose to accept, the [onboarding](#onboarding) procedure is followed.
 Team members may retire at any time by emailing [the team][team].
 
 Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team].
-For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any changes to this and related documents any more.
-Multiple members can be removed at the same time, but every vote can only cover one single person.
-For parallel, distinct removal votes every team member retains full voting rights and quorum relevance except their own.
+For this vote, the member in question is not eligible to vote and does not count towards the quorum.
+Any removal vote can cover only one single person.
 
 Upon death of a member, their team membership ends automatically.
 
@@ -97,7 +96,8 @@ Material changes to this document are discussed publicly on the [developer maili
 
 ### Editorial changes
 
-Editorial changes are changes which fix spelling or grammar.
+Editorial changes are changes which fix spelling or grammar, update work affilition, or similar; they update style or reflect an outside and obvious reality.
+They do not change the intention or meaning of anything in this document.
 They must be made via PR, be left unmerged for at least 48 hours, and need the approval of at least one other [team member](#team-members).
 
 ### Other matters
--- a/content/governance.md
+++ b/content/governance.md
@@ -174,7 +174,7 @@ The ex-member is
 * announced as removed to CNCF. We will explicitly request reconfirmation of removal from CNCF.
 * removed from shared password storage. All passwords, API tokens, etc are cycled.
 * not allowed to call themselves an active team member any more, nor allowed to imply this to be the case.
-* added to a list of alumnae if they agree to it.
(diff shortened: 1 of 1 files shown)


RichiH:
I will let this sit here for feedback before calling for a formal vote.

content/governance.md
```diff
@@ -34,19 +34,16 @@ New members may be proposed by any existing member by email to [prometheus-team]
...
+If they choose to accept, the [onboarding](#onboarding) procedure is followed.
 
 Team members may retire at any time by emailing [the team][team].
 
-Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote and does not count towards the quorum.
+Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any [editorial](#editorial-changes) changes to this and related documents any more.
 
 Upon death of a member, their team membership ends automatically.
 
+In case the member leaves, the [offboarding](#offboarding) procedure is applied.
```
roidelapluie: Should that include stepping down as maintainer of repos?

content/governance.md
```diff
@@ -34,19 +34,16 @@ New members may be proposed by any existing member by email to [prometheus-team]
...
-* Team members are added to the [GitHub organization][gh] as _Owner_.
-* Team members are added to the [team mailing list][team].
-* Team members are added to the list of team members in this document.
-* New team members are announced on the [developers mailing list][devs] by an existing team member.
+If they choose to accept, the [onboarding](#onboarding) procedure is followed.
 
 Team members may retire at any time by emailing [the team][team].
 
-Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote and does not count towards the quorum.
+Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any [editorial](#editorial-changes) changes to this and related documents any more.
```
brian-brazil: Why is this duplicating the removal docs? It seems very odd to call out this particular privilege of being a team member.

RichiH: To fix the race condition of them making "editorial" changes to the offboarding process.

content/governance.md
```diff
@@ -145,6 +147,32 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
+
+### Offboarding
+
+The ex-member is
+* removed from the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR. In case of forced removal, no approval is needed.
+* removed from the [GitHub organization][gh].
+* removed from the [team mailing list][team].
+* announced as removed to CNCF via Amye Scavarda Perrin and Chris Aniszczyk, or via a role account once existing. We will explicitly request reconfirmation of removal from CNCF.
+* removed from shared password storage. All passwords, API tokens, etc are cycled.
+* not allowed to call themselves an active team member any more, nor allowed to imply this to be the case.
```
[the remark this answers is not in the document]
RichiH: Will add.

content/governance.md
```diff
@@ -145,6 +147,32 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
+## On- / Offboarding
+
+Changes in the On- /Offboarding section can be taken by [lazy consensus](#consensus) unless challenged.  If no consensus can be reached, the matter may be resolved by [majority vote](#majority-vote).
+
+### Onboarding
+
+The new member is
+* added to the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR.
+* announced on the [developers mailing list][devs] by an existing team member. Ideally, the new member replies in this thread.
+* added to the [GitHub organization][gh] as _Owner_.
```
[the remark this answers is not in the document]
RichiH: I considered this, but think that's in the social, not the technical scope of the project maintenance. No strong feelings either way, so will add and see what others think.

content/governance.md
```diff
@@ -145,6 +147,32 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
+* added to the [team mailing list][team].
+* announced to CNCF via Amye Scavarda Perrin and Chris Aniszczyk, or via a role account once existing.
```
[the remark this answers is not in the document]
RichiH: I didn't ask and will remove their names.

content/governance.md
```diff
@@ -145,6 +147,32 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
 
+## On- / Offboarding
+
+Changes in the On- /Offboarding section can be taken by [lazy consensus](#consensus) unless challenged.  If no consensus can be reached, the matter may be resolved by [majority vote](#majority-vote).
+
+### Onboarding
+
+The new member is
+* added to the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR.
+* announced on the [developers mailing list][devs] by an existing team member. Ideally, the new member replies in this thread.
```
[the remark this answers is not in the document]
RichiH: They should acknowledge them becoming maintainers on the -dev list

content/governance.md
RichiH: That is a valid question and should see broader discussion. In the case of forced removal, most certainly.

We always have the mechnanism of team votes which can force anything, including removal as repo maint.

RichiH: Actually, I changed to

`* removed from the [GitHub organization][gh] and related organizations and repositories.
`

which covers pretty much anything. Default to save and re-add as needed.

content/governance.md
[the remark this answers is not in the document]
RichiH: Leaving this as unresolved on purpose; feel free to close or suggest different wording.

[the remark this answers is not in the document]
RichiH: Just updated anyway.

content/governance.md
brian-brazil: This is still very weird, this sentence is about the special rules of a removal vote - why is is talking about changing this doc? The governance already covers changes to it, it's not like someone can unilaterally change it.

content/governance.md
```diff
@@ -34,19 +34,16 @@ New members may be proposed by any existing member by email to [prometheus-team]
...
-* Team members are added to the [GitHub organization][gh] as _Owner_.
-* Team members are added to the [team mailing list][team].
-* Team members are added to the list of team members in this document.
-* New team members are announced on the [developers mailing list][devs] by an existing team member.
+If they choose to accept, the [onboarding](#onboarding) procedure is followed.
 
 Team members may retire at any time by emailing [the team][team].
 
-Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote and does not count towards the quorum.
+Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any changes to this and related documents any more.
```
beorn7: Perhaps clarify that the procedure is separate for each individual member? I vaguely remember the joke at the dev summit that went like: "I want to vote everybody out except me. I'm the only one eligible to vote, and I vote for it."

content/governance.md
```diff
@@ -145,6 +147,34 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
+
+Changes in the On- /Offboarding section can be taken by [lazy consensus](#consensus) unless challenged.  If no consensus can be reached, the matter may be resolved by [majority vote](#majority-vote).
+
+### Onboarding
+
+The new member is
+* added to the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR.
+* announced on the [developers mailing list][devs] by an existing team member. Ideally, the new member replies in this thread, acknowleding team membership.
+* added to the [GitHub organization][gh] as _Owner_.
+* optinally added to the community, junkyard, and related repositories.
```
beorn7: optionally

content/governance.md
```diff
@@ -145,6 +147,34 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
 
+## On- / Offboarding
+
+Changes in the On- /Offboarding section can be taken by [lazy consensus](#consensus) unless challenged.  If no consensus can be reached, the matter may be resolved by [majority vote](#majority-vote).
+
+### Onboarding
+
+The new member is
+* added to the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR.
+* announced on the [developers mailing list][devs] by an existing team member. Ideally, the new member replies in this thread, acknowleding team membership.
```
beorn7: acknowledging

content/governance.md
```diff
@@ -145,6 +147,34 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
+* optinally added to the community, junkyard, and related repositories.
+* added to the [team mailing list][team].
+* announced to CNCF.
+* given access to the shared password storage.
+
+### Offboarding
+
+The ex-member is
+* removed from the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR. In case of forced removal, no approval is needed.
+* removed from the [GitHub organization][gh] and related organizations and repositories.
```
beorn7: As we explicitly allow repository maintainers that are not -team members, we should also allow the transition from -team member to maintainer-only. This could come in handy if somebody wants to retire but is willing to stay a maintainer for a transition period. And it's even more relevant should we decide to sanction against too many -team members from the same company (in which case -team membership becomes a possible liability as you could "steal" voting power from your colleagues, which might make you prefer to be a maintainer without being a -team member).

content/governance.md
```diff
@@ -145,6 +147,34 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
+* removed from the [team mailing list][team].
+* announced as removed to CNCF. We will explicitly request reconfirmation of removal from CNCF.
+* removed from shared password storage. All passwords, API tokens, etc are cycled.
+* not allowed to call themselves an active team member any more, nor allowed to imply this to be the case.
+* added to an alumni list if they agree to it.
```
beorn7: _alumni_ is the masculine form. Suggestion: "added to a list of alumnae and alumni if they agree to it."

content/governance.md
```diff
@@ -145,6 +147,34 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
+
+If needed, we reserve the right to publicly announce removal.
+
 ## FAQ
 
 This section is informational. In case of disagreement, the rules above overrule any FAQ.
```
beorn7: Now that we have a section that is marked explicitly as changeable by lazy consensus (or a normal vote if needed), perhaps we could do the same here? Or just cut the whole doc into the actual governance and and informational appendix (with onboarding/offboarding list, FAQ, and possibly more).

content/governance.md
roidelapluie: It seems like changing the document is the least damage an harmful member could do. Maybe it would be better in case of a force removal vote, that the team member gets its owner rights dropped in github and the requester also drops his owner rights at the same time. Or just leave it as it.

content/governance.md
RichiH: I RAN :set spell ON THE INITIAL COMMIT, BUT NOT THIS ONE :(

Thanks

RichiH: (New computer, vimrc not copied yet)

content/governance.md
RichiH: The implied statement is that we can always decide to do this anyway and have a safe default.

This question is more around "what is the default, especially when we disagree", not "what can we do if we agree".

Should I put in verbiage to make this clearer?

content/governance.md
RichiH: Boilerplated.

content/governance.md
RichiH: Never had Latin; rephrased in the plural (I think/hope).

content/governance.md
RichiH: Good point; I will add that. For now, I would prefer one doc, but no strong opinion.

content/governance.md
RichiH: We now allow lazy consensus in On-/Offboarding, and in the FAQ. That's what I want to prevent.

content/governance.md
beorn7: I phrased it in an unclear way. I didn't want to create two documents, but have a separator in the one doc that says something like "everything from here in is informational and can be changed by lazy consensus etc."

content/governance.md
beorn7: Yes. I think we should distinguish between "this has to happen in any case" and "this can be done differently if we agree to".

content/governance.md
```diff
@@ -145,9 +147,37 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
+### Offboarding
+
+The ex-member is
+* removed from the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR. In case of forced removal, no approval is needed.
+* removed from the [GitHub organization][gh] and related organizations and repositories.
+* removed from the [team mailing list][team].
+* announced as removed to CNCF. We will explicitly request reconfirmation of removal from CNCF.
+* removed from shared password storage. All passwords, API tokens, etc are cycled.
+* not allowed to call themselves an active team member any more, nor allowed to imply this to be the case.
+* added to a list of alumnae if they agree to it.
```
beorn7: This is now only the feminine form. Personally, I'm fine with it, definitely better than using the masculine form in an anyway all-male club. But my original suggestion was including both genders, which is probably the best in general.

content/governance.md
beorn7: Unresolved to make my original suggestion visible again. Resolve again if you really want to use the feminine form only.

beorn7: Context: _alumni_ is plural masc., _alumnae_ is plural fem.

content/governance.md
```diff
@@ -34,19 +34,16 @@ New members may be proposed by any existing member by email to [prometheus-team]
...
-* Team members are added to the [GitHub organization][gh] as _Owner_.
-* Team members are added to the [team mailing list][team].
-* Team members are added to the list of team members in this document.
-* New team members are announced on the [developers mailing list][devs] by an existing team member.
+If they choose to accept, the [onboarding](#onboarding) procedure is followed.
 
 Team members may retire at any time by emailing [the team][team].
 
-Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote and does not count towards the quorum.
+Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any changes to this and related documents any more. In case more than one removal vote runs at the same time, these limitations apply only to to respective removal votes, not globally.
```
beorn7: "to to respective removal votes" → "to each relevant removal vote individually"

This still doesn't clarify that the vote "I want to remove Alice, Bob, and Chris" is invalid.

content/governance.md
RichiH: Added.

content/governance.md
RichiH: I was not aware that this was a thing (and the Internet lied to me about it), thanks!

"ex-member list"? I would prefer to avoid gendered terms at all, if possible.

content/governance.md
RichiH: See above; "ex-members" ok? I don't like gendered terms at all for brevity and as they exclude non-binary people.

content/governance.md
RichiH: Rephrased completely to be more explicit.

roidelapluie:
Previous members is nicer

content/governance.md
RichiH: Should I make this more explicit? Reword? Happy?

brian-brazil: Leave the wording as it was, end the sentence at quorum.

content/governance.md
```diff
@@ -34,19 +34,19 @@ New members may be proposed by any existing member by email to [prometheus-team]
...
 
-Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote and does not count towards the quorum.
+Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team].
+For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any changes to this and related documents any more.
+Multiple members can be removed at the same time, but every vote can only cover one single person.
+For parallel, distinct removal votes every team member retains full voting rights and quorum relevance except their own.
 
 Upon death of a member, their team membership ends automatically.
 
+In case the member leaves, the [offboarding](#offboarding) procedure is applied.
```
brian-brazil: What if they die? They don't leave in that case, but their membership ends.

content/governance.md
```diff
@@ -95,6 +95,11 @@ Decisions are usually made by [lazy consensus](#consensus). If no consensus can
 
 Material changes to this document are discussed publicly on the [developer mailing mailing list][devs]. Any change requires a [supermajority](#supermajority-vote) in favor. Editorial changes may be made by [lazy consensus](#consensus) unless challenged.
 
+### Editorial changes
+
+Editorial changes are changes which fix spelling or grammar.
```
brian-brazil: This is far more restrictive than what we have done in practice. For example someone changing company would now need a super-majority vote to update this doc.

content/governance.md
```diff
@@ -34,19 +34,19 @@ New members may be proposed by any existing member by email to [prometheus-team]
...
-* Team members are added to the [team mailing list][team].
-* Team members are added to the list of team members in this document.
-* New team members are announced on the [developers mailing list][devs] by an existing team member.
+If they choose to accept, the [onboarding](#onboarding) procedure is followed.
 
 Team members may retire at any time by emailing [the team][team].
 
-Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team]. For this vote, the member in question is not eligible to vote and does not count towards the quorum.
+Team members can be removed by [supermajority vote](#supermajority-vote) on [the team mailing list][team].
+For this vote, the member in question is not eligible to vote, does not count towards the quorum, and may not make any changes to this and related documents any more.
```
brian-brazil: You can simplify this by stating that a given vote for removal can only cover one person, the rest then follows from that automatically.

content/governance.md
RichiH: If the membership ends by automation, they leave, so the procedure applies.

content/governance.md
RichiH: Happy?

content/governance.md
RichiH: Valid point, will re-word.

content/governance.md
brian-brazil: Different terminology is used, so it's plausible that these are different things where different rules applies. If you want to make all these things explicit, it should be watertight.

RichiH: Re-worded. Should we also take certified mental degradation into account. Think not.

content/governance.md
```diff
@@ -95,6 +94,12 @@ Decisions are usually made by [lazy consensus](#consensus). If no consensus can
 
 Material changes to this document are discussed publicly on the [developer mailing mailing list][devs]. Any change requires a [supermajority](#supermajority-vote) in favor. Editorial changes may be made by [lazy consensus](#consensus) unless challenged.
 
+### Editorial changes
+
+Editorial changes are changes which fix spelling or grammar, update work affilition, or similar; they update style or reflect an outside and obvious reality.
+They do not change the intention or meaning of anything in this document.
+They must be made via PR, be left unmerged for at least 48 hours, and need the approval of at least one other [team member](#team-members).
```
roidelapluie: nip: 48h after the last edit (commit change)

RichiH: I rephrased and linked back to lazy consensus; that, by definition, can't be rushed.

RichiH:
Very valid points; I honestly had forgotten about Conor. Last I talked to
him, Prometheus was a job, nothing more.

Should we ask Matt?

beorn7:
> Should we ask Matt?

Matt is kind of an easy case as he is explicitly acknowledged in the Acknowledgement section on https://prometheus.io/community/ . I'm more concerned about other people that were never on prometheus-team because they contributed before prometheus-team existed. Might become a rabbit hole.

I suggest to think about it once a team member has left and actually wishes to be listed. Then we can decide about the correct format of the alumni/ae list.

content/governance.md
RichiH: This was a serious question, btw. Still think we don't need to, just went exploring corner cases as "someone in a coma" would mean someone else gaining access to their accounts, passwords, etc.

roidelapluie:
On 17 Feb 02:31, Richard Hartmann wrote:
> RichiH commented on this pull request.
> 
> 
> 
> >  
>  Upon death of a member, their team membership ends automatically.
>  
> +In case the member leaves, the [offboarding](#offboarding) procedure is applied.
> 
> This was a serious question, btw. Still think we don't need to, just went exploring corner cases as "someone in a coma" would mean someone else gaining access to their accounts, passwords, etc.

I think that this should not be automated. A forced removal vote would be enough.



-- 
 (o-    Julien Pivotto
 //\    Open-Source Consultant
 V_/_   Inuits - https://www.inuits.eu

content/governance.md
RichiH: Added verbiage.

content/governance.md
RichiH: Beorn approved PR in full.

content/governance.md
RichiH: Beorn approved PR in full.

content/governance.md
RichiH: Beorn approved PR in full.

content/governance.md
```diff
@@ -46,6 +46,9 @@ Upon death of a member, they leave the team automatically.
 
 In case a member leaves, the [offboarding](#offboarding) procedure is applied.
 
+In case of certified mental incapacitation of a member, the [offboarding](#offboarding) procedure can be applied partially or in full to ensure account and access security and voting rights and quorum count will be suspended.
```
roidelapluie: Without a vote?

RichiH: Yes. I like to automate cases which emotional load; and it would be emotional for us.

NB: We can always vote to override this case-by-case.

brian-brazil: This varies a lot based on jurisdiction, and do you mean permanently or just until they're no longer sectioned? This seems very easy to abuse, if it comes to it we can call a vote.

beorn7: I think this is way to blurry to have it in here as anything we can do without a vote.

And if we have to vote on it anyway, we don't really need a separate section.

WRT account and access security: If an account is compromised, it has to be protected. But we don't need to spell that out in the governance.

content/governance.md
```diff
@@ -145,9 +150,37 @@ A vote on a single proposal is considered successful if at least two thirds of t
...
+
+The On- / Offboarding section is informational and can be changed by [lazy consensus](#consensus) unless challenged.  If no consensus can be reached, the matter may be resolved by [majority vote](#majority-vote).
+
+### Onboarding
+
+The new member is
+* added to the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR.
+* announced on the [developers mailing list][devs] by an existing team member. Ideally, the new member replies in this thread, acknowledging team membership.
+* added to the [GitHub organization][gh] as _Owner_.
+* optionally added to the community, junkyard, and related repositories.
```
juliusv: repositories -> organisations?

content/governance.md
```diff
@@ -95,6 +94,12 @@ Decisions are usually made by [lazy consensus](#consensus). If no consensus can
 
 Material changes to this document are discussed publicly on the [developer mailing mailing list][devs]. Any change requires a [supermajority](#supermajority-vote) in favor. Editorial changes may be made by [lazy consensus](#consensus) unless challenged.
 
+### Editorial changes
+
+Editorial changes are changes which fix spelling or grammar, update work affilition, or similar; they update style or reflect an outside and obvious reality.
+They do not change the intention or meaning of anything in this document.
+They must be made via PR and accepted via [lazy consensus](#consensus).
```
brian-brazil: Could this prevent a majority vote on these?

RichiH:
No as that is always the fallback. I had half a mind to restructure the
whole consensus and voting references, but decided to tackle that in the
future.

content/governance.md
roidelapluie: are we follwing rfc2119 ?
1. MUST   This word, or the terms "REQUIRED" or "SHALL", mean that the
   definition is an absolute requirement of the specification.

roidelapluie: @brian-brazil actually it is also said: Editorial changes *may* be made by lazy consensus unless challenged.

brancz: I do think the challenge part is important because whether something is editorial or not may be subjective.

roidelapluie: @brancz I do not understand your concern. The challenge part is covered by the lazy consensus rules.

roidelapluie: And the challenged part is also present in this new document, 3 lines above. I was quoting it.

brancz: Sorry it seems I updated my comment before you submitted yours. We are in agreement.

content/governance.md
RichiH: Done.

content/governance.md
RichiH: Removed that section.

RichiH:
For back-reference, we are now voting on this in "[VOTE] New governance document" on prometheus-dev

RichiH:
Quorum has been reached and 16 in favour votes are documented at https://groups.google.com/forum/#!topic/prometheus-developers/4mHqIUlloA0 .

23 * 0.66.. = 15.3.., so 16 is a 2/3 majority.

PR #1569, opened 2020-03-03
Add link to ircd_exporter

dgl:
Signed-off-by: David Leadbeater <dgl@dgl.cx>


content/docs/instrumenting/exporters.md
```diff
@@ -85,6 +85,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [EMQ exporter](https://github.com/nuvo/emq_exporter)
    * [Gearman exporter](https://github.com/bakins/gearman-exporter)
    * [IBM MQ exporter](https://github.com/ibm-messaging/mq-metric-samples/tree/master/cmd/mq_prometheus)
+   * [IRCd exporter](https://github.com/dgl/ircd_exporter)
```
brian-brazil: I think this name here might be a bit confusing, as it's not for a particular IRC server. "IRC network" maybe?

IRC is also not a queuing system, which the rest of these are, so it should go down in misc.

dgl: I've moved it, I think IRCd exporter is reasonable, it can actually work in two modes; `--stats.local-only` enables a single server exporter mode, so it can used as an "IRCd" exporter; when I write "IRC network" exporter it somehow sounds like something it isn't...

PR #1570, opened 2020-03-05
Changed guide to follow best practices 

vonneudeck:
These changes are done to resolve issues Brian Brazil pointed out.
I reworded it to:
- be clearer
- make it less likely that readers form bad habits
- correct language mistakes

Signed-off-by: Nicolai von Neudeck <nicolai@vonneudeck.com>

### DIFF
--- a/content/docs/guides/multi-target-exporter.md
+++ b/content/docs/guides/multi-target-exporter.md
@@ -22,13 +22,13 @@ By multi-target [exporter](/docs/instrumenting/exporters/) pattern we refer to a
 * the exporter subsequently starts the scrape after getting Prometheus’ GET requests and once it is done with scraping.
 * the exporter can query multiple targets.
 
-This pattern is only used for certain exporters, namely the [blackbox](https://github.com/prometheus/blackbox_exporter) and the [SNMP exporter](https://github.com/prometheus/snmp_exporter).
+This pattern is only used for certain exporters, such as the [blackbox](https://github.com/prometheus/blackbox_exporter) and the [SNMP exporter](https://github.com/prometheus/snmp_exporter).
 
 The reason is that we either can’t run an exporter on the targets, e.g. network gear speaking SNMP, or that we are explicitly interested in the distance, e.g. latency and reachability of a website from a specific point outside of our network, a common use case for the [blackbox](https://github.com/prometheus/blackbox_exporter) exporter.
 
 ## Running multi-target exporters
 
-Multi-target exporters are flexible regarding their environment and can be run in many ways. As regular programs, in containers, as background services, on baremetal, virtual machines. Because they are queried and do query over network they do need appropriate open ports. Otherwise they are frugal.
+Multi-target exporters are flexible regarding their environment and can be run in many ways. As regular programs, in containers, as background services, on baremetal, on virtual machines. Because they are queried and do query over network they do need appropriate open ports. Otherwise they are frugal.
 
 Now let’s try it out for yourself!
 
@@ -302,10 +302,11 @@ NOTE: If you use Docker for Mac or Docker for Windows, you can’t use `localhos
 `prometheus.yml` for Linux:
 
 ```yaml
-scrape_configs:
-
-- job_name: blackbox-exporter-metrics
+global:
   scrape_interval: 5s
+
+scrape_configs:
+- job_name: blackbox # To get metrics about the exporter itself
   metrics_path: /metrics
   static_configs:
     - targets:
@@ -315,10 +316,11 @@ scrape_configs:
 `prometheus.yml` for macOS and Windows:
 
 ```yaml
-scrape_configs:
-
-- job_name: blackbox-exporter-metrics
+global:
   scrape_interval: 5s
+
+scrape_configs:
+- job_name: blackbox # To get metrics about the exporter itself
   metrics_path: /metrics
   static_configs:
     - targets:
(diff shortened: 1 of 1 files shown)


content/docs/guides/multi-target-exporter.md
```diff
@@ -399,17 +401,15 @@ The details are complicated and out of scope for this guide. Hence we will limit
 Here is the config you will use to do that. Don’t worry if this is a bit much at once, we will go through it step by step:
 
 ```yaml
+global:
+  scrape_interval:
```
brian-brazil: You're missing the value here

content/docs/guides/multi-target-exporter.md
```diff
@@ -399,17 +401,15 @@ The details are complicated and out of scope for this guide. Hence we will limit
...
+global:
+  scrape_interval:  
 scrape_configs:
-
 - job_name: blackbox-exporter-metrics
-  scrape_interval: 5s
   metrics_path: /metrics
-  static_configs:
-    - targets:
+  static_configs: # to get metrics about the exporter itself - targets:
```
brian-brazil: Unsplit this

content/docs/guides/multi-target-exporter.md
```diff
@@ -399,17 +401,15 @@ The details are complicated and out of scope for this guide. Hence we will limit
...
       - localhost:9115   # for Windows and macOS replace with - host.docker.internal:9115
 
-- job_name: 'blackbox'
-  scrape_interval: 5s
+- job_name: blackbox # to get metrics about targets we want the exporter to query for us
```
brian-brazil: You still need to distinguish the types of probe jobs

content/docs/guides/multi-target-exporter.md
```diff
@@ -494,4 +494,4 @@ That is it. Restart the Prometheus docker container and look at your [metrics](h
 
 # Summary
 
-In this guide, you learned how the multi-target exporter pattern works and how to run a blackbox exporter with a customised module and to configure Prometheus using relabeling to scrape metrics with prober labels.
+In this guide, you learned how the multi-target exporter pattern works, how to run a blackbox exporter with a customised module and to configure Prometheus using relabeling to scrape metrics with prober labels.
```
brian-brazil: We use the Oxford comma in our docs, so you need a comma before the "and" too

content/docs/guides/multi-target-exporter.md
```diff
@@ -315,10 +316,11 @@ scrape_configs:
...
 
 ```yaml
-scrape_configs:
-
-- job_name: blackbox-exporter-metrics
+global:
   scrape_interval: 5s
+
+scrape_configs:
+- job_name: blackbox-exporter # to get metrics about the exporter itself
```
brian-brazil: It might be better to put these above rather than beside

vonneudeck: It is already not very good to even have a comment instead of a more explicit name. If we move it up the connection will be completely lost.

brian-brazil: Having a comment before the block it refers to is pretty normal, I'm more thinking about the horizontal space this takes up.

content/docs/guides/multi-target-exporter.md
vonneudeck: I know why you want that and I think this is a terrible idea for teaching.

vonneudeck: You want that because it fits into a bigger scheme.
But the bombardment with bigger scheme everywhere is the reason why people don’t understand what is going on and need guidance in the first place.

content/docs/guides/multi-target-exporter.md
vonneudeck: This 72 chars (with spaces) in total. Small enough for an old terminal.

And "normal" was not my argument. I want to teach something. I needs to be there so people can make the connection when only focusing on one line. Two lines are far more complicated and won’t work as much.

content/docs/guides/multi-target-exporter.md
brian-brazil: I'm not saying you have to go into great depth, but the example should at least hint at how to handle this. blackbox on its own is not a good job name, unless it refers to the exporter itself.

brian-brazil: One option that just occurred to me is blackbox_e2e, which has the distinction without mentioning the module name.

vonneudeck: I understand your concern and I have changed both job names the way you proposed in the first place. `blackbox_e2e` would in my opinion be even more confusing and I want to keep the guide as clear as possible. I just want you to know and understand that while your naming scheme is great for the daily praxis, it is not necessarily ideal for learning. Making things unnecessarily hard to understand in guides for beginners is a form of gate-keeping and we have more than enough of that already. For successfully teaching physics it is better to start with classical mechanics. Quantum physics and relativity theory have to come later.

content/docs/guides/multi-target-exporter.md
```diff
@@ -351,24 +353,24 @@ docker \
...
 
 ```yaml
-scrape_configs:
-
-- job_name: blackbox-exporter-metrics
+global:
   scrape_interval: 5s
+
+scrape_configs:
+- job_name: blackbox # to get metrics about the exporter itself
```
brian-brazil: Be consistent in capitalisation of your comments, they've varying across the file.

brian-brazil:
Thanks!

PR #1577, opened 2020-03-13
Link to Prometheus::Tiny client library for Perl

guillemj:
Add a reference to this Perl library.

### DIFF
--- a/content/docs/instrumenting/clientlibs.md
+++ b/content/docs/instrumenting/clientlibs.md
@@ -32,6 +32,7 @@ Unofficial third-party client libraries:
 * [.NET / C#](https://github.com/prometheus-net/prometheus-net)
 * [Node.js](https://github.com/siimon/prom-client)
 * [Perl](https://metacpan.org/pod/Net::Prometheus)
+* [Perl](https://metacpan.org/pod/Prometheus::Tiny)
 * [PHP](https://github.com/endclothing/prometheus_client_php)
 * [Rust](https://github.com/pingcap/rust-prometheus)
 


brian-brazil:
Thanks for your PR. To encourage the community to work together we only list one entry for a given language/runtime. I'd suggest trying to first improve the existing entry, before looking to replace it.

guillemj:
Hmm, these two Perl modules serve different audiences I guess, so I'm not sure there's much point in trying to unify them. The tiny one implies no or little dependencies and being very lightweight. The existing entry has more dependencies, and contains explicit type support, and seems catered at instrumenting native services, while it makes using it to instrument external services more cumbersome. And while the latter could be improved, I'd personally still want to use a tiny variant f.ex.

Personally I find it most useful to have a list of all available and current client libraries to choose from, so that one does not need to look for alternatives or consider implementing their own, or assume there's nothing else to use. But if that's not your policy then I guess disregard this PR.  :)

brian-brazil:
> seems catered at instrumenting native services, while it makes using it to instrument external services more cumbersome. 

Most client libraries don't have sufficient support to correctly write exporters, so that's not that big of a deal. Direct instrumentation is the primary use case.

> Personally I find it most useful to have a list of all available and current client libraries to choose from,

You can get that from your favourite search engine, I try to curate a bit. Otherwise we'd have 5-10 client libraries listed per language/runtime, each reinventing the wheel, most abandoned, and end users no good way to evaluate which is the "best" one.

jan--f:
Hello from the bug scrub!
The proposed project seems to be rather inactive these days, so I'll close this for now. Please re-open, if this should still be added.

PR #1579, opened 2020-03-15
Adding link to R Prometheus client library

cfmack:
Signed-off-by: cfmack <cfmack@gmail.com>

### DIFF
--- a/content/docs/instrumenting/clientlibs.md
+++ b/content/docs/instrumenting/clientlibs.md
@@ -33,8 +33,8 @@ Unofficial third-party client libraries:
 * [Node.js](https://github.com/siimon/prom-client)
 * [Perl](https://metacpan.org/pod/Net::Prometheus)
 * [PHP](https://github.com/endclothing/prometheus_client_php)
-* [Rust](https://github.com/pingcap/rust-prometheus)
 * [R](https://github.com/cfmack/pRometheus)
+* [Rust](https://github.com/pingcap/rust-prometheus)
 
 When Prometheus scrapes your instance's HTTP endpoint, the client library
 sends the current state of all tracked metrics to the server.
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for sharing. R goes before Rust.


I've only used R as a CLI tool, and application library - what sort of things does it typically do as a daemon and how would it be deployed?

Looking at the code itself:

I'm not seeing escaping in help or label values.
A client library should only be storing things in memory, something like Redis is just going to be too slow as it involves hitting the kernel and network stack. There's no need to make where the metrics are stored pluggable - and looking at the code it's doing things like marshalling JSON on every observation, which is not going to be performant either. Generally the one of the official clients is a better base to work from.
I'm not seeing any locking, is R always single-threaded and single-process?
Are you checking that the passed label values match up to the label names?
Are you checking that label values are valid UTF-8?
It's odd to have namespace without subsystem, and these days we're tending towards having neither and just relying on name.
You shouldn't need a `getGauge` call, that's an anti-pattern. Have users remember the gauge's address when they create it: https://www.robustperception.io/label-lookups-and-the-child
I'm not sure you're setting the content type to a correct value.
What is type used for? It seemed unused, and is confusing naming in that Prometheus client libraries already have types such as gauge and counter.

I'd generally recommend a read through https://prometheus.io/docs/instrumenting/writing_clientlibs/

cfmack:
@brian-brazil ,  great feedback and I will get to it.

Plumber rstudio supported way of making APIs in front of R, including models and other systems that already use R.   R is indeed single threaded.  The way around this people use docker containers do to this limitation.  I have experience using this in production environments and docker makes this pretty slick (as you can imagine).   [Read more](https://www.rplumber.io/docs/hosting.html#docker)

Additionally, there are some language specific actions that may require tweaks between R and other languages.   I'll start taking action on your questions and notes.  Feedback appreciated.  Thank you.

cfmack:
@brian-brazil , alphabetical order corrected, btw.  Thanks again.

brian-brazil:
Thanks!

> Additionally, there are some language specific actions that may require tweaks between R and other languages.

That's always the case, there's still various design goals that should be aimed for as much as is practical.

PR #1580, opened 2020-03-17
Fix list formatting in Governance doc

juliusv:
Lists have to be in a paragraph of their own to be detected as lists by
the Markdown processor we use.

### DIFF
--- a/content/governance.md
+++ b/content/governance.md
@@ -156,6 +156,7 @@ The On- / Offboarding section is informational and can be changed by [lazy conse
 ### Onboarding
 
 The new member is
+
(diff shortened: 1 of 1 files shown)


juliusv:
@RichiH small editorial (formatting fix) change

RichiH:
Thanks! I actually saw that, then read up on how much time I need to cool-off a editorial change and then forgot to send the PR...

Waiting for netlify to finish.

juliusv:
@RichiH Yep - and always a good idea to look at the Netlify preview (or run things locally) to check that things are actually formatted correctly :)

RichiH:
Point taken ;)

PR #1581, opened 2020-03-17
Add ex-member section

RichiH:

### DIFF
--- a/content/governance.md
+++ b/content/governance.md
@@ -71,6 +71,10 @@ The current team members are:
 * Tobias Schmidt ([SoundCloud](https://soundcloud.com/))
 * Tom Wilkie ([Grafana Labs](https://grafana.com/))
 
+The ex-members who are publicly listed are:
+
+* Max Inden
+
 ### Maintainers
 
 Maintainers lead one or more project(s) or parts thereof and serve as a point of conflict resolution amongst the contributors to this project. Ideally, maintainers are also team members, but exceptions are possible for suitable maintainers that, for whatever reason, are not yet team members.


content/governance.md
```diff
@@ -71,6 +71,10 @@ The current team members are:
 * Tobias Schmidt ([SoundCloud](https://soundcloud.com/))
 * Tom Wilkie ([Grafana Labs](https://grafana.com/))
 
+The ex-members who are publicly listed are:
```
beorn7: Perhaps better "The ex-members that elected to be publicly listed are:"

Everybody can see that they are publicly listed, so there is not much of an information in there right now.

One might argue the reader will know what is implied. But even then, I'd use more concise wording: "Publicly listed es-members:"

However, there more problems with this list. First of all, we never even asked @conr if he wants to be listed (which is easy to solve, we just do so: @conr, would you like to be included in this list? :o).

A more subtle problem is how to deal with all the significant contributors to the project that have never been team members because prometheus-team didn't exist when they were active. Formally, that's not a problem, but it might send a message of ingratitude. How about adding a line describing that fact? "Note that the project had many significant contributions before a formal team membership was established on 201x-xx-xx."

RichiH: Then maybe `ex-members` is simply the most concise.

I had planned to contact everyone who helped a lot in the past if we should list them. E.g. @matttproud

beorn7: Matt certainly has a very special place in the project. (As already acknowledged in https://prometheus.io/docs/introduction/faq/#who-wrote-prometheus and in the bottom section of  https://prometheus.io/community/ . But he was never member of prometheus-team because prometheus-team as a formal entity was only introduced when he already had become an alumnus.

We could give him a shoutout as a founder here again (but only with his consent, of course). However, if we start to list people as ex-members that never have been formal members of the formal entity prometheus-team, we would put ourselves into a position where we have to judge all contributors that had been active before prometheus-team and make the call if they qualify to be listed here or not. I don't think that's feasible.

conr: >  @conr, would you like to be included in this list?
Sure! Hope you're all doing well 👍

beorn7: Added you in #1583 .

PR #1583, opened 2020-03-17
Improve wording around former team membership

beorn7:
- Use "Previous members" rather than "Ex-members" in agreement with
  the wording in the offboarding list.

- Add a paragraph noting that there were significant contributors
  before prometheus-team existed.

### DIFF
--- a/content/governance.md
+++ b/content/governance.md
@@ -72,6 +72,7 @@ The current team members are:
 
 Previous members:
 
+* Conor Broderick
 * Max Inden
 * Stuart Nelson
 
--- a/content/governance.md
+++ b/content/governance.md
@@ -70,10 +70,12 @@ The current team members are:
 * Tobias Schmidt ([SoundCloud](https://soundcloud.com/))
 * Tom Wilkie ([Grafana Labs](https://grafana.com/))
 
-The ex-members who are publicly listed are:
+Previous members:
 
 * Max Inden
 * Stuart Nelson
+
+_Note that Prometheus has received significant contributions from a number of unlisted individuals before this governance document and thus formal team membership was created._
 
 ### Maintainers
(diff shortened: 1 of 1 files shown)


content/governance.md
```diff
@@ -70,11 +70,13 @@ The current team members are:
 * Tobias Schmidt ([SoundCloud](https://soundcloud.com/))
 * Tom Wilkie ([Grafana Labs](https://grafana.com/))
 
-The ex-members who are publicly listed are:
+Previous members:
```
beorn7: Note that I removed the shenanigans about "publicly listed" as per the discussion on #1581 .

content/governance.md
```diff
@@ -70,11 +70,13 @@ The current team members are:
...
 
 * Max Inden
 * Stuart Nelson
 
+_Note that Prometheus has received significant contributions from a number of unlisted individuals before this governance document and thus formal team membership was created._
```
beorn7: If Matt gives us his agreement, we could add something like: "This includes one of the founders of the project, Matt T. Proud."

beorn7:
Now also including @conr .

beorn7:
Can I get a review please? Pinging @roidelapluie @bwplotka @gouthamve @juliusv who were on the "listing ex-members" mail thread. (Not pinging RichiH because he took the day off today and I don't want to disturb him.)

I don't think this is overly controversial, and we can still amend later, in particular in case we get consent from Matt to give him credits here.

content/governance.md
bwplotka: This is really nice, well put :+1:

brian-brazil:
:+1:

beorn7:
Thanks everyone.

PR #1585, opened 2020-03-17
Add Google Cloud Spanner Adapter to Prometheus integration doc.

codeactivist:
Adds a remote read/write storage adapter for Google Cloud Spanner.


brian-brazil:
Thanks!

I'd suggest not depending on common/model, I'm hoping to remove most of that at some point.

codeactivist:
Thank you Brian, I will look into removing common/model.

PR #1586, opened 2020-03-21
BigBlueButton exporter

greenstatic:
I have created a Prometheus exporter for BigBlueButton that the community can use along with a Grafana dashboard.

Signed-off-by: greenstatic <gregor@krmelj.xyz>


content/docs/instrumenting/exporters.md
```diff
@@ -224,6 +224,7 @@ separate exporters are needed:
    * [App Connect Enterprise](https://github.com/ot4i/ace-docker)
    * [Ballerina](https://ballerina.io/)
    * [BFE](https://github.com/baidu/bfe)
+   * [BigBlueButton](https://github.com/greenstatic/bigbluebutton-exporter)
```
brian-brazil: This looks to be an exporter for an API, rather than an application itself

brian-brazil:
In relation to the code, it seems it's polling on a (rather tight) loop and updating direct instrumentation rather than using a custom collector with GaugeMetricFamily - which would be simpler and less load.

content/docs/instrumenting/exporters.md
greenstatic: > This looks to be an exporter for an API, rather than an application itself

Yes that is correct, the exporter uses the applications [official API](https://docs.bigbluebutton.org/dev/api.html) that is documented on their website for integrations.

brian-brazil: Yes, you've put it in the wrong section.

greenstatic:
> In relation to the code, it seems it's polling on a (rather tight) loop and updating direct instrumentation rather than using a custom collector with GaugeMetricFamily - which would be simpler and less load.

Thank you for your feedback. This is my first Prometheus exporter, so I am quite new at this. If I understand correctly `GaugeMetricFamily` is used when you expose metrics within the application itself (but the metrics you are exposing are not directly under your control; e.g. a library). Like you have noticed already this exporter uses the applications official API to gather information, thus making `GaugeMetricFamily` not useful in this case, correct?

I agree that performance wise it is better to implement the functionality in the application itself, however that was not a possibility in my case due to tight time constraints. The project (BigBlueButton) is being heavily adopted currently in education due to the COVID-19 virus and the need to have online lectures. My organization needed to have monitoring up and running ASAP and I wanted to share this with others who are in similar situation.

I suspect interfacing with applications via its API and exposing metrics is not common for exporters?

content/docs/instrumenting/exporters.md
greenstatic: My apologies, which section is appropriate?

brian-brazil:
> I understand correctly GaugeMetricFamily is used when you expose metrics within the application itself (but the metrics you are exposing are not directly under your control; e.g. a library)

It's for when they're coming from a 3rd party. That's sometimes a library, but usually a different process/system as is the case here.

> Like you have noticed already this exporter uses the applications official API to gather information, thus making GaugeMetricFamily not useful in this case, correct?

It's exactly where GaugeMetricFamily should be used.

> I suspect interfacing with applications via its API and exposing metrics is not common for exporters?

It's common, we've an entire section for exporters like this: https://prometheus.io/docs/instrumenting/exporters/#apis

greenstatic:
Ah I see. It took me a while to fully understand your concerns. 

For anybody else reading who comes across this, the main discussion is having a process poll indefinitely an API to collect metrics and then update it directly opposed to using the built-in Prometheus client collector technique - where you create a custom class with the `collect` method and return a list (usually using the generator pattern and `yield`). The latter is more efficient because the `collect` method is only triggered once a HTTP request comes to the exporter opposed to the polling method which isn't dependent on the number/frequency of HTTP requests to the exporter.

One of my main concerns regarding the exporter was the time it takes for the service (BigBlueButton) API to respond. In total if I perform all the required API calls sequentially this can easily be 5 seconds. For this reason I decided to constantly poll (with the option of specifying the delay time) and update the metrics manually. Albeit this could cause the exporter to serve a few seconds of stale metrics (hopefully this isn't a big issue in the metrics world), it was my opinion at the time that this acceptable since the metrics endpoint of the exporter would respond without long delays - I ignored the tiny race condition where a response would come while I updated the metrics manually. This method would also allow for multiple clients to perform metrics requests without triggering additional service API calls, since the values are sort of being cached by manually updating the metrics. Also since Prometheus would scrape at a regular (e.g. 5 second) interval which would trigger service requests anyway I thought polling wouldn't matter as much.

I agree this method isn't particularly efficient and may serve a few second stale metrics, however I think this is better than having long (5 seconds, and possibly longer) response times if I perform the service API calls "live" (by using the custom collector method you propose). I am open for discussion and all ears if you have any suggestions or tips on an improved implementation with the issue of long request delays in mind - this surely has already come up before?

I will move my entry in the documentation to the APIs section after you deem the project ready.

brian-brazil:
> think this is better than having long (5 seconds, and possibly longer) response times if I perform the service API calls "live" (by using the custom collector method you propose).

Typical scrape intervals are more than 5s, so it's not a problem for an exporter. That's only really a concern when you are getting towards a minute.

greenstatic:
> Typical scrape intervals are more than 5s, so it's not a problem for an exporter. That's only really a concern when you are getting towards a minute.

So if I understand correctly, it is common for the metrics endpoint to take 5 seconds to respond? I see the default scrape timeout is 10 sec, which I guess is a worse case scenario. Are there any other projects that you know of that require more than 5 sec to respond to a metrics request?

roidelapluie:
> > Typical scrape intervals are more than 5s, so it's not a problem for an exporter. That's only really a concern when you are getting towards a minute.
> 
> So if I understand correctly, it is common for the metrics endpoint to take 5 seconds to respond? I see the default scrape timeout is 10 sec, which I guess is a worse case scenario. Are there any other projects that you know of that require more than 5 sec to respond to a metrics request?

that's frequent with jmx_exporter or snmp_exporter

greenstatic:
I have reimplemented the exporter using `GaugeMetricFamily` like you suggested, as well as moved the entry in the docs to the `APIs` section.

brian-brazil:
Thanks!

If you don't have labels on a MetricFamily then you can set the value in the constructor and immediately yield it.

PR #1587, opened 2020-03-23
Move Fabian to ex-members

RichiH:


RichiH:
@fabxc please confirm.

PR #1451, opened 2019-10-07
Add Awesome Prometheus collection of related resources

avivace:
Signed-off-by: Antonio Vivace <avivace4@gmail.com>

PR #1590, opened 2020-03-24
Add Awesome Prometheus

beorn7:
Supersedes #1451, picking up the review comments from there.

### DIFF
--- a/content/docs/introduction/media.md
+++ b/content/docs/introduction/media.md
@@ -10,6 +10,8 @@ collecting all Prometheus-related resources on the internet.
 
 The following selection of resources are particularly useful to get started
 with Prometheus.
+[Awesome Prometheus](https://github.com/roaldnefs/awesome-prometheus)
+contains a more comprehensive community-maintained list of resources.
 
 ## Blogs
 


roidelapluie:
"High Availability" does not seem to do what it looks like. We put a lot of effort reviewing this list, I don't know if we should point to a 3rd party resource like that.

brian-brazil:
:+1: 

> "High Availability" does not seem to do what it looks like. We put a lot of effort reviewing this list, I don't know if we should point to a 3rd party resource like that.

This is an ancient list, not much curation was involved.

beorn7:
No list will ever be perfect. I don't think a disputed terminology usage should keep us from linking it.

I like the general approach of the "Awesome …" lists. The one linked in this PR is the only Prometheus list in the [central meta list](https://github.com/sindresorhus/awesome). My optimistic approach is that by linking it here, it will only get better.

@roidelapluie is your comment just a general doubt, or is it intended to block merging this PR?

roidelapluie:
> No list will ever be perfect. I don't think a disputed terminology usage should keep us from linking it.
> 
> I like the general approach of the "Awesome …" lists. The one linked in this PR is the only Prometheus list in the [central meta list](https://github.com/sindresorhus/awesome). My optimistic approach is that by linking it here, it will only get better.
> 
> @roidelapluie is your comment just a general doubt, or is it intended to block merging this PR?

general doubt as I know that we generally avoid linking exporter with secrets on command line etc, linking such a list here could expose those unsecure exporters to the community.

but I am fine merging it for the good of the community :)

PR #1592, opened 2020-03-30
Add link to powershell client

jobec:


brian-brazil:
Hmm, there looks to be at least 3 of these out there. Do you know which of these is the most popular?

jobec:
I only today published mine.

As far as I could see, there's none out there that's can act as a real exported. All of them seem to use the push gateway and don't have a webserver built in.

What others did you find?

brian-brazil:
https://github.com/rgl/PowerShellExporter and https://github.com/guhuajun/prometheus_posh_client

Given what powershell is, pgw support would seem to be the main thing. I'd also hope that the community would work towards one integration, rather than frequently starting from scratch.

jobec:
I found both a they didn't do what I needed in a simple way. Hence my new implementation.

Closing as I don't want to get into a discussion which one is better.

And no, push gateway is not "the way" for powershell. Why depend on yet another component when it can run a webserver just as well. It's the same for Python, Perl, PHP,...

PR #1594, opened 2020-03-31
official rabbitmq exporter

eez0:
@juliusv 

The following PR replaces the link to third party RabbitMQ exporters with a link to the official exporter provided by RabbitMQ.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -90,8 +90,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [NSQ exporter](https://github.com/lovoo/nsq_exporter)
    * [Mirth Connect exporter](https://github.com/vynca/mirth_exporter)
    * [MQTT blackbox exporter](https://github.com/inovex/mqtt_blackbox_exporter)
-   * [RabbitMQ exporter](https://github.com/kbudde/rabbitmq_exporter)
-   * [RabbitMQ Management Plugin exporter](https://github.com/deadtrickster/prometheus_rabbitmq_exporter)
+   * [RabbitMQ exporter](https://github.com/rabbitmq/rabbitmq-prometheus)
    * [RocketMQ exporter](https://github.com/apache/rocketmq-exporter)
    * [Solace exporter](https://github.com/dabgmx/solace_exporter)
 


brian-brazil:
We do have a preference for listing official exporters, however some due diligence is required as with all cases where we replace a listing. Does this have a superset of the features of the listed exporters? Is it doing anything practices wise that's not as good as the existing ones? Have attempts been made to work with the existing communities?

roidelapluie:
Is is linked in the section with native support with this link too https://www.rabbitmq.com/prometheus.html

Is this redundant?

eez0:
@roidelapluie the link I used is specific about the rabbitmq-prometheus plugin. The documentation link you mentioned is rather an overview of the whole RabbitMQ monitoring setup using Prometheus + Grafana.

@brian-brazil to be honest, I don't know what differs between the official and the other exporters. Just from a quick sight, it looks the official provides way more metrics https://github.com/rabbitmq/rabbitmq-prometheus/blob/master/metrics.md.

brian-brazil:
> Is is linked in the section with native support with this link too

That would have been an oversight on my part then. We should confirm that that supersedes both of the listed exporters, and then remove them.

roidelapluie:
The exporter proposed in this PR is used in the link we have in the native instrumentation list.

eez0:
Ok just checked it out. It's indeed later in the site, in the [Software exposing Prometheus metrics section](https://prometheus.io/docs/instrumenting/exporters/#software-exposing-prometheus-metrics).

So the question now is:
- Should we consider third party exporters different from native exporters (which are third party from the Prometheus' point of view in any case)?
- Should we move up the native section to take precedence over third party?
- Should we do nothing and close this PR?

brian-brazil:
A given piece of software should (almost always) only get one entry on this page. If a piece of software produces good Prometheus metrics itself there's no need to list an exporter, and if it has poor metrics then we would only list an exporter.

Without digging, I'd guess here that we should be removing the exporter entries.

gerhard:
As one of the [rabbitmq/rabbitmq-prometheus](https://github.com/rabbitmq/rabbitmq-prometheus) authors, while I would like to say deprecate [kbudde/rabbitmq_exporter](https://github.com/kbudde/rabbitmq_exporter) which uses RabbitMQ's HTTP API, I can't because rabbitmq-prometheus does not have feature parity with rabbitmq_exporter (one of the biggest wants is per queue metrics). Until the missing functionality is implemented, I would keep **kbudde/rabbitmq_exporter** in this list.

On a related subject, watch [Observe and understand RabbitMQ - Gerhard Lazu & Michal Kuratczyk](https://www.youtube.com/watch?v=L-tYXpirbpA&t=72) for 3 minutes to understand why all monitoring solutions that use the RabbitMQ's HTTP API for metrics are problematic (this includes @kbudde's **rabbitmq_exporter**). While we want to eventually remove metrics from RabbitMQ's HTTP API (a.k.a. rabbitmq-management), it's unlikely to happen in 2020, and predicting anything beyond this timeline would be guessing - your guess is just as good as mine.

cc @michaelklishin @mkuratczyk

brian-brazil:
> Until the missing functionality is implemented, I would keep kbudde/rabbitmq_exporter in this list.

That sounds reasonable then, let's wait for that and then clean things up.

PR #1596, opened 2020-04-08
Fix Apdex score query

garrettthomaskth:
Fix score query according to cited formula http://en.wikipedia.org/wiki/Apdex
@brian-brazil


brian-brazil:
The query is correct, keep in mind that the values are cumulative so the satisfied queries are double counted.

PR #1597, opened 2020-04-09
Copy over Cortex changes

RichiH:

### DIFF
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -12,12 +12,13 @@ environments.
 This page describes the general security assumptions of Prometheus and the
 attack vectors that some configurations may enable.
 
-As with any complex system, it is not possible to guarantee that there are no
-bugs. If you find a security bug please report it privately to the maintainers
-listed in the MAINTAINERS.md of the relevant repository and CC
-prometheus-team@googlegroups.com. We will fix the issue and coordinate a
-release date with you, acknowledging your effort and mentioning you by name
-if you want.
+As with any complex system, it is certain that bugs will be found, some of
+them security-relevant. If you find a _security bug_ please report it
+privately to the maintainers listed in the MAINTAINERS of the relevant
+repository and CC cortex-team@googlegroups.com. We will fix the issue as soon
+as possible and coordinate a release date with you. You will be able to choose
+if you want public acknowledgement of your effort and if you want to be
+mentioned by name.
 
 ## Prometheus
 
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -12,7 +12,7 @@ environments.
 This page describes the general security assumptions of Prometheus and the
 attack vectors that some configurations may enable.
 
-As with any complex system, it is certain that bugs will be found, some of
+As with any complex system, it is near certain that bugs will be found, some of
 them security-relevant. If you find a _security bug_ please report it
 privately to the maintainers listed in the MAINTAINERS of the relevant
 repository and CC prometheus-team@googlegroups.com. We will fix the issue as soon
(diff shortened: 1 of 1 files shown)


brian-brazil:
I don't think this is the change you meant to propose.

content/docs/operating/security.md
```diff
@@ -12,12 +12,13 @@ environments.
...
-As with any complex system, it is not possible to guarantee that there are no
-bugs. If you find a security bug please report it privately to the maintainers
-listed in the MAINTAINERS.md of the relevant repository and CC
-prometheus-team@googlegroups.com. We will fix the issue and coordinate a
-release date with you, acknowledging your effort and mentioning you by name
-if you want.
+As with any complex system, it is certain that bugs will be found, some of
+them security-relevant. If you find a _security bug_ please report it
+privately to the maintainers listed in the MAINTAINERS of the relevant
+repository and CC cortex-team@googlegroups.com. We will fix the issue as soon
```
roidelapluie, line 18:
```suggestion
repository and CC prometheus-team@googlegroups.com. We will fix the issue as soon
```

RichiH:
@brian-brazil Do your concerns go beyond what @roidelapluie pointed out?

content/docs/operating/security.md
```diff
@@ -12,12 +12,13 @@ environments.
 This page describes the general security assumptions of Prometheus and the
 attack vectors that some configurations may enable.
 
-As with any complex system, it is not possible to guarantee that there are no
-bugs. If you find a security bug please report it privately to the maintainers
-listed in the MAINTAINERS.md of the relevant repository and CC
-prometheus-team@googlegroups.com. We will fix the issue and coordinate a
-release date with you, acknowledging your effort and mentioning you by name
-if you want.
+As with any complex system, it is certain that bugs will be found, some of
```
brian-brazil, line 15: It's not certain, just extremely likely.

RichiH: Without descending into the halting problem and the differences of having and finding: Do you care for this to be changed or can it be merged as-is?

RichiH:
The motivation is
* I think the new text is better
* I prefer to have similar texts be as similar as possible to spot deliberate differences more quickly

content/docs/operating/security.md
brian-brazil: I care, as the change as proposed is incorrect.

brian-brazil:
That's not really answering my question. Why do you think this new process would be better?

RichiH:
There is no new process that I can see. It clears up wording, nothing more.

brian-brazil:
It's not just a cleanup though, the meaning is being changed.

content/docs/operating/security.md
juliusv: Is there a misunderstanding here regarding "it is certain that bugs will be found"? Because the first time I read it, I parsed it as "any bugs that exist will be found", but then I realized what's meant is probably that people will find bugs, but obviously not all of them (which makes much more sense).

brian-brazil: My reading of it is that it is 100% certain that bugs will be found (the exact sense of that doesn't matter) - and you can't ever be 100% certain of things, particularly into the future.

RichiH:
@brian-brazil I take it that you agree with the changes as-is?

brian-brazil:
I'm still waiting for an explanation for the new process, as this is not a mere rewording.

RichiH:
I'm still waiting for an explanation for what new process you are seeing.

Stretching it: We don't force people to go publicly on the record that they were the ones reporting it?

brian-brazil:
:+1: 

Yes, that was the main meaning change - claiming something is a cleanup change when it's not is to be avoided. DCO still needs fixing.

RichiH:
If that's what you meant, then yes, the reason I think it's better not to force people to go public is to not force them. This is hardly new process as we would have honored that request before, so it's clarification.

RichiH:
@brian-brazil happy as-is?

brian-brazil:
I already :+1:ed. Just give it a better description when you squash&merge.

RichiH:
Clicked too fast; not worth reverting for, sorry.

PR #1598, opened 2020-04-13
Add link to Connect Box exporter

mbugert:
Hi,

I wrote an exporter for Compal CH7465LG cable modems. These are marketed by several carriers across Europe under different names, most commonly under the name "Connect Box".

I wrote the exporter as a custom collector since one cannot run code on the modem itself. It's highly unlikely that users will have multiple modems running at home. I therefore read the modem IP from an external YAML config file instead of following the service discovery approach used in the IPMI or SNMP exporters. I hope that's fine for you.

I'd love to see this get accepted here, so please let me know if there is anything you would like to see fixed. :blush:


brian-brazil:
Does this work only one one specific model of modem? I'm not sure of the wisdom of listing an exporter for each individual type of network device out there that doesn't support SNMP.

connectbox_uptime_seconds_total is a gauge.

It'd be simpler to fail the scrape rather than have connectbox_scrape_success.

roidelapluie:
> It'd be simpler to fail the scrape rather than have connectbox_scrape_success.

https://github.com/prometheus/haproxy_exporter/blob/master/haproxy_exporter.go#L186

Or call it connectbox_up, like we do in haproxy exporter?

mbugert:
> Does this work with only one one specific model of modem? I'm not sure of the wisdom of listing an exporter for each individual type of network device out there that doesn't support SNMP.

Yes, it supports only one model albeit one which is owned by many cable customers. Whether it makes sense listing such an exporter in the docs is your choice.

> connectbox_uptime_seconds_total is a gauge.

Ok, I'll change that.

> Or call it connectbox_up, like we do in haproxy exporter?

Ok.

jan--f:
Hello from the bug scrub!
The proposed addition seems inactive nowadays, so I'll close this for now. Please re-open this if you still want to add the exporter.

PR #1600, opened 2020-04-15
Add Teamspeak3 exporter to exporter list

hikhvar:
Signed-off-by: Christoph Petrausch <chrobbert@gmail.com>


brian-brazil:
Thanks for sharing, this exporter takes a password on the commandline which is not secure - can you fix that please?

The bandwidth metrics are missing units, and usually max would go at the end of the metric name so that it'll end up beside the metric for the current value (where there is one).
This exporter doesn't appear to be collecting the data synchronously on every scrape, and I'm not sure it'd handle concurrent scrapes correctly.

hikhvar:
Hello,
thank you for the review. I removed the option to set the password via command line. Now the exporter supports only a password file.  
I renamed the bandwidth metrics and added the unit. 

Furthermore I refactored the scraping process. Every scrape fetches the information from the teamspeak server.

PR #1606, opened 2020-04-22
Added druid exporter in the list

iamabhishek-dubey:
Signed-off-by: Abhishek Dubey abhishekbhardwaj510@gmail.com

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -35,6 +35,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Consul exporter](https://github.com/prometheus/consul_exporter) (**official**)
    * [Couchbase exporter](https://github.com/blakelead/couchbase_exporter)
    * [CouchDB exporter](https://github.com/gesellix/couchdb-exporter)
+   * [Druid Exporter](https://github.com/opstree/druid-exporter)
    * [ElasticSearch exporter](https://github.com/justwatchcom/elasticsearch_exporter)
    * [EventStore exporter](https://github.com/marcinbudny/eventstore_exporter)
    * [Memcached exporter](https://github.com/prometheus/memcached_exporter) (**official**)
@@ -54,7 +55,6 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [SQL exporter](https://github.com/free/sql_exporter)
    * [Tarantool metric library](https://github.com/tarantool/prometheus)
    * [Twemproxy](https://github.com/stuartnelson3/twemproxy_exporter)
-   * [Druid Exporter](https://github.com/opstree/druid-exporter)
 
 ### Hardware related
    * [apcupsd exporter](https://github.com/mdlayher/apcupsd_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. We try to avoid duplication in this list, what does this do that the JMX exporter can't?

roidelapluie:
By looking at the README I see druid_segement_* metrics. Is that a typo? segement vs segment.

iamabhishek-dubey:
Hey @roidelapluie 
Thanks for the comment, yes it was a typo and I have fixed :)

iamabhishek-dubey:
Hey @brian-brazil 
Thanks for the comment, yes initially we tried JMX exporter with druid but we were not getting the supervisors, task metrics. That's why we created this

brian-brazil:
https://druid.apache.org/docs/latest/operations/metrics.html seems to have task metrics

iamabhishek-dubey:
Yes but supervisor metrics are not there so shall I modify exporter for supervisors metrics only?

brian-brazil:
It's more understanding if this is a new thing that deserves a listing, or if it's only mildly different. I have a sense that this is more akin to kube-state-metrics, but the docs of both this exporter and druid itself aren't very clear.

iamabhishek-dubey:
Okay @brian-brazil I am closing the PR but thanks for the feedbacks

iamabhishek-dubey:
One last question, can we list this under miscellaneous section like kube-state-metrics or do you think this needs more maturity?

brian-brazil:
It's not a question of maturity, it's a question of if this is doing things that are fundamentally different enough from what is possible with the jmx exporter.

iamabhishek-dubey:
Understood, so I will again compare the metrics between these two and then reopen the PR if needed until then I am closing the PR. Thanks for feedbacks

iamabhishek-dubey:
I have tested the druid with JMX exporter but it only gives the JVM data, we are not actually getting the metrics of the druid. Here are the example metrics:-
https://gist.github.com/iamabhishek-dubey/5ef19d3db9deb25475a80c9ff5c79262

Again re-opening the PR :-)

iamabhishek-dubey:
Also, I have updated the documentation for better clarity.

brian-brazil:
That's still not very clear, and it makes it seem as though druid itself doesn't expose anything via JMX.

The key question for me is is this monitoring druid itself, or the data inside druid? In general an exporter should only be doing one of these things.

iamabhishek-dubey:
Yes, druid doesn't expose its metrics via JMX, and this exporter is to monitor druid and its services not the data inside it.

content/docs/instrumenting/exporters.md
```diff
@@ -54,6 +54,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [SQL exporter](https://github.com/free/sql_exporter)
    * [Tarantool metric library](https://github.com/tarantool/prometheus)
    * [Twemproxy](https://github.com/stuartnelson3/twemproxy_exporter)
+   * [Druid Exporter](https://github.com/opstree/druid-exporter)
```
brian-brazil, line 57: Keep it sorted.

iamabhishek-dubey: Done

iamabhishek-dubey:
I have made the suggested changes

brian-brazil:
Thanks!

PR #1608, opened 2020-04-28
Add bind_query and dhcpd_leases exporters

DRuggeri:
Stir crazy during COVID-19 lockdown? Write some exporters for managing your home network better!

This PR adds a few links to the community list of exporters.


brian-brazil:
Thanks for your PR. We try to avoid duplication in this list, so that the community tends to work together rather than continuously reinventing the wheel.

What does the bind query exporter do that the grok exporter can't?

The dhcp lease exporter is taking a password on the command line, which is insecure and will need to be fixed.
You are also bind mounting a file, which won't work out if a new file is created and then renamed to it. I'd suggest mounting the directory instead to avoid this.
You should use MustNewConstMetric rather than direct instrumentation, currently your implementation is racy.
Would it be practical to use the grok exporter here too? Given what I recall of the semantics of dhcpd.leases it may be doable.

DRuggeri:
Thanks, @brian-brazil for the quick review. I've taken a stab at replies here

What does the bind query exporter do that the grok exporter can't?
 - Aside from being FAR simpler to config and use <g>, bind_query_exporter supports reverse lookups of the client. This is a must-have if clients are DHCP assigned.

The dhcp lease exporter is taking a password on the command line, which is insecure and will need to be fixed.
 - I assume you mean the `web.auth.password`. I'm surprised other exporters don't make use of creds for scraping, but in any event I've changed this on both exporters.

You are also bind mounting a file, which won't work out if a new file is created and then renamed to it. I'd suggest mounting the directory instead to avoid this.
 - Easy enough to change. This is done for both exporters.

You should use MustNewConstMetric rather than direct instrumentation, currently your implementation is racy.
 - This is new to me, so I've given it a swag. I've changed this in the dhcpd exporter - does that fit with what you're describing? I can do the same for bind_query next.

Would it be practical to use the grok exporter here too? Given what I recall of the semantics of dhcpd.leases it may be doable.
 - It wouldn't work in this case since the exporter differentiates active versus expired leases by parsing the balanced data within the brackets. It also accounts for the fact that leases in the file can be "overwritten" (it's a journal) with a new lease (new client getting a previously used IP or a client refreshing its lease).

DRuggeri:
To follow up on the previous response, I'm not sure how an exporter should be modified to not use direct instrumentation when it has a counter (or CounterVec) and the exporter is responsible for collecting the metrics. In those cases, the exporter isn't mirroring metrics from what another system is gathering - the exporter is doing the gathering itself. Looking at several official examples confirms the mirroring pattern. Looking at the grok exporter, for CounterVec cases, it looks like it's also using direct instrumentation.

brian-brazil:
> Aside from being FAR simpler to config and use 

That's not considered sufficient reason, a config file is simpler than duplicating logic into an entire software project.

> supports reverse lookups of the client. This is a must-have if clients are DHCP assigned.

I'm not following you here. I'd expect either that the reverse lookups would always return the same response in which case that's not a great justification, or they do often return different results (I guess due to dynamic DNS updates) in which case that could pose cardinality issues.

>  I'm not sure how an exporter should be modified to not use direct instrumentation when it has a counter (or CounterVec) and the exporter is responsible for collecting the metrics.

You'd be correct, it's very rare that such an approach is correct in an exporter though. Here it is correct for scrapesTotalMetric and scrapeErrorsTotalMetric as they're about the exporter itself, however those metrics likely aren't of much use and it'd be better to fail the scrape when there's an error. The scrape duration is typically the only one you'd have, as combined with scrape_duration_seconds you can measure network latency with it.

The issue is more that your code is racy, which me suggesting adding MustNewConstMetric is usually enough to fix. In particular https://github.com/DRuggeri/dhcpd_leases_exporter/blob/master/collectors/stats_collector.go#L97 could be called concurrently by different scrapes.

DRuggeri:
That's not considered sufficient reason, a config file is simpler than duplicating logic into an entire software project.
 - Of course. It looks like github stripped the < g > tag ( a grin) that followed the cheeky statement.

I'm not following you here. I'd expect either that the reverse lookups would always return the same response in which case that's not a great justification, or they do often return different results (I guess due to dynamic DNS updates) in which case that could pose cardinality issues.
 - Yes, that is exactly the case. In one of the use cases documented in the README, that is the primary goal - figure out which client on a private network managed by some dynamic mechanism made a request. That dynamic mechanism could be something simple like DHCP or something more complicated like a container orchestrator that moves workloads around on a whim. In either case, for  such a network, the IP address has no useful meaning.
 - I understand the cardinality problem - this is documented also in the README and is why the Names collector is not enabled by default. The administrator should do something (either by a whitelist or a blacklist) to limit the number of matches. Sample use case: I want to get rid of the "foo.barcorp.local" DNS record, but my orchestrator moves workloads on a whim. The vector `name="foo.barcorp.local", client="baz.service.barcorp.local"` actually avoids the cardinality problem rather than creating an explosion of `name="foo.barcorp.local", client="1.2.3.4", name="foo.barcorp.local", client="1.2.3.5", name="foo.barcorp.local", client="1.2.3.5" ...`




The issue is more that your code is racy, which me suggesting adding MustNewConstMetric is usually enough to fix. In particular https://github.com/DRuggeri/dhcpd_leases_exporter/blob/master/collectors/stats_collector.go#L97 could be called concurrently by different scrapes.
 - Yes, the initial response would have been more helpful to know what portion was considered racy moreso than pointing at changing how metrics are implemented.
 - To the specific example, I'm not sure there's a race condition. That call just signals [dhcpleasesreader](https://github.com/DRuggeri/dhcpdleasesreader) to check the file and re-read only it if it's changed. That guy (now) has a [built-in mutex](https://github.com/DRuggeri/dhcpdleasesreader/blob/master/dhcpd_info.go#L55), but previously relied on a stat of the file to determine if it should do work, and exit right away if not.

DRuggeri:
Was reviewing various PRs I have open and realized I haven't checked in about this.

Hopefully the summary above helps - any feedback?

brian-brazil:
> In either case, for such a network, the IP address has no useful meaning.

A bit of a weird setup, but I can see it where the DHCP pool is smaller than the number of devices.

> That guy (now) has a built-in mutex,

The writes have a mutex, the reads do not. I recommend not keeping any state, but rather just feeding data to the channel.

DRuggeri:
I'm not sure I follow. The reads don't need a mutex - the DHCP daemon only writes its state occasionally. Therefore, this library only reads the DHCP daemon state when it has changed on the filesystem.

This is done under the principle that a single initial read (and holding the parsed data until it changes) coupled with subsequent `stat` of the file (until it changes) is less expensive than always reading the file X times every Y seconds (where X and Y are the number of times this is scraped from various prometheus daemons).

Further, should the behavior of the daemon change, the library can be updated - but the exporter is decoupled from those details since it just calls the `Read()` function.

brian-brazil:
> The reads don't need a mutex - the DHCP daemon only writes its state occasionally. 

That's not relevant. If the reads aren't also protected by a mutex then that's a data race.

DRuggeri:
That doesn't help at all. Can you *specifically* point to the data race??? The entire `Read()` method is wrapped in a static mutex. Entrance to the function is ungoverned, but all of the internals are forced to be serialized.

There's even [an explicit test case](https://github.com/DRuggeri/dhcpdleasesreader/blob/master/dhcpd_info_test.go#L32) in the library that demonstrates there is no data race for the read operation when executed 10 times in the background simultaneously when being executed 10 times in the foreground.

brian-brazil:
https://github.com/DRuggeri/dhcpd_leases_exporter/blob/master/collectors/lease_collector.go#L83 accesses the data without holding the mutex.

DRuggeri:
I see. Thank you - I thought you were still referring to the library code rather than the collector code.

While I don't see a risk here (we may get newer results?... but that's the idea behind caching and using the latest cache), I've committed [a change](https://github.com/DRuggeri/dhcpd_leases_exporter/commit/2ac57e86b59a408ac4237c563e487d72d8059ca8) to just put the topic to bed.

brian-brazil:
Thanks!

> While I don't see a risk here (we may get newer results?

In the more general case could mix results from different reads, not a risk here but it's best to play these things safe. A mutex is cheap, against trying to figure this out after the exporter has evolved to a point where things like panics may also be a risk.

> I've committed a change to make all interactions with the library serialized to put the topic to bed.

It's the library where the fix should be made, as it does not offer a safe interface to read its data. I'd suggest a getter method that uses the same lock as the update method.

PR #1611, opened 2020-04-29
Add TLS and basic auth policy

roidelapluie:
All of this comes from the dev summit 2018 notes:

https://docs.google.com/document/d/1-C5PycocOZEVIPrmM1hn8fBelShqtqiAmFptoG4yK70/edit#

cc @RichiH @juliusv @brian-brazil @SuperQ

### DIFF
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -20,7 +20,7 @@ release date with you, acknowledging your effort and mentioning you by name
 if you want.
 
 Prometheus is maintained by volunteers, not by a company. Therefore, fixing on
-security issues is done on a best-effort basis. We thrive to fix security issues
+security issues is done on a best-effort basis. We strive to fix security issues
 in less than 7 days.
 
 ## Prometheus
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -168,6 +168,10 @@ environment variable as used by EC2 service discovery) may end up exposed due to
 code outside of our control or due to functionality that happens to expose
 wherever it is stored.
 
+The web security configuration file, with TLS and HTTP Basic Authentication
+settings, can be changed on the fly. It is your responsibility to protect that
+file from unwanted reads/writes.
+
 ## Denial of Service
 
 There are some mitigations in place for excess load or expensive queries.
@@ -209,10 +213,6 @@ Passwords are hashed with the [bcrypt](https://en.wikipedia.org/wiki/Bcrypt)
 algorithm, and it is your responsibility to pick the number of rounds that
 matches your security standards. More rounds make brute-force more complicated
 at the cost of more CPU power and more time to authenticate the requests.
-
-The web security configuration file, with users and path to certificates, can be
-changed on the fly, with the exception of enabling/disabling TLS entirely. It is
-your responsibility to protect that file from unwanted reads/writes.
 
 ## Libraries
 
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -130,7 +130,9 @@ TLS client-side authentication with certificates will also be supported.
 The Go projects will share the same TLS library, which will be based on the
 Go vanilla TLS library. We keep [Go default TLS
 parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
-support only TLSv1.2 and higher.
+support only TLSv1.2 and higher. This decision is based upon current industry
+standards (notes from [Google][tlsgoogle], [Apple][tlsapple],
+[Mozilla][tlsmozilla], and [Microsoft][tlsmicrosoft]).
 
 TLS will be added to Java exporters in the future.
 
@@ -232,3 +234,8 @@ pre-built binaries provided by the project.
 
 For more details, please read the
 [final report of the audit](/assets/downloads/2018-06-11--cure53_security_audit.pdf).
+
+[tlsgoogle]:https://security.googleblog.com/2018/10/modernizing-transport-security.html
+[tlsapple]:https://webkit.org/blog/8462/deprecation-of-legacy-tls-1-0-and-1-1-versions/
+[tlsmozilla]:https://blog.mozilla.org/security/2018/10/15/removing-old-versions-of-tls/
+[tlsmicrosoft]:https://security.googleblog.com/2018/10/modernizing-transport-security.html
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -132,7 +132,7 @@ Go vanilla [crypto/tls](https://golang.org/pkg/crypto/tls) library.
 We default to TLS 1.2 as minimum version. Our policy regarding this is based on
 [Qualys SSL Labs](https://www.ssllabs.com/) recommendations, where we strive to
 achieve a grade 'A' with a default configuration and correctly provided
-certificates, while sticking as closely as possible to the upstream go defaults.
+certificates, while sticking as closely as possible to the upstream Go defaults.
 Achieving that grade provides a balance between perfect security and usability.
 
 TLS will be added to Java exporters in the future.
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -193,10 +193,6 @@ environment variable as used by EC2 service discovery) may end up exposed due to
 code outside of our control or due to functionality that happens to expose
 wherever it is stored.
 
-The web security configuration file, with TLS and HTTP Basic Authentication
-settings, can be changed on the fly. It is your responsibility to protect that
-file from unwanted reads/writes.
-
 ## Denial of Service
 
 There are some mitigations in place for excess load or expensive queries.
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -20,9 +20,9 @@ release date with you, acknowledging your effort and mentioning you by name
 if you want.
 
 Prometheus is maintained by volunteers, not by a company. Therefore, fixing on
-security issues is done on a best-effort basis. We strive to fix security issues
-in less than 7 days for the following repositories: prometheus, alertmanager,
-node_exporter and blackbox_exporter.
+security issues is done on a best-effort basis. We strive to release security
+fixes within 7 days for: prometheus, alertmanager, node_exporter,
+blackbox_exporter and pushgateway.
 
 ## Prometheus
 
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -203,7 +203,7 @@ If you have special TLS needs, like a different cipher suite or older TLS
 version, the TLS support on the projects should enable you to build a secure
 tunnel between the servers and reverse proxies with special settings.
 
-Server-side Basic Authentication is also supported. In such a case, usernames
+HTTP Basic Authentication is also supported. In such a case, usernames
 and passwords are provided in the same configuration file as the TLS
 configuration. Basic Authentication can be used without TLS, but it will then
 expose usernames and passwords over the network.
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -133,6 +133,7 @@ We default to TLS 1.2 as minimum version. Our policy regarding this is based on
 [Qualys SSL Labs](https://www.ssllabs.com/) recommendations, where we strive to
 achieve a grade 'A' with a default configuration and correctly provided
 certificates, while sticking as closely as possible to the upstream go defaults.
+Achieving that grade provides a balance between perfect security and usability.
 
 TLS will be added to Java exporters in the future.
 
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -135,8 +135,8 @@ support only TLSv1.2 and higher.
 TLS will be added to Java exporters in the future.
 
 If you have special TLS needs, like a different cipher suite or older TLS
-version, the TLS support on the projects should enable you to build a secure
-tunnel between the servers and reverse proxies with special settings.
+version, the TLS support enables you to build a secure tunnel between the
+servers and reverse proxies with special settings.
 
 HTTP Basic Authentication will also be supported. In such a case, usernames
 and passwords are provided in the same configuration file as the TLS
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -140,15 +140,15 @@ If you have special TLS needs, like a different cipher suite or older TLS
 version, the TLS support enables you to build a secure tunnel between the
 servers and reverse proxies with special settings.
 
-HTTP Basic Authentication will also be supported. In such a case, usernames
-and passwords are provided in the same configuration file as the TLS
-configuration. Basic Authentication can be used without TLS, but it will then
-expose usernames and passwords over the network.
+HTTP Basic Authentication will also be supported. Basic Authentication can be
+used without TLS, but it will then expose usernames and passwords in cleartext
+over the network.
 
-Passwords are hashed with the [bcrypt](https://en.wikipedia.org/wiki/Bcrypt)
-algorithm, and it is your responsibility to pick the number of rounds that
-matches your security standards. More rounds make brute-force more complicated
-at the cost of more CPU power and more time to authenticate the requests.
+On the server side, basic authentication passwords are hashed on disk with the
+[bcrypt](https://en.wikipedia.org/wiki/Bcrypt) algorithm on disks, and it is
+your responsibility to pick the number of rounds that matches your security
+standards. More rounds make brute-force more complicated at the cost of more CPU
+power and more time to authenticate the requests.
 
 Various Prometheus components support client-side authentication and
 encryption. If TLS client support is offered, there is often also an option
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -21,7 +21,8 @@ if you want.
 
 Prometheus is maintained by volunteers, not by a company. Therefore, fixing on
 security issues is done on a best-effort basis. We strive to fix security issues
-in less than 7 days.
+in less than 7 days for the following repositories: prometheus, alertmanager,
+node_exporter and blackbox_exporter.
 
 ## Prometheus
 
@@ -185,19 +186,29 @@ automatically restart on failure.
 
 In the future, server-side TLS support will be rolled out to the different
 projects under the Prometheus organization. Those projects include Prometheus,
-Alertmanager, Pushgateway and the official exporters. The projects will share
-the same TLS library, which will be based on the Golang vanilla TLS library. We
-keep [Golang's default TLS
+Alertmanager, Pushgateway and the official exporters.
+
+The golang projects will share the same TLS library, which will be based on the
+Golang vanilla TLS library. We keep [Golang's default TLS
 parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
 explicitly disable TLSv1.0 and TLSv1.1.
 
-There is no plans at the moment to extend those settings. If you need older TLS
-versions or different cipher suite, we recommend you to setup a reverse proxy in
-front of your services, which will handle those settings.
+Research will be made later on the non-golang exporters to see how to best
+address server-side TLS, but you can expect a policy close to this one.
 
-Changes regarding those matters need to reach consensus within the
-Prometheus-team as they impact all the projects. Pull requests made without
-prior consensus from -team will be automatically rejected.
+The TLS support on the projects should enable you to build a secure tunnel
+between the servers and your reverse proxies if you have more specialized TLS
+needs, like a cipher suite or older TLS version.
+
+Server-side Basic Authentication is also supported. In such a case, usernames
+and passwords are provided in the same configuration file as the TLS
+configuration. Basic Authentication can be used without TLS, but it will expose
+usernames and passwords over the network.
+
+Passwords are hashed with the [bcrypt](https://en.wikipedia.org/wiki/Bcrypt)
+algorithm, and it is your responsibility to pick the number of rounds that
+matches your security standards. More rounds make brute-force more complicated
+at the cost of more CPU power and more time to authenticate the requests.
 
 ## Libraries
 
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -19,10 +19,10 @@ prometheus-team@googlegroups.com. We will fix the issue and coordinate a
 release date with you, acknowledging your effort and mentioning you by name
 if you want.
 
-Prometheus is maintained by volunteers, not by a company. Therefore, fixing on
+Prometheus is maintained by volunteers, not by a company. Therefore, fixing
 security issues is done on a best-effort basis. We strive to release security
-fixes within 7 days for: prometheus, alertmanager, node_exporter,
-blackbox_exporter, and pushgateway.
+fixes within 7 days for: Prometheus, Alertmanager, Node_exporter,
+Blackbox_exporter, and Pushgateway.
 
 ## Prometheus
 
@@ -192,6 +192,9 @@ Secrets from other sources used by dependencies (e.g. the `AWS_SECRET_KEY`
 environment variable as used by EC2 service discovery) may end up exposed due to
 code outside of our control or due to functionality that happens to expose
 wherever it is stored.
+
+Special care must be taken to prevent unauthorized reads or writes to
+configuration files that contain secrets.
 
 ## Denial of Service
 
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -128,19 +128,20 @@ Pushgateway and the official exporters.
 Authentication of clients by TLS client certs will also be supported.
 
 The Go projects will share the same TLS library, which will be based on the
-Go vanilla TLS library. We keep [Go default TLS
-parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
-support only TLSv1.2 and higher. This decision is based upon current industry
-standards (notes from [Google][tlsgoogle], [Apple][tlsapple],
-[Mozilla][tlsmozilla], and [Microsoft][tlsmicrosoft]). We strive to keep TLS
-compatibility with official [client libraries](../instrumenting/clientlibs)
-runtimes that were released less than two years ago.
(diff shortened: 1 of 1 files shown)


content/docs/operating/security.md
```diff
@@ -19,6 +19,10 @@ prometheus-team@googlegroups.com. We will fix the issue and coordinate a
 release date with you, acknowledging your effort and mentioning you by name
 if you want.
 
+Prometheus is maintained by volunteers, not by a company. Therefore, fixing on
+security issues is done on a best-effort basis. We thrive to fix security issues
```
gouthamve, line 23: s/thrive/strive

content/docs/operating/security.md
```diff
@@ -177,6 +181,24 @@ and bandwidth.
...
+
+In the future, server-side TLS support will be rolled out to the different
+projects under the Prometheus organization. Those projects include Prometheus,
+Alertmanager, Pushgateway and the official exporters. The projects will share
+the same TLS library, which will be based on the Golang vanilla TLS library. We
+keep [Golang's default TLS
+parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
+explicitly disable TLSv1.0 and TLSv1.1.
+
+There is no plans at the moment to extend those settings. If you need older TLS
```
brian-brazil, line 194: The first sentence is a bit wishy washy, I'd remove it.

content/docs/operating/security.md
```diff
@@ -177,6 +181,24 @@ and bandwidth.
...
+versions or different cipher suite, we recommend you to setup a reverse proxy in
```
brian-brazil, line 195: a different

content/docs/operating/security.md
```diff
@@ -177,6 +181,24 @@ and bandwidth.
 It is recommended to monitor all components for failure, and to have them
 automatically restart on failure.
 
+## Server-side TLS and authentication
```
brian-brazil, line 184: This doesn't mention about basic-auth, and we should probably put in a warning about using that over the clear at least.

content/docs/operating/security.md
```diff
@@ -177,6 +181,24 @@ and bandwidth.
...
+
+In the future, server-side TLS support will be rolled out to the different
+projects under the Prometheus organization. Those projects include Prometheus,
+Alertmanager, Pushgateway and the official exporters. The projects will share
+the same TLS library, which will be based on the Golang vanilla TLS library. We
+keep [Golang's default TLS
+parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
+explicitly disable TLSv1.0 and TLSv1.1.
```
brancz, line 192: I feel this is worth re-discussing. At the time we didn't have TLSv1.3, I think we should default to that. Bonus points we get from that is that there are no configurable cipher suites with 1.3 yet anyways.

brian-brazil: That'd in addition to already breaking client_java and client_python compared to Go defaults, would also break client_golang (and possibly also client_ruby), plus I imagine a non-trivial amount of users.

content/docs/operating/security.md
```diff
@@ -19,6 +19,10 @@ prometheus-team@googlegroups.com. We will fix the issue and coordinate a
 release date with you, acknowledging your effort and mentioning you by name
 if you want.
 
+Prometheus is maintained by volunteers, not by a company. Therefore, fixing on
+security issues is done on a best-effort basis. We thrive to fix security issues
+in less than 7 days.
```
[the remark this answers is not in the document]
SuperQ: How about:

```
We strive to release security fixes within 7 days for (list of projects).
```

content/docs/operating/security.md
SuperQ: I would word this as "we support TLSv1.2 and higher".

content/docs/operating/security.md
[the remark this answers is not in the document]
brian-brazil: That's a question for @beorn7 really as the maintainer.

The reason that blackbox is in there is that Ben asked for it in that dev summit, and I said sure.

content/docs/operating/security.md
roidelapluie: There is a difference between default to that and that only. I think that TLS1.3 is not spread enough to disable 1.2 entirely. Imagine that you run RHEL and you can't use curl to debug.

content/docs/operating/security.md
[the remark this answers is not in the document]
beorn7: I think the wording is sufficiently vague that I'd be happy to do this kind of commitment for any repository I'm maintaining. (I mean, what would it mean to not do that? "I won't even _try_ to fix a security issue in a week time.")

content/docs/operating/security.md
roidelapluie: Added basic auth

content/docs/operating/security.md
roidelapluie: Rephrased that.

content/docs/operating/security.md
roidelapluie: Rephrased that.

content/docs/operating/security.md
```diff
@@ -177,6 +181,24 @@ and bandwidth.
...
+the same TLS library, which will be based on the Golang vanilla TLS library. We
+keep [Golang's default TLS
+parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
+explicitly disable TLSv1.0 and TLSv1.1.
+
+There is no plans at the moment to extend those settings. If you need older TLS
+versions or different cipher suite, we recommend you to setup a reverse proxy in
+front of your services, which will handle those settings.
+
+Changes regarding those matters need to reach consensus within the
```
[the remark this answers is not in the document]
roidelapluie: I have removed it. It was to address your concerns about the PR's in common.

content/docs/operating/security.md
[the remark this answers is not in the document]
roidelapluie: Rephrased it.

content/docs/operating/security.md
```diff
@@ -177,6 +181,24 @@ and bandwidth.
 It is recommended to monitor all components for failure, and to have them
 automatically restart on failure.
 
+## Server-side TLS and authentication
+
+In the future, server-side TLS support will be rolled out to the different
+projects under the Prometheus organization. Those projects include Prometheus,
+Alertmanager, Pushgateway and the official exporters. The projects will share
+the same TLS library, which will be based on the Golang vanilla TLS library. We
```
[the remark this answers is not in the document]
roidelapluie: I have now tried to address non-go exporters

content/docs/operating/security.md
brancz: "TLSv1.2 and higher" I'm happy with as well. Needs to be configurable then though.

roidelapluie: rephrased to use 1.2 and higher. The configuration bit I have not added as I see no consensus.

I think that we can discuss the configuration part at the next dev summit to reach a consensus.

brian-brazil: If you run RHEL when you need TLS 1.0 :)

content/docs/operating/security.md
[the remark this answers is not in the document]
roidelapluie: I take that as a yes, the reason behind explicitly covering it is that it is one that is exposed to more than just  the prometheus server.

We can find a consensus to cover more or all our repos in the coming dev summit.

roidelapluie:
Added a note about security of the config file, too

content/docs/operating/security.md
```diff
@@ -177,6 +182,38 @@ and bandwidth.
...
+
+The golang projects will share the same TLS library, which will be based on the
+Golang vanilla TLS library. We keep [Golang's default TLS
+parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
+support only TLSv1.2 and higher.
+
+Research will be made later on the non-golang exporters to see how to best
+address server-side TLS, but you can expect a policy close to this one.
+
+The TLS support on the projects should enable you to build a secure tunnel
```
brian-brazil, line 199: I'd predicate this on the special needs first, as currently written it makes it sounds like the primary purpose of this is the tunnel rather than this being primarily intended to be used standalone.

content/docs/operating/security.md
```diff
@@ -177,6 +182,38 @@ and bandwidth.
...
+between the servers and your reverse proxies if you have more specialized TLS
+needs, like a cipher suite or older TLS version.
+
+Server-side Basic Authentication is also supported. In such a case, usernames
```
brian-brazil, line 203: HTTP Basic Authentication

content/docs/operating/security.md
```diff
@@ -177,6 +182,38 @@ and bandwidth.
...
+In the future, server-side TLS support will be rolled out to the different
+projects under the Prometheus organization. Those projects include Prometheus,
+Alertmanager, Pushgateway and the official exporters.
+
+The golang projects will share the same TLS library, which will be based on the
+Golang vanilla TLS library. We keep [Golang's default TLS
+parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
+support only TLSv1.2 and higher.
+
+Research will be made later on the non-golang exporters to see how to best
```
brian-brazil, line 196: Java may end up being weird (e.g. how do you only support 1.2+ and the various client auth options in a JVM that doesn't support it? Will it even compile?), so I'd phrase it more as "TLS will be added to Java exporters in the future".

We've no Python or Ruby exporters, so we can limit this to Java.

content/docs/operating/security.md
```diff
@@ -177,6 +182,38 @@ and bandwidth.
...
+expose usernames and passwords over the network.
+
+Passwords are hashed with the [bcrypt](https://en.wikipedia.org/wiki/Bcrypt)
+algorithm, and it is your responsibility to pick the number of rounds that
+matches your security standards. More rounds make brute-force more complicated
+at the cost of more CPU power and more time to authenticate the requests.
+
+The web security configuration file, with users and path to certificates, can be
+changed on the fly, with the exception of enabling/disabling TLS entirely. It is
+your responsibility to protect that file from unwanted reads/writes.
```
brian-brazil, line 215: Hmm, this final bit would be better covered up in the Secrets section I think.

content/docs/operating/security.md
brian-brazil: Yeah, configuration is something that is to be avoided - that's what a reverse proxy is for as it stands.

This doc still doesn't explain why we stick with defaults, or why we make an exception for 1.2. Defaults was for general sanity/maintainability, but a general basis for cherry-picking versions (and why only versions, not ciphers or the other 10+ settings) needs to be elucidated.

content/docs/operating/security.md
roidelapluie: moved

content/docs/operating/security.md
roidelapluie: Rephrased

content/docs/operating/security.md
roidelapluie: fixed

content/docs/operating/security.md
roidelapluie: addressed

content/docs/operating/security.md
```diff
@@ -177,6 +182,38 @@ and bandwidth.
...
 automatically restart on failure.
 
+## Server-side TLS and authentication
+
+In the future, server-side TLS support will be rolled out to the different
+projects under the Prometheus organization. Those projects include Prometheus,
+Alertmanager, Pushgateway and the official exporters.
+
+The golang projects will share the same TLS library, which will be based on the
+Golang vanilla TLS library. We keep [Golang's default TLS
```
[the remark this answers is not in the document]
roidelapluie: renamed

content/docs/operating/security.md
```diff
@@ -177,6 +182,38 @@ and bandwidth.
 It is recommended to monitor all components for failure, and to have them
 automatically restart on failure.
 
+## Server-side TLS and authentication
```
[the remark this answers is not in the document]
roidelapluie: changed line 119

content/docs/operating/security.md
```diff
@@ -19,6 +19,11 @@ prometheus-team@googlegroups.com. We will fix the issue and coordinate a
 release date with you, acknowledging your effort and mentioning you by name
 if you want.
 
+Prometheus is maintained by volunteers, not by a company. Therefore, fixing on
+security issues is done on a best-effort basis. We strive to release security
+fixes within 7 days for: prometheus, alertmanager, node_exporter,
+blackbox_exporter and pushgateway.
```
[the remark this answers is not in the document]
roidelapluie: Added comma before and

content/docs/operating/security.md
```diff
@@ -177,6 +186,33 @@ and bandwidth.
...
+support only TLSv1.2 and higher.
+
+TLS will be added to Java exporters in the future.
+
+If you have special TLS needs, like a different cipher suite or older TLS
+version, the TLS support on the projects should enable you to build a secure
+tunnel between the servers and reverse proxies with special settings.
+
+HTTP Basic Authentication is also supported. In such a case, usernames
+and passwords are provided in the same configuration file as the TLS
```
brian-brazil, line 207: Hmm, this is the first mention of a configuration file which makes it seem as though we forgot that - though I don't think we need this level of detail in this doc (that's more for the usage docs).

content/docs/operating/security.md
```diff
@@ -177,6 +186,33 @@ and bandwidth.
...
+
+The Go projects will share the same TLS library, which will be based on the
+Go vanilla TLS library. We keep [Go default TLS
+parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
+support only TLSv1.2 and higher.
+
+TLS will be added to Java exporters in the future.
+
+If you have special TLS needs, like a different cipher suite or older TLS
+version, the TLS support on the projects should enable you to build a secure
```
brian-brazil, line 203: This reads oddly. Maybe:

the TLS support enables you to

roidelapluie: done

content/docs/operating/security.md
roidelapluie: that config file has security implications -- if people give access to anyone it is possible to change certs, disable client side auth, add users..

content/docs/operating/security.md
```diff
@@ -163,6 +168,10 @@ environment variable as used by EC2 service discovery) may end up exposed due to
 code outside of our control or due to functionality that happens to expose
 wherever it is stored.
 
+The web security configuration file, with TLS and HTTP Basic Authentication
```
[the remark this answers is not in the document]
roidelapluie: I have removed that

content/docs/operating/security.md
```diff
@@ -19,6 +19,11 @@ prometheus-team@googlegroups.com. We will fix the issue and coordinate a
 release date with you, acknowledging your effort and mentioning you by name
 if you want.
 
+Prometheus is maintained by volunteers, not by a company. Therefore, fixing on
```
csmarchbanks, line 22: `fixing on security issues` should perhaps just be `fixing security issues`.

content/docs/operating/security.md
```diff
@@ -19,6 +19,11 @@ prometheus-team@googlegroups.com. We will fix the issue and coordinate a
...
+security issues is done on a best-effort basis. We strive to release security
+fixes within 7 days for: prometheus, alertmanager, node_exporter,
```
csmarchbanks, line 24: These project names are capitalized in the `Authentication, Authorization, and Encryption` section, consistency would be nice.

roidelapluie: done

content/docs/operating/security.md
roidelapluie: fixed

content/docs/operating/security.md
brian-brazil: What goes in what config files is too detailed to go here I think, plus other config files also have security implications depending on your setup.

content/docs/operating/security.md
```diff
@@ -163,6 +193,9 @@ environment variable as used by EC2 service discovery) may end up exposed due to
 code outside of our control or due to functionality that happens to expose
 wherever it is stored.
 
+Special care must be taken to prevent unauthorized reads or writes to
+configuration files that contain secrets.
```
brian-brazil, line 197: That's not really Prometheus specific, nor secret specific. I'd add something two paragraphs up that protecting the files on disk is the user's responsibility.

content/docs/operating/security.md
```diff
@@ -116,9 +121,38 @@ resulting from additional load and failed scrapes.
...
+Passwords are hashed with the [bcrypt](https://en.wikipedia.org/wiki/Bcrypt)
+algorithm, and it is your responsibility to pick the number of rounds that
+matches your security standards. More rounds make brute-force more complicated
+at the cost of more CPU power and more time to authenticate the requests.
+
+Various Prometheus components support client-side authentication and
+encryption. If TLS client support is offered, there is often also an option
+called `insecure_skip_verify` which skips SSL verification.
+
+# API calls
```
brian-brazil, line 155: "API security" maybe?

content/docs/operating/security.md
```diff
@@ -116,9 +121,38 @@ resulting from additional load and failed scrapes.
...
 ## Authentication, Authorization, and Encryption
 
-Prometheus and its components do not provide any server-side
-authentication, authorization or encryption. If you require this, it is
-recommended to use a reverse proxy.
+In the future, server-side TLS support will be rolled out to the different
+Prometheus projects. Those projects include Prometheus, Alertmanager,
+Pushgateway and the official exporters.
+
+TLS client-side authentication with certificates will also be supported.
```
brian-brazil, line 128: This seems misleading, we already support that for clients. "Authentication of clients by TLS client certs" would be clearer.

content/docs/operating/security.md
```diff
@@ -116,9 +121,38 @@ resulting from additional load and failed scrapes.
...
+
+The Go projects will share the same TLS library, which will be based on the
+Go vanilla TLS library. We keep [Go default TLS
+parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
+support only TLSv1.2 and higher.
```
brian-brazil, line 133: This still needs a justification of some form, this appears completely arbitrary currently which will make it more challenging to deal with requests to change it good/bad reasons.

roidelapluie: I have added links.

content/docs/operating/security.md
roidelapluie: fixed wording

content/docs/operating/security.md
roidelapluie: good idea

content/docs/operating/security.md
roidelapluie: ack

content/docs/operating/security.md
brian-brazil: That's a good start, we need something to cover why we're not changing anything else - otherwise we could have fun when any of those 4 organisations propose e.g. going to TLS 1.3 only or disabling specific ciphers. For example the current standing policy is to stick to the Go defaults, so that the decision is completely out of our hands and we thus can keep it simple and not worry about any of this.

What I'm looking for here (and have been asking for in relation to TLS&friends for many years) is essentially a plan that doesn't boil down to making it up as we go along. There's various hurdles we will inevitably have to deal with as things naturally evolve over time, and I don't want us to have a half-baked approach to TLS that considers the feature done and dusted merely because some code has been written. The code is the easy bit, and a justification that only covers why we have these particular settings right now is not enough. Users requesting changes to the hard coding, vulnerabilities, increases in computing power, geopolitics, changes in dependencies, and new standards are all going to happen.
The policy needs to be able to help answer questions like when we would disable 1.2, reenable 1.0 (and yes, it has happened in the past that an older version was suddenly more secure than a newer version), switch to/from whatever the cipher flavour of the month is, and what happens when Go changes their defaults. So things to be considered might include how much breakage of existing users is okay, is it okay to create a situation where the latest releases of official Prometheus components don't work with each other, how far back are we willing to have such breakage, how perfect should our ssllabs rating be, how good the actual security is for typical users, how often we need to reevaluate/update this configuration, how much effort we're realistically able/willing to put into this, and how we try to balance all these concerns.

Now all of that doesn't belong in this document as this document is more a what than why, but if you want to have a consensus to change our policy to one that allows for TLS 1.0/1.1 to be disabled then this needs to exist in some form.

content/docs/operating/security.md
```diff
@@ -116,9 +121,40 @@ resulting from additional load and failed scrapes.
...
+If you have special TLS needs, like a different cipher suite or older TLS
+version, the TLS support enables you to build a secure tunnel between the
+servers and reverse proxies with special settings.
+
+HTTP Basic Authentication will also be supported. Basic Authentication can be
+used without TLS, but it will then expose usernames and passwords in cleartext
+over the network.
+
+On the server side, basic authentication passwords are hashed on disk with the
+[bcrypt](https://en.wikipedia.org/wiki/Bcrypt) algorithm on disks, and it is
```
simonpasquier, line 148: "are hashed on disk with ... algorithm~~ on disks~~,"

SuperQ, line 148:
```suggestion
On the server side, basic authentication passwords are stored as hashes with the
[bcrypt](https://en.wikipedia.org/wiki/Bcrypt) algorithm. It is
```

roidelapluie:
Added:

"We strive to keep TLS
compatibility with official [client libraries](../instrumenting/clientlibs)
runtimes that were released less than two years ago."

content/docs/operating/security.md
roidelapluie: thx!

roidelapluie: Thx for noticing.

content/docs/operating/security.md
roidelapluie: "We strive to keep TLS
compatibility with official client libraries
runtimes that were released less than two years ago."

content/docs/operating/security.md
```diff
@@ -116,9 +121,42 @@ resulting from additional load and failed scrapes.
...
+
+Authentication of clients by TLS client certs will also be supported.
+
+The Go projects will share the same TLS library, which will be based on the
+Go vanilla TLS library. We keep [Go default TLS
+parameters](https://golang.org/pkg/crypto/tls/#Config), with one exception: we
+support only TLSv1.2 and higher. This decision is based upon current industry
+standards (notes from [Google][tlsgoogle], [Apple][tlsapple],
+[Mozilla][tlsmozilla], and [Microsoft][tlsmicrosoft]). We strive to keep TLS
+compatibility with official [client libraries](../instrumenting/clientlibs)
```
brian-brazil, line 136: client library runtimes

This still isn't covering all the other settings

content/docs/operating/security.md
```diff
@@ -156,7 +190,9 @@ secret. Throughout the Prometheus system, metrics are not considered secret.
 Fields containing secrets in configuration files (marked explicitly as such in
 the documentation) will not be exposed in logs or via the HTTP API. Secrets
 should not be placed in other configuration fields, as it is common for
-components to expose their configuration over their HTTP endpoint.
+components to expose their configuration over their HTTP endpoint. It is the
+responsibility of the user to protect files on disks from unwanted reads and
```
brian-brazil, line 194: on disk

roidelapluie:
To lift some of the concerns about what we can and can't accept as contributions from users, and I think that whatever we write here we will get those questions, I proposed the following pull request in the node_export https package, covering a wider range of TLS settings that can be set:
https://github.com/prometheus/node_exporter/pull/1695

That way we enable users to pick their TLS ciphers.

brian-brazil:
I think that level of change would require a dev summit discussion, as any configurability was seen as off the table at the last one - particularly as that PR has made some interesting choices in what it does and doesn't allow to be configured.

SuperQ:
@brian-brazil We do not require dev summits for making decisions. A dev summit is simply a way to get the same lazy consensus and/or vote done in real time.

brian-brazil:
You're correct, however given all the back and forth on this issue a dev summit would seem the most expedient way of moving from the standing plan of pure defaults to anything else.

content/docs/operating/security.md
brian-brazil: Doing some research Go was the last to release TLS 1.3 support in February 2019, so 2 years means we would drop 1.2 support in February 2021.

content/docs/operating/security.md
```diff
@@ -19,6 +19,11 @@ prometheus-team@googlegroups.com. We will fix the issue and coordinate a
 release date with you, acknowledging your effort and mentioning you by name
 if you want.
 
+Prometheus is maintained by volunteers, not by a company. Therefore, fixing
+security issues is done on a best-effort basis. We strive to release security
+fixes within 7 days for: Prometheus, Alertmanager, Node_exporter,
```
juliusv, line 24: I'd either say "node_exporter" or "Node Exporter" (probably the latter in a normal English sentence), but mixing snake_case with CamelCase seems odd (same for Blackbox Exporter).

content/docs/operating/security.md
```diff
@@ -116,9 +121,44 @@ resulting from additional load and failed scrapes.
...
+[Qualys SSL Labs](https://www.ssllabs.com/) recommendations, where we strive to
+achieve a grade 'A' with a default configuration and correctly provided
+certificates, while sticking as closely as possible to the upstream Go defaults.
+Achieving that grade provides a balance between perfect security and usability.
+
+TLS will be added to Java exporters in the future.
+
+If you have special TLS needs, like a different cipher suite or older TLS
+version, you can tune the minimum TLS version and the ciphers, as long as the
+cipher are not [marked as insecure](https://golang.org/pkg/crypto/tls/#InsecureCipherSuites)
```
juliusv, line 142:
```suggestion
cipher is not [marked as insecure](https://golang.org/pkg/crypto/tls/#InsecureCipherSuites)
```

juliusv:
:+1: besides editorial nits

content/docs/operating/security.md
roidelapluie: ack

content/docs/operating/security.md
roidelapluie: ack

roidelapluie:
Thanks all!

PR #1612, opened 2020-04-30
Fixed broken link

danitfk:
There is one broken link in the writing exporter page.


brian-brazil:
The current link works fine for me on prometheus.io

danitfk:
> The current link works fine for me on prometheus.io

My bad, I was checking on Github itself :)

PR #1614, opened 2020-05-04
Add prometheus_client package as an unofficial Dart client

Fox32:
Signed-off-by: Oliver Sand <oliver.sand@tentaclelabs.com>


brian-brazil:
Thanks!

This looks like a port of the Java client. The only thing I noticed as being off is that the Content-Type header is never sent in either the examples or Shelf integration.

Fox32:
Thanks for the feedback. Yes, it's very similar to the Java client, where possible.

The content-type header is set here in the shelf integration: https://github.com/tentaclelabs/prometheus_client/blob/master/prometheus_client_shelf/lib/shelf_handler.dart#L20

brian-brazil:
So it is, it's still missing in the two examples. That'll break at some future point when openmetrics is presumed default format.

Fox32:
Thanks for pointing that out. The example that isn't using shelf is indeed missing it. I will change that.

PR #1615, opened 2020-05-04
Add libvirtd_exporter to list of exporters

biox:


brian-brazil:
Thanks for your PR. Why should we list this one, and not https://github.com/zhangjianweibj/prometheus-libvirt-exporter which appears to be both older and has an established community?

biox:
> Thanks for your PR. Why should we list this one, and not https://github.com/zhangjianweibj/prometheus-libvirt-exporter which appears to be both older and has an established community?

I was under the impression that both could be listed, since they'll likely be diverging.

Our use-cases are different, as zhangjianwebj's has an Openstack Nova component, and ours will contain some code specific to our own VPS platform.

If the intention of the page of exporters is to only list generic exporters and not tailored ones, feel free to decline this PR, as ours will likely be tailored & specific.

brian-brazil:
I aim to avoid duplication, and encourage the community to build around one great exporter rather than reinventing the wheel. Thus two libvirt exporters are unlikely to be listed, and a more generic exporter is preferred to a specialised one.

biox:
Sounds good! I'll close this up.

PR #1616, opened 2020-05-05
Thanos supports remote read now.

bwplotka:
Thanks to: https://github.com/thanos-io/thanos/pull/2566

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>


roidelapluie:
Actually this is a 3rd party component, maybe we should link to it directly?

bwplotka:
We link to it in official docs.

brian-brazil:
This is where we list these.

PR #1618, opened 2020-05-06
Addition of Commercial Support Provider

Rucknar:
Adding in Infinity Works, we do a lot with Prometheus for our customers and have a very strong understanding of how Prometheus can be used best to provide value. Happy to discuss offline if there is an required process to follow here or if there are any questions.

### DIFF
--- a/content/community.html
+++ b/content/community.html
@@ -105,6 +105,7 @@ title: Community
       <li><a href="https://container-solutions.com">Container Solutions</a></li>
       <li><a href="https://credativ.com">credativ</a></li>
       <li><a href="https://grafana.com">Grafana Labs</a></li>
+      <li><a href="https://www.infinityworks.com">Infinity Works</a></li>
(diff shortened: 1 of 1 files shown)


Rucknar:
^ Force pushes relate to updating sign-off to pass DCO checks.

brian-brazil:
Do you have a more specific Prometheus services-related page? It would make it easier for potential customers to understand what you can do for them.

Rucknar:
@brian-brazil we don’t as of yet, we’ve just relaunched our site and are in the process of adding in some of the service propositions now. We’re not totally sure we’ll drill down to such a level for Prometheus as a stand-alone service on our offerings page as there’s a good risk there it becomes over-crowded. I think we might instead look to shape an insights piece and use that to steer incoming visitors.

brian-brazil:
This list has moved, it's over on the support/training page now.

grobie:
@Rucknar If you want to get listed please update your PR as the list of providers has moved.

PR #1620, opened 2020-05-07
Add nftables exporter to the monitoring list

ynsta:
Hello,

I propose to add this exporter to the monitoring list.

It uses only pure go (no external lib or binary) to create metrics with the number of netfilter rules loaded, labeled with the domain, table and chain.

It is useful to trigger alerts on firewall changes.

Signed-off-by: Stany MARCEL <stanypub@gmail.com>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -194,6 +194,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Locust Exporter](https://github.com/ContainerSolutions/locust_exporter)
    * [Meteor JS web framework exporter](https://atmospherejs.com/sevki/prometheus-exporter)
    * [Minecraft exporter module](https://github.com/Baughn/PrometheusIntegration)
+   * [nftables exporter](https://github.com/Intrinsec/nftables_exporter)
    * [OpenStack exporter](https://github.com/openstack-exporter/openstack-exporter)
    * [OpenStack blackbox exporter](https://github.com/infraly/openstack_client_exporter)
    * [oVirt exporter](https://github.com/czerwonk/ovirt_exporter)


content/docs/instrumenting/exporters.md
```diff
@@ -158,6 +158,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Munin exporter](https://github.com/pvdh/munin_exporter)
    * [Nagios / Naemon exporter](https://github.com/Griesbacher/Iapetos)
    * [New Relic exporter](https://github.com/mrf/newrelic_exporter)
+   * [nftables exporter](https://github.com/Intrinsec/nftables_exporter)
```
brian-brazil, line 161: Firewalls are not monitoring systems.

ynsta: > This looks like there might be some overlap with the node exporter, but it seems that it also requires privs - @SuperQ?

Hello, yes I started this as an collector in node_exporter for iptables and was rejected as it required some additional capabilities.

> It'd be better to fail the scrape if there's a problem, rather than giving the user an extra metric to have to check.

I made it return nothing when it fail like for example for missing capabilities, after the initial review back then is it not the best practice ?

> Firewalls are not monitoring systems.
Yes my  bad, should it be put in Miscellaneous ?

brian-brazil:
> I made it return nothing when it fail like for example for missing capabilities, after the initial review back then is it not the best practice ?

You've no useful metrics that you can return on failure, so it's best to send back a 500 and fail the whole scrape.

> Yes my bad, should it be put in Miscellaneous ?

Yes.

ynsta:
Ok I'll make the changes and update this PR.

ynsta:
I moved it to Miscellaneous list.

brian-brazil:
Thanks!

PR #1621, opened 2020-05-07
Add Honeypot exporter to the misc list

ynsta:
Hello,

I propose to add this exporter to the misc list.

It is a small honeypot that exports the total number of connections (or packets for udp) in a configured list of proto/port. The metric is labeled with proto, src, dst, and port.

We use it to combined with a firewall to detect anomalies on our scan activities.

Signed-off-by: Stany MARCEL <stanypub@gmail.com>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -183,6 +183,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [eBPF exporter](https://github.com/cloudflare/ebpf_exporter)
    * [Ethereum Client exporter](https://github.com/31z4/ethereum-prometheus-exporter)
    * [JFrog Artifactory Exporter](https://github.com/peimanja/artifactory_exporter)
+   * [Honeypot exporter](https://github.com/Intrinsec/honeypot_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
This sounds like it's going to have cardinality issues given IP addresses are one of the labels, I'm not sure it makes sense to list that given this.

ynsta:
The monitoring done with this exporter is to verify connection frequency of an host on a given port or verify that some "blacklisted" hosts do not connect to the honey pot.

It is still in early stage and don't know yet if this logic will be performed in the alerting stage or if I provide the rules in the exporter configuration to expose in other metrics

brian-brazil:
That sounds a bit niche, and that is only safe to use in very limited circumstances. Thus I don't think it belongs on this list. Also, for general connectivity testing there's already the blackbox exporter.

ynsta:
Yes we already use the blackbox exporter to monitor exposed services but it does not solve this issue that is kind of the reversed one.

I you think it is to specific, thank you anyway for you comment on the potential cardinality issue.

ynsta:
I close this PR for now and maybe re-propose it later when the cardinality issue will be solved.

PR #1622, opened 2020-05-08
Add ITM exporter to the list

rafal-szypulka:
@brian-brazil 
I would like to add an exporter that integrates a couple of IBM monitoring products with Prometheus:
- IBM Tivoli Monitoring
- IBM Application Performance Management (on-prem version)
- IBM Tivoli Composite Application Manager for Applications
- IBM Tivoli Composite Application Manager for Transactions
- IBM OMEGAMON (monitoring suite for IBM z/OS systems and middleware https://www.ibm.com/it-infrastructure/z/omegamon#2306732)

The exporter collects the last snapshot of selected metric values from ITM metric API and exposes them as Prometheus gauges.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -153,6 +153,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Heka exporter](https://github.com/imgix/heka_exporter)
    * [Huawei Cloudeye exporter](https://github.com/huaweicloud/cloudeye-exporter)
    * [InfluxDB exporter](https://github.com/prometheus/influxdb_exporter) (**official**)
+   * [ITM exporter](https://github.com/rafal-szypulka/itm_exporter)
    * [JavaMelody exporter](https://github.com/fschlag/javamelody-prometheus-exporter)
    * [JMX exporter](https://github.com/prometheus/jmx_exporter) (**official**)
    * [Munin exporter](https://github.com/pvdh/munin_exporter)


brian-brazil:
Thanks for your PR. You're taking a password on the command line, that should be fixed and you'll also need to add the DCO.

If you're not 100% certain that all the metrics are gauges, use untyped rather than gauges.

`itm_scrape_duration_seconds_total` is a gauge, so should not end in `_total` as that's for counters.
It'd be better to fail the scrape with a 500 when there's an error.

rafal-szypulka:
Thank you for your feedback. 
Password as a CLI argument is optionally allowed only for [ad-hoc commands](https://github.com/rafal-szypulka/itm_exporter#itm-exporter-cli-options) that list specific items from the API (same way as curl which also allows for password via CLI argument ) and not allowed for the `export mode`, so it won’t be visible in the process table when exporter is running. The password always can be specified in the config file if we don't want to have it in the shell history. Is it acceptable?

I think I fixed DCO and I renamed `itm_scrape_duration_seconds_total` to `itm_scrape_duration_seconds_all`.

Regarding 500, [I used `itm_up 0`]( https://github.com/rafal-szypulka/itm_exporter/blob/0924505bc86ec36b1395e12964b81a75b5851038/main.go#L323) in case of problems with scrape as described here: https://prometheus.io/docs/instrumenting/writing_exporters/#failed-scrapes

brian-brazil:
> Password as a CLI argument is optionally allowed only for ad-hoc commands that list specific items from the API

A password on the command line is a password on the command line.

> Regarding 500, I used itm_up 0 in case of problems with scrape as described here

This exporter can't provide useful metrics when there's a failure though, so a 500 would be better.

rafal-szypulka:
OK, I get it :) Now password can be set only in the config file. In the case of a general API connection/authentication/collection failure exporter now fails with 500.

brian-brazil:
Thanks!

PR #1625, opened 2020-05-11
Add spacing to first row

ddavison:
Re-use the existing `top-hr` class which applies a margin to the first row.  

```css
.top-hr {
    margin-top: 30px;
}
```

### Before / After

prom

### DIFF
--- a/content/index.html
+++ b/content/index.html
@@ -13,7 +13,7 @@ layout: jumbotron
 </div>
 
 <div class="container">
-  <div class="row">
+  <div class="row top-tr">
     <div class="col-md-3 col-sm-6 col-xs-12 feature-item">
       <a href="/docs/concepts/data_model/">
         <h2><i class="fa fa-flask"></i> Dimensional data</h2>


ddavison:
> /cc @juliusv
> 
> I'm fighting with some GPG signing stuff, will get back to you.

:tada: This is ready now

juliusv:
Thanks! We actually originally had a spacing in there by having a margin under the jumbotron (see https://web.archive.org/web/20160416134055/https://prometheus.io/), but it got removed when we inserted an intermediary header that had its own margin, and then we didn't think of adding it back when that intermediate header was removed. I was always wondering recently where that spacing went until I researched it now :) But the `top-hr` on the first row way also sounds good, thanks!

juliusv:
@ddavison Oh, this added `top-tr`, but I guess it should've been `top-th`? (you'll see the site still looks the same :))

juliusv:
It should be `top-hr` I mean of course, apparently that class name is hard to get right :laughing:

ddavison:
Oh my goodness @juliusv You are absolutely right :facepalm:

ddavison:
➡️ https://github.com/prometheus/docs/pull/1628

PR #1626, opened 2020-05-12
Add exporter for KDB+

sshanks-kx:
Add exporter for KDB+

Ref: 
https://github.com/KxSystems/prometheus-kdb-exporter
https://kx.com/

Signed-off-by: Simon Shanks <sshanks@kx.com>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -38,7 +38,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Druid Exporter](https://github.com/opstree/druid-exporter)
    * [ElasticSearch exporter](https://github.com/justwatchcom/elasticsearch_exporter)
    * [EventStore exporter](https://github.com/marcinbudny/eventstore_exporter)
-   * [KDB+ exporter](https://github.com/KxSystems/prometheus-kdb-exporter) (**official**)
+   * [KDB+ exporter](https://github.com/KxSystems/prometheus-kdb-exporter)
    * [Memcached exporter](https://github.com/prometheus/memcached_exporter) (**official**)
    * [MongoDB exporter](https://github.com/dcu/mongodb_exporter)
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/exporters.md
```diff
@@ -38,6 +38,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Druid Exporter](https://github.com/opstree/druid-exporter)
    * [ElasticSearch exporter](https://github.com/justwatchcom/elasticsearch_exporter)
    * [EventStore exporter](https://github.com/marcinbudny/eventstore_exporter)
+   * [KDB+ exporter](https://github.com/KxSystems/prometheus-kdb-exporter) (**official**)
```
brian-brazil, line 41: This is not an official exporter.

brian-brazil:
A few comments to help improve the exporter:

Gauges should not end in `_total`. `summary` and `histogram` should not be part of metric names - and there's no need to expose both for the same data, just a histogram will do. The memory metrics sound kdb specific rather than the process_ metrics coming from a client library, so should be prefixed with kdb_. 
memory_physical_bytes is a machine-level metric, and the node exporter should be relied upon for that.
You should seem to be setting the http content-type.

content/docs/instrumenting/exporters.md
sshanks-kx: Does official mean that its an official exporter that's from Prometheus, or the official exporter from the product that's it is to be used with?
I had interpreted that its official if its created by the company upon which its used.

Thanks

sshanks-kx: Sorry, just noticed on the page it had 
'Some of these exporters are maintained as part of the official Prometheus GitHub organization, those are marked as official, others are externally contributed and maintained.'
Will change it.

brian-brazil:
Can you fix the DCO? Squashing is usually easiest.

brian-brazil:
Thanks!

PR #1627, opened 2020-05-13
Add some missing Pagerduty optional fields

grdryn:
These fields are all available on the PagerdutyConfig type:
https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L213..L215

I took the descriptions from the PagerDuty events-api-v2 docs:
https://developer.pagerduty.com/docs/events-api-v2/overview/

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -72,9 +72,9 @@ global:
   [ smtp_auth_password: <secret> ]
   # SMTP Auth using PLAIN.
   [ smtp_auth_identity: <string> ]
-  # SMTP Auth using CRAM-MD5. 
+  # SMTP Auth using CRAM-MD5.
   [ smtp_auth_secret: <secret> ]
-  # The default SMTP TLS requirement. 
+  # The default SMTP TLS requirement.
   # Note that Go does not support unencrypted connections to remote SMTP endpoints.
   [ smtp_require_tls: <bool> | default = true ]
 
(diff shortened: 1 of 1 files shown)


grdryn:
@simonpasquier Please take a look when you have a moment, thanks! :slightly_smiling_face:

content/docs/alerting/configuration.md
```diff
@@ -442,6 +442,15 @@ links:
 
 # The HTTP client's configuration.
 [ http_config: <http_config> | default = global.http_config ]
+
+# The part or component of the affected system that is broken.
+[ component: <string> ]
```
brian-brazil, line 447: These look like tmpl_string to me

grdryn: @brian-brazil I've updated them now to be `tmpl_string`. I don't fully understand the difference, or how to determine when it should be `string` or `tmpl_string`, so any information or links you can provide on that would be appreciated!

brian-brazil: You need to dig through to where they're used, and see if they're templated - which all 3 of these are.

content/docs/alerting/configuration.md
```diff
@@ -442,6 +442,15 @@ links:
...
 [ http_config: <http_config> | default = global.http_config ]
+
+# The part or component of the affected system that is broken.
+[ component: <tmpl_string> ]
+
+# A cluster or grouping of sources.
+[ group: <tmpl_string> ]
+
+# The class/type of the event.
+[ class: <tmpl_string> ]
```
simonpasquier, line 453: lets move it before the `http_config` entry to be consistent with the other sections.

grdryn: I've done that now. There's another problem though, this file was deleted from the repo in #1657 and doesn't seem to exist anywhere else. Per that change, it looks like these AlertManager docs might be moving to the alertmanager repo, and pulled in here during build. I can't seem to find it there either though. 

@roidelapluie any thoughts or suggestions?

simonpasquier: @grdryn indeed you need to move your changes here:
https://github.com/prometheus/alertmanager/blob/master/docs/configuration.md
sorry for the annoyance

PR #1628, opened 2020-05-13
Fix row to use correct margin class

ddavison:
`top-tr` is an invalid CSS class.  The appropriate class is `top-hr` to apply a margin.


juliusv:
:+1: Thanks :)

PR #1604, opened 2020-04-21
[WIP]Add a training & support page

celestehorgan:
Closes #1599.

Todo:
- [ ] Finalize criteria (if any) around PR-ing to this page and let @celestehorgan know.

PR #1629, opened 2020-05-13
Training & Support page fixups / remove marketing blurbs

juliusv:
I just wanted to remove the marketing blurbs as requested in
https://github.com/prometheus/docs/pull/1604, but while doing that I
noticed that the current HTML wasn't using Bootstrap validly (the only
valid children of rows are cols, not containers), so I changed that
around everywhere as well, and also fixed the footer, which wasn't
captured in its usual non-fluid container.

training_page

### DIFF
--- a/content/support-training.html
+++ b/content/support-training.html
@@ -4,129 +4,122 @@ layout: jumbotron
 ---
 
 <div class="container-fluid training">
-  
+
   <div class="row gray-bg">
-    <div class="container">
-    <div class="col-md-12 doc-content">
-      <h1>Support & Training</h1>
-      <p>This list is provided in alphabetical order. If you'd like to add your training resource, see <a href="#add">Add a resource</a>.
-    </div>
+    <div class="col-md-12">
+      <div class="container doc-content">
+        <h1>Support & Training</h1>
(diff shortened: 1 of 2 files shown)


content/support-training.html
```diff
@@ -4,129 +4,122 @@
...
-          <div class="panel panel-default">
-            <div class="panel-body">
-              <a href="https://training.robustperception.io/" target="_blank">
-                <h3>Learn how to Monitor with Prometheus</h3>
-              </a>
-              <p>Delivered by <a href="https://training.robustperception.io/" target="_blank">Robust Perception Training</a>.
-            </div>
+        <div class="row equal">
+          <div class="col-md-3 col-sm-6 doc-content tc-col">
+            <a href="https://training.robustperception.io/" class="btn btn-training btn-default btn-large" target="_blank">
```
brian-brazil, line 24: Hmm, wouldn't L go before R or are we sorting on the The?

juliusv: Ah yeah, fixed.

juliusv:
(Forgot to git-add the LF logo file, I just pushed it directly to master)

PR #1632, opened 2020-05-14
Aerospike official Prometheus exporter

realmgic:
This is a link to the official Aerospike Official Prometheus Exporter maintained by Aerospike engineering team

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -30,7 +30,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
 [Cassandra](http://cassandra.apache.org/).
 
 ### Databases
-   * [Aerospike exporter](https://github.com/alicebob/asprom)
+   * [Aerospike exporter](https://github.com/aerospike/aerospike-prometheus-exporter)
    * [ClickHouse exporter](https://github.com/f1yegor/clickhouse_exporter)
    * [Consul exporter](https://github.com/prometheus/consul_exporter) (**official**)
    * [Couchbase exporter](https://github.com/blakelead/couchbase_exporter)


brian-brazil:
While there is a preference for vendor exporters, there's still due diligence to be done. Does this do/not-do anything that the existing exporter does? Have you tried to work to work with the existing more popular exporter?

realmgic:
The new exporter support all of the metrics being emitted by the database (including Enterprise Edition metrics specific metrics, and newly released metrics from newest version - 4.9 and 5.0).
The open source does not support some of the EE features and metrics.

brian-brazil:
The existing exporter appears to be actively maintained, did anyone try adding those metrics? I'd like to avoid situations where an existing community is lost.

Also why is the main collection logic in functions called Refresh? That hints collection may not be synchronous and race-free.

realmgic:
I'm not sure if the community would be lost, especially if the vendor officially maintain the exporter as part of its [CNCF ](https://www.cncf.io/) commitment (see here, for example - https://www.aerospike.com/products/cloud-foundations/ ).

Regarding the technical question - I'd need to ask @spkesan to comment.

brian-brazil:
The existing exporter has had multiple contributors from different companies, the proposed one has none outside Aerospike. I'd rather not completely throw away the small community that @alicebob has built.

realmgic:
We can also ask @Xaelias, @Alb0t for their opinion

Xaelias:
Hi 👋 
I won't speak directly for @alicebob although I'm pretty sure I know their opinion.

For me personally (which is only a small contributor of asprom, alicebob did all the hard work :-D), it makes a lot of sense to fully replace asprom with aeroprom. With one caveat: older CE versions.
We know that some people use older versions of aerospike for instance because they need more namespaces than are allowed in current CE versions. asprom still serves a purpose there because aeroprom is never going to really support these old versions.

For newer versions though, aeroprom just makes more sense. We struggle to keep up with the updates. I'm the only one with access to the EE (through my work) and I add metrics whenever I need them or find that I'm missing them, and while in theory you could just parse the `info` reference and the `metrics` reference from aerospike to generate the list of metrics, it just makes sense to use the exporter from aerospike themselves.
They know better. They will test it more extensively than we ever could. They have more metrics (and will always have more metrics). They keep up to date release with releases. And talking and working with their engineers, they are serious and good people. I have 0 worries.

While I will definitely miss contributing to asprom, I will be deploying aeroprom for my work either this week or the next. There is just no good reason to use an exporter that two people struggle to maintain (we don't even cover half the use cases you can have with aerospike, we don't use index on disk, none of us have access to pmem or these kind of things, I added secondary indexes metrics a couple weeks ago XD, when 5.0 is released we won't be doing multi-site clusters, the list goes on). There are a lot of metrics we don't bother parsing because they are not numbers (booleans, states, ...)...

I would definitely redirect people to asprom for older versions (and I'm sure the Aerospike folks will be more than happy to mention asprom on their readme/github page).
But anyone using anything 4.5+ should really just use the aerospike version.

And from previous discussion (some of them in public), I think I know @alicebob is more than happy to leave the job of supporting the aerospike exporter to people that have way more resources, and way easier access to knowledge/information than we do :-)

Their exporter will be better and more consistent/up to date in every way :-)
It's a good thing IMO.
But I do appreciate that this discussion is taking place. It means a lot about prometheus and I'm all for it!

PS: As far as the change goes, like I mentioned, we should leave asprom and point people using early 4.x or 3.x to asprom still. aeroprom is not going to support them well or at all. And they are not going to get support.

PPS: You should probably mention that the aerospike version is still technically in beta, although I'm told it will be GA soon.

Xaelias:
I won't speak directly for @alicebob although I'm pretty sure I know their opinion.

For me personally (which is only a small contributor of asprom), it makes a lot of sense to fully replace asprom with aeroprom. With one caveat: older CE versions.
We know that some people use older versions of aerospike for instance because they need more namespaces than are allowed in current CE versions. asprom still serves a purpose there because aeroprom is never going to really support these old versions.

For newer versions though, aeroprom just makes more sense. We struggle to keep up with the updates. I'm the only one with access to the EE (through my work) and I add metrics whenever I need them or find that I'm missing them, and while in theory you could just parse the `info` reference and the `metrics` reference from aerospike to generate the list of metrics, it just makes sense to use the exporter from aerospike themselves.
They have more metrics (and will always have more metrics). They keep up to date release with release. And talking and working with their engineers, they are serious and good people.

While I will definitely miss contributing to asprom, I will be deploying aeroprom for my work either this week or the next. There is just no good reason to use an exporter that a couple people struggle to maintain (we don't even cover half the use cases you can have with aerospike, we don't use index on disk, none of us have access to pmem or these kind of things, when 5.0 is released we won't be doing multi-site clusters, the list goes on). There are a lot of metrics we don't bother parsing because they are not numbers (booleans, states, ...)...

I would definitely redirect people to asprom for older versions (and I'm sure the Aerospike folks will be more than happy to mention asprom on their readme/github page).
But anyone using anything 4.5+ should really just use the aerospike version.

And from previous discussion (some of them in public), I think I know @alicebob is more than happy to leave the job of supporting the aerospike exporter to people that have way more resources, and way easier access to knowledge/information than we do :-)

Their exporter will be better and more consistent/up to date in every way :-)
It's a good thing IMO.

alicebob:
Hi! Thanks for linking me here. Yeah, I'm also happy to hand it over to the core team. I made asprom back when Prometheus was still new, and I needed the AS stats in there. But that was years ago, and Prometheus is a bit more common now.

I'll add a note on the asprom readme linking to the new exporter as a better option for new installs.
A lot of thanks to both @Xaelias and @Alb0t for all their updates the last year or two!

brian-brazil:
Great, that all sounds good. Can you fix the DCO?

realmgic:
Done

brian-brazil:
Thanks!

Alb0t:
> A lot of thanks to both @Xaelias and @Alb0t for all their updates the last year or two!
Thanks @alicebob for putting this together for us to use up to this point. We've been testing the vendor-provided exporter for a little while now and giving initial feedback, and plan to totally replace asprom with the vendor provided product at some point soon. This will be a pretty painless change due to the fact we can simply drop in the new metric names replacing what we were using from asprom.

I'm looking forward to bothering the vendor to add features instead of rolling up my own sleeves, so personally I'm very happy about this change.

Appreciate you trying to hold things together @brian-brazil and thanks again for the work you put into this @alicebob

PR #1636, opened 2020-05-18
Add prometheus-version docsearch meta tag to pages

juliusv:
I hope this will cause the Aloglia indexing engine to add a
"prometheus-version" indexing facet to our pages, so that when adding
Algolia search to the site in a subsequent step (I can already test it
locally), we can use this to search only for pages that are either the
latest Prometheus version (otherwise we get 10x the same-looking
result), or have no Prometheus version at all (the doc pages outside of
the versioned Prometheus sub-tree). I need to explicitly add "none" as a
version for the non-versioned pages, as Algolia search doesn't allow
facet filtering on null / undefined values.


juliusv:
@bwplotka @roidelapluie

bwplotka:
> I need to explicitly add "none" as a version for the non-versioned pages,

Just to update your description: "unversioned"  at the end (: 

Nice, thanks! :+1:

juliusv:
Ah thanks, I actually meant for it to read "none" in the code as well but didn't add to git. Fixed!

PR #1639, opened 2020-05-22
Add BMC exporter to list

gebn:
This is a pure-Go alternative to the existing `ipmi_exporter`. It doesn't expose the entire SDR, but the metrics it does export follow Prometheus conventions and are normalised across BMC vendors, making it practical to aggregate them across a large estate. I hope to add more coverage over time. Any comments much appreciated!

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -59,6 +59,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
 ### Hardware related
    * [apcupsd exporter](https://github.com/mdlayher/apcupsd_exporter)
    * [BIG-IP exporter](https://github.com/ExpressenAB/bigip_exporter)
+   * [BMC exporter](https://github.com/gebn/bmc_exporter)
    * [Collins exporter](https://github.com/soundcloud/collins_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for sharing your exporter. We try to avoid duplication in this list to encourage the community to work towards one tool. This exporter appears to duplicate the IMPI exporter, without all of its features. I would suggest trying to work with the existing listed exporter to see if you can improve that.

gebn:
Thanks for your consideration Brian. I don't really see them merging, as they approach the problem from opposite angles - there isn't really any scope for incremental improvement. I'll close this MR.

brian-brazil:
I'd suggest talking to the impi exporter about switching over to whatever Go library you're using for IPMI, as shelling out isn't the best way of doing things - though I'm not sure how that'd work for local IPMI.

gebn:
The whole thing actually started after a few too many segfaults from `ipmi_exporter`, and crazy amounts of time spent in kernel mode! That's not a bad idea - I wrote the [underlying library](https://github.com/gebn/bmc). Retrieving the metrics is one thing, being able to normalise them and make them useful is another. `ipmi_exporter` prioritised quantity; I had different requirements and went for quality.

PR #1643, opened 2020-05-25
Add chief alert executor alertmanager webhook receiver

pcarranza:


brian-brazil:
Thanks for sharing. We generally only list one type of integration, and it looks like both https://github.com/imgix/prometheus-am-executor and https://github.com/adnanh/webhook have been around for longer and have established communities so one of those should be listed.

pcarranza:
ACK

PR #1646, opened 2020-05-26
Add Cryptowat exporter to the list

nbarrientos:


brian-brazil:
Thanks!

This exporter doesn't look like it'll handle a pair/exchange disappearing or the API going down gracefully - instead it'll silently provide stale data. Scraping synchronously and failing the scrape on error would handle this.

`crypto_last_value` is a bit redundant, the value of a gauge is always the most recent value. Just `crypto_value` would be enough - presuming that `value` is the unit here. The `value` in `crypto_change_24h_percent_value` is redundant as the unit is already percent, and ratios are preferred.

nbarrientos:
Thanks for your comments, I'll patch as suggested and release a new version.

> This exporter doesn't look like it'll handle a pair/exchange disappearing or the API going down gracefully - instead it'll silently provide stale data. Scraping synchronously and failing the scrape on error would handle this.

Yup, that's correct. If a given market disappears then the exporter will stop doing `.Set()`s on the corresponding metric. I'll give it a think.

> crypto_last_value is a bit redundant, the value of a gauge is always the most recent value. Just crypto_value would be enough - presuming that value is the unit here

Yup, Makes sense. The `last` comes directly from how the API calls the data point. The unit is tricky though because it depends on the pair selected. If the pair scraped is say `btcusd` then the unit would be `usd`. Any suggestion on how to make this better?

> The value in crypto_change_24h_percent_value is redundant as the unit is already percent, and ratios are preferred.

Ack :)

brian-brazil:
> I'll give it a think.

If you need the async due to throttling, one option is to have a metric indicating the last successful full update.

> If the pair scraped is say btcusd then the unit would be usd. Any suggestion on how to make this better?

That wouldn't really be workable for graphing etc. as the metric name would vary. You don't have to have a unit, but just `crypto` on its own would be weirdly short.

PR #1647, opened 2020-05-27
Add network exporter

syepes:
Addition of the Network Exporter


brian-brazil:
Thanks for your PR. I aim to avoid duplication in this list, and this network exporter seems to duplicate both the blackbox and especially smokeping exporters which are already listed. If there's ways in which you feel those could be improved, I'd encourage you to talk to their maintainers.

syepes:
@brian-brazil  Actually is a mix of both blackbox and smokeping but with more metrics (jitter stats) and features but specially the [MTR](https://en.wikipedia.org/wiki/MTR_(software)) module (traceroute)

brian-brazil:
At the end of the day it's filling the same base use cases that already listed exporters do, so I don't think it makes sense to list this here. You are free to announce your exporter on the -users list.

syepes:
@brian-brazil, Ok no problem

PR #1648, opened 2020-05-28
Make celsius recommendation more nuanced

beorn7:
Triggered by https://github.com/prometheus/client_golang/pull/761

### DIFF
--- a/content/docs/practices/naming.md
+++ b/content/docs/practices/naming.md
@@ -75,7 +75,7 @@ The list is not exhaustive.
 | Family | Base unit | Remark | 
 | -------| --------- | ------ |
 | Time   | seconds   |        |
-| Temperature | celsius | _celsius_ is preferred over _kelvin_ for practical reasons. |
+| Temperature | celsius | _celsius_ is preferred over _kelvin_ for practical reasons. _kelvin_ is acceptable as a base unit in special cases like color temperature or where temperature has to be absolute. |
 | Length | meters | |
 | Bytes  | bytes | | 
 | Bits   | bytes | To avoid confusion combining different metrics, always use _bytes_, even where _bits_ appear more common. |


brian-brazil:
I strong disagree with this PR. The whole point of base units is that there is exactly one of them. If someone wishes to not follow best practice we can't stop them, but that doesn't mean we should then explicitly endorse breaking of best practices.

For example should we allow bits, given that far more network engineers who typically work in bits use Prometheus, and their use cases are more in line with the general category systems monitoring that Prometheus is intended for?

beorn7:
I think bits/bytes is a borderline case, too, but for other reasons. For one, it _is_ in fact a constant source of confusion because there are many cases where you have to ask yourself if that number is bits or bytes. That's different for Kelvin/Celsius, where the domains are usually clearly separated (nobody would assume Celsius if talking about color temperature or if calculating a Boltzmann factor).

Also, bit/byte is converted by the multiplication of a number everyone knows from the top of their head, but Kelvin/Celsius is an _addition_ by a number that only a small fraction of Prometheus users know from the top of their head. But I wrote that already in https://github.com/prometheus/client_golang/pull/761 . There are only so many arguments to make here, and we are repeating them already.  Time to vote.

content/docs/practices/naming.md
```diff
@@ -75,7 +75,7 @@ The list is not exhaustive.
 | Family | Base unit | Remark | 
 | -------| --------- | ------ |
 | Time   | seconds   |        |
-| Temperature | celsius | _celsius_ is preferred over _kelvin_ for practical reasons. |
+| Temperature | celsius | _celsius_ is preferred over _kelvin_ for practical reasons. _kelvin_ is acceptable in special cases like color temperature and scientific usage. |
```
brian-brazil, line 78: This contradicts the docs at https://prometheus.io/docs/instrumenting/writing_exporters/#naming 6th paragraph. The simplest fix would be replacing must with should, though I think it'd be better to expand that paragraph a little explaining where it can be okay to do things like this (i.e. extremely unlikely to end up mixing units down the line) rather than changing here.

"Scientific usage" is also a bit vague - does that include all environmental/weather monitoring for example? I'd leave things to the concrete example of monitor color temperature.

roidelapluie: Can we maybe have a category "Color temperature and Science"

brian-brazil: For such a relatively rare case, that seems a bit much. Science is also still a bit vague, e.g. I've attempted science many times with machine temperature (turns out it was dodgy RAM).

beorn7: It's not really a contradiction because _kelvin_ would be the proper base unit in those examples.  I changed the remark in this PR to use the word "base unit", although I'm torn if this formal correctness is really worth making the remark more verbose.

A verbose explanation when multiple base units are appropriate is IMHO more confusing than helpful, as the case is very rare (so far, temperature is the only one). It would probably just trigger people to shoehorn the case of their own favorite unit into this line of argument. I suggest to just leave it with temperature here. Should anyone ever bring up the temperature example as a reason to push another alternative base unit, we could still expand on the underlying reasoning.

I deliberately added the "scientific usage" because I think it's important. I kept it deliberately vague for the sake of brevity, given that this is merely inside a remark. I don't think there is any danger that anybody will feel compelled to convert the readings from their weather monitoring device into K. However, there is a danger that somebody that needs to calculate Boltzmann factors will feel compelled to use °C if we only mention the color temperature case.

In the hope to address your comments, I have extended the remark in my latest push. My problem with it is that the remark is now much longer without any practical impact. It is mostly there to address your rather formal concerns, which in my personal opinion have little impact on what people will actually do. However, I'm willing to take the leap of faith that this a longer remark is here more beneficial than harmful.

brian-brazil: This seems to be internally-self consistent now.

> It would probably just trigger people to shoehorn the case of their own favorite unit into this line of argument. 

That's my primary concern with all of this.

brian-brazil:
:+1:

PR #1653, opened 2020-06-01
Fixes #1638 - routing tree editor not working

### LINKED ISSUE #1638: Routing Tree Editor website not accurately simulating routes
What happened:

I have been working on implementing the notifications in our prometheus install. I wanted to test the routes before I started blasting everyone with notifications so I went to the routing tree editor to test my config. I would apply labels and the alert never moved past the initial default receiver. I loaded my config into the website at https://prometheus.io/webtools/alerting/routing-tree-editor/

I suspected it was an error in my layout and spent time testing it and modifying it and nothing ever changed. To test my theory I used the following config file: "https://github.com/prometheus/alertmanager/blob/master/doc/examples/simple.yml"
In testing that file in the routing tree editor the same behaviour was observed. No matter what labels I used or modified the route never seemed to move beyond the default route.

What was expected:
I expected the route to follow and change receiver based on changed labels to test an alerting path.

adamshone:
One for @juliusv 

The `input` selector was finding the search box that has been added to the page header rather than the correct input.

### DIFF
--- a/static/routing-tree.js
+++ b/static/routing-tree.js
@@ -57,7 +57,7 @@ d3.select(".js-parse-and-draw").on("click", function() {
 
 // Click handler for input labelSet
 d3.select(".js-find-match").on("click", function() {
-  var searchValue = document.querySelector("input").value
+  var searchValue = document.querySelector(".js-label-set-input").value
   var labelSet = parseSearch(searchValue);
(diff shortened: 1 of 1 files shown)


juliusv:
:+1: Thanks, good catch, just selecting a random input on the page sounds like a bad idea indeed!

PR #1654, opened 2020-06-04
Document new webhook max_alerts config field

juliusv:
See https://github.com/prometheus/alertmanager/pull/2274

### DIFF
--- a/content/docs/alerting/configuration.md
+++ b/content/docs/alerting/configuration.md
@@ -698,14 +698,14 @@ endpoint:
 ```
 {
   "version": "4",
-  "groupKey": <string>,    // key identifying the group of alerts (e.g. to deduplicate)
-  "truncatedAlerts": <uint64>, // how many alerts have been truncated due to "max_alerts"
+  "groupKey": <string>,              // key identifying the group of alerts (e.g. to deduplicate)
+  "truncatedAlerts": <uint64>,       // how many alerts have been truncated due to "max_alerts"
   "status": "<resolved|firing>",
   "receiver": <string>,
   "groupLabels": <object>,
   "commonLabels": <object>,
   "commonAnnotations": <object>,
-  "externalURL": <string>,  // backlink to the Alertmanager.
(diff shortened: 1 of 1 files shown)


roidelapluie:
Can you define uint64 at the top?

juliusv:
@roidelapluie Done. FWIW we also have a bunch of `int` fields in the main Prometheus configuration at https://prometheus.io/docs/alerting/configuration/#configuration-file that we don't define at the top. But I haven't checked yet if they are all different types of ints, etc.

content/docs/alerting/configuration.md
```diff
@@ -46,6 +46,7 @@ Generic placeholders are defined as follows:
 * `<secret>`: a regular string that is a secret, such as a password
 * `<tmpl_string>`: a string which is template-expanded before usage
 * `<tmpl_secret>`: a string which is template-expanded before usage that is a secret
+* `<uint64>`: an unsigned 64-bit integer value
```
brian-brazil, line 49: We use `int` over in the Prometheus config docs.

juliusv: Yeah, that's what I was wondering about. Should we really just use int everywhere in the docs, or should we also specify it more clearly in the main Prometheus docs? To make sure people don't set values that overflow or underflow.

brian-brazil: I'd say keep it simple with just int.

juliusv: Ok, changed to int.

content/docs/alerting/configuration.md
```diff
@@ -693,6 +699,7 @@ endpoint:
 {
   "version": "4",
   "groupKey": <string>,    // key identifying the group of alerts (e.g. to deduplicate)
```
brian-brazil, line 701: Can we line up the comments?

juliusv: Done.

PR #1657, opened 2020-06-06
Version Alertmanager

### LINKED ISSUE #857: Extract alertmanager documentation
The alertmanager pages should be moved to prometheus/alertmanager eventually.

roidelapluie:
cc @simonpasquier @juliusv is that something we want?

Closes #857


juliusv:
Here is some historical context which I had to dig out again to remember:

https://github.com/prometheus/prometheus/issues/4216
https://github.com/prometheus/docs/issues/857
https://github.com/prometheus/alertmanager/pull/1062

As far as I can see, the PR by @josedonizetti back then was waiting on @grobie doing some implementation changes to the repo versioning (https://github.com/prometheus/alertmanager/pull/1062#issuecomment-360799938), but that ultimately never happened. I'm not sure what those required changes were. Maybe it was around how the current system of pulling repos is very slow (so we already had to limit the number of Prometheus versions included in the docs so that CI doesn't time out), but not sure.

But generally yes, we want something like this.

roidelapluie:
Yes I did read this. The timeouts do not come from the coming of the repos, but from the fact that each page takes 1s to render. 1 version=printing of new pages. Improving the repo clones will not help.

juliusv:
Yeah, Ruby / nanoc is really slow. So if we wanted AM versioning we could either store fewer versions of Prom/AM, look into increasing CI timeouts (though I think even with longer timeouts, the current feedback cycle length is already too long), or move to a faster framework like Hugo (where I was worried so far about how to replicate all of our custom filter functionality).

roidelapluie:
Maybe creating a checksum method within repo_docs that would render e.g. the lists of data sources+shasum of the repo would help.

juliusv:
:+1: Nice, thanks for the optimizations and making it work :)

juliusv:
(please change from your fork to the official repo before merging, of course)

PR #1659, opened 2020-06-07
Fix Dockerfile

roidelapluie:
Ruby 2.4 is needed and the encoding is mising on some data sources.

I have moved encoding next to `type:`, so it will less likely be removed
accidentaly.


SuperQ:
We should probably do a nanoc update to get to a current (Ruby 2.6 or 2.7) release after this.

roidelapluie:
Very naively I picked up ruby 2.4 because it is the last version with a -onbuild docker image.

I will look for the update.

PR #1662, opened 2020-06-08
Speedup build with caching navbar

roidelapluie:


roidelapluie:
There is a bug still

roidelapluie:
There is a bug with the versioned content

SuperQ:
Oof, this stuff is overly complicated to work around how slow generating takes.

roidelapluie:
> Oof, this stuff is overly complicated to work around how slow generating takes.

The gain is ~50% so I think it is worth it. With that merged, I think we can now version alertmanager docs.

SuperQ:
I agree, it's worth it. But, rather than spend too much time trying to speedup this build, we should continue the Hugo investigation.

PR #1670, opened 2020-06-11
Add web exporter link to exporters page.

huseyinyilmaz:
Signed-off-by: Huseyin Yilmaz <yilmazhuseyin@gmail.com>

Hello,
I wrote a new exporter that can count elements on a page with css queries. It also exposes request duration and response size as metrics. I have been using it on production for around couple weeks and I want to share it with community.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -117,6 +117,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Tinyproxy exporter](https://github.com/igzivkov/tinyproxy_exporter)
    * [Varnish exporter](https://github.com/jonnenauha/prometheus_varnish_exporter)
    * [WebDriver exporter](https://github.com/mattbostock/webdriver_exporter)
+   * [Web exporter](https://github.com/huseyinyilmaz/web_exporter)
 
 ### APIs
    * [AWS ECS exporter](https://github.com/slok/ecs-exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for sharing. Knowing the number of DOM elements before any JS is applied seemed a bit niche, and general HTTP monitoring is already covered by the blackbox and webdriver exporters. We try to avoid duplication in this list, so the community tends to work towards improving existing exporters rather than continuously reinventing the wheel.

huseyinyilmaz:
Hey @brian-brazil, you are right that this one is close to blackbox exporter and webdriver exporter. Problem I was having before implementing this was checking the actual content of the page. For instance, I want to know if page has a header, footer and certain components on the page. This is specially helpful if some of the components depends on a database table that can be changed from an interface.

As far as I am aware there is only single way of achieving this right now. Blackbox exporter provides a configuration like this: 

```  
     fail_if_body_not_matches_regexp:
        - '"class":"page-header"'
        - '"class":"page-footer"'
```

There are 2 problems with this. First one is, I am just running a regex match on page as text. It does not let me look for a specific element in specific location. It might even match with something else on the page by mistake. Second thing is it creates a metric on failure. I do not necessarily want to create an alert from this even though this is the main usecase. I also graph those for elements that is variable in numbers.

I guess my point is we already have exporters that check network health better than web_exporter. I am actually using blackbox exporter for most of health checks myself. But point of web_exporter is actually check the content of the page more than the network. I throw in network health info as an extra because I already had the data in hand and they create pretty sweet graphs :)

@brian-brazil let me know if you still think web exporter does not cover different enough need from webdriver exporter and blackbox exporter. If you still think it is too similar, I will close the pr.

brian-brazil:
I think this is much too niche to list, the blackbox exporter already sufficiently covers this use case between regex matching and response size. In general I'd treat blackbox monitoring as a smoketest, not try to use it to indirectly do SQL queries.

huseyinyilmaz:
Got it. Closing the pr.

PR #1673, opened 2020-06-15
Add Sysdig to Remote Endpoints and Storage

dmitsh:
Signed-off-by: Dmitry Shmulevich <dmitry.shmulevich@sysdig.com>

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -55,7 +55,7 @@ data volumes.
   * [QuasarDB](https://doc.quasardb.net/master/user-guide/integration/prometheus.html): read and write
   * [SignalFx](https://github.com/signalfx/metricproxy#prometheus): write
   * [Splunk](https://github.com/kebe7jun/ropee): read and write
-  * [Sysdig](https://sysdig.com/opensource/prometheus): read
+  * [Sysdig](https://sysdig.com/products/monitor/prometheus-monitoring): read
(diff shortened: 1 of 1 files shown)


content/docs/operating/integrations.md
```diff
@@ -55,6 +55,7 @@ data volumes.
   * [QuasarDB](https://doc.quasardb.net/master/user-guide/integration/prometheus.html): read and write
   * [SignalFx](https://github.com/signalfx/metricproxy#prometheus): write
   * [Splunk](https://github.com/kebe7jun/ropee): read and write
+  * [Sysdig](https://sysdig.com/opensource/prometheus): read
```
brian-brazil, line 58: To be clear a normal Prometheus can read data from Sysdig via the remote read protocol?

We've not had one of those before for this list, and I'm not seeing any relevant docs on the linked page so I'm just double checking.

dmitsh: Sysdig does not support Prometheus Remote Read API, but it exposes Prometheus HTTP API endpoint for querying the data.

dmitsh: I've also updated the link

brian-brazil: This list is for remote storage integrations, we don't have a list for things that provide the HTTP API I'm afraid.

dmitsh:
Got it, thanks!

PR #1676, opened 2020-06-17
Add bitcoin exporter to the list

mmilata:
Permission given by the maintainer in https://github.com/jvstein/bitcoin-prometheus-exporter/issues/14

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -177,6 +177,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
 ### Miscellaneous
    * [ACT Fibernet Exporter](https://git.captnemo.in/nemo/prometheus-act-exporter)
    * [BIND exporter](https://github.com/prometheus-community/bind_exporter)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for you PR, we try to avoid duplication on this list and this appears to be for the same software as the bitcond exporter we already list. Is this actually something significantly different?

mmilata:
It's a different software (with different metrics) for same bitcoin daemon.

brian-brazil:
We generally only list one exporter for a given piece of software. I'd suggest working with @LePetitBloc to add whatever features you feel that exporter is missing so we end up with one really good exporter.

mmilata:
That sounds great in theory but I doubt it's going to happen. I'd suggest mentioning on the exporter list that it is first-come-first-served only and that users should perhaps search the web to see if there's another exporter that is more actively maintained or one that is more suitable to their needs.

SuperQ:
@mmilata It's not first-come-first-served. If an exporter is abandoned and or a new one comes along that greatly improves on the old one, we are willing to change the canonical listing.

What we are trying to accomplish is avoiding too many simultaneous active forks. Competition in the ecosystem is fine, but we need justification for why we should change our recommendation.

PR #1677, opened 2020-06-18
Update PgBouncer exporter link

SuperQ:
Replace link to original Python PgBouncer exporter.
* The Python exporter hasn't been updated since 2017.
* The new Go exporter has many more metrics available.
* The new Go exporter is in the Prometheus Community.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -46,7 +46,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [MySQL server exporter](https://github.com/prometheus/mysqld_exporter) (**official**)
    * [OpenTSDB Exporter](https://github.com/cloudflare/opentsdb_exporter)
    * [Oracle DB Exporter](https://github.com/iamseth/oracledb_exporter)
-   * [PgBouncer exporter](http://git.cbaines.net/prometheus-pgbouncer-exporter/about)
+   * [PgBouncer exporter](https://github.com/prometheus-community/pgbouncer_exporter)
    * [PostgreSQL exporter](https://github.com/wrouesnel/postgres_exporter)
    * [Presto exporter](https://github.com/yahoojapan/presto_exporter)
    * [ProxySQL exporter](https://github.com/percona/proxysql_exporter)


SuperQ:
Should we add a new `(**community**)` tag? :grin:

brian-brazil:
Hmm, the existing one does look pretty dead however this proposed exporter is a bit broken and not following some best practices. I try to ensure that a replacement is at least as good in terms of best practices. So there's some things I'd like to see fixed and improved before replacement.

It's using direct instrumentation in a racey way in places. The up metric should use ConstMetric, and I'd also question how useful scrapes_total is. In addition the last_scrape_duration_seconds and last_scrape_error do not seem useful as implemented and are more likely to cause confusion than anything else as they're not necessarily about the current scrape, and having two error/up metrics is confusing. This either needs to be simplified down to one error indicator metric, or the difference between these two error metrics made clear to users. In any case, the error metric(s) need to be correctly implemented in a non-racey way via ConstMetric.

An entire class of metrics the existing listing has that the current one doesn't via the ListsCollector: https://git.cbaines.net/prometheus-pgbouncer-exporter/tree/prometheus_pgbouncer_exporter/collectors.py#n63
Missing a few individual metrics that are easily added later is okay when considering a replacement, but an entire collector worth of presumably useful metrics isn't something we should lose.

In terms of general comments for improvement:

The mutex isn't doing anything, it should be removed - not that there should be mutexes on the collection path anyway.

client_wait_seconds and server_in_transaction_seconds are missing a _total.

reserve_pool should have connections in the name I think, similarly for pool_size.

max/current_connections would be better as connections_current/max so they'll end up beside each other. Similarly for the pool metrics.

Should the connection pools metrics use a label?

If queries_duration_seconds is indeed a counter, it should end in _total.

https://github.com/prometheus-community/pgbouncer_exporter/blob/783865ba593961fa7bbe99406d60beba3efedbd0/pgbouncer_exporter.go#L80 - copy&paste error

> Should we add a new (**community**) tag? grin

I'd rather not start to try classifying 3rd party exporter orgs, it'd be noisy and make it incorrectly seem as though community exporters are either preferred or required.

SuperQ:
I've added 3 PRs to address the issues:
* https://github.com/prometheus-community/pgbouncer_exporter/pull/34
* https://github.com/prometheus-community/pgbouncer_exporter/pull/35
* https://github.com/prometheus-community/pgbouncer_exporter/pull/36

PR #1679, opened 2020-06-19
Use current Docker registry

dhoppe:
The following Docker registry is obsolete and provides only version 0.33.0, while 0.35.0 is the last release.


brian-brazil:
Thanks!

PR #1681, opened 2020-06-23
content/docs/instrumenting: Rust client moved to tikv org

mxinden:
Signed-off-by: Max Inden <mail@max-inden.de>


brian-brazil:
Thanks!

PR #1683, opened 2020-06-24
NetApp Solifire exporter

mjavier2k:
- Added link to NetApp Solidfire exporter


brian-brazil:
Thanks for sharing. Your exporter has secrets on the command line which is insecure, please fix this.


In terms of general comments to help improve the exporter:

I'd suggest failing the scrape when there's an error rather than making the user have to check an additional metric.

Several of the types seem off. For example node_name is an info metric and should thus be a Gauge and be called node_info. node_total_memory_gb is a gauge, and should not duplicate all the labels that are on the info metric. 
drives_status looks like it might be a enum - I presume the status can change over time? In which case have a series per potential status with a 1 for the current status and 0 for the rest. drives_capacity should not have a status label, as it'll break things every time the status changes.

cluster_node_stats_mbytes_in sounds like a counter rather than a gauge, and should thus also end in _total. 
For cluster_node_stats_load_histogram there's no need to mention histogram in the metric name and it's not clear what the unit (if any) is. 
For cluster_node_stats_count _count indicates part of a summary/histogram so should be avoided as a suffix for gauges.
For cluster_node_stats_cpu_total_seconds total and seconds should be the other way around.
For the various threshold metrics, threshold should be at the end of the metric name. The general idea is that if all the metrics were sorted, the related metrics would be together so the threshold would be near the thing it's a threshold for.
For cluster_threshold_sum_total_cluster_bytes sum and total are a bit redundant.

Bytes are also the base unit that should be used rather than a mix of megabytes (or is that millibits?) and gigabytes, microseconds&milliseconds should be seconds, and percentages to ratios (presuming that the raw data isn't available for PromQL to calculate the ratio itself - in which case don't expose the ratio). For cluster_node_stats_read_latency_usec_total the metric name and help disagree about what the unit is. cluster_node_stats_used_memory is missing a unit.

cluster_stats_utilization - it's not clear what this is the utilisation of.

mjavier2k:
Thanks Brian for the detailed review. Appreciate it.

ill review and make the necessary improvements.

neufeldtech:
@brian-brazil @beorn7 - we've implemented the recommended fixes and improvements over the past few months. Let us know if there are other changes required before solidfire-exporter can be added to this list. 

Thanks

neufeldtech:
Bump @beorn7

beorn7:
I'm not a qualified reviewer for this.

@RichiH could you make a call here? Or delegate as you see fit.

jan--f:
Hello from the bug scrub!
Apologies for the long delay here. @mjavier2k is you still want to get this merged, can you please rebase?

jan--f:
Hello from the bug scrub!
Closing this for now, please feel free to reopen.

PR #1687, opened 2020-06-27
Add Power == watts base unit to naming.md

mhansen:
I'm just making an IoT power device and I thought this might be nice to standardize in the docs.

### DIFF
--- a/content/docs/practices/naming.md
+++ b/content/docs/practices/naming.md
@@ -83,5 +83,5 @@ The list is not exhaustive.
 | Voltage | volts | |
 | Electric current | amperes | |
 | Energy | joules | |
+| Power  | | Prefer exporting a counter of joules, then `rate(joules[5m])` gives you power in Watts. |
 | Mass   | grams | _grams_ is preferred over _kilograms_ to avoid issues with the _kilo_ prefix. |
- 


brian-brazil:
Thanks for your PR. The preferred way to export this is joules, as when you have a counter in joules and take a rate on it you end up with watts - but the underlying unit in the metric names should still be the joule.

mhansen:
Great point. Watts only really makes sense for instantaneous gauges. How
about I amend this to spell out that you should probably just use joules?

mhansen:
Just writing this up, realised perhaps amperes should be rate(coulombs) by
the same logic?

mhansen:
I've updated it to provide your guidance for power, WDYT? (I'm OK to drop
this if this is over-the-top documentation, or perhaps not in keeping with
the rest of this section).

content/docs/practices/naming.md
```diff
@@ -84,4 +84,4 @@ The list is not exhaustive.
 | Electric current | amperes | |
 | Energy | joules | |
 | Mass   | grams | _grams_ is preferred over _kilograms_ to avoid issues with the _kilo_ prefix. |
- 
+| Power  | | Prefer exporting a counter of joules, then `rate(joules)` gives you power in Watts.  |
```
brian-brazil, line 87: Let's use rate(joules[5m]) so the syntax is valid.

brian-brazil:
> Perhaps amperes should be rate(coulombs) by the same logic?

That's be consistent, however I don't think I've come across a device that ever exposes that. Whereas both the joule and watt variants exist. I'd not generally try to cover everything, bytes and seconds are the main ones and the examples will hopefully provide enough reasoning so that developers end up largely doing something sane.

mhansen:
Yeah, I’ve also never found something exposing raw coulombs either. Sgtm.
I’ll adapt the syntax soon.

content/docs/practices/naming.md
```diff
@@ -84,4 +84,4 @@ The list is not exhaustive.
 | Electric current | amperes | |
 | Energy | joules | |
 | Mass   | grams | _grams_ is preferred over _kilograms_ to avoid issues with the _kilo_ prefix. |
- 
+| Power  | watts | Prefer exporting a counter of joules, then `rate(joules[5m])` gives you power in Watts. |
```
brian-brazil, line 87: You added a base unit since the last push, watts still aren't the base unit. If we were to put anything in here it would be joules.

Also this should go beside the energy entry - like we do with bits&bytes.

mhansen:
Moved besides energy, and removed the unit (sorry, thought it might be
useful to say watts is the extension but prefer joules).

brian-brazil:
Thanks!

brian-brazil:
> sorry, thought it might be useful to say watts is the extension but prefer joules

It's a bit of a grey area. Things get a little messy once you move beyond bytes and seconds.

mhansen:
Thanks for your help on this, I appreciate the guidance and I’m going to
miss a lot less detail exporting joules :-)

PR #1701, opened 2020-07-26
Refactor download release selection

grobie:
* Add test
* Use Ruby's comparable module to implement sort behavior

---

Thanks for the fix @roidelapluie! I added a quick test so that the behavior can be migrated easily in case the docs page gets ported to hugo or such in the future.

### DIFF
--- a/lib/helpers/download.rb
+++ b/lib/helpers/download.rb
@@ -7,7 +7,7 @@ module Downloads
   # repositories returns a list of all repositories with releases.
   def self.repositories
     @_repositories ||= begin
-      repos = Dir.glob('downloads/*').map { |dir| Repository.new(dir) }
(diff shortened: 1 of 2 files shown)


roidelapluie:
Could you add v1.19.0-rc too, in the test?

grobie:
@roidelapluie like this?

roidelapluie:
Yes, but is -rc valid? or is it going to `rescue` and should be `-rc.0` ?

grobie:
Semver defines a release candidate by "... a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]". So `-rc` is a valid pre-release identifier. I added an explicit `#version` test for that test case.

roidelapluie:
Thanks!

PR #1705, opened 2020-07-29
Add New Relic as remote write endpoint

B1XR11L729w583adTl5899556EJ5322a:


brian-brazil:
Can you add the DCO?

Also remote write URLs are not considered secret, so this will expose the license key.

aepuls:
@brian-brazil It looks like John included the `signed off by` line in the commit. Do you need more for the DCO than that?

aepuls:
We are aware that the remote write URLs are not secret but do plan to use this mechanism for now.

brian-brazil:
Thanks!

It'd be better to use basic auth or bearer tokens, which would keep the secret secret. URLs in general are not secret.

jgals:
For future reference, New Relic has since added support for remote write authentication via bearer tokens.

PR #1706, opened 2020-07-29
Add a shortcut for /storage

roidelapluie:
to merge before https://github.com/prometheus/prometheus/pull/7691

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -32,6 +32,9 @@
 /docs/alerting/             /docs/alerting/overview/                    302
 /docs/practices/            /docs/practices/naming/                     302
 
+# Short-URL to be used in the Prometheus logs.
+/storage/ /docs/prometheus/latest/storage/    302
+
 # Redirects for index.hml pages.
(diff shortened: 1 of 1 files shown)


RichiH:
Should this go to `/nfs` or even `/go/nfs`?

roidelapluie:
> Should this go to `/nfs` or even `/go/nfs`?

While now it is for NFS it could be extended in the future, WDYT? I have no strong opinion about the /go part.

brian-brazil:
I don't think we should start creating a redirect farm, let's just link to the doc page directly.

roidelapluie:
It is short. If an operator sees this on a vnc or bmc, without copy/paste, they can quickly type the link in a browser.

I also anticipate that we would need a redirect anyway when the docs would move and need a redirect. As people use old version of prometheus, that link will be around for a long time.

brian-brazil:
As-is searching for "prometheus nfs" will find it, I don't feel we should be over-thinking this.

PR #1709, opened 2020-08-01
Add nvme_exporter to the list

yongseokoh:
Hi maintainers,

I would like to add the nvme_export (https://github.com/yongseokoh/nvme_exporter) link to the page. Please, have it included in the project.

Thanks

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -107,6 +107,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Hadoop HDFS FSImage exporter](https://github.com/marcelmay/hadoop-hdfs-fsimage-exporter)
    * [Lustre exporter](https://github.com/HewlettPackard/lustre_exporter)
    * [ScaleIO exporter](https://github.com/syepes/sio2prom)
+   * [NVMe SSD exporter](https://github.com/yongseokoh/nvme_exporter)
 
 ### HTTP
    * [Apache exporter](https://github.com/Lusitaniae/apache_exporter)


brian-brazil:
Thanks for your PR. This sounds like something that should be part of the node exporter rather than its own exporter - and already seems to be covered by https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/blob/master/smartmon.py

yongseokoh:
> Thanks for your PR. This sounds like something that should be part of the node exporter rather than its own exporter - and already seems to be covered by https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/blob/master/smartmon.py

Some metrics may be duplicated between the node exporter and nvme exporter. But, it differs that it seems that it has been focussed on SATA SSDs, while the nvme exporter deals with NVMe metrics. I think that the NVMe has a lot of metrics in comparison of SATA SSDs. The node exporter cannot cover all metrics from NVMe SSDs.

brian-brazil:
I'd suggest working to add those metrics to the node exporter then, as they're fundamentally machine metrics.

yongseokoh:
Thanks for your suggestion. I will extend my project so that it can handle various NVMe metrics not covered in the node exporter. You can find metrics in the NVMe specification https://nvmexpress.org/wp-content/uploads/NVM-Express-1_4-2019.06.10-Ratified.pdf.

brian-brazil:
My point was that these metrics seem to belong in the node exporter, so it'd be best to add them there rather than requiring users to run an additional daemon. If the node exporter rejects these metrics, we can discuss this at that point.

yongseokoh:
Okay. I see. Thanks for your valuable feedback.

PR #1712, opened 2020-08-04
Strengten marketing claim on download page

RichiH:


roidelapluie:
I don't think we should make people download promql.. that just sounds sooo strange

roidelapluie:
.. and the claim you make is probably wrong

content/download.html
```diff
@@ -13,6 +13,11 @@ <h1>Download</h1>
       </ul>
     </div>
 
+    <p>
+      Download Prometheus, the monitoring framework and TSDB which redefined the
+      observability industry. This includes support for PromQL, the powerful query language for
+      accessing your monitoring data at a scale which was not possible before Prometheus and PromQL.
```
juliusv, line 19: How about this (just sliiightly less obnoxious?) variant:

> Download Prometheus, the leading open-source monitoring framework and TSDB. This includes support for PromQL, the powerful Prometheus query language for processing your monitoring data in a flexible way which was not possible before Prometheus and PromQL.

roidelapluie:
Why do we need this? You also download an awesome set of service discoveries, that many other monitoring systems don't offer out of the box, and many other things.

juliusv:
:+1:

PR #1716, opened 2020-08-08
Add alertmanager-gotify-bridge project to alertmanager integrations

DRuggeri:
Add another webhook receiver for alertmanager integrations


DRuggeri:
Will reopen against a corrected branch without junk commits...

PR #1717, opened 2020-08-08
Add alertmanager-gotify-bridge project to integrations

DRuggeri:
Add another webhook receiver for alertmanager integrations


brian-brazil:
Thanks for your PR. This takes a secret on the command line, which is not secure. Can you fix that please, and then we can list it?

DRuggeri:
Ah, yes, that's right. I've fixed that now in https://github.com/DRuggeri/alertmanager_gotify_bridge/commit/18914ae2ab79504f7d21f351684e82d8216ab397

DRuggeri:
Good to merge?

brian-brazil:
Thanks!

The readme makes it seem as though the flag would work, which is a bit confusing.

DRuggeri:
Oops - thanks, @brian-brazil. Fixed that in the README just now.

PR #1719, opened 2020-08-08
Add Mikrotik exporter

hatamiarash7:


brian-brazil:
Thanks for you PR.

This exporter is taking a secret on the commandline which is insecure, can you fix that please?

What does this cover that the snmp exporter doesn't, and vice versa? Are there any key metrics missing?

I'd suggest a read of https://prometheus.io/docs/instrumenting/writing_exporters/ to help improve your exporter, for example:

Service discovery is the responsibility of Prometheus, not exporters. You should take the target from a URL parameter like the snmp exporter does, and not offer features like custom DNS resolution. I'd personally also suggest removing the single device mode as the multiple device mode should be sufficient, and make things a bit simpler for you.

I'm not sure of the wisdom of providing route-based BGP, the global routing table is not something you should try to expose to Prometheus as that's basically a database dump semantically plus it'll have cardinality issues. This is worth a warning in the docs at least.

Many metrics seem to have an incorrect  type. For example updates-sent sounds like a counter, but is exposed as a gauge. If you're not use, use untyped. Counters should also end in _total, and _count should be avoided for gauges as that indicates one of the counters of a summary/histogram.

Many metrics seem to be missing units. Some units also seem to not be base units. E.g. in w60g signal is a percentage rather than a ratio, frequency is in MHz rather than Hz, and txDistance lacks a unit.

In the dhcpcollector, expiresafter and status should be gauges - not labels. The main metric here should also have a _info suffix, and be a gauge.

On a parse error I'd recommend failing the whole scrape, rather than silently returning no or partial data.

The wlanif collector seems to be excluding disabled interfaces. In general it's best to fetch everything, so that there's no gaps if someone disables something or it otherwise breaks.

In resource collector (and other places) there are more labels on metrics than they should be. Boardname and version belong in an info metric, not on metrics such as memory usage. Name and address sounds like target labels, and should be handled on the Prometheus end (and may also be on an info metric).

PR #1722, opened 2020-08-11
Add PostgreSQL remote endpoint adapter

jkatz:
Hi @brian-brazil,

For your consideration, a URL to a [Prometheus PostgreSQL Adapter](https://github.com/CrunchyData/postgresql-prometheus-adapter) on the remote endpoints and storage list.

Please let me know if you have any questions.

Thanks!

Jonathan

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -53,6 +53,7 @@ data volumes.
   * [MetricFire](https://www.hostedgraphite.com/docs/prometheus/prometheus.html): read and write
   * [New Relic](https://docs.newrelic.com/docs/set-or-remove-your-prometheus-remote-write-integration): write
   * [OpenTSDB](https://github.com/prometheus/prometheus/tree/master/documentation/examples/remote_storage/remote_storage_adapter): write
+  * [PostgreSQL](https://github.com/CrunchyData/postgresql-prometheus-adapter): read and write
   * [PostgreSQL/TimescaleDB](https://github.com/timescale/prometheus-postgresql-adapter): read and write
   * [QuasarDB](https://doc.quasardb.net/master/user-guide/integration/prometheus.html): read and write
   * [SignalFx](https://github.com/signalfx/metricproxy#prometheus): write


brian-brazil:
We already list a generic Postgresql integration, is this doing something so fundamentally different that it makes sense to list both?

brancz:
Correct me if I'm wrong but the other integration appears to have added this notice 4 months ago:

> SUNSET NOTICE
> We'll be sunsetting this project in the coming months as we focus on a new implementation with additional functionality and better support for new TimescaleDB features (such as compression). You can find the new project at https://github.com/timescale/timescale-prometheus.
> 
> More details can be found in our design document for the new project.
> 
> This project will continue only in maintenance mode.

Sounds to me like it would make sense to list both for some time and then eventually remove the one we have currently.

brian-brazil:
In that case the usual thing to try and do would be to try to continue on that project elsewhere, rather than creating something new from scratch.

@cevian can you comment?

jkatz:
Hi @brian-brazil,

I appreciate the review and response. Yes, the remote endpoint adapter for PostgreSQL referenced in this pull request is substantively different.

This remote endpoint adapter only leverages upstream PostgreSQL.  The existing adapter listed is actually a connector to TimescaleDB, which is an extension to PostgreSQL, and thus requires the TimescaleDB extension to PostgreSQL as a dependency.  

From the [Timescale-Prometheus](https://github.com/timescale/timescale-prometheus) docs:

> This project connects Prometheus to TimescaleDB, creating an analytical and long-term storage platform for Prometheus metrics.

and

> TimescaleDB is a distributed time-series database built on PostgreSQL

As TimescaleDB (the extension) is maintained by Timescale (the company) and there is a requirement for a PostgreSQL Adapter that is not dependent on the TimescaleDB extension, this new Prometheus PostgreSQL Adapter is actually a substantively different project that we believe brings new functionality to the Prometheus community.

Happy to answer additional questions.

Thanks,

Jonathan

brian-brazil:
> From the Timescale-Prometheus docs:

Timescale-Prometheus is not what we link to, https://github.com/timescale/prometheus-postgresql-adapter is what the comparison is to (otherwise this would be much simpler). It is not TimeScale specific by my reading of it nor does it require the TimescaleDB extension.

A 3-year old Postgres integration with 16 contributors from a variety of companies is much preferable to a 3 month only one with only one commit from one company, as I want to try and build communities around existing integrations rather than having the wheel continuously reinvented and the community split.

jkatz:
Hi @brian-brazil,

Thanks for your input and clarification!

There are several substantial differences between this project and the currently linked to adapter, which drove our decision to create our own adapter vs. leverage the existing adapter:

- Dependencies. It is mentioned in both projects documentation that Timescale's [prometheus-postgresql-adapter](https://github.com/timescale/prometheus-postgresql-adapter#prometheus-remote-storage-adapter-for-postgresql) is dependent on [pg_prometheus](https://github.com/timescale/pg_prometheus#prometheus-metrics-for-postgresql) and vice versa. Thus it is necessary to install the `pg_prometheus` extension in your PostgreSQL database for the project to work. The adapter being presented in this pull request does not require any dependencies to be installed within your PostgreSQL database.

- CLA requirement. The existing project requires that you sign a copyright assignment in order to contribute.

We of course welcome additional community contribution and are open to discussion of how best to facilitate such contribution. Crunchy Data is quite active in the PostgreSQL community, but is less familiar with the approaches associated with the Prometheus community. We would welcome any guidance on how best to engage in a community-friendly way here.

Happy to answer more questions if you have any.

Thanks,

Jonathan

brian-brazil:
> Thus it is necessary to install the pg_prometheus extension in your PostgreSQL database for the project to work.

I'm afraid I don't know Postgres enough to know how big of a deal this is. I'd honestly be surprised if any stock RDBMS was suitable for the sort of data volumes that something like Prometheus can produce, and my general impression is that Postgres extensions overall are pretty mature. Overall I'd expect that an extension is inevitable on performance grounds, though I know that Postgres is also pretty sophisticated internally.

> CLA requirement. The existing project requires that you sign a copyright assignment in order to contribute.

This is not a problem per-se, as we don't require listings here to even have public source code (though that's only come up so far for SaaS). Also considering the current listing appears to be deprecated the first question is more which repo the codebase and community would be moving to, which I would imagine wouldn't have such an issue.

If it turns out that what we currently list is only useful with Timescale in practice, noone uses it independently, nor wants to continue maintaining it, then replacing this and also adding the new Timescale system could make sense. Some comment from the Timescale maintainers would be useful here.

> We would welcome any guidance on how best to engage in a community-friendly way here.

You're always free to publicise on the user and dev mailing lists. If the community switches to yours, then replacing the existing listing isn't a big problem. Replacing when the existing community is much larger requires much more due diligence, as I don't want to be replacing a given listing every few months.

Given that there's already been 3 years worth of community growth and development, I'd rather not lose that just because the main developers are stepping back. It wouldn't be the first time that that had happened, and the community took over. The particular circumstances here are a bit unusual though.

jkatz:
Hi @brian-brazil,

I hope this note finds you well. I apologize for my slow reply on the above, as we entered the final portion of the major release cycle of PostgreSQL 13 (which coincidentally shipped today :smile:) which kept me rather distracted.

Since we last communicated, we did put out an announcement over the [PostgreSQL -announce mailing list](https://www.postgresql.org/about/news/2070/) and have received some more downloads and contributions.

I have been trying to figure out which are the best [mailing lists to discuss this on](https://prometheus.io/community/). Would prometheus-users be the right place? Given my role managing similar things in the PostgreSQL community, I am particularly sensitive to ensuring I'm interacting with community resources properly :smile:

Thanks,

Jonathan

brian-brazil:
-users would be the best place to announce, -developers if you want to discuss development of the adapter itself.

@cevian can you comment on the plans for the timescale postgres adapter?

akulkarni:
Hi @brian-brazil - Sorry for not responding sooner. Not sure how we missed this. I'll ping @cevian tomorrow (time zones!) and get back to you in the next day or so. Thanks!

cevian:
Hi @Brian-brazil - 

Sorry for missing the previous messages.

That prometheus-postgresql-adapter has been replaced with promscale ([https://github.com/timescale/promscale](https://github.com/timescale/promscale)), which started off as a fork with performance improvements, and now also offers more functionality (e.g., native PromQL). (In reality promscale is version 2.0 of the prometheus-postgresql-adapter, and in retrospect we should have kept it in the same repo to make this clear, but for some reason we didn’t.)

To give a little color, we replaced the prometheus-postgresql-adapter project with promscale for two reasons that are relevant here:

- Its schema design (which the Crunchy seems to also use) was quite limited for query speed and operations. For example, putting everything in a single table doesn’t give fine-grained control over things like data retention and also couples the performance characteristics of metrics with vastly different cardinalities. The Crunchy project being proposed in this PR would suffer from the same issues since it uses an almost identical schema design. (I wish Crunchy had reached out to us beforehand so we could have warned them.)

- The dependency on pg_prometheus was indeed a hindrance to some users, so that has been removed. 

Promscale is a more “prometheus-native” approach that fixes both of these issues. The schema design is vastly different and shows much faster performance in our benchmarks. The schema design (unlike that with the initial prometheus-postgresql-adapter or Crunchy’s version) also de-couples individual metrics, allowing for collection of metrics with vastly different cardinalities and retention periods. 

There has been a lot of thought and community discussion into optimizing this schema for Prometheus data, as can be seen this [design doc](https://tsdb.co/prom-design-doc).

Also, as mentioned earlier, Promscale adds native PromQL support.

Our intention is to support Promscale running on either TimescaleDB or vanilla PostgreSQL with no additional extensions (Vanilla PostgreSQL support is slated for release in the next 1-2 weeks). 

The main advantage of the Crunchy adapter is that it supports Postgres 12 partitioning. We would be happy to accept a PR to add that (should only be a few lines of code).

One possibility is to update the docs link from the current adapter to Promscale (and we are happy to submit a PR to that effect). Users are already moving from the current adapter to Promscale (including some large production users that I can share if you would like).

As an aside - this situation feels a little weird to us. It seems like the adapter from Crunchy is either a fork of our project, or that it borrows heavily from our source code:

TimescaleDB prometheus-postgresql-adapter https://github.com/timescale/prometheus-postgresql-adapter/blob/master/pkg/postgresql/client.go#L586
CrunchyData postgresql-prometheus-adapter https://github.com/CrunchyData/postgresql-prometheus-adapter/blob/master/pkg/postgresql/client.go#L561

But we don’t think they ever reached out to us for improving the project in any way. We would love to collaborate with them.

brian-brazil:
> Our intention is to support Promscale running on either TimescaleDB or vanilla PostgreSQL with no additional extensions (Vanilla PostgreSQL support is slated for release in the next 1-2 weeks).

Thanks for the information. Based on this it currently sounds like we should be updating the current link once that lands.

cevian:
We have finished making Promscale compatible with vanilla PostgreSQL with no additional extensions. For those interested, the work is in https://github.com/timescale/promscale/pull/279. This will be released with version 0.1.0 (scheduled for tomorrow).

I will submit a separate PR to update the link shortly.

brian-brazil:
Superceded by #1758 so closing.

PR #1724, opened 2020-08-17
Add link to Nagios section like others in this page for consistency

showuon:
In the comparison page, each time, we mentioned an previously described scope, we'll put a link there for reference. Just like this:
image

But we didn't add the link in the Sensu's scope section, which should link to Nagios section. 

Signed-off-by: Luke Chen <showuon@gmail.com>


brian-brazil:
Thanks!

PR #1725, opened 2020-08-17
Update the old/wrong link in FAQ page

showuon:
1. `[the section about configuring the local storage](/docs/prometheus/1.8/storage/)`
-> We should point to the latest version of storage doc

2. `[scraping TLS- and auth-enabled targets](/docs/operating/configuration/#%3Cscrape_config%3E)`
-> The link will redirect to `/docs/prometheus/latest/configuration/configuration` page, and the anchor for `scrape_config` is different now. Fix it!

### DIFF
--- a/content/docs/introduction/faq.md
+++ b/content/docs/introduction/faq.md
@@ -201,7 +201,7 @@ to configure Prometheus for the amount of memory you have available.
 ### My Prometheus 1.x server reports to be in “rushed mode” or that “storage needs throttling”.
 
 Your storage is under heavy load. Read
-[the section about configuring the local storage](/docs/prometheus/1.8/storage/)
+[the section about configuring the local storage](/docs/prometheus/latest/storage/)
 to find out how you can tweak settings for better performance.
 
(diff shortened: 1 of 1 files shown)


content/docs/introduction/faq.md
```diff
@@ -201,7 +201,7 @@ to configure Prometheus for the amount of memory you have available.
 ### My Prometheus 1.x server reports to be in “rushed mode” or that “storage needs throttling”.
 
 Your storage is under heavy load. Read
-[the section about configuring the local storage](/docs/prometheus/1.8/storage/)
```
roidelapluie, line 204: The question is about "My Prometheus 1.x server" so the current link is correct.

showuon: @roidelapluie , you're right! I've reverted my change in this commit: https://github.com/prometheus/docs/pull/1725/commits/8fcadfbd343728a209d46c976fbd94b36ee1060f. Thanks.

roidelapluie:
Thanks!

PR #1730, opened 2020-08-20
Update the supported list for alert manager

showuon:
1. Remove the unsupported HipChat
2. Add 2 more supported systems: VictorOps, WeChat
Ref: https://prometheus.io/docs/alerting/latest/configuration/


brian-brazil:
Thanks!

PR #1734, opened 2020-08-26
Adding Hazelcast Management Center to the exporter list

erosb:
Adding Hazelcast Management Center to the exporter list

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -193,6 +193,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [eBPF exporter](https://github.com/cloudflare/ebpf_exporter)
    * [Ethereum Client exporter](https://github.com/31z4/ethereum-prometheus-exporter)
    * [JFrog Artifactory Exporter](https://github.com/peimanja/artifactory_exporter)
+   * [Hazelcast Management Center exporter](https://docs.hazelcast.org/docs/management-center/latest/manual/html/index.html#management-center-prometheus-exporter)
    * [Hostapd Exporter](https://bitbucket.i2cat.net/users/miguel_catalan/repos/hostapd_prometheus_exporter)
    * [IRCd exporter](https://github.com/dgl/ircd_exporter)
    * [Linux HA ClusterLabs exporter](https://github.com/ClusterLabs/ha_cluster_exporter)


brian-brazil:
Is this really an exporter? It seems to be the application itself exposing metrics.
Where can I find the relevant source code?

erosb:
Hello @brian-brazil , the Management Center is a monitoring tool that gathers metrics from Hazelcast nodes, and exports the metrics to Prometheus.

The relevant source code is in this package: https://github.com/hazelcast/management-center/tree/master/src/main/java/com/hazelcast/webmonitor/service/prometheus

Also the relevant integration test: https://github.com/hazelcast/management-center/blob/master/src/test/java/com/hazelcast/webmonitor/service/prometheus/PrometheusExportIntegrationTest.java

brian-brazil:
Those links don't work. Is this basically a proxy for JMX?

> the Management Center is a monitoring tool that gathers metrics from Hazelcast nodes,

That's not the preferred architecture for Prometheus, which should scrape processes directly. This is both more efficient, more scalable, and avoids artifacts and weirdness.

brian-brazil:
Also, do you have a link to an architecture diagram? Most of what I can find about Hazelcast is marketing material, so I can't figure out how this all fits together.

erosb:
Hello,

you may look into the Architecture section of the HZ reference manual, though it doesn't include management center:  https://docs.hazelcast.org/docs/latest-dev/manual/html-single/#hazelcast-imdg-architecture

brian-brazil:
Thanks, but that appears to be software architecture rather than systems architecture. I want to understand what sort of inter-process communication is going on here.

erosb:
Briefly, Hazelcast nodes (cluster members) can be connected to a Management Center application, which is partly a monitoring tool. The nodes stream metrics to MC using Hazelcast's own server-client communication protocol. Most of the metrics are related to the distributed objects (operation latencies, data structure sizes, cache hit ratios, etc).

The list of metrics can be found here: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/internal/metrics/MetricDescriptorConstants.java

brian-brazil:
Okay, so this isn't an exporter then but software exposing Prometheus metrics. Can these metrics be obtained more directly from the nodes, such as via JMX?

jgardiner68:
It is an exporter as Management Center is exporting our internal metrics into a Prometheus end point.

It is possible to obtain metrics directly from the nodes using JMX from the nodes but there are a number of important advantages by using the Management Center exporter:

- Management Center provides a Clustered overview as part of metrics 
- Hazelcast is designed to allow nodes to be added/removed from the cluster which would require the Prometheus config to be updated if pulling the metrics directly.  Using the Management Center Prometheus exporter removes that requirement
- Management Center supports connections to multiple clusters
- As requested by our customers, having a Prometheus exporter within Management Center removes an additional hop

brian-brazil:
> It is an exporter as Management Center is exporting our internal metrics into a Prometheus end point.

That doesn't make it an exporter automatically, by that definition the Alertmanager is an exporter - which it isn't. Exporters exist with a primary aim of producing metrics from 3rd party systems, and this is all 1st party.

> Management Center provides a Clustered overview as part of metrics
> Hazelcast is designed to allow nodes to be added/removed from the cluster which would require the Prometheus config to be updated if pulling the metrics directly. Using the Management Center Prometheus exporter removes that requirement
> Management Center supports connections to multiple clusters

These are all considered anti-patterns, service discovery, scraping, and aggregation are the responsibility of Prometheus - not of targets.

> As requested by our customers, having a Prometheus exporter within Management Center removes an additional hop

This is actually adding a hop, adds a SPOF, latency, and can cause artifacts. The Prometheus way is to talk to targets directly.

> It is possible to obtain metrics directly from the nodes using JMX 

Our general policy is not to try not to list duplicates, so as these metrics are already available in a more Promethean way via an existing exporter so this PR is not the way to go and I'm going to close this for noe.

If your nodes gain a Prometheus endpoint at some point, we can list them under software exporting Prometheus metrics.

PR #1737, opened 2020-08-28
Add c-lightning exporter to exporters.md

cdecker:
The exporter was already added to the port-reservations list, and exports a variety of metrics from the Lightning Network client.


brian-brazil:
Thanks!

Some suggestions to improve your exporter:

_total is a suffix for counters, it shouldn't be used with gauges. Some metrics like lightning_channel_in_msatoshi_offered also sound like they might be counters rather than gauges.

Units are inconsistent, sometimes missing and conflicting with the help text. You should consistently use satoshis, and also include satoshi as the unit in all relevant metric names for clarity.

id is a bit of a generic label, peer or peer_id would be clearer.

cdecker:
Thanks for the review @brian-brazil and a huge thanks for the suggestions, we'll implement them asap :+1:

PR #1738, opened 2020-08-31
Dovecot 2.3.11+ can expose metrics natively

mrueg:
https://doc.dovecot.org/configuration_manual/stats/openmetrics/
https://dovecot.org/doc/NEWS

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -254,6 +254,7 @@ separate exporters are needed:
    * [Diffusion](https://docs.pushtechnology.com/docs/latest/manual/html/administratorguide/systemmanagement/r_statistics.html)
    * [Docker Daemon](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-metrics)
    * [Doorman](https://github.com/youtube/doorman) (**direct**)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for sharing, this is not valid OpenMetrics format so that inaccuracy would need to be fixed first. Please check your output with https://www.robustperception.io/checking-openmetrics-output-is-valid

In addition there's several aspects of the metrics that should be improved, so as timestamps not being exposed (and in the wrong unit).

dovecot_stats_uptime_seconds is a gauge not a counter, and would be better as the more usual process_start_time_seconds. dovecot_build_info _count is redundant on dovecot_client_connections_count. dovecot_client_connections_duration_usecs_sum should be in seconds, and the _sum shouldn't be there - did you mean to use a summary? dovecot_auth_success_count should not have a success label, also where are the failures and/or total? dovecot_imap_command_count has inconsistent labels, as does dovecot_smtp_command_duration_usecs_sum. dovecot_smtp_command_histogram should not contain the word histogram, that's redundant.

mrueg:
@brian-brazil Thanks for looking at the metrics, I've incorporated your feedback into a PR to the upstream project.

roidelapluie:
Thank you @mrueg , I have tested this locally.

PR #1742, opened 2020-09-07
added climate monitor exporter

branttaylor:
Signed-off-by: Brant Taylor <brant.taylor@lifeway.com>

@brian-brazil

Added a Prometheus exporter to the docs that I wrote for some climate monitors that I have here at my house.


brian-brazil:
Thanks!

Some suggestions to improve your exporter:

If this is multi-threaded then it is racy, as you generally shouldn't be resetting metrics on every scrape as a concurrent scrape will see partial data.

The deviceName label on all metrics is really a target label, which it is the responsibility of service discovery/relabelling on the Prometheus side to handle. If you wish to expose this do so as a single info metric. In general the same label pair should never appear on all metrics from a target.

These metrics have missing and/or non-base units. Celcius is the base unit for temperature, the unit of humidity is unclear (it should be a ratio), and I've no idea what the units are for airflow, light, or sound are here. Cubic metres per second, candelas, and bels? Make sure the unit is part of the metric name so it's clear to anyone using the metrics.

branttaylor:
Thank you, @brian-brazil ! Great suggestions, I will work on improving it.

PR #1743, opened 2020-09-10
fix typo: Nan -> NaN

ledmonster:
@juliusv @brian-brazil I assume this should be "NaN", thanks.


brian-brazil:
Can you add the DCO?

Technically Nan is also valid.

brian-brazil:
Superseded by #1744.

PR #1747, opened 2020-09-18
Add vmware vcenter/esx exporter

ynsta:
Hello,

I made a prometheus stand alone exporter for VCenter metrics. Metrics are fetched by govmomi api.

It also works with stand alone ESX without VCenter.

| Collectors          | Description |
| ------------------- | ----------- |
| `collector.ds`      | Datastore metrics collector |
| `collector.esx`     | ESX (HostSystem) metrics collector |
| `collector.respool` | ResourcePool metrics collector |
| `collector.spod`    | Datastore Cluster (StoragePod) metrics collector |
| `collector.vm`      | VirtualMachine metrics Collector |

It is functional but I still have some tests to perform in order to verify that all metrics units are correct as it is not very well documented in vmware api.

Regards,

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -177,6 +177,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [StatsD exporter](https://github.com/prometheus/statsd_exporter) (**official**)
    * [TencentCloud monitor exporter](https://github.com/tencentyun/tencentcloud-exporter)
    * [ThousandEyes exporter](https://github.com/sapcc/1000eyes_exporter)
+   * [vmware vcenter / esx exporter](https://github.com/Intrinsec/govc_exporter)
 
 ### Miscellaneous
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. I try to follow the community when select comes to which exporter to list for a given system, how does this differ from https://github.com/pryorda/vmware_exporter ?

ynsta:
Hello,

Last time I tried this python exporter it crashed with my setup, also some information was missing for my use cases so I made a new one based on some code already done for an exporter to splunk.

Also it is implemented in go so have a smaller footprint.

For my exporter I still have some work on tests to check all units (mib vs bytes) as it is not very well documented on vmware api side.


Regards,

brian-brazil:
To avoid continuously reinventing the wheel and splitting the community, I like to encourage working with the existing exporter rather than creating something new from scratch. I'd ask you to engage with the existing exporter, as if we're going to list one that's the one we'd currently list. Given that I'm going to close this off for now.

PR #1748, opened 2020-09-18
tarantool/metrics instead of tarantool/prometheus

ZolotarevIgor-mail-ru:
https://github.com/tarantool/prometheus is deprecated. tarantool/metrics contains an actual version of Prometheus plugin.


brian-brazil:
Thanks!

PR #1751, opened 2020-09-26
Add Caddy

roidelapluie:
Caddy has native prom metrics now.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -247,6 +247,7 @@ separate exporters are needed:
    * [App Connect Enterprise](https://github.com/ot4i/ace-docker)
    * [Ballerina](https://ballerina.io/)
    * [BFE](https://github.com/baidu/bfe)
+   * [Caddy](https://caddyserver.com/docs/metrics) (**direct**)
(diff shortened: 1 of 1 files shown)


roidelapluie:
PR: https://github.com/caddyserver/caddy/pull/3709
It is already released.

brian-brazil:
:+1: 

Given the cardinality issues that have been seen with other proxy systems, it'd be wise to allow the user to configure which caddy_admin_http_requests_total labels they want.

roidelapluie:
> Given the cardinality issues that have been seen with other proxy systems, it'd be wise to allow the user to configure which caddy_admin_http_requests_total labels they want.

cc @hairyhenderson

hairyhenderson:
Thanks for adding this, @roidelapluie!

@brian-brazil the `caddy_admin_http_requests_total` metric in particular is internal and the `path` label (I expect that's what you're referring to) has limited cardinality (i.e. it's not just using `req.URL.Path`).

But, I do agree in general, and I'm planning on adding some more configurability. The initial goal was to just get some basic instrumentation enabled.

brian-brazil:
Ah indeed, it was the non-admin one I meant. Assembling metric names form different places makes it hard to figure out what they are from simple reading/grepping of the code.

PR #1755, opened 2020-10-03
Add 'Better Uptime' to list of webhook receivers

Daniel15:
Documents the fact that Better Uptime has Prometheus webhook support.


brian-brazil:
Thanks!

PR #1757, opened 2020-10-05
add bsbmp-exporter

David-Igou:
This exporter is for several models of Bosch Sensortec environmental sensors. Tested on Raspberry Pi Zero and 4, but should also work on Banana and Orange Pi's (I plan on verifying this)

Feedback is welcome.


brian-brazil:
Thanks!

Some suggestions for the exporter:

The mutex in Collect seems to be redundant, you're not keeping any state (which is good).

Ratios should be exposed rather than percentages. It's redundant to expose pressure in two slightly different units, pascals are the more base one. I'd suggest failing the scrape rather than crashing the exporter if there's an error.

PR #1758, opened 2020-10-05
Change the integration link for PostgreSQL

cevian:
Promscale is the v2 version of the prometheus-postgresql-adapter
project. It uses a much more efficient schema and is compatible
with vanilla PostgreSQL without additional extensions.

Previous discussion in #1722.

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -53,7 +53,7 @@ data volumes.
   * [MetricFire](https://www.hostedgraphite.com/docs/prometheus/prometheus.html): read and write
   * [New Relic](https://docs.newrelic.com/docs/set-or-remove-your-prometheus-remote-write-integration): write
   * [OpenTSDB](https://github.com/prometheus/prometheus/tree/master/documentation/examples/remote_storage/remote_storage_adapter): write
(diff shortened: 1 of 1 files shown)


content/docs/operating/integrations.md
```diff
@@ -53,7 +53,7 @@ data volumes.
   * [MetricFire](https://www.hostedgraphite.com/docs/prometheus/prometheus.html): read and write
   * [New Relic](https://docs.newrelic.com/docs/set-or-remove-your-prometheus-remote-write-integration): write
   * [OpenTSDB](https://github.com/prometheus/prometheus/tree/master/documentation/examples/remote_storage/remote_storage_adapter): write
-  * [PostgreSQL/TimescaleDB](https://github.com/timescale/prometheus-postgresql-adapter): read and write
+  * [PostgreSQL/TimescaleDB/Promscale](https://github.com/timescale/promscale): read and write
```
brian-brazil, line 56: No need to put Promscale in the text, we write the name of what is being integrated with as that's what the user if looking for it they're on this page.

cevian: Ok changed back.

brian-brazil:
Thanks!

PR #1761, opened 2020-10-13
Add `alertify` to the "Gotify" section of the "Alertmanager Webhook …

scottwallacesh:
…Receiver" docs

@brian-brazil

Signed-off-by: Scott Wallace <scott@wallace.sh>


brian-brazil:
Thanks for your PR. To avoid splitting the community and continuously re-inventing the wheel, we generally will only list one entry for a given integration. If there are features you feel the existing listing is missing, please work with them to improve it.

scottwallacesh:
No problem.

PR #1762, opened 2020-10-14
Automate repository docs inclusion

### LINKED ISSUE #1702: Automatically generate releases numbers
When a new PR is out, we update nanoc.yaml.

In the datasource there: we have:

- LATEST (points to the latest stabe release)
- RC (if there is an ongoing rc, with release branch number  in URL, e.g. 2.21) -- they have a notice on top to say that it is a pre_release
- the stable release (same as latest but with release number in URL, e.g. 2.20)
- the old relases (we keep ~9 of them, they have a note on top saying they are old)

Some particularities:
- Only latest is included in algolia search (see docsearch:include in the header)
- We do have Prom + alertmanager docs versioned now

grobie:
So far every prometheus/alertmanater/... release branch had to be
manually configured in the nanoc.yaml config file. With this change the
most recent release branches will be checked out automatically if a
corresponding semver tag exists.

As the Prometheus git repository includes several hundreds of megabytes
of vendored assets, the repository is cloned bare and all blobs are
filtered by default. Each version is then checked out in an individual
working tree and git's spare-chekcout feature is used to reduce the
checkout to the `docs/` folder. The git data is cached in
`tmp/repo_docs/` and will be recreated automatically if removed.

Fixes  #1702.


lib/data_sources/repo_docs.rb
```diff
@@ -1,60 +1,158 @@
-# TODO(ts): Rewrite data source and use one single instance to combine all
-# different versions for a given path.
-class RepoDocsDataSource < ::Nanoc::DataSources::Filesystem
+require 'uri'
+
+# The RepoDocs datasource provices items sourced from other Git repositories.
```
juliusv, line 3:
```suggestion
# The RepoDocs datasource provides items sourced from other Git repositories.
```

lib/data_sources/repo_docs.rb
```diff
@@ -1,60 +1,158 @@
...
+# For a given repository_url, all git version tags are fetched and for the most
+# recent (in order to save compilation time) tags the `docs/` folder in the
+# respective `release-<version>` is checked out and its content mounted under
+# the given `items_root`.
+#
+# As the Prometheus git repository includes several hundreds of megabytes of
+# vendored assets, the repository is cloned bare and all blobs are filtered by
+# default. Each version is then checked out in an individual working tree and
+# git's spare-chekcout feature is used to reduce the checkout to the `docs/`
```
juliusv, line 12:
```suggestion
# git's sparse-checkout feature is used to reduce the checkout to the `docs/`
```

juliusv (approved):
:+1: Awesome, thanks! I didn't analyze every nanoc code bit in full detail, but the site seems to work great and the overall changes look and sound amazing as well. Yay for bare repos and sparse checkouts!

lib/data_sources/repo_docs.rb
grobie: I should really enable the spell checker plugin more often ...

grobie:
Note that the previous implementation already used sparse-checkouts. But git's new `partialclonefilter = blob:none` wasn't available back then, so I used `depth=1` and individual checkouts. That doesn't work particularly well when also fetching tags/branches though. With the new approach only one git repo is necessary for all versions which should be dramatically faster on slower connections (not really relevant for netlify builds though obviously).

juliusv:
:+1: nice!

roidelapluie:
Thanks!!!!!!

PR #1764, opened 2020-10-15
Improve robustness of git working tree checkout

### LINKED ISSUE #1763: Deploy fails after last commit
https://app.netlify.com/sites/prometheus-docs/deploys/5f8744af5515264191c4cf97

cc @grobie

grobie:
For unknown reasons the netlify pipeline ends up in a state where a git
worktree is not created after a fresh clone of the repository. The
previous implementation checked for the actual checked out working tree
and assumed the administrative git worktree files have to exist as well
in that case. This change explicitly verifies the existence of both path
structures and handles more cases of inconsistencies gracefully.

Hopefully resolves #1763. Will let the pr pipeline run a few times. The master branch pipeline succeeded once last night as well with the new setup.

### DIFF
--- a/lib/data_sources/repo_docs.rb
+++ b/lib/data_sources/repo_docs.rb
@@ -99,19 +99,21 @@ class RepoDocsDataSource < ::Nanoc::DataSource
   end
 
   # git_checkout checks out the directory in the specified branch using git's
-  # sparse checkout and returns the path to the location in the workingtree.
+  # sparse checkout and returns the path to the location in the working tree.
   def git_checkout(branch, directory)
-    worktree = File.absolute_path(File.join(git_dir.delete_suffix('.git'), branch))
-    if !File.exist?(File.join(worktree, '.git'))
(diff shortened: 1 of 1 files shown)


grobie:
Interesting, for some reason it indeed appears that checked out working trees under `tmp/repo_docs/prometheus/release-<version>/` are cached on netlify, but not the bare git repository itself in `tmp/repo_docs/prometheus.git/`. I'll make sure to handle this case as well then.

```
10:34:00 AM: Loading site… Cloning into bare repository 'tmp/repo_docs/prometheus.git'...
10:34:01 AM: Cloning into bare repository 'tmp/repo_docs/alertmanager.git'...
10:34:01 AM: fatal: '/opt/build/repo/tmp/repo_docs/prometheus/release-2.22' already exists
10:34:01 AM: Preparing worktree (checking out 'release-2.22')
```

grobie:
Alright, this seems to handle all edge cases. Still curious to learn what causes this behavior on netlify. I suspect something related to absolute paths referencing a different location which gets cached across builds, but I don't have an immediate mental model for such setup.

PR #1769, opened 2020-10-15
Add Rediseen as one option for Redis exporter

XD-DENG:
Starting from 2.4.0, [Rediseen](https://github.com/XD-DENG/rediseen) starts to support exposing Redis INFO command output in Prometheus-compatible format. This provides an easy way for Redis users to use Prometheus with the metrics they are familiar with.

@brian-brazil  Hope you can consider merging this. Please let me know if any clarification is needed. Thanks!


brian-brazil:
This does not appear to be an exporter, as its primary purpose is not producing metrics. It's not the responsibility of an application to fetch metrics from its backends, it should only be providing metrics about itself.

In addition I try to only list one exporter for a given thing, to avoid splitting the community and continuously reinventing the wheel.

Accordingly this doesn't belong on this list.

XD-DENG:
Understand, and thanks for the clarification. All makes sense to me. Will close this.

Thanks!

PR #1771, opened 2020-10-16
Fix broken url in docs

saintmalik:
### Summary

Easy navigation and easy reading for others

### DIFF
--- a/content/docs/guides/query-log.md
+++ b/content/docs/guides/query-log.md
@@ -39,7 +39,7 @@ scrape_configs:
   - targets: ['localhost:9090']
 ```
 
-Then, [reload](../prometheus/latest/management_api/#reload) the Prometheus configuration:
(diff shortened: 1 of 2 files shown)


grobie (commented):
Thanks! Please sign the DCO as well.

content/docs/guides/query-log.md
```diff
@@ -39,7 +39,7 @@ scrape_configs:
   - targets: ['localhost:9090']
 ```
 
-Then, [reload](../prometheus/latest/management_api/#reload) the Prometheus configuration:
+Then, [reload](./prometheus/latest/management_api/#reload) the Prometheus configuration:
```
grobie, line 42: My comment was forgotten. I don't think this will work either. Laura use a link with an absolute path `/docs/prometheus/...`.

grobie:
Thanks!

PR #1775, opened 2020-10-27
Show .zip archive only for Windows

grobie:
If a release archive is available in both .tar.gz and .zip format, only
show the .zip archive in the download list.

### DIFF
--- a/lib/helpers/download.rb
+++ b/lib/helpers/download.rb
@@ -125,10 +125,15 @@ module Downloads
       @data['assets']
     end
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
:+1: 

This looks okay, but I'm not a ruby person.

grobie:
Thanks, I tested it locally and it works fine. The last missing part for Windows zip releases is a new release of promu https://github.com/prometheus/promu/pull/196 and updating all Makefiles.

PR #1777, opened 2020-10-29
Added SIGNL4 as Alertmanager Webhook Receiver

rons4:
The initial idea was to add a deeper, code-based integration. However, I was told to use the simple webhook right now.

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -84,7 +84,7 @@ For notification mechanisms not natively supported by the Alertmanager, the
   * [Rocket.Chat](https://rocket.chat/docs/administrator-guides/integrations/prometheus/)
   * [ServiceNow](https://github.com/FXinnovation/alertmanager-webhook-servicenow)
   * [Signal](https://github.com/dgl/alertmanager-webhook-signald)
-  * [SIGNL4](https://www.signl4.com/blog/portfolio_item/prometheus-alertmanager-mobile-alert-notification-duty-schedule-escalation/)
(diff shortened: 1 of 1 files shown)


content/docs/operating/integrations.md
```diff
@@ -84,6 +84,7 @@ For notification mechanisms not natively supported by the Alertmanager, the
   * [Rocket.Chat](https://rocket.chat/docs/administrator-guides/integrations/prometheus/)
   * [ServiceNow](https://github.com/FXinnovation/alertmanager-webhook-servicenow)
   * [Signal](https://github.com/dgl/alertmanager-webhook-signald)
+  * [SIGNL4 – Critical Mobile Alerting](https://www.signl4.com/blog/portfolio_item/prometheus-alertmanager-mobile-alert-notification-duty-schedule-escalation/)
```
brian-brazil, line 87: Just the basic name of the product is sufficient. This list is to help existing users of your product to find the integration.

In your example also the repeat_interval is less than the group_interval, which doesn't make sense. Resolve timeout should also be removed, that setting is the default and has no effect for Prometheus alerts - the setting does however cause confusion.

brian-brazil (commented):
Can you also add the DCO?

content/docs/operating/integrations.md
rons4: Thanks and I did the changes as requested.

rons4: The DCO check still fails. I am not sure why but I try to find out.

brian-brazil:
The DCO needs to be on all commits, squashing is generally easiest.

rons4:
I think it looks fine now. Thanks @brian-brazil for all your support.

brian-brazil:
Thanks!

PR #1780, opened 2020-10-29
Add xMatters as an Alertmanager Webhook Receiver

ipugh-xm:
This allows users to send alerts into xMatters.

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -89,6 +89,7 @@ For notification mechanisms not natively supported by the Alertmanager, the
   * [SNMP traps](https://github.com/maxwo/snmp_notifier)
   * [Squadcast](https://support.squadcast.com/docs/prometheus)
   * [Telegram bot](https://github.com/inCaller/prometheus_bot)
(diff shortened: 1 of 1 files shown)


roidelapluie:
Thanks!

Please note that the URL is not considered secret and is displayed in the web UI. You could use the http_config token to get API keys, instead of using the URL.

ipugh-xm:
Is there a way in http_config to set a parameter, or would we have to go with bearer token or basic auth instead?

roidelapluie:
You would need basic auth ir bearer token. Note that I am pretty sure that the `username` of basic auth is not mandatory.

ipugh-xm:
I'll add some details on API Key security for our users who want that. Is there anything else I need to do?

ipugh-xm:
@roidelapluie I've added a note telling our users the lack of security with the way the API Key is stored and mentioned the option of basic authentication. Is there anything else I need to do for our integration to be in the docs?

roidelapluie:
Thanks! It is unclear how to use basic auth however, maybe you should consider adding an exemple.

PR #1784, opened 2020-11-05
add mail-archive.com links, for DeGoogle

GETandSELECT:
Hi

Sorry, I don't know what DCO is. Do you wish real name?

### DIFF
--- a/content/community.html
+++ b/content/community.html
@@ -16,8 +16,8 @@ title: Community
     <p>
       <strong>User mailing lists:</strong>
       <ul>
-        <li><a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a> <a href="https://www.mail-archive.com/prometheus-announce@googlegroups.com/">mirror</a> – low-traffic list for announcements like new releases.</li>
-        <li><a href="https://groups.google.com/forum/#!forum/prometheus-users">prometheus-users</a> <a href="https://www.mail-archive.com/prometheus-users@googlegroups.com/">mirror</a> – for discussions around Prometheus usage and community support. Announcements are <em>not</em> generally mirrored from <a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a>.</li>
+        <li><a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a> <a href="https://www.mail-archive.com/prometheus-announce@googlegroups.com/">(mirror)</a> – low-traffic list for announcements like new releases.</li>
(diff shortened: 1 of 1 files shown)


grobie:
Thank you for your contribution. For the DCO, please have a look at the command-line steps explained here https://github.com/prometheus/docs/pull/1784/checks. Your commits are required to include a line "Signed-off-by: [Name] <[Email]>" in the commit message, see https://github.com/prometheus/docs/commit/d8827bc26ca75ceab5d3f488c38373c6c6fa96e8 for an example.

GETandSELECT:
Thanks. It's now all green.

brian-brazil:
Can you explain what the goal of this PR is? Is something broken with the existing links? I'd rather nor add clutter here.

The goal of these links is primarily so users can subscribe/post.

GETandSELECT:
There is a growing population that try to avoid Google (see https://en.wikipedia.org/wiki/DeGoogle and https://en.wikipedia.org/wiki/Surveillance_capitalism).

It's a mirror, an alternative to Google Groups, at least for browsing/searching.

content/community.html
```diff
@@ -16,8 +16,8 @@ <h1>Community</h1>
     <p>
       <strong>User mailing lists:</strong>
       <ul>
-        <li><a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a> – low-traffic list for announcements like new releases.</li>
-        <li><a href="https://groups.google.com/forum/#!forum/prometheus-users">prometheus-users</a> – for discussions around Prometheus usage and community support. Announcements are <em>not</em> generally mirrored from <a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a>.</li>
+        <li><a href="https://groups.google.com/forum/#!forum/prometheus-announce">prometheus-announce</a> <a href="https://www.mail-archive.com/prometheus-announce@googlegroups.com/">(mirror)</a> – low-traffic list for announcements like new releases.</li>
```
grobie, line 19: nit: Please move the parentheses from the link to regular text, so that it becomes

```
... (<a href="...">mirror</a>) ...
```

grobie (approved):
I don't see adding a single word link to each mailing list paragraph crosses the threshold of "adding clutter". The wish to have a non-Google option to access the mailing list archive has been voiced by multiple community users and I'm in favor of supporting an independent access as long as there is no or minimal maintenance work for the Prometheus project.

GETandSELECT:
@grobie done. Thank you

SuperQ (approved):
LGTM

SuperQ:
Without Google, we wouldn't have Prometheus. But without mirroring, we don't have backup access to the lists. This is is a generally good practice.

PR #1785, opened 2020-11-05
Updated comparison of Prometheus vs Sensu Go

calebhailey:
Hi there @juliusv @brian-brazil  :wave: first time contributor to Prometheus here, with hopefully more contributions to come (we're using Prometheus a fair bit at Sensu, including standardizing on Prometheus for [instrumenting the Sensu platform](https://docs.sensu.io/sensu-go/latest/api/metrics/#get-sensu-metrics)). 

We have wanted to update the "Prometheus vs Sensu" content in the Prom docs for a _long time_, but never got around to it until now. We'd like to be a good citizen in the Prom community, so please let me know if there are any changes I ought to make to the content here and I'll be happy to adjust it accordingly. 

[#monitoringlove](https://twitter.com/hashtag/monitoringlove)

Signed-off-by: Caleb Hailey <caleb@sensu.io>


brian-brazil:
Thanks for your PR, I understand Sensu has changed a bit since this was first written so could probably do with some freshening up. A lot of the proposed text comes across as marketing copy, whereas this is intended to be an impartial high-level technical comparison. Can you try to adjust this to make it more like the others, and stick within the categories?

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,38 @@ environment, then Prometheus is a good choice.
 
 ## Prometheus vs. Sensu
 
-[Sensu](https://sensu.io) is a composable monitoring pipeline that can reuse existing Nagios checks.
+[Sensu](https://sensu.io) is an open source monitoring and observabliity pipeline with a commercial distribution for scaling, added security, and enterprise integrations. Sensu's monitoring-as-code solution and horizontally scalable event processing backend are designed for larger environments, with several active installations exceeding 50,000 nodes under management.
```
SuperQ, line 249:
```suggestion
[Sensu](https://sensu.io) is an open source monitoring and observabliity pipeline with a commercial distribution. 
```

Remove non-technical verbage.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,38 @@ environment, then Prometheus is a good choice.
...
+Sensu processes events via the [Sensu Observablity Pipeline](https://docs.sensu.io/sensu-go/latest/observability-pipeline/), where events are deduplicated/correlated (via [EventFilters](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-filter/)), transformed (via [Mutators](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-transform/)), and processed (via [Handlers](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-process/)). 
 
-There is also a [client socket](https://docs.sensu.io/sensu-core/latest/reference/clients/#what-is-the-sensu-client-socket) permitting ad-hoc check results to be pushed into Sensu. 
+Common use cases include: 
+
+- Eliminate data silos by integrating with existing systems of record, and data platforms like InfluxDB, ElasticSearch, Splunk
+- Consolidate monitoring tools with support for existing plugins & exporters (Nagios, StatsD, Telegraf, Prometheus)
+- Automate diagnosis & self-healing with built-in auto-remediation or integrations with Ansible Tower, RunDeck, and SaltStack
```
SuperQ, line 259:
```suggestion
```
This is out of scope for a technical comparison

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,38 @@ environment, then Prometheus is a good choice.
...
 
 ### Data model
 
-Sensu has the same rough data model as [Nagios](/docs/introduction/comparison/#prometheus-vs-nagios).
+Sensu primarily focuses on normalizing and processing observability data as [Events](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/), identified by an event name, entity (e.g. node, container, or service), and optional [key-value metadata](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metadata-attributes) called "labels" or "annotations". Events can also optionally contain one or more metric data points. Sensu Agents extract metrics from known observability formats at the edge (i.e. metric data is normalized prior to transmission to Sensu backends) including Prometheus Exposition Format, InfluxDB line protocal, Graphite plain text, OpenTSDB line protocol, and Nagios Perfdata. 
+
+Prometheus ["stores all data as time series", which are uniquely identified by "metric name and optional key-value pairs called labels.](https://prometheus.io/docs/concepts/data_model/).
```
SuperQ, line 265:
```suggestion
```

No need to mention Prometheus here.

SuperQ (changes requested):
A few suggestions.

juliusv:
Thanks from me as well! I agree with the general sentiment that it would be great to update the Sensu section (and others as well), but parts of the wording also seem a bit too much like marketing copy to me. Generally the suggestions by @SuperQ seem to go in the right direction.

calebhailey:
Thanks, @brian-brazil, @juliusv, and @SuperQ :+1: 

We'll take another stab (starting with @SuperQ's suggestions) and come back with a fresh revision. 

Cheers! 🍻

calebhailey:
@brian-brazil @juliusv @SuperQ we updated the content per your requests/recommendations. Let me know if this works. 

Cheers 🍻

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,30 @@ environment, then Prometheus is a good choice.
 
 ## Prometheus vs. Sensu
 
-[Sensu](https://sensu.io) is a composable monitoring pipeline that can reuse existing Nagios checks.
+[Sensu](https://sensu.io) is an open source monitoring and observabliity pipeline with a commercial distribution. It can reuse existing Nagios plugins for compatibility.
```
brian-brazil, line 249: just monitoring pipeline is enough.

The "for compatibility" doesn't add anything.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,30 @@ environment, then Prometheus is a good choice.
...
 
 ### Scope
 
-The same general scope differences as in the case of [Nagios](/docs/introduction/comparison/#prometheus-vs-nagios) apply here.
-
-There is also a [client socket](https://docs.sensu.io/sensu-core/latest/reference/clients/#what-is-the-sensu-client-socket) permitting ad-hoc check results to be pushed into Sensu. 
+Sensu processes events via the [Sensu Observablity Pipeline](https://docs.sensu.io/sensu-go/latest/observability-pipeline/), where events are deduplicated/correlated (via [EventFilters](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-filter/)), transformed (via [Mutators](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-transform/)), and processed (via [Handlers](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-process/)). Sensu also collects and routes metrics to a number of backend time-series databases such as Prometheus, InfluxDB, etc.
```
brian-brazil, line 253: This is implementation, not scope. This should be about at high level what the problem space is that it is addressing, without mentioning specific internals.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,30 @@ environment, then Prometheus is a good choice.
...
 
 ### Data model
 
-Sensu has the same rough data model as [Nagios](/docs/introduction/comparison/#prometheus-vs-nagios).
+Sensu primarily focuses on normalizing and processing observability data as [Events](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/), identified by an event name, entity (e.g. node, container, or service), and optional [key-value metadata](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metadata-attributes) called "labels" or "annotations". The Sensu agents can also extract and route metrics from common formats such as Prometheus Exposition Format, InfluxDB line protocal, Graphite plain text, OpenTSDB line protocol, StatsD, and Nagios Perfdata and rewrite them into any backend storage format (Elastic, Influx, Wavefront, Prometheus).
```
brian-brazil, line 257: Normalising, processing, and extraction are not data model so this can be trimmed down quite a bit.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,30 @@ environment, then Prometheus is a good choice.
...
 
 ### Storage
 
-Sensu uses Redis to persist monitoring data, including the Sensu client registry, check results, check execution history, and current event data.
+Sensu _caches_ event status information and real-time inventory data in an embedded database (etcd) or an external RDBMS (Postgres). All event and time series data is sent to one or more specialized datastores for long term storage and analysis (e.g. Prometheus, InfluxDB, Wavefront, TimescaleDB, Elasticsearch, Splunk).
```
brian-brazil, line 261: That smells like it might be more like storage than a cache to me. If the cache was flushed or down, would that affect results?

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,30 @@ environment, then Prometheus is a good choice.
...
 
 ### Architecture
 
-Sensu has a [number of components](https://docs.sensu.io/sensu-core/latest/overview/architecture/). It uses
-RabbitMQ as a transport, Redis for current state, and a separate server for
-processing and API access.
+Sensu is a primarily agent-based monitoring solution (agentless applications are also supported) ([reference](https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/deployment-architecture/)). The Sensu agents automatically connect to Sensu backends over an encrypted pub/sub transport (TLS-encrypted websockets), and handle automated registration/de-registration, and local service discovery. Sensu's use of a pub/sub transport allows it to securely work over complex network topologies including NATs, VPNs, and "air-gapped" environments (useful in IoT deployments).
```
brian-brazil, line 265: This section is to talk about clustering/HA, and right now only the very last sentence of this section covers that. So mention what the components are, and then that sentence up to the and.

The rest is not relevant.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,30 @@ environment, then Prometheus is a good choice.
...
 
-All components of a Sensu deployment (RabbitMQ, Redis, and Sensu Server/API) can be clustered for highly available and redundant configurations.
+The entire Sensu observabliity pipeline configuration can be defined as code (declarative JSON or YAML template files). All components of a Sensu deployment can be clustered for highly available and scalable configurations beyond 50,000 servers (or entities).
 
 ### Summary
-If you have an existing Nagios setup that you wish to scale as-is, or want to take advantage of the automatic registration feature of Sensu, then Sensu is a good choice.
 
-If you want to do whitebox monitoring, or have a very dynamic or cloud based environment, then Prometheus is a good choice.
+If you have an existing Nagios (or similar) monitoring solution in a heterogeneus ephemeral cloud environment (e.g. EC2, GCP, Azure, Kubernetes, etc), or if you wish to consolidate and integrate existing best-of-breed monitoring solutions, then you should consider Sensu with its automatic registration/de-registration and data transformation features.
```
brian-brazil, line 273: Is this not suitable for Nagios in homogeneous, persistent, or on-prem environments? I think the previous wording covers this part better, as you seem to be over-constraining where Sensu may make sense.

The 2nd half reads as marketing, can you propose something more concrete?

calebhailey:
Thanks for your feedback, @brian-brazil I hope you had a nice holiday and that your 2021 is off to an excellent start! 

All of your comments have been very helpful so far – much appreciated! I generally agree with your suggestions to trim our edits, and I think our most recent update will address most of that feedback. The only bit of constructive criticism I'm confused about is the comment regarding the "architecture" section. Even if we avoid anything that remotely resembles "marketing", I would expect that the most technically accurate way to describe Sensu's architecture is to list all of the moving parts and their high level roles/functions (tl;dr: backends provide api/scheduler; agents connect to the backend via pub/sub transport; backends can be clustered for HA and federated configurations are supported, as needed); but your comment suggests that the architecture subheading "is to talk about clustering/HA" and "the rest is not relevant". 

I'm assuming the non-relevant parts were not technical enough and smelled too much like marketing (a fair criticism), for example: _"Sensu is a primarily agent-based monitoring solution (agentless applications are also supported)"_?? I've rewritten this section to just list the components and their technical role/function. I hope this works, but I'll be happy to strip away everything that's not just about clustering/HA, as needed. Like I explained above, I'm not trying to work around your recommendations here, I just didn't understand this one bit of feedback. 

I also reworked the summary section in a similar style as the [InfluxDB section](https://prometheus.io/docs/introduction/comparison/#summary-0) to (hopefully) highlight the few similarities between Sensu and Prometheus, as well as the use cases (which are quite different IMO). I think this is a better framework for the summary, but let me know if you want me to change it. 

Cheers 🍻 #monitoringlove 

image

calebhailey:
@SuperQ @brian-brazil I don't want to presumptively press the "resolve conversation" buttons on your suggestions, but I believe all of them have been addressed and are ready for a fresh round of feedback.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,43 @@ environment, then Prometheus is a good choice.
 
 ## Prometheus vs. Sensu
 
-[Sensu](https://sensu.io) is a composable monitoring pipeline that can reuse existing Nagios checks.
+[Sensu](https://sensu.io) is an open source monitoring and observability pipeline with a commercial distribution which offers additional features for scalability. The entire Sensu observabliity pipeline configuration can be defined as code (declarative JSON or YAML template files). It can reuse existing Nagios plugins.
```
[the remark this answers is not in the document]
calebhailey: :+1: done

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,43 @@ environment, then Prometheus is a good choice.
...
 
 ### Data model
 
-Sensu has the same rough data model as [Nagios](/docs/introduction/comparison/#prometheus-vs-nagios).
+In Sensu, events (which can include service health and/or [metric](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metric-attributes) data) are identified by an [entity](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-entities/entities/) name (e.g. server, cloud compute instance, container, or service), an event name, and optional [key-value metadata](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metadata-attributes) called "labels" or "annotations".
```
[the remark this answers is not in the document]
calebhailey: Great point. I linked to the [`metrics`](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metric-attributes) specification documentation but didn't actually explain how Sensu represents metric [`points`](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#points-attributes). I'll update that now.

To be clear, because Sensu is an event-based system, the "data model" is oriented around events, some of which may optionally contain metrics. The updated description will explain this better: 

> In Sensu, events (which can include service health and/or [metric](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metric-attributes) data) are identified by an [entity](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-entities/entities/) name (e.g. server, cloud compute instance, container, or service), an event name, and optional [key-value metadata](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metadata-attributes) called "labels" or "annotations". The Sensu Event payload may include one or more metric [`points`](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#points-attributes), represented as a JSON object containing a `name`, `tags`, `timestamp`, and `value` (always represented as a float).

I won't add an example in the PR, but the Sensu event payload itself contains a number of numerical data type fields (not just strings), specifically including the above-mentioned `metrics` field; here's a pseudo event example: 

```json
{
  "check": {},
  "entity": {},
  "metrics": {
    "handlers": [
      "influxdb"
    ],
    "points": [
      {
        "name": "sensu-go-sandbox.curl_timings.time_total",
        "tags": [
          {
            "name": "region",
            "value": "us-west-1"
          }
        ],
        "timestamp": 1552506033,
        "value": 0.005
      },
      {
        "name": "sensu-go-sandbox.curl_timings.time_namelookup",
        "tags": [
          {
            "name": "region",
            "value": "us-west-1"
          }
        ],
        "timestamp": 1552506033,
        "value": 0.004
      }
    ]
  },
  "metadata": {
    "name": "example",
    "namespace": "example",
    "labels": {},
    "annotations": {}
  },
  "timestamp": 1552506033,
}
```

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,43 @@ environment, then Prometheus is a good choice.
...
+All components of a Sensu deployment can be clustered for high availability and improved event-processing throughput. Sensu also provides built-in support for federation (including RBAC and other API resource replication) which is recommended for deployments larger than 30,000 nodes (or entities) under management.
 
 ### Summary
-If you have an existing Nagios setup that you wish to scale as-is, or want to take advantage of the automatic registration feature of Sensu, then Sensu is a good choice.
 
-If you want to do whitebox monitoring, or have a very dynamic or cloud based environment, then Prometheus is a good choice.
+Sensu and Prometheus have a few capabilities in common, but they take very different approaches to monitoring. Both offer extensible discovery mechanisms for dynamic cloud-based environments and ephemeral compute platforms (though the underlying mechanisms are quite different). Both provide support for collecting multi-dimensional metrics via labels (and annotations). Both have extensive integrations (and Sensu natively supports collecting metrics from all Prometheus exporters). Where Sensu and Prometheus differ the most is in their use cases.
```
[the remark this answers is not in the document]
calebhailey: Haha, yes. I'm terrible with the parens. Such a bad habit! 🤦

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,43 @@ environment, then Prometheus is a good choice.
...
 
 ### Data model
 
-Sensu has the same rough data model as [Nagios](/docs/introduction/comparison/#prometheus-vs-nagios).
+In Sensu, events (which can include service health and/or [metric](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metric-attributes) data) are identified by an [entity](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-entities/entities/) name (e.g. server, cloud compute instance, container, or service), an event name, and optional [key-value metadata](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metadata-attributes) called "labels" or "annotations".
 
 ### Storage
 
-Sensu uses Redis to persist monitoring data, including the Sensu client registry, check results, check execution history, and current event data.
+Sensu stores current and recent event status information and real-time inventory data in an embedded database (etcd) or an external RDBMS (Postgres). All event and time series data is sent to one or more external data platforms for long term storage and analysis (e.g. Prometheus, InfluxDB, TimescaleDB, Elasticsearch, Splunk, Wavefront, etc).
```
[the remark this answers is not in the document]
calebhailey: Sensu itself is a pipeline, which is designed to forward data to external data platforms. Can we at least say that without name-dropping what those external data platforms are? Is it the reference to the third-party tools which is objectionable here?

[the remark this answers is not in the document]
brian-brazil: This section is about Sensu's own storage, not 3rd party integrations that it can output to.

content/docs/introduction/comparison.md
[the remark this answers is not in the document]
brian-brazil: The data model here isn't just about metrics. If the event payload is basically an arbitrary set of structured data, you can say that.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,43 @@ environment, then Prometheus is a good choice.
...
-processing and API access.
+Sensu is a primarily agent-based monitoring solution ([reference](https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/deployment-architecture/)). Sensu agents automatically connect to one or more Sensu backends over an encrypted pub/sub transport (TLS-encrypted websockets), and handle automated registration/de-registration of entities and local service discovery. Sensu [backends](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-schedule/backend/) provide the control plane, including the configuration [API](https://docs.sensu.io/sensu-go/latest/api/) and [scheduler](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-schedule/). 
 
-All components of a Sensu deployment (RabbitMQ, Redis, and Sensu Server/API) can be clustered for highly available and redundant configurations.
+All components of a Sensu deployment can be clustered for high availability and improved event-processing throughput. Sensu also provides built-in support for federation (including RBAC and other API resource replication) which is recommended for deployments larger than 30,000 nodes (or entities) under management.
```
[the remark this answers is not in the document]
calebhailey: Sensu's [federation](https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/use-federation/) is effectively a built-in support for clustering Sensu clusters; our customers use it specifically for improved performance in large scale deployments or geographically disparate environments, but I'll strike it if you still think it doesn't apply.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,43 @@ environment, then Prometheus is a good choice.
...
+
+Where Sensu is better: 
+
+- If you're collecting and processing hybrid observability data (including metrics and events)
+- If you're monitoring hybrid infrastructure (including on-prem servers, public/private cloud, and cloud-native/Kubernetes)
+- If you're consolidating mulitple monitoring tools and need support for metrics _and_ traditional monitoring Nagios-style plugins or check scripts
+- If you are already invested in one or more data platforms like InfluxDB, Elasticsearch, Splunk, TimescaleDB, Wavefront, etc (Sensu's primary function as an observability pipeline means it is designed to route data to the data platform(s) of your choosing)
```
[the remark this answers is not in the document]
calebhailey: To be frank, Sensu probably wouldn't exist if there weren't gaps in these tools. Our worldview (and one of our biggest challenges from a marketing perspective, quite frankly) is that Sensu exists to fill these gaps without introducing yet another data silo... thus the pipeline approach. But I like your suggestion to focus on the pipeline for events. 

How about merging the first and fourth bullets to say:

> If you're collecting and processing hybrid observability data (including metrics and events) and you are already invested in one or more data platforms like InfluxDB, Elasticsearch, Splunk, TimescaleDB, Wavefront, etc (Sensu's primary function as a pipeline provides a lot of flexibility to route observability data to the platform of your choosing)

??

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,43 @@ environment, then Prometheus is a good choice.
...
 
 ### Scope
 
-The same general scope differences as in the case of [Nagios](/docs/introduction/comparison/#prometheus-vs-nagios) apply here.
-
-There is also a [client socket](https://docs.sensu.io/sensu-core/latest/reference/clients/#what-is-the-sensu-client-socket) permitting ad-hoc check results to be pushed into Sensu. 
+As an observability _pipeline_, Sensu focuses on processing and alerting of observability data as a stream of [Events](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/). It handles deduplication, correlation, and other filtering logic and then processes events by sending alerts, creating and resolving tickets in incident management systems, storing event & telemetry data in any number of data platforms, and managing automated remediation workflows. Sensu also provides a lightweight [monitoring agent](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-schedule/agent/) (event producer) for monitoring of ephemeral cloud compute and distributed systems.
```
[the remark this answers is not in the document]
calebhailey: Good point RE: Alertmanager. :+1: Will trim this down a bit.

content/docs/introduction/comparison.md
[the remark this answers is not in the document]
calebhailey: Actually, I think I should just drop this bullet point and mention that both Sensu [and Prometheus](https://prometheus.io/docs/prometheus/latest/storage/#remote-storage-integrations) share the ability to send observability data to remote data platforms. The only real difference there is in the contents (metrics vs events).

content/docs/introduction/comparison.md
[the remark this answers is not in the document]
calebhailey: Removed. :+1:

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,43 @@ environment, then Prometheus is a good choice.
...
+
+Where Sensu is better: 
+
+- If you're collecting and processing hybrid observability data (including metrics and events)
+- If you're monitoring hybrid infrastructure (including on-prem servers, public/private cloud, and cloud-native/Kubernetes)
+- If you're consolidating mulitple monitoring tools and need support for metrics _and_ traditional monitoring Nagios-style plugins or check scripts
+- If you are already invested in one or more data platforms like InfluxDB, Elasticsearch, Splunk, TimescaleDB, Wavefront, etc (Sensu's primary function as an observability pipeline means it is designed to route data to the data platform(s) of your choosing)
+- More powerful alerting and event-processing platform
```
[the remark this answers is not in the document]
calebhailey: It's a fair question RE: alerts, and the differences RE: alerting are probably so nuanced that it's not worth mentioning here. There are far more similarities than differences when it comes to basic alerting (e.g. both offer "silencing", "grouping" is called "aggregation" in Sensu, and "inhibition" is called "dependency filtering" in Sensu), so I'll drop that mention. 

As an aside – Sensu Go does achieve a degree of "in-built alert generation" now by means of our Asset framework. In practice Sensu shares some concepts with the Kubernetes platform in that many of the "built-in" components (e.g. kube-dns) are packaged as containers (tarballs w/ metadata) and updated via K8s resource definitions which pull & run those containers on demand; in Sensu Go integrations are packaged as Assets (tarballs w/ metadata) and updated via Sensu resource definitions with pull & run those integrations on demand. If I understand how AlertManager works, I think the difference is that instead of having to ship new AlertManager releases to add new integrations or modify existing ones, we ship new Assets and update the resource definition templates so existing deployments can easily take advantage of new features. 

TL;DR - AlertManager offers more built-in integrations out of the box (a handful of hard-coded integrations, plus webhooks), but they still have to be configured. You need that same configuration in Sensu, but the Sensu config includes references to integrations that can be pulled and run on-demand; and OSS/third-party integrations can be added with the exact same UX.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,43 @@ environment, then Prometheus is a good choice.
...
 ### Summary
-If you have an existing Nagios setup that you wish to scale as-is, or want to take advantage of the automatic registration feature of Sensu, then Sensu is a good choice.
 
-If you want to do whitebox monitoring, or have a very dynamic or cloud based environment, then Prometheus is a good choice.
+Sensu and Prometheus have a few capabilities in common, but they take very different approaches to monitoring. Both offer extensible discovery mechanisms for dynamic cloud-based environments and ephemeral compute platforms (though the underlying mechanisms are quite different). Both provide support for collecting multi-dimensional metrics via labels (and annotations). Both have extensive integrations (and Sensu natively supports collecting metrics from all Prometheus exporters). Where Sensu and Prometheus differ the most is in their use cases.
+
+Where Sensu is better: 
+
+- If you're collecting and processing hybrid observability data (including metrics and events)
+- If you're monitoring hybrid infrastructure (including on-prem servers, public/private cloud, and cloud-native/Kubernetes)
```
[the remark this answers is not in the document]
calebhailey: Different means to a common end I suppose. I don't want to argue this point so I've dropped this from the summary section. Ultimately our goal with this PR is to more accurately tell Sensu's story while being respectful of the Prom project. 

As an aside... my personal thoughts on the matter are that Prometheus is a better experience in a _homogeneous_ K8s environment because the discovery mechanism and scraping "just work" in K8s. On paper you could argue that the same is true with Sensu – our approach to discovery is agent sidecars and/or daemonsets, and data collection "just works" because of the pub/sub transport – but in practice Prometheus is just more optimized for K8s, and visa versa since the K8s platform itself has standardized on Prom for exposing platform metrics. 

Prom does offer SD for basically all-the-platforms out of the box, but the configuration varies from platform-to-platform; Sensu SD works the same on every platform - just run the agent and it auto-registers endpoints. Both have their pros/cons. 

For collecting data, a pull/scrape-based collection workflow isn't quite as seamless outside of K8s as it is on K8s; I suspect many users leverage Prom federation to help here, but opening ports to scrap endpoints across hybrid cloud infrastructure is a complex effort. Conversely, Sensu data collection works the same on every platform thanks to the agent and pub/sub transport, easily traversing complex network topologies including NATs and VPNs. Again you could argue that both have their pros/cons.

TL;DR - if it's true that Prometheus (the most mature CNCF project outside of Kubernetes itself) is optimized for cloud-native/Kubernetes, then by the very nature of optimization it might not necessarily work as seamlessly in other platforms. Conversely, Sensu's more generic approach is not as optimized for Kubernetes, but it works more consistently across all platforms.

calebhailey:
@brian-brazil thanks again for the feedback. I have completed another round of changes in between my morning of meetings. I _think_ 🤞  I have made all of your recommended changes, less one. I updated the description of Sensu's federation feature as it is part of our clustering/HA story (it's literally a technique to cluster Sensu clusters), but if you still don't think it fits I'll go ahead and remove it. 

Let me know if I missed anything else.

Cheers 🍻

content/docs/introduction/comparison.md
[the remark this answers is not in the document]
brian-brazil: My point was around the "one", Sensu makes more sense if you've "more". So wording along the lines you suggest seems okay, if it's talking about having multiple outputs. It's not really needed to list them though.

content/docs/introduction/comparison.md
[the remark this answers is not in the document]
brian-brazil: Sounds like this is a terminology thing. Alerting for Prometheus is the bit in Prometheus. What you're talking about here is notifications, which we do in the Alertmanager - and it sounds like the functionality is basically the same so there's nothing to contrast there.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,38 @@ environment, then Prometheus is a good choice.
...
 
 ### Data model
 
-Sensu has the same rough data model as [Nagios](/docs/introduction/comparison/#prometheus-vs-nagios).
+Sensu [Events](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/) represent service health and/or [metrics](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metric-attributes) in a structured data format identified by an [entity](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-entities/entities/) name (e.g. server, cloud compute instance, container, or service), an event name, and optional [key-value metadata](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metadata-attributes) called "labels" or "annotations". The Sensu Event payload may include one or more metric [`points`](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#points-attributes), represented as a JSON object containing a `name`, `tags`, `timestamp`, and `value` (always a float).
```
brian-brazil, line 257: At the end, it should probably be mentioned that tags are key-value.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,38 @@ environment, then Prometheus is a good choice.
...
+Sensu stores current and recent event status information and real-time inventory data in an embedded database (etcd) or an external RDBMS (Postgres). 
 
 ### Architecture
 
-Sensu has a [number of components](https://docs.sensu.io/sensu-core/latest/overview/architecture/). It uses
-RabbitMQ as a transport, Redis for current state, and a separate server for
-processing and API access.
-
-All components of a Sensu deployment (RabbitMQ, Redis, and Sensu Server/API) can be clustered for highly available and redundant configurations.
+All components of a Sensu deployment can be clustered for high availability and improved event-processing throughput. Sensu also provides built-in support for [cluster federation](https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/use-federation/#configure-federation), including RBAC and other [API resource replication](https://docs.sensu.io/sensu-go/latest/operations/deploy-sensu/etcdreplicators/), which is recommended for deployments larger than 30,000 nodes under management.
```
brian-brazil, line 265: This federation doesn't seem relevant here, as it's not about scaling the processing abilities of the system, so should be removed. Specific operational advice (30k nodes) also shouldn't be here, and neither should configuration deployment.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,38 @@ environment, then Prometheus is a good choice.
...
 
 ### Summary
-If you have an existing Nagios setup that you wish to scale as-is, or want to take advantage of the automatic registration feature of Sensu, then Sensu is a good choice.
 
-If you want to do whitebox monitoring, or have a very dynamic or cloud based environment, then Prometheus is a good choice.
+Sensu and Prometheus have a few capabilities in common, but they take very different approaches to monitoring. Both offer extensible discovery mechanisms for dynamic cloud-based environments and ephemeral compute platforms, though the underlying mechanisms are quite different. Both provide support for collecting multi-dimensional metrics via labels and annotations. Both have extensive integrations, and Sensu natively supports collecting metrics from all Prometheus exporters. Both are capable of forwarding observability data to third-party data platforms (e.g. InfluxDB, Elasticsearch, Splunk). Where Sensu and Prometheus differ the most is in their use cases.
```
brian-brazil, line 269: I'm not sure it make sense to mention specific 3rd party things to forward to, as there's not going to be much overlap between ones where it'd make sense to send both types of data to I suspect.

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,38 @@ environment, then Prometheus is a good choice.
...
+
+Where Sensu is better: 
+
+- If you're collecting and processing hybrid observability data (including metrics _and_ events)
+- If you're consolidating mulitple monitoring tools and need support for metrics _and_ traditional monitoring (e.g. Nagios-style plugins or check scripts)
```
brian-brazil, line 274: "traditional monitoring" could be a bit loaded, maybe just keep it to what's in the parens as that's what we're actually talking about?

content/docs/introduction/comparison.md
```diff
@@ -246,31 +246,38 @@ environment, then Prometheus is a good choice.
...
 
 ### Summary
-If you have an existing Nagios setup that you wish to scale as-is, or want to take advantage of the automatic registration feature of Sensu, then Sensu is a good choice.
 
-If you want to do whitebox monitoring, or have a very dynamic or cloud based environment, then Prometheus is a good choice.
+Sensu and Prometheus have a few capabilities in common, but they take very different approaches to monitoring. Both offer extensible discovery mechanisms for dynamic cloud-based environments and ephemeral compute platforms, though the underlying mechanisms are quite different. Both provide support for collecting multi-dimensional metrics via labels and annotations. Both have extensive integrations, and Sensu natively supports collecting metrics from all Prometheus exporters. Both are capable of forwarding observability data to third-party data platforms (e.g. InfluxDB, Elasticsearch, Splunk). Where Sensu and Prometheus differ the most is in their use cases.
+
+Where Sensu is better: 
+
+- If you're collecting and processing hybrid observability data (including metrics _and_ events)
```
brian-brazil, line 273: Hmm, wouldn't Sensu also make sense if you're only doing events?

brian-brazil (commented):
I think we're getting pretty close now.

content/docs/introduction/comparison.md
calebhailey: how about a generic description: "forwarding data to third-party data platforms (e.g. other event stores or TSDBs)". Neither platform is limited in terms of the different types of data platforms that data can be forwarded to, but a few categorical examples might help the reader to understand the common use cases.

content/docs/introduction/comparison.md
calebhailey: Yes. I'll change this to an "and/or"

calebhailey:
Thanks again, @brian-brazil. Sorry for the slow follow-up here. You were being very responsive last week, but then I got derailed by the political news last week... 

Cheers 🍻

brian-brazil:
Thanks!

> You were being very responsive last week, but then I got derailed by the political news...

I think we're all still a bit off-kilter from that even on this side of the ocean, no need to apologise.

calebhailey:
🎉

PR #1786, opened 2020-11-06
Add GPFS exporter

treydock:
Adds link to GPFS exporter.  This filesystem is now called IBM Spectrum Scale though the actual filesystem type is GPFS, and is commonly found at HPC centers.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -108,6 +108,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Ceph exporter](https://github.com/digitalocean/ceph_exporter)
    * [Ceph RADOSGW exporter](https://github.com/blemmenes/radosgw_usage_exporter)
    * [Gluster exporter](https://github.com/ofesseler/gluster_exporter)
+   * [GPFS exporter](https://github.com/treydock/gpfs_exporter)
    * [Hadoop HDFS FSImage exporter](https://github.com/marcelmay/hadoop-hdfs-fsimage-exporter)
    * [Lustre exporter](https://github.com/HewlettPackard/lustre_exporter)
    * [ScaleIO exporter](https://github.com/syepes/sio2prom)


brian-brazil:
This looks to be collecting filesystem metrics, is there a way to get these from the kernel without privileged access so it could be added to the node exporter?
mmdf appears to be directly duplicating node exporter functionality.

Some of the same comments for the other exporters also apply here.

treydock:
The `df` output from a GPFS filesystem is not guaranteed to be the same as `mmdf` for block storage and it's often the case that `mmdf` is more accurate.  Also the metadata space available on GPFS is only accessible via `mmdf`.  The inodes from `df` don't tell you how much metadata space is still available and if metadata space is exhausted you've bricked your filesystem, so getting that information is really important.

The metrics from `mmpmon` which is like read/write performance of a client is not exposed via sysfs or procfs or anything that would be possible to scrape with node exporter.  Unfortunately the only way I know to get performance metrics out of GPFS for the client is to execute `mmpmon`.  Other tools like PCP (Performance Co Pilot) use things like `mmpmon`, doesn't necessarily mean it's greatest solution but as far as I'm aware it's the only solution for getting this information.

The rest of the metrics are mostly health or status and per my other comments those items were cached because it's not entirely uncommon for GPFS collection to have hiccups if the system is busy and I don't want an alert to be falsely resolved because for 2 minutes there was no data due to an issue collecting the information.  We ran into a lot of alertmanager spam when filesystem was busy and some of the storage nodes would have gaps in data and constantly spam us with alert / resolve / alert / resolve simply because there was gaps in data and not actual changes in state.  If there are other ways to solve that problem that don't involve the use of `absent()` then I'd love to know about them but thus far I've not come across a solution.  

Also the caching in this exporter and my other exporters is totally opt-in.  If a user does not pass the necessary flags, the caching is bypassed and I have tests for that.

treydock:
@brian-brazil I've made what I hope to be the necessary adjustments to the GPFS exporter.  I removed all caching logic and the caching CLI argument.  I also renamed several metrics to match some of the documented naming conventions like changing `gpfs_perf_read_bytes` (counter) to `gpfs_perf_read_bytes_total` and shuffling a few to put units last like `gpfs_fs_inodes_allocated` to become `gpfs_fs_allocated_inodes`.  These changes are only on master, I have not tagged these changes.

Let me know if more adjustments are needed on this particular exporter.

brian-brazil:
In mmpmon, what is the NodeName? That smells like it might be a target label.

In mmhealth, the metric looks to be mixing labels and values. Similarly in mmces.

In mmfd avoid "total" in metric names, that's implied. For example total_bytes doesn't tell me anything, size_bytes would be more informative. Percentages should also be exposed as ratios - though in the case of both metrics here you should drop them as they are redundant. Last execution could be removed, the node exporter's mtime metric should cover you there.

treydock:
> In mmpmon, what is the NodeName? That smells like it might be a target label.

So the NodeName is the name used to add a node to the GPFS cluster but doesn't necessarily match the name used by Promemtheus.  So for example we have majority of our nodes with a name like `o0001.ten.osc.edu` inside Prometheus for the target but the node name in GPFS is `ib-o0001`.  The host hostname with `ib-` prefix is not used by all our systems, but most of them.  Some systems main hostname is used , just depends on the networking of the system and how they are setup.  Also the NodeName is what one would see in GPFS commands and logs so having that be 100% exactly what GPFS reports is very useful, and not relying on relabeling that could easily be screwed up.

> 
> In mmhealth, the metric looks to be mixing labels and values. Similarly in mmces.

I'm not sure I follow.  The mmhealth and mmces metrics for most part are state metrics and I do not know all the possible state values, the man pages for those commands do not describe all possible states.  I know some common ones I've seen before but that's by no means an exhaustive list.  I know for example that mmces states can be `DEGRADED` but that state string isn't even present in the man page.

The online documentation is the same as man pages:

https://www.ibm.com/support/knowledgecenter/STXKQY_5.0.5/com.ibm.spectrum.scale.v5r05.doc/bl1adm_mmhealth.htm
https://www.ibm.com/support/knowledgecenter/STXKQY_5.0.5/com.ibm.spectrum.scale.v5r05.doc/bl1adm_mmces.htm

> 
> In mmfd avoid "total" in metric names, that's implied. For example total_bytes doesn't tell me anything, size_bytes would be more informative. Percentages should also be exposed as ratios - though in the case of both metrics here you should drop them as they are redundant. Last execution could be removed, the node exporter's mtime metric should cover you there.

I'll make the adjustments.

treydock:
> though in the case of both metrics here you should drop them as they are redundant.

What metrics are redundant? The only percentages were gpfs_fs_metadata_free_percent and gpfs_fs_free_percent which are two very different things.  Or is the redundant because there are used/free/total metrics?

treydock:
> Last execution could be removed, the node exporter's mtime metric should cover you there.

I would like to remove this but the logic in the exporter adds the label `collector="mmdf-$fsname"` where `$fsname` is either passed as argument or discovered if not specified.  So we have one system where it queries 2 different GPFS filesystems via same cron execution.  I do not know of a way to decouple that because I'm using `prometheus.WriteToTextfile` and so one one of my hosts that textfile has the data for 2 filesystems and it's useful to be able to know which mmdf is stale vs assuming all of them are stale.

brian-brazil:
> So the NodeName is the name used to add a node to the GPFS cluster but doesn't necessarily match the name used by Promemtheus.

That's a target label then. If you want to expose it it should be as an Info metric, not as an instrumentation label across all metrics. It's not the place of an exporter to decide what target labels are applied.

> I'm not sure I follow.

The `status` label means that the series appears and disappears, and the value also changes from 0 to 1. That's mixing different ways of exposing things in one metric, which would thus be quite challenging to use. A single metric should be using only one of those.

> The only percentages were gpfs_fs_metadata_free_percent and gpfs_fs_free_percent which are two very different things. Or is the redundant because there are used/free/total metrics?

Exactly, you can calculate them from the gauges in PromQL.

treydock:
I am removing the `nodename` label and replacing with `gpfs_perf_info` metric that like this:

```
gpfs_perf_info{fs="project",nodename="ib-pitzer-rw02.ten"} 1
```

For the status metrics I made it so that the value is always 1 and the label for `status` is only thing that changes.

I removed the percent metrics.

brian-brazil:
Thanks!

brian-brazil:
Thanks!

PR #1787, opened 2020-11-06
Add TSM exporter

treydock:
Adds link to TSM exporter (Tivoli Storage Manager) which is also called IBM Spectrum Protect.  This is a common enterprise backup and archival solution.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -111,6 +111,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Hadoop HDFS FSImage exporter](https://github.com/marcelmay/hadoop-hdfs-fsimage-exporter)
    * [Lustre exporter](https://github.com/HewlettPackard/lustre_exporter)
    * [ScaleIO exporter](https://github.com/syepes/sio2prom)
+   * [Tivoli Storage Manager/IBM Spectrum Protect exporter](https://github.com/treydock/tsm_exporter)
 
 ### HTTP
    * [Apache exporter](https://github.com/Lusitaniae/apache_exporter)


brian-brazil:
Most of my comments on improving the Netapp exporter also apply here.

This exporter is putting passwords on the commandline, this is insecure - please fix this. If there's a way to avoid shelling out that'd be even better.

The timeouts here also seem more complicated than they need to be, a single overall timeout should cover things.

The mix of literal values and regexes in the config to limit returned values is a little inconsistent, have you considered only having regexes?

For libvolume_media, how many statuses are these and will 0s be returned? libvolume_scratch also seems redundant, as that information is already in libvolume_media.

Usually boolean metrics are positive, so I'd call things completed rather than not_completed to avoid double negatives.

`name` is a bit generic as a label name in this context, where these devices seem to have at least four different types of names. Be explicit and use drive_name etc. (or just drive) to avoid confusion.

content/docs/instrumenting/exporters.md
```diff
@@ -111,6 +111,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Hadoop HDFS FSImage exporter](https://github.com/marcelmay/hadoop-hdfs-fsimage-exporter)
    * [Lustre exporter](https://github.com/HewlettPackard/lustre_exporter)
    * [ScaleIO exporter](https://github.com/syepes/sio2prom)
+   * [TSM exporter](https://github.com/treydock/tsm_exporter)
```
brian-brazil, line 114: I don't know what TSM is, spell it out.

treydock:
> Most of my comments on improving the Netapp exporter also apply here.

So this exporter appears to be a bit more aggressive with caching than the others where others were just state. I think I tried to only cache things in this exporter I was alerting on, and again per other comments the cache is opt-in and something done out of necessity to avoid alert spam.

One thing I forgot to mention about caching in other comments was the  `collect_timeout` and `collect_error` metrics offer a way to have insight when caching is used. The caching is only used on timeout or error occurs and so if you alert on errors or timeouts you will know if you enabled caching your getting stale data.  Personally I much prefer that over a possible alert spam.

> 
> This exporter is putting passwords on the commandline, this is insecure - please fix this. If there's a way to avoid shelling out that'd be even better.

I will have to see if maybe there is an environment variable I can use of something.  Unfortunately the only way to run SQL queries against TSM is dsmadmc command, but the point about not exposing the password is something I will have to investigate.

> 
> The timeouts here also seem more complicated than they need to be, a single overall timeout should cover things.
> 

I opted for multiple timeouts because some of these SQL queries can run longer than others so I felt it would be worth allowing someone to tune when the exporter times out one query vs another query.  Does that seem valid?

> The mix of literal values and regexes in the config to limit returned values is a little inconsistent, have you considered only having regexes?


> 
> For libvolume_media, how many statuses are these and will 0s be returned? libvolume_scratch also seems redundant, as that information is already in libvolume_media.

Looks like the status is either `private` or `scratch`.  And I think I see your point about duplication as I could easily do this:

```
sum(tsm_libvolume_media{status="scratch"})
```

Would this be better to move towards maybe `tsm_libvolume_private_count` and `tsm_libvolume_scratch_count` or just keep `tsm_libvolume_media` and drop the `_scratch` metric?

> 
> Usually boolean metrics are positive, so I'd call things completed rather than not_completed to avoid double negatives.
> 

So this is not a boolean, the `tsm_replication_not_completed` is a count that can be 0+ and not necessarily true/false.  If replication is delayed for several days and the replication is daily the count can be like 2 or 3 or even higher I believe.

> `name` is a bit generic as a label name in this context, where these devices seem to have at least four different types of names. Be explicit and use drive_name etc. (or just drive) to avoid confusion.

I'll make the necessary adjustments to avoid generic `name`.

content/docs/instrumenting/exporters.md
treydock: So should I call it the official product name that was a result of recent re-branding by IBM, which is `IBM Spectrum Protect` or the more common name used by people using the product that is `Tivoli Storage Manager`?  The product itself didn't change when named changed, was just a rebranding.

brian-brazil: What would a user looking for this call it? If both are plausible, both can be listed.

brian-brazil:
> One thing I forgot to mention about caching in other comments was the collect_timeout and collect_error metrics offer a way to have insight when caching is used. The caching is only used on timeout or error occurs and so if you alert on errors or timeouts you will know if you enabled caching your getting stale data. Personally I much prefer that over a possible alert spam.

That's pushing a lot of complexity, additional alerts, and thus gotchas, on users for them to use the exporter correctly. It's much better to write robust alerts and fail the scrape when there's any hiccups, as there's only one failure mode users have to handle: `up == 0`.

> Does that seem valid?

Considering the whole scrape should be failing when it happens it seems like overcomplicating things to me, but it is a valid tradeoff.

> Would this be better to move towards maybe tsm_libvolume_private_count and tsm_libvolume_scratch_count or just keep tsm_libvolume_media and drop the _scratch metric?

With only two, I can't say as I lack domain knowledge. What are these volumes? Would you ever want to sum both the scratch and private together? Might more types be added?

> So this is not a boolean, the tsm_replication_not_completed is a count that can be 0+ and not necessarily true/false.

Is it a counter then? This sounds like an unusual metric, it might be best to coerce it to 0/1.

treydock:
> 
> That's pushing a lot of complexity, additional alerts, and thus gotchas, on users for them to use the exporter correctly. It's much better to write robust alerts and fail the scrape when there's any hiccups, as there's only one failure mode users have to handle: `up == 0`.

I would try this but I am not aware of how to make a scrape fail at the exporter level.  The interface for collectors don't lend themselves to return errors and thus fail the HTTP scrape request.  I would really like to keep these timeout/error metrics as I already have to check for errors/timeouts to know if there is data to parse so doesn't add a lot of code complexity to convert those existing variables into metrics.  I think I would save like 5 lines of code per collector to remove these metrics, so I don't see as a real benefit in the removal.

> 
> > Would this be better to move towards maybe tsm_libvolume_private_count and tsm_libvolume_scratch_count or just keep tsm_libvolume_media and drop the _scratch metric?
> 
> With only two, I can't say as I lack domain knowledge. What are these volumes? Would you ever want to sum both the scratch and private together? Might more types be added?

According to docs for TSM these volumes are either scratch or private.  These volumes are actually physical tape library tapes I believe.  I'm not a TSM expert, much of the queries I am using I pulled from code a coworker wrote ages ago to get information out of TSM. 

> 
> > So this is not a boolean, the tsm_replication_not_completed is a count that can be 0+ and not necessarily true/false.
> 
> Is it a counter then? This sounds like an unusual metric, it might be best to coerce it to 0/1.

This is not a counter.  It is the number of replication tasks that have not completed in the last 24 hours.  This value can be a problem if >0 or in some cases we have some that are a problem if >1 , it all depends on each replication tasks' configured frequency I believe.

content/docs/instrumenting/exporters.md
treydock: I went ahead and listed both since the rename is somewhat recent and so people might be looking for either old or new name.

treydock:
Looks like there is no way to pass the password for the command via things like environment variables or stdin.  I think the only potion and the only thing IBM documents is passing the password at command line so I think what this exporter is doing at this time is all that is possible with the application without writing like a wrapper script.

treydock:
So far I've modified exporter to remove metric caching and replace `name` labels with more specific labels.  The caching that remains is actually necessary to get the metrics due to how TSM database exposes data about scheduled things like events and replication and is done to make comparisons between completed events, it's not actually caching the final metrics.

treydock:
Also I found a good example of an exporter that has metrics about exporter errors and that's the node exporter with the `node_scrape_collector_success`, I just did the inverse of `collector_error` but same idea.  I may have borrowed the code from there, I know the collector code this and others has around defining collectors was borrowed from node exporter.

brian-brazil:
> I would try this but I am not aware of how to make a scrape fail at the exporter level.

https://www.robustperception.io/failing-a-scrape-with-the-prometheus-go-client

> I think I would save like 5 lines of code per collector to remove these metrics, so I don't see as a real benefit in the removal.

The issue isn't the code complexity, it's the complexity it forces on users of the exporter.

> This is not a counter. It is the number of replication tasks that have not completed in the last 24 hours.

That would be good to put into the help then.

> Looks like there is no way to pass the password for the command via things like environment variables or stdin.

That would be a first, can you point me at the relevant docs?

> The caching that remains is actually necessary

I'm confused, caching should never be necessary for correctness.

> Also I found a good example of an exporter that has metrics about exporter errors and that's the node exporter 

The node exporter is not a good base to follow when writing exporters, as it's far more complex than a typical exporter needs to be - and accordingly makes things quite complex for users.

content/docs/instrumenting/exporters.md
hoffie: > What would a user looking for this call it? If both are plausible, both can be listed.

Some input from a TSM user: We usually call it "TSM" despite knowing that it has been renamed. Filesystem paths still refer to the old name, while license stuff uses the new name. Users will recognize both, but in order to make it simpler to find such an exporter, it would make sense to list both names in my opinion.

treydock:
> > I would try this but I am not aware of how to make a scrape fail at the exporter level.
> 
> https://www.robustperception.io/failing-a-scrape-with-the-prometheus-go-client
> 
> > I think I would save like 5 lines of code per collector to remove these metrics, so I don't see as a real benefit in the removal.
> 
> The issue isn't the code complexity, it's the complexity it forces on users of the exporter.
> 

The rational behind doing that failed scrape vs my already coded error metrics is it's assumed there are already `up` alerts.  What if I document clearly the alert that can be added to make use of the error and timeout metrics?  We have found these extremely useful at our site to have a quick sense of what is failing vs a generic `up == 0` tells us very little about what actually broke.  I feel like the argument of extra complexity can be worked around with proper documentation on my part, which I'm fine doing, I just know from experience how useful these error and timeout alerts have been for our large install.

> > This is not a counter. It is the number of replication tasks that have not completed in the last 24 hours.
> 
> That would be good to put into the help then.

The help already has this for both metrics where it's relevant:

```
# HELP tsm_schedule_not_completed Number of scheduled events not completed for today
# TYPE tsm_schedule_not_completed gauge
tsm_schedule_not_completed{schedule="BCPDB2_DAILY"} 0
```

> 
> 
> That would be a first, can you point me at the relevant docs?

Available environment variables:

https://www.ibm.com/support/knowledgecenter/en/SSEQVQ_8.1.4/client/c_cfg_spevunix.html

Docs for command executed:

https://www.ibm.com/support/knowledgecenter/SSEQVQ_8.1.2/srv.reference/r_cmdline_adclient_options.html

Error if I omit `-ID` and `-PAssword`:

```
bash-4.2$ /usr/bin/dsmadmc -servername=OMIT  "SELECT START_TIME,END_TIME,NODE_NAME,FSNAME,COMP_STATE,TOTFILES_REPLICATED,TOTBYTES_REPLICATED FROM replicationview"
ANS8019E Id and password must be supplied in Batch mode.
```

I think the best I could do is a wrapper script but that seems ugly.  I think documenting security of the host running this exporter might be another solution.

> I'm confused, caching should never be necessary for correctness.

It has to do with the query executed that might not show results for today's execution if run at the wrong time, it's just a quirk with the query being executed: `SELECT schedule_name,status,actual_start,completed FROM events WHERE DATE(scheduled_start) BETWEEN '2020-11-06' AND '2020-11-07'`.  If some events haven't been initiated recently, like if this gets executed in the morning before events have triggered, then no data would return and the metrics are not designed to be current values, the metric is the duration of last scrape so the metric is by design going to be stale after the last completed event.  If you have an event happen every 24 hours it doesn't make sense to have 1 point on a graph for that one event and then gaps, because the metric is not the current runtime of events but the duration of the last completed event for things that are spaced far apart.  This is not the same caching that was previously removed and I feel like caching is a bad term, it's the nature of the metric that past values must be stored.  The query has to be scoped by a time frame to avoid a massive dump of data going back to whenever TSM purges event data.

Maybe the metric names need to be renamed to make it clear the value is last occurrence and not some instant metric.  These metrics in their current form are doing exactly what they need to do and taking away the saving of past data would make these metrics no longer useful. I realize the immediate reaction to what appears like metric caching but these metrics have to store past values or else they can't even exist in a meaningful way, which might just be something to make clearer to people using this tool via help docs , metric name or README updates because from working with my local TSM expert, these queries are behaving exactly as designed.

content/docs/instrumenting/exporters.md
```diff
@@ -109,8 +109,10 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Ceph RADOSGW exporter](https://github.com/blemmenes/radosgw_usage_exporter)
    * [Gluster exporter](https://github.com/ofesseler/gluster_exporter)
    * [Hadoop HDFS FSImage exporter](https://github.com/marcelmay/hadoop-hdfs-fsimage-exporter)
+   * [IBM Spectrum Protect exporter](https://github.com/treydock/tsm_exporter)
```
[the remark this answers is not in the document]
treydock: How would that look exactly?  Do like `IBM Spectrum Protect/Tivoli Storage Manager exporter`?  Not sure  I follow how to put both in a single link and make it presentable.

brian-brazil:
> The rational behind doing that failed scrape vs my already coded error metrics is it's assumed there are already up alerts.

The problem isn't that there are already up alerts - the problem is that up alerts are no longer sufficient. This causes continual confusion with the blackbox exporter for example, so you have to be really really sure that having these metrics is vital and failure normal. Are you aware of scrape_duration_seconds?

> I think the best I could do is a wrapper script but that seems ugly.

My reading of the docs is that if it's not provided on the command line, you can pass them in on stdin instead. Have you tried passing just ID? Per https://www.ibm.com/support/knowledgecenter/en/SSEQVQ_8.1.4/client/r_opt_password.htmlit it can also go in the options file. Failing that, a pseudo terminal should do the trick.

> I think documenting security of the host running this exporter might be another solution.

I'm hoping to avoid that, it's a last resort that has not been needed thus far.

> I feel like caching is a bad term, it's the nature of the metric that past values must be stored.

It's not caching, it's depending on in-exporter state for correctness and thus wrong - as you don't produce the desired result on a clean start. It's also a memory leak.

> the metric is the duration of last scrape so the metric is by design going to be stale after the last completed event.

That's not right, the last completion is always the value of the last completion. That that may have been weeks ago doesn't make it stale, only the schedule being deleted could do that.

Is this full SQL? If so you should be able to get what you want with something like `SELECT schedule_name, max(completed), max(actual_start) FROM events WHERE status = "COMPLETED" GROUP BY schedule_name` presuming a given schedule can't run concurrently.

content/docs/instrumenting/exporters.md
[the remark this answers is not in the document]
brian-brazil: Something like that would do it. I'd put whichever it's most known as first.

treydock:
> > The rational behind doing that failed scrape vs my already coded error metrics is it's assumed there are already up alerts.
> 
> The problem isn't that there are already up alerts - the problem is that up alerts are no longer sufficient. This causes continual confusion with the blackbox exporter for example, so you have to be really really sure that having these metrics is vital and failure normal. Are you aware of scrape_duration_seconds?

I have a similar metric in most my exporters. The issue with failing with 500 on any error is most these exporters of mine are like node exporter in they use many collectors and one collector failure can not and should not invalidate other collectors that were successful. For a simple exporter with like 1 command executed to collect data, a failed scrape makes sense. I feel like failing entire scrape for one collector error when others don't fail is a bad design choice which again is likely why I feel like this aligns more with node exporter.

> 

> My reading of the docs is that if it's not provided on the command line, you can pass them in on stdin instead. Have you tried passing just ID? Per https://www.ibm.com/support/knowledgecenter/en/SSEQVQ_8.1.4/client/r_opt_password.htmlit it can also go in the options file. Failing that, a pseudo terminal should do the trick.
> 
> I'm hoping to avoid that, it's a last resort that has not been needed thus far.
> 

I'll try stdin and see how that works. Not sure what is meant by pseudo terminal or how to implement that in Go.

> It's not caching, it's depending on in-exporter state for correctness and thus wrong - as you don't produce the desired result on a clean start. It's also a memory leak.
> 

I agree this is a bad approach but best I can do to get desired information.  I will play with code some to see if unavoidable but these metrics serve a mission critical alert for us and making them less useful is not an option for me.

> > the metric is the duration of last scrape so the metric is by design going to be stale after the last completed event.
> 
> That's not right, the last completion is always the value of the last completion. That that may have been weeks ago doesn't make it stale, only the schedule being deleted could do that.
> 
> Is this full SQL? If so you should be able to get what you want with something like `SELECT schedule_name, max(completed), max(actual_start) FROM events WHERE status = "COMPLETED" GROUP BY schedule_name` presuming a given schedule can't run concurrently.

I will experiment with different queries to see if avoid the state being saved. If I can't I know event collector can have duration metric removed as we don't actively use it but absolutely need not_completed metrics.

treydock:
Upon further inspection I think dsmadmc somehow sanitizing the process output because in one shell I ran the exporter in background and then in another shell queried and get this in ps output:

```
tsm_exp+  77891  0.1  0.0 713884 11704 ?        Ssl  10:37   0:00 /var/tmp/tsm_exporter --config.file=/etc/tsm_exporter.yaml --path.dsm_log.dir=/tmp/tsm_exporter --web.listen-address=:9310 --collector.db.timeout=20 --collector.drives.timeout=20 --collector.events.timeout=20 --collector.libvolumes.timeout=20 --collector.log.timeout=20 --collector.replicationview.timeout=20 --collector.volumes.timeout=20
tsm_exp+  77903 20.0  0.0 194292 16944 ?        Sl   10:37   0:00  \_ dsmadmc -SERVERName=sp02.OMIT -ID         -PAssword              -DATAONLY=YES -COMMAdelimited SELECT FREE_SPACE_MB,TOTAL_SPACE_MB,USED_SPACE_MB FROM log
```

Both the ID and Password values were stripped. I think something like dsmadmc is stripping sensitive information. Not entirely sure how that's possible but based on my tests both the ID and Password are not exposed in process output so I think there is no need to change the code WRT hiding the password.

I could not get stdin or options file to work.  However I think I've found another way if necessary.  Instead of executing the direct command instead I found I could set an environment variable like `DSM_PASSWORD` before executing a command with Go os.Exec like this:

```
	cmd := execCommand(ctx, "/bin/bash", "-c", fmt.Sprintf("dsmadmc %s %s -PAssword=$DSM_PASSWORD -DATAONLY=YES -COMMAdelimited \"%s\"", servername, id, query))
```

Without using `/bin/bash -c` to run the command, the environment variable would not get expanded.  But I don't think this will be necessary per testing what ps output actually shows.

content/docs/instrumenting/exporters.md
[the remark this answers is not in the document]
treydock: I updated the entry in markdown to be just 1 item with both names.

brian-brazil:
> Upon further inspection I think dsmadmc somehow sanitizing the process output

That's something that some things to, however changing argv after the process starts doesn't fix the security vulnerability as the password was still visible in argv to everyone on the machine.

treydock:
> > Upon further inspection I think dsmadmc somehow sanitizing the process output
> 
> That's something that some things to, however changing argv after the process starts doesn't fix the security vulnerability as the password was still visible in argv to everyone on the machine.

I'm not sure I have any other option.  Both STDIN and option file attempts have failed, the only way this particular command seems to work when the query is the last argument is passing password at command line. How would one even access argv if it's sanitized?  I don't see the password in /proc status either:

```
# cat /proc/$(pgrep dsmadmc | tail -n 1)/cmdline
dsmadmc-SERVERName=sp02.OMIT-ID        -PAssword             -DATAONLY=YES-COMMAdelimitedSELECT COMP_STATE,END_TIME,FSNAME,NODE_NAME,START_TIME,TOTBYTES_REPLICATED,TOTFILES_REPLICATED FROM replicationview WHERE DATE(START_TIME) BETWEEN '2020-11-07' AND '2020-11-08' order by START_TIME
```

If there is some way for me to check if some argv exploit is even possible, I'd like to know how to test that, because it's looking like the current method of executing this command is the only option with this particular executable without writing complicated expect logic into the code.

brian-brazil:
> How would one even access argv if it's sanitized?

It reads the value, and then changes it. This means it's now a race for an attacker, but not a particularly hard one to win - particularly if the command is being run on a very regular schedule as is the case here.

Passwords are not safe to have on the command line.

It sounds like a pseudo terminal is needed, https://github.com/creack/pty looks like one option.

treydock:
> It sounds like a pseudo terminal is needed, https://github.com/creack/pty looks like one option.

I'm not sure I follow how a pty code could help.  Is that sort of like an expect type application?  I couldn't tell from the README how this would help, but maybe I'm missing a fundamental feature of PTYs.  Currently I am trying to get one of the many expect Go modules to work, but if PTY offers some simpler solution that would be great but just not sure how PTY module would help.

brian-brazil:
> Is that sort of like an expect type application?

Yes, a pty is how expect "works" as it looks like a terminal.

treydock:
So I've managed to get this kind of working with https://github.com/google/goexpect, BUT the TSM queries that return more than a few lines are timing out.  The actual queries run outside Go or with the batch mode and password in command take roughly 0.5 seconds.  I've opened an issue with goexpect on this: https://github.com/google/goexpect/issues/59.  I am not sure expect is going to work if there is just some issue with how the TSM queries work.  I've written a test expect program that SSHes to another host using expect and runs a shell script that dumps the output that would come from TSM and that has no issues, the only issue is when running expect with TSM command.

This program works: https://gist.github.com/treydock/240167c26bfa858fc8c656f89197f056

It dumps the output of what one would get from the TSM query, so the actual output is not the issue so it's seeming like this could be an issue with how TSM operates under certain conditions.

treydock:
I think something with TSM might change behavior based on size of data returned.  I made `LIMIT` a flag and if I do `LIMIT 21` the expect works if I do `LIMIT 22` the query times out.  The query outside Go takes the same amount of time regardless so this might just be some quirk with the `dsmadmc` command when some threshold of data is returned.

treydock:
I think I found a solution but it's not great.  Rather than having TSM write the query output to the shell, which wasn't working with expect, instead use `ioutil.TempFile` to create a temporary file for each query, instruct dsmadmc to use `-OUTfile=<TempFile Name()>` then read the temp file for output.  It's really ugly but is the only solution I could get to work with expect.

https://github.com/treydock/tsm_exporter/pull/20

Main change was to how dsmadmc queries are performed: https://github.com/treydock/tsm_exporter/blob/password/collector/collector.go#L107-L171

brian-brazil:
> The query outside Go takes the same amount of time regardless so this might just be some quirk with the dsmadmc command when some threshold of data is returned.

Most likely a buffer for the output filled up on the expect side of things.

treydock:
> > The query outside Go takes the same amount of time regardless so this might just be some quirk with the dsmadmc command when some threshold of data is returned.
> 
> Most likely a buffer for the output filled up on the expect side of things.

I thought that too but I can run the exact same code with same output I'd expect to get with TSM query but instead dump the results with SSH command and shell script and I get no timeouts.  The only time I get timeouts is when number of rows with TSM is greater than around 21, but dumping 31 lines of the exact same output with SSH and expect does not have issues.

Let me know if having TSM write the results to a TempFile and then reading that file is a viable option. It's not the greatest solution but it's the only way I can get rid of the password in CLI.  The code is not merged, my last comment has links.

brian-brazil:
Anything that avoids putting the password on the commandline works.

treydock:
I've updated the exporter to use expect for passing the password and TSM query.

brian-brazil:
Giving a quick look over:

In volumes and stgpools there's percentage metrics. 

In status the metric is mixing label values and metric values, and in this case the reason is entirely coming from your code so I'd suggest logging rather than exposing as a metric.

Time metrics in replicationview and db are lacking a seconds unit.

In db as a counter TotalBuffReq should have a _total suffix, not infix.

treydock:
> Giving a quick look over:
> 
> In volumes and stgpools there's percentage metrics.

I'll make that adjustment to convert them to ratios or 0.0-1.0.

> In status the metric is mixing label values and metric values, and in this case the reason is entirely coming from your code so I'd suggest logging rather than exposing as a metric.

This is sort of like the overall health of the ability to query TSM.  I use this in alertmanager inhibit rules, but it's just a simple query that can determine if TSM is responding to queries.  I went ahead and replaced the `reason` label with just a log message but kept the value as 0/1 based on query health.  I also removed `servername` label because it's sort of like the Prometheus target.

> 
> Time metrics in replicationview and db are lacking a seconds unit.

So in replicationview the `start_time` and `end_time` and db `last_backup_time` metrics are actually unix timestamps.  It's not clear to me from Prometheus metric documentation the best practice for unix timestamps in terms of metric naming.  Should `tsm_replication_start_time` become `tsm_replication_start_timestamp_seconds` or use some other suffix to indicate unix timestamp?

> 
> In db as a counter TotalBuffReq should have a _total suffix, not infix.

I am making this change.

brian-brazil:
> Should tsm_replication_start_time become tsm_replication_start_timestamp_seconds or use some other suffix to indicate unix timestamp?

There's no single practice, both would work (or start_time_seconds).

treydock:
I made those metrics so the suffix is `_timestamp_seconds`

brian-brazil:
> I'll make that adjustment to convert them to ratios or 0.0-1.0.

The metric names in those two still need a _ratio rather than _percent  (and the help is a bit confusing, something can't be both a ratio and percentage).

treydock:
Renamed those metrics to end with _ratio and updated help content.

brian-brazil:
Thanks!

PR #1788, opened 2020-11-06
Add cgroup exporter

treydock:
Add link to cgroup exporter.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -190,6 +190,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [cAdvisor](https://github.com/google/cadvisor)
    * [Cachet exporter](https://github.com/ContaAzul/cachet_exporter)
    * [ccache exporter](https://github.com/virtualtam/ccache_exporter)
+   * [cgroup exporter](https://github.com/treydock/cgroup_exporter)
    * [c-lightning exporter](https://github.com/lightningd/plugins/tree/master/prometheus)
    * [DHCPD leases exporter](https://github.com/DRuggeri/dhcpd_leases_exporter)
    * [Dovecot exporter](https://github.com/kumina/dovecot_exporter)


brian-brazil:
Thanks for your PR. I aim to avoid duplication in this list, to discourage splitting the community and continuously reinventing the wheel. We already list cAdvisor which provides extensive cgroup metrics, so if there's something that's missing you should work with the maintainers of that tool.

In this particular case, I believe there is also a desire from some quarters to replace cAdvisor with something simpler and more performant. It may be worth announcing this exporter on the developers mailing list, and if it becomes more popular it could potentially replace cAdvisor in this listing.

treydock:
@brian-brazil This is for cgroups outside containers, so like `user.slice` or `system.slice` for systemd cgroups or job cgroups from HPC jobs setup by Torque or SLURM.  These things do not involve tools like Docker so not sure cadvisor is really applicable, as these are just pure cgroups and not actual containers.

What developers mailing list are you referring to, the Prometheus developers list?

brian-brazil:
cAdvisor covers all cgroups, for example on my desktop it shows up all the cgroups that systemd has setup.

> What developers mailing list are you referring to, the Prometheus developers list?

Yes.

SuperQ:
If you're looking for systemd cgroup data, there is the [systemd_exporter](https://github.com/povilasv/systemd_exporter).

SuperQ:
It would be nice to have something more Prometheus-specific than cAdvisor. It has a number of drawbacks and bugs. But given the popularity and development community around it, I still prefer to have cAdvisor be the canonical cgroup exporter.

treydock:
This exporter was written to parse SLURM and Torque cgroups, which if you're not familiar are resource managers for HPC jobs.  I wonder if remaining the exporter to something more specific to those applications?  There is logic in this exporter to parse the HPC job ID out of the cgroup path, which is necessary for how I'm using this exporter, so while it will work with any cgroup structure (or should), it has extra logic specific to HPC cgroups.  If a rename doesn't seem like a viable option then I guess my only avenue would be posting to mailing list?  I don't see us switching to cAdvisor due to us needing to deploy something on our HPC compute environment that is as light weight as possible.

brian-brazil:
> There is logic in this exporter to parse the HPC job ID out of the cgroup path

A cgroup exporter shouldn't really have business logic like that, but if you want it you should talk to the cAdvisor maintainers. Relabelling would also likely cover that.

PR #1789, opened 2020-11-06
Add SSH exporter

treydock:
Adds SSH exporter that is mostly used for monitoring purposes in testing that SSH is healthy but does produce some timing data.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -226,6 +226,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Smokeping prober](https://github.com/SuperQ/smokeping_prober)
    * [SMTP/Maildir MDA blackbox prober](https://github.com/cherti/mailexporter)
    * [SoftEther exporter](https://github.com/dalance/softether_exporter)
+   * [SSH exporter](https://github.com/treydock/ssh_exporter)
    * [Teamspeak3 exporter](https://github.com/hikhvar/ts3exporter)
    * [Transmission exporter](https://github.com/metalmatze/transmission-exporter)
    * [Unbound exporter](https://github.com/kumina/unbound_exporter)


brian-brazil:
Thanks for your PR. I aim to avoid duplication in this list, to discourage splitting the community and continuously reinventing the wheel. This seems to be fundamentally doing the same thing as the script exporter which we already list, and if you just want to test if SSH is responding across many machines the blackbox exporter covers that.

treydock:
The blackbox exporter doesn't actually log in with SSH, we tried using that and it failed to detect when logins where broken despite SSH still listening on port 22.  The script exporter might work to replace some of this functionality but even that script exporter warns against a large number of executions (at bottom of README) which we would do as we have around 200 systems monitored for healthy SSH and forking out to SSH 200 times every 60 seconds would be really ugly.  This exporter uses Go's builtin SSH interface.

brian-brazil:
I think this is a bit niche to list here and encourages poor practices. I don't see this as testing SSH - it's testing running commands over what happens to be SSH. In general I'd try to discourage monitoring that involves SSHing into boxes to execute commands, as that's basically nagios-style checks and quite inefficient on both ends due to all the ssh overhead. Dedicated agents (e.g. the node exporter) are a better way to collect metrics.

I'd expect something called an ssh exporter to be doing extensive testing of SSH itself (e.g. what auth methods were offered?), and limited to that scope.

treydock:
This is pretty much inspired from a need we had to fill from migrating from Nagios to Prometheus, where we had to ensure that SSH into a box was healthy because we have interactive SSH nodes where if SSH is broken we have hundreds of really unhappy users.  The only reliable way to test SSH is from an external system that is connecting into SSH.  This does have timing collection of how long it takes to login with SSH and the command execution is optional, and not required.  My examples show that because that's what we've deployed but if the commands are omitted it's just checking that an SSH session could be established.  Having a locally run SSH exporter seems pretty much useless to me as SSH is a means to get into a box and knowing about the local daemon serves very little purpose.  It's sort of like checking a website via localhost rather than remotely via the actual website address, it doesn't really monitor much other than the local state which is not always the state you care about.

Would there maybe be a possibility of expanding on blackbox exporter to have a more robust SSH specific collector?  Maybe not something that executes a command but something that makes an actual SSH session either via password of SSH key and verifies external access to SSH is actually healthy.  The current SSH solution for blackbox exporter can very easily report success of an unhealthy SSH daemon as it doesn't fully exercise the SSH login process, just checks SSH is listening essentially.

brian-brazil:
> Would there maybe be a possibility of expanding on blackbox exporter to have a more robust SSH specific collector? 

As the maintainer of it, I don't want to get into specific application protocols. Line-based TCP is as far as it'll go.

> Having a locally run SSH exporter seems pretty much useless to me as SSH is a means to get into a box and knowing about the local daemon serves very little purpose. It's sort of like checking a website via localhost rather than remotely via the actual website address, it doesn't really monitor much other than the local state which is not always the state you care about.

It's not one or the other. You can do both; one to check the network access, the other to check the login.

Often you can rely on the node exporter being up to cover ssh also, as ssh being down but the node exporter being up should be quit rare -  if either is not responding over the network it's likely because the whole machine is down.

While this is of use to you, in the context of this list l don't think this makes sense to have as an entry. You are of course free to announce it on the -users mailing list.

SuperQ:
Since this is actually doing a full SSH login probe, which by design the blackbox exporter doesn't do, this would be a very good addition to the list.

SuperQ (approved):

roidelapluie:
I have not looked at the technical details, but local SSSD (ldap) crashing or other issues can not easily be caught by node or blackbox exporter. I also think it would fit in.

RichiH:
It's super useful

brian-brazil:
Thanks!

I note that this exporter has hardcoded disabling of host key checks, which may not be the best approach.

PR #1793, opened 2020-11-09
docs(guides): add windows exporter guide

noelmcloughlin:
This PR introduces a Windows Exporter Guide (mirrors node_exporter) so people can find the procedure here.  

Otherwise, you need to search the internet for Blog on Windows and prometheus .

### DIFF
--- a/content/docs/guides/windows-exporter.md
+++ b/content/docs/guides/windows-exporter.md
@@ -1,0 +1,68 @@
+---
+title: Monitoring Windows host metrics with Windows Exporter
+---
+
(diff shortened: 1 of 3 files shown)


brian-brazil:
I'm not sure it makes sense to have docs for unofficial tools as part of the core docs, we're not generally in a position to review such things as we don't maintain them. This would be better placed as part of the docs for this integration.

noelmcloughlin:
Sure, if this PR is misplaced then can be closed and I'll see windows_exporter readme can be updated.

Issue #1799, opened 2020-11-19
Add windows_exporter do Downloads


roidelapluie:
Even if it is not official, it would be worth it to list the windows_exporter in https://prometheus.io/download/
We could add a note saying that it is a community project.

PR #1800, opened 2020-11-19
Support download list for community repositories

grobie:
In order to display windows_exporter releases and the available
artifacts on the downloads page, this change adds the option to fetch
release information for all organizations on Github. As the
windows_exporter artifact naming doesn't follow the established format
unfortunately, a few additional special format options were added.

---

Solves #1799. FYI @carlpett @roidelapluie

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
-NANOC      = bundle exec nanoc
-GUARD      = bundle exec guard
-DOWNLOADS := prometheus alertmanager blackbox_exporter consul_exporter graphite_exporter haproxy_exporter memcached_exporter mysqld_exporter node_exporter pushgateway statsd_exporter
+NANOC     ?= bundle exec nanoc
+GUARD     ?= bundle exec guard
+DOWNLOADS := $(shell cat downloads.cfg)
 
 build: clean downloads compile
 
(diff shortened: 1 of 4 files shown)


brian-brazil:
> As the windows_exporter artifact naming doesn't follow the established format unfortunately

Should we ask them to?

grobie:
> Should we ask them to?

I think this would be helpful, but would also result in a breaking change for anyone who has automated the download process. I'm worried about having to support more variations though, as the artifact naming format is already pretty ambiguous with a varying number of `-` and `.` delimiters used in standard vs. pre-release versions.

The current differences in the windows_exporter release artifacts are:

* artifact names dropped the `.<os>` part from the `<name>-<version>.<os>-<arch>.<ext>` format
* sha256sums.txt filenames include an `<arch>\` prefix (probably the result of storing artifacts in separate folders)

brian-brazil:
I think long term consistency for users would be more beneficial, and I don't think it'd be unreasonable to ask 3rd party projects we list here to follow our artifact naming conventions.

grobie:
Makes sense to me! I guess one could argue that the `.msi` / `.exe` artifacts are so unique that consolidation isn't necessarily important, but the `<arch>` prefix in the sha256sums.txt is quite odd.

@carlpett @martinlindhe Any comments / input from your side?

brian-brazil:
Yeah, I don't see a need to force tarballs but everything before the extension could be consistent. The arch in the shasums looks like a bug to me.

carlpett:
As you guessed @grobie, the sha256sums is because of generating them from subdirs. I no longer recall what exactly that was a workaround for, something about promu on Windows I think. Generally, it's been a somewhat demanding project to maintain, and CI has been one of the things that hasn't gotten a lot of love. I'll try to fix the shasums thing asap.

On the naming topic:
- I'd be fine with changing the artifact names to cause less pain elsewhere, eg here, as long as we don't need to package tarballs. Since there's no built-in support for neither tar nor gz, that'd be a significant user friction.
- We'll need to sync with the packaging for Chocolatey (~homebrew for Windows) that it won't break them, because that is a fairly common way people get the exporter. @andrewmostello, you'd need to change this, correct?
- We could add the os bit in there, for obvious reasons it hasn't been hugely useful for the project itself.

RichiH:
Minor point: If your build chain could be modified to support all compression/package formats that would increase consistency even more, but again, this is minor.

Thanks for being willing to rename :)

roidelapluie:
> * We could add the os bit in there, for obvious reasons it hasn't been hugely useful for the project itself.

no freedos version of the exporter? :smile:

grobie:
> I'd be fine with changing the artifact names to cause less pain elsewhere, eg here, as long as we don't need to package tarballs. Since there's no built-in support for neither tar nor gz, that'd be a significant user friction.

I don't see any need for building tar.gz packages for Windows. The feedback that this isn't available on Windows has been raised before, all releases using promu >= v0.7.0 (e.g. Prometheus v2.23.0 will be the first) will produce .zip packages for Windows. I haven't found any other major software releasing tar.gz archives for windows either. Using the most common and appropriate package formats for different platforms makes sense to me.

Makefile
```diff
@@ -1,6 +1,6 @@
-NANOC      = bundle exec nanoc
-GUARD      = bundle exec guard
-DOWNLOADS := prometheus alertmanager blackbox_exporter consul_exporter graphite_exporter haproxy_exporter memcached_exporter mysqld_exporter node_exporter pushgateway statsd_exporter
+NANOC     ?= bundle exec nanoc
+GUARD     ?= bundle exec guard
+DOWNLOADS := $(shell cat downloads.cfg)
```
roidelapluie, line 3: can that be downloads.cfg and scripts / ruby reading the files directly?

andrewmostello:
> * We'll need to sync with the packaging for Chocolatey (~homebrew for Windows) that it won't break them, because that is a fairly common way people get the exporter. @andrewmostello, you'd need to change this, correct?

That's right, but it should be a quick adjustment. I'd just need to change the release URL here: https://github.com/andrewmostello/prometheus-windows-exporter.install/blob/master/tools/chocolateyinstall.ps1#L5

On a related note, the Chocolatey install is based on the msi installer. If that were ever dropped it'd require a much larger change to the install script.

jan--f:
Hello from the bug scrub!
@grobie  Is this a change we should still add? If so would you kindly rebase this?

jan--f:
Hello from the bug scrub!

I'll close this for now as it seems abandoned. Please feel free to reopen.

PR #1801, opened 2020-11-20
dockerswarm: Add label-based port example

nivekuil:
I think this is a more solid, simpler approach for making a service visible than publishing a port just for prometheus. Would have saved me a few hours to have it in here :)


nivekuil:
see also: https://groups.google.com/g/prometheus-users/c/tLdIOXElmuQ/m/BTaNmfJsAgAJ

roidelapluie:
Can you add white lines before and after the examples, and add DCO? Thanks

nivekuil:
Done; only added a \n after, doesn't seem like there should be 2 newlines between this and the last example?

content/docs/guides/dockerswarm.md
```diff
@@ -199,6 +199,15 @@ Global tasks run on every daemon.
   target_label: docker_node
 ```
 
+### Scrape a service based on the `prometheus.port` label to avoid having to publish the port
```
roidelapluie, line 202:
```suggestion
### Scraping a service based on the `prometheus.port` label to avoid having to publish the port
```

one last nit

jan--f:
Hello from the bug scrub!
Thanks for your contribution. Would you please rebase the PR? This would still be great to have.

jan--f:
Hello from the bug scrub!
I'll close this for now, but please feel free to re-open in the future.

PR #1808, opened 2020-12-05
Add Vector to tools that export Prometheus metrics

jamtur01:
Vector has both a Prometheus source and sink and a `remote_write` source and sink.

@brian-brazil

Signed-off-by: James Turnbull <james@lovedthanlost.net>


brian-brazil:
Can you add the DCO?

https://vector.dev/docs/reference/sinks/prometheus_remote_write/ also says that Prometheus ingests remote write, which is incorrect. Can you fix that please?

jamtur01:
The DCO is fixed.

roidelapluie:
thanks!

PR #1815, opened 2020-12-14
Add IoTDB exporter link

fagnercarvalho:
I created an exporter for the IoTDB database and I wanted to add to the exporters page.

Let me know what you think.

Thanks!

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -57,6 +57,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [SQL exporter](https://github.com/free/sql_exporter)
    * [Tarantool metric library](https://github.com/tarantool/metrics)
    * [Twemproxy](https://github.com/stuartnelson3/twemproxy_exporter)
+   * [IoTDB exporter](https://github.com/fagnercarvalho/prometheus-iotdb-exporter)
 
 ### Hardware related
    * [apcupsd exporter](https://github.com/mdlayher/apcupsd_exporter)


brian-brazil:
This exporter is taking a password on the commandline, which is not secure. Can you fix this please?

Some advice to improve your exporter:

Some of the metric names could be claner, for example iotdb_timeSeries_timeSeriesCount could just be iotdb_timeseries. The filesizt metrics are also lacking a unit in their metric name.

The up and last_scrape_error metrics are not correctly implemented due to race conditions. I'd suggest keeping it simple and removing these metrics in favour of failing the scrape when there's an error.

fagnercarvalho:
Hey @brian-brazil, thanks, I really appreciate the review!

Here are my considerations:

> This exporter is taking a password on the commandline, which is not secure. Can you fix this please?

Sure. This is fixed. Now the IoTDB instance password should be set as an environment variable and the option was removed from the available parameters.

> Some of the metric names could be claner, for example iotdb_timeSeries_timeSeriesCount could just be iotdb_timeseries. The filesizt metrics are also lacking a unit in their metric name.

Okay, I simplified the count metrics and added the unit to the file size metrics.

> The up and last_scrape_error metrics are not correctly implemented due to race conditions. I'd suggest keeping it simple and removing these metrics in favour of failing the scrape when there's an error.

Could you explain this? For these two metrics I took inspiration from the [mysqld_exporter](https://github.com/prometheus/mysqld_exporter).

brian-brazil:
> Sure. This is fixed. Now the IoTDB instance password should be set as an environment variable and the option was removed from the available parameters.

There's still a password on the commandline in example down the bottom of the readme.

> Okay, I simplified the count metrics and added the unit to the file size metrics.

There's more metrics that could be simplified, and the suffix should be _bytes rather than Bytes. For example  iotdb_system_file_size_bytes and iotdb_storage_groups.

> Could you explain this? For these two metrics I took inspiration from the mysqld_exporter.

See https://prometheus.io/docs/instrumenting/writing_exporters/#failed-scrapes
Unless you have a really good reason otherwise, it's best to return a 500.

fagnercarvalho:
> There's still a password on the commandline in example down the bottom of the readme.

Oh no, I see what you meant now. I removed the password parameter from the README.

> There's more metrics that could be simplified, and the suffix should be _bytes rather than Bytes. For example iotdb_system_file_size_bytes and iotdb_storage_groups.

Understood. I simplified the file size metrics too and changed the metrics to use snake case.

> See https://prometheus.io/docs/instrumenting/writing_exporters/#failed-scrapes
> Unless you have a really good reason otherwise, it's best to return a 500.

Got it, it is not returning 500 in case the exporter is not able to open a connection or fails when scraping any of the metrics.

And thanks for the fast replies, I know it is late in Ireland now.

brian-brazil:
Thanks!

You still have three gauges ending in _count, which is a suffix used by summaries and historgrams. This should be removed to avoid confusion, and because it doesn't really add information.

fagnercarvalho:
Just removed the `_count` suffix from these metrics, thanks!

PR #1816, opened 2020-12-15
Minor: Add youtube to the group list

roidelapluie:
I know that it says that we do not list them all but it is easier if we
actually list the ones we know about.


brian-brazil:
:+1:

PR #1818, opened 2020-12-16
adding faust-prometheus exporter

skecskes:
@juliusv
@brian-brazil
I added library that can be plugged into FAUST, the kafka streaming framework written in python by RobinHood and will directly expose Faust metrics of the app in prometheus format.


brian-brazil:
This isn't an application it's a library, so it'd go down in the list at the bottom.

In terms of improving your library, I have the following suggestions:

Don't have 3 different metrics only differing in labels. messages_received_per_topics_partition is sufficient for example, and can then be called faust_received_messages.

Latencies should be in seconds, not milliseconds. 

Avoid "total" in metric names, it is redundant.

Naming is generally inconsistent, e.g. topic_messages_sent vs total_sent_messages. In that case a single metric called faust_sent_messages with all the labels would suffice.

Rebalances sounds like a counter, not a gauge.

This code shouldn't be touching the GC and other default metrics the Python client provides, it should only be a collector and should definitely not be poking around in REGISTRY and other implementation internals of the client library. As-is this code will silently break any other Prometheus instrumentation that is loaded in before it.
Nor should a collector like this be doing exposition - that's a separate concern that's up to the application developer.

jan--f:
Hello from the bug scrub!
The proposed link is a 404 now. Closing...

PR #1819, opened 2020-12-16
Add Amazon Managed Service for Prometheus (AMP)

matanbaruch:
AWS added Manage Prometheus as read and write service.
https://aws.amazon.com/prometheus/


brian-brazil:
Thanks for your PR. As far as I'm aware this is Cortex, and we list software here rather than every company running that software as a service. We don't list Grafana Cloud either for example, which is also Cortex.

matanbaruch:
Agreed.

But it should be mentioned somewhere, as it is an easy way to implement Prometheus on the cloud. 

Maybe a new managed service section?

brian-brazil:
Listing companies who provide Cortex as a service seems more something for Cortex docs than Prometheus docs.

matanbaruch:
Got it.

PR #1823, opened 2020-12-21
[exporters] File statistics exporter

michael-doubez:
Simple exporter to gather simple metrics about files.


SuperQ:
Neat, one metric naming suggestion, `file_glob_match_number` becomes `file_glob_matches`.

SuperQ:
Please fix the DCO with `git commit -s --amend`

michael-doubez:
> Please fix the DCO with `git commit -s --amend`

Done

michael-doubez:
> Neat, one metric naming suggestion, `file_glob_match_number` becomes `file_glob_matches`.

Is there any rule of thumb for gauges as there is `_total` for counters ?

michael-doubez:
Should I still hope for integration ?
Note: Debian package exists https://packages.debian.org/source/sid/prometheus-exporter-exporter

juliusv:
Thanks, I think this just got lost in time, merging it now.

> Is there any rule of thumb for gauges as there is _total for counters (mandatory in OpenMetrics IIRC) ?

There's no explicit suffix convention for gauges, so instead of `_number`, one generally just uses the plural of whatever is being tallied / measured, and that's it.

PR #1824, opened 2020-12-21
Add tookit doc

roidelapluie:
WIP

### DIFF
--- a/nanoc.yaml
+++ b/nanoc.yaml
@@ -85,6 +85,14 @@ data_sources:
       entrypoint: /overview/
       repository_url: https://github.com/prometheus/alertmanager.git
   -
+    type: repo_docs
+    encoding: utf-8
+    identifier_type: legacy
+    items_root: /docs/toolkit/
+    config:
+      entrypoint: /introduction/
+      repository_url: https://github.com/roidelapluie/exporter-toolkit
+  -
     type: filesystem
     items_root: /assets
     content_dir: static
--- a/content/docs/guides/index.md
+++ b/content/docs/guides/index.md
@@ -1,5 +1,5 @@
 ---
 title: Guides
-sort_rank: 9
+sort_rank: 10
 nav_icon: map-o
 ---
--- a/content/docs/practices/index.md
+++ b/content/docs/practices/index.md
@@ -1,5 +1,5 @@
 ---
 title: Best practices
-sort_rank: 8
+sort_rank: 9
 nav_icon: thumbs-o-up
 ---
--- a/content/docs/toolkit/index.md
+++ b/content/docs/toolkit/index.md
@@ -1,0 +1,5 @@
+---
+title: Toolkit
+sort_rank: 8
+nav_icon: toolbox
+---


nanoc.yaml
```diff
@@ -84,6 +84,14 @@ data_sources:
...
       entrypoint: /overview/
       repository_url: https://github.com/prometheus/alertmanager.git
+  -
+    type: repo_docs
+    encoding: utf-8
+    identifier_type: legacy
+    items_root: /docs/toolkit/
+    config:
+      entrypoint: /introduction/
+      repository_url: https://github.com/roidelapluie/exporter-toolkit
```
brian-brazil, line 94: This doesn't seem right

content/_redirects
```diff
@@ -31,6 +31,7 @@
 /docs/instrumenting/        /docs/instrumenting/clientlibs/             302!
 /docs/operating/            /docs/operating/security/                   302!
 /docs/alerting/             /docs/alerting/latest/overview/             302!
+/docs/tookit/               /docs/toolkit/latest/introduction/          302!
```
grobie, line 34:
```suggestion
/docs/toolkit/               /docs/toolkit/latest/introduction/          302!
```

grobie (commented):
Hm, I'm not sure it's good idea adding yet another top-level menu item, we already have too many. Best practice is to limit the number of menu items per level to 7, to not confuse people too much. The exporter toolkit is only relevant for a very small number of people using our docs.

I'd suggest mounting the docs below Instrumenting instead. Not sure whether that requires some changes to the navigation code to allow repodocs on that level.

roidelapluie:
> Hm, I'm not sure it's good idea adding yet another top-level menu item, we already have too many. Best practice is to limit the number of menu items per level to 7, to not confuse people too much. The exporter toolkit is only relevant for a very small number of people using our docs.

This is the end user documentation, so the TLS config that will be used by all the exporters and Prometheus / AM / Pushgateway. it is not developers doc.

roidelapluie:
I can't make the navbar to work but here is the relevant page for end users https://5fe0c8bd64952c00087ea9b1--prometheus-docs.netlify.app/docs/toolkit/latest/web-configuration/

grobie:
> This is the end user documentation, so the TLS config that will be used by all the exporters and Prometheus / AM / Pushgateway. it is not developers doc.

Ah, I wasn't sure what documentation you plan to publish there. Still it seems this fits well in the already existing "Instrumenting" section.

brian-brazil:
Currently where we have duplicated functionality (i.e. http_config) we duplicate the docs in each place in Prometheus, AM, and blackbox where it occurs.

jan--f:
Hello from the bug scrub!
@roidelapluie https://github.com/prometheus/exporter-toolkit is still around. This still seems worth adding, wdyt?

jan--f:
Hello from the bug scrub!
I will close this for now due to lack of feedback or response. Please feel free to reopen at any time, I think this would still be good to have.

PR #1827, opened 2021-01-06
Update docs for TLS

roidelapluie:

### DIFF
--- a/content/docs/introduction/faq.md
+++ b/content/docs/introduction/faq.md
@@ -222,21 +222,14 @@ second, will only run into precision issues after over 285 years.
 
 ### Why don't the Prometheus server components support TLS or authentication? Can I add those?
 
-Note: The Prometheus team has changed their stance on this during its development summit on
-August 11, 2018, and support for TLS and authentication in serving endpoints is now on the
-[project's roadmap](/docs/introduction/roadmap/#tls-and-authentication-in-http-serving-endpoints).
-This document will be updated once code changes have been made.
+TLS and basic authentication is gradually being rolled out to the different
+components. Please follow the different releases and changelogs to know which
+components have already implemented it.
 
-While TLS and authentication are frequently requested features, we have
-intentionally not implemented them in any of Prometheus's server-side
-components. There are so many different options and parameters for both (10+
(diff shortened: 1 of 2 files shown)


content/docs/introduction/faq.md
```diff
@@ -222,21 +222,14 @@ second, will only run into precision issues after over 285 years.
...
-authentication solutions in every server component.
-
-If you need TLS or authentication, we recommend putting a reverse proxy in
-front of Prometheus. See, for example [Adding Basic Auth to Prometheus with
-Nginx](https://www.robustperception.io/adding-basic-auth-to-prometheus-with-nginx/).
+TLS and basic authentication is gradually being rolled out to the different
+components. Please follow the different releases and changelogs to know which
+components have already implemented it.
+
+The components currently supporting TLS and authentication are:
```
brian-brazil, line 229: I'm not sure it makes sense to try and keep a list like this up to date.

SuperQ: I mostly agree, but if we limit it to official components, we only have 15 things to list here. (13 exporters, Prometheus, Alertmanager).

That's a maintainable list.

content/docs/introduction/faq.md
```diff
@@ -222,21 +222,14 @@ second, will only run into precision issues after over 285 years.
...
+
+- Prometheus 2.24.0 and later
+- Node Exporter 1.0.0 and later
```
SuperQ, line 232:
```suggestion
- node\_exporter 1.0.0 and later
```

content/docs/introduction/faq.md
roidelapluie: This is not really something to maintain, it is a one-shot.

roidelapluie: on 15 times a one-shot, and we can even batch them when adding them.

brian-brazil:
:+1:

grobie (approved):

content/docs/introduction/faq.md
roidelapluie: We use Node Exporter in the security part, and Prometheus is with a capital lette

SuperQ (approved):

PR #1829, opened 2021-01-07
Added Volume Exporter

mnadeem:
https://github.com/mnadeem/volume_exporter

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -237,6 +237,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Unbound exporter](https://github.com/kumina/unbound_exporter)
    * [WireGuard exporter](https://github.com/MindFlavor/prometheus_wireguard_exporter)
    * [Xen exporter](https://github.com/lovoo/xenstats_exporter)
+   * [Volume Exporter](https://github.com/mnadeem/volume_exporter)
 
 
 When implementing a new Prometheus exporter, please follow the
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. I am to avoid duplication in this list to encourage the community to work together and avoid continuously reinventing the wheel. This exporter appears to duplicate existing basic functionality of the node/windows exporters, so accordingly doesn't make sense to list. I would recommend using one of those exporters for these and other kernel metrics.

mnadeem:
This is useful to monitor disk/volume/PV storage, for various reasons

* Host path cannot be mounted to container / Deamonset (This is what happens in cooperates, In nodes multiple projects run on same or different nodes, and hence host path mount cannot be granted, further node scapping cannont be done as well, since it requires cluster role, and in a cluster multiple projects [namspaces] runs, and hence cluster role is not granted to projects)
* [In ability](https://bugzilla.redhat.com/show_bug.cgi?id=1373288) of Cloud provider to gather information about PV

Above I have listed some of the use cases ( real world) where in node exporter cannot be deployed, should this stop us from monitoring pv/volumes/disk? 

**This is altogether a different exporter and does not duplicate any existing exporters. Its area of focus (Where ever node exporter cannot be deployed) is altogether different.**

**It just fills the vacuum**

Hope it helps

brian-brazil:
If you've questions on how to use the node exporter I'd recommend taking them to the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users), if you feel there's features it's missing please work with its maintainers.

I do not think it makes sense to list more than one exporter that does `statsfs` for a given operating system.

PR #1830, opened 2021-01-08
Added additional Elixir library

akoutmos:
PromEx is a library that I have been working on for some time now, and from the users that have adopted it, it has become a great observability solution using Prom+Grafana. Figured I would add it to the documentation list here :).

Thanks in advance!!


brian-brazil:
Thanks for your PR. To encourage the community to work on one solution and continuously reinventing the wheel, we only list one client library for a given runtime/language. As far as I can see the existing listing is more popular, predates the one in this PR by 4 years, and has a much bigger community. If you feel there's things the existing listing could do better, I'd recommend discussing it with them.

akoutmos:
Thanks for the quick reply! The two libraries have two very different design philosophies behind them, so I don't think library consolidation is necessarily an option. But totally understandable that you want to keep 1 library per language :).

Thanks again!

jan--f:
Hello from the bug scrub!
I think adding a second alternative is certainly valid. Could you please rebase your PR and perhaps add a one liner what distinguishes the two projects?

jan--f:
Hello from the bug scrub!

Closing this for now as abandoned. Please feel free to reopen at any time.

PR #1847, opened 2021-01-26
Add Iptables exporter

madron:
@brian-brazil

I added [Iptables exporter](https://github.com/madron/iptables-exporter) link.
Let me know if it fullfils prometheus standards and minimum requirements.
Thanks

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -203,6 +203,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Ethereum Client exporter](https://github.com/31z4/ethereum-prometheus-exporter)
    * [JFrog Artifactory Exporter](https://github.com/peimanja/artifactory_exporter)
    * [Hostapd Exporter](https://github.com/Fundacio-i2CAT/hostapd_prometheus_exporter)
+   * [Iptables exporter](https://github.com/madron/iptables-exporter)
    * [IRCd exporter](https://github.com/dgl/ircd_exporter)
    * [Linux HA ClusterLabs exporter](https://github.com/ClusterLabs/ha_cluster_exporter)
    * [JMeter plugin](https://github.com/johrstrom/jmeter-prometheus-plugin)


brian-brazil:
Thanks for your PR. While these are host-based from the kernel metrics, it looks like the Node exporter developers believe this is best as a separate exporter so that's fine. More interestingly there are other iptables exporters out there, but they either look dead or seem to be doing something different, and none have any community of note. So this could be listed.


Looking at the code I see an issue though, you're using direct instrumentation and thus will produce incorrect results. Plus poking around the internals of a client library isn't exactly kosher. Please switch to using CounterMetricFamily.

I also note that you have a custom MetricsHandler, which adds no functionality on top of the base one provided by the client library. This however won't support OpenMetrics, so you should use the standard one.

From a general exporter design standpoint, this exporter seems to require you to explicitly add a comment to every single rule you want a metric for. I'd personally expect an exporter like this to expose everything by default, and the naming feature to be an extra thing you could optionally do if you wanted.

madron:
Thanks for interesting hints !  I'll take some time to elaborate/fix your suggestions.

madron:
> Looking at the code I see an issue though, you're using direct instrumentation and thus will produce incorrect results. Plus poking around the internals of a client library isn't exactly kosher. Please switch to using CounterMetricFamily.
> 
> I also note that you have a custom MetricsHandler, which adds no functionality on top of the base one provided by the client library. This however won't support OpenMetrics, so you should use the standard one.

I tried to address both your suggestions:  https://github.com/madron/iptables-exporter/blob/5a423c16aaaaf17216f0c87f08b2c495929a6dad/iptables_exporter/main.py

> From a general exporter design standpoint, this exporter seems to require you to explicitly add a comment to every single rule you want a metric for. I'd personally expect an exporter like this to expose everything by default, and the naming feature to be an extra thing you could optionally do if you wanted.

Without explicit comments it expose only the number of rules in every table/chain. Maybe it seems useless but I use a lot just to check if someone manually add/delete some rules and forget to restore the original configuration.

I am not really convinced about exposing everything by default for these reasons:
- I didn't find a good tag or set of tags to describe every possible iptables rule.
- In all my kubernetes nodes i have at least 6000 rules and I'm afraid it can pose some pressure on prometheus servers. 

 I'm open for suggestions. !  :-)

jan--f:
Hello from the bug scrub!

PR #1851, opened 2021-01-28
Add starlette-prometheus

mvolfik:


jan--f:
Hello from the bug scrub!
Thank you for your contribution. I'm afraid we need to decline this for now. We are happy to feature exporters that are actively developed and used. Unfortunately this one seems no longer under maintenance. Please feel free to reopen at a later stage.

PR #1858, opened 2021-01-29
Update naming.md

elmariofredo:
We have run with our users to metrics with more than 300 characters, so suggestion about length would be nice in docs.

Suggestion is based on https://docs.datadoghq.com/developers/guide/what-best-practices-are-recommended-for-naming-metrics-and-tags/#rules-and-best-practices-for-naming-metrics


content/docs/practices/naming.md
```diff
@@ -26,6 +26,7 @@ A metric name...
    (exported by many client libraries)
  * <code><b>http</b>\_request\_duration\_seconds</code>
    (for all HTTP requests)
+* ...should not exceed 200 characters ( less than 100 is preferred )
```
grobie, line 29: Thanks for this suggestion, I agree it makes sense to provide some guidance here.

Can you provide some examples for the long metric names you have come across? I assume users are encoding information in metric names, which should be set as labels instead. In that case, I'd suggest to change the wording accordingly to something like "...should not encode information, use labels for that" for example.

jan--f:
Hello from the bug scrub!
Sorry for the delay. Lots of things have changed, I think the metrics name length is perhaps less of an issue then it was.
Please reopen this if you still think we should mention to avoid long metric names and outline why.

PR #1864, opened 2021-02-09
Add Prom-migrator to list of integrations

cevian:
Prom-migrator allows users to migrate existing data between various long-term storage systems. In particular it can migrate data from any system that support remote_read to any system that supports remote_write. I think it would be useful to users and this seems like a natural place to mention this.

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -63,6 +63,8 @@ data volumes.
   * [VictoriaMetrics](https://github.com/VictoriaMetrics/VictoriaMetrics): write
   * [Wavefront](https://github.com/wavefrontHQ/prometheus-storage-adapter): write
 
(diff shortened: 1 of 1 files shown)


cevian:
cc: @juliusv and thanks for the suggestion.

juliusv:
:1: from my side, but /cc @roidelapluie for one more opinion, since I was the one suggesting this.

roidelapluie:
I have no strong opinion on this but I would have hoped that such a tool would use more of the power of the current upstream clients (e.g. using the same config format instead of passing password in the command line), which does not seem the case.

cevian:
@roidelapluie we are more than happy to implement any suggestions. Can you please expand on which config format you mean?

roidelapluie:
As said, I have no strong opinion but I think it could be useful to be able to reuse prometheus remote_write and remote_read config format for such a tool.  But that is not a blocker to me, just a suggestion, and I do not know if that would be practical at all for your tool. This is 1000 feet comment :)

PR #1865, opened 2021-02-10
Update integrations.md

RichiH:
Thoughts @cevian @juliusv @roidelapluie ?

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -35,6 +35,8 @@ is primarily intended for long term storage. It is recommended that you perform
 careful evaluation of any solution in this space to confirm it can handle your
 data volumes.
 
+### Storage backends
+
   * [AppOptics](https://github.com/solarwinds/prometheus2appoptics): write
   * [AWS Timestream](https://github.com/dpattmann/prometheus-timestream-adapter): read and write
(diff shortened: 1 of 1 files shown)


roidelapluie:
We have a warning in the page:

> The remote write and remote read features of Prometheus allow transparently sending and receiving samples. This is primarily intended for long term storage. It is recommended that you perform careful evaluation of any solution in this space to confirm it can handle your data volumes.


Somehow I think we want to make something similar for the tools, mentioning that they are third party and that we recomment users to evaluate them carefully.

RichiH:
That's a fair point; in a call now, do you want to put something in?

cevian:
I think this change makes sense. Truthfully, I wasn't sure if the place I put prom-migrator originally fit but didn't see a better place and didn't want to create a new section myself.

As for the disclaimer, I wonder if there is /any/ section here this doesn't apply to -- and if not why not put it at the top of the page instead of for each section. I think it's fine either way --- just wanted to point out putting at at the top may have less duplication.

jan--f:
Hello from the bug scrub!
The URL proposed here returns a 404 right now and the project is deprecated. Hence I'll close this for now. Please feel free to reopen if still needed.

PR #1871, opened 2021-02-15
first commit to add sighup as partner

mbuccarello:
Hi, maintainers

I'm trying to create a pull request to add SIGHUP to the partner list. Please @RichiH let me know if everything is fine for you.

I have tried to run locally the site but there is an issue with the alertmanager-0.19 tag, because the script seems unable to download that tag.

Thanks for your time and help.

Michele


content/support-training.html
```diff
@@ -90,6 +90,13 @@ <h2>Commercial support</h2>
               </a>
           </div>
 
+          <div class="col-md-3 col-sm-6 doc-content tc-col">
+
+            <a href="https://sighup.io/" class="btn btn-training btn-default btn-large" target="_blank">
```
juliusv, line 95: Is there a connection to Prometheus support here that could be clarified through a more specific link? From just looking at the homepage, it seems all about a Kubernetes distribution / Kubernetes management.

I realize some others in this list are also a bit unspecific in their links, but I'm wondering whether we should ask for / require more Prometheus-focussed info, since this is what people looking at this page will be wondering about as well.

PR #1876, opened 2021-02-17
Update integrations.md

luebken:
Adding Instana as we've added support for remote_write.


juliusv:
Sorry, I'm late :grimacing: Thanks!

PR #1878, opened 2021-02-18
Blog Post - Introducing the '@' Modifier

codesome:
To keep the momentum from the 2.25 release, it would be nice if this can go in today.

### DIFF
--- a/content/blog/2021-02-18-introducing-the-@-modifier.md
+++ b/content/blog/2021-02-18-introducing-the-@-modifier.md
@@ -1,0 +1,28 @@
+---
+title: Introducing the '@' Modifier
+created_at: 2021-02-18
+kind: article
+author_name: Ganesh Vernekar
+---
+
+In Prometheus v2.25.0, we have introduced a new PromQL modifier `@`. Similar to how `offset` modifier lets you offset the evaluation of vector selector, range vector selector, and subqueries by a fixed duration relative to the evaluation time, the `@` modifier lets you fix the evaluation for those selectors irrespective of the query evaluation time.
+
+    <vector-selector> @ <timestamp>
+    <range-vector-selector> @ <timestamp>
+    <subquery> @ <timestamp>
+
+The `<timestamp>` is a unix timestamp and described with a float literal.
+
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
LGTM

roidelapluie (approved):

RichiH (changes requested):
As a reader: I don't know why you told me this. What problem do I have which you are solving? This does not need to be a full in-depth exploration of what use cases you're solving, but it needs more substance as to "why do I care?" to capture the reader.

codesome:
@RichiH how about now?

gouthamve (approved):

RichiH (changes requested):
This is looking good. Moving the new section to the top as an intro makes sense.

The basic recipe to lead in with in such posts is "Up to now when using Prometheus, you couldn't do X. Now you can do X because we did Y. This solves Y. Let's look at how we did Z"

codesome:
@RichiH updated

content/blog/2021-02-18-introducing-the-@-modifier.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Introducing the '@' Modifier
+created_at: 2021-02-18
+kind: article
+author_name: Ganesh Vernekar
+---
+
+Currently, the `topk()` query only makes sense as an instant query where you get exactly `k` results, but when you run it as a range query, you can get much more than `k` results since every step is evaluated independently. This `@` modifier lets you fix the ranking for all the steps in a range query.
```
RichiH, line 8:
```suggestion
Have you ever selected the top 10 time series for something, but instead of 10 you got 100? If yes, this one is for you. Let me walk you through what the underlying problem is and how I fixed it.

Currently, the `topk()` query only makes sense as an instant query where you get exactly `k` results, but when you run it as a range query, you can get much more than `k` results since every step is evaluated independently. This `@` modifier lets you fix the ranking for all the steps in a range query.
```

RichiH (approved):
Would suggest the intro, else ship it.

PR #1880, opened 2021-02-20
Add new exporter to exporters.md

italux:
## Description
Add [Sentry Exporter](https://github.com/italux/sentry-prometheus-exporter) into exporters list


italux:
@RichiH could you please check this out

italux:
@juliusv @RichiH Could you please review this doc update?

juliusv:
@italux Hi, sorry for the radio silence!

We usually just check whether added exporters follow best practices such as https://prometheus.io/docs/practices/naming/#metric-and-label-naming and a few more general best practices.

Would it be possible to get some of the metrics in the exporter to follow the conventions a bit better?

* `sentry_rate_limit_events_sec`: Clarify and spell out unit: `sentry_rate_limit_events_per_second`
* `sentry_issues`: As far as I can tell by looking at the code, this is using a histogram metric type under the hood, but it's not used like a valid Prometheus histogram, as it manually sets the `le` labels to values that cannot be parsed as a float (e.g. `24h` instead of just the number of seconds corresponding to that). I would recommend doing either of:
  * Switching it to a normal gauge metric family instead and not use the special `le` histogram label for non-numeric values.
  * Or you could also keep it a histogram and use the `le` label as intended, but then you'd also have to put the time unit (seconds) into the metric name somewhere (so the user knows that the duration in the `le` label is in seconds). It also seems from the code like there's `_gcount` and `_gsum` time series that I guess should just be called `_sum` and `_count`?

Overall it would be very helpful to see a sample metrics output (either in the exporter docs or in the PR) to get an easier idea of what the metrics actually look like.

jan--f:
Hello from the bug scrub!
Thank you for your contribution. I'm afraid we need to decline this for now. We are happy to feature exporters that are actively developed and used. Unfortunately this one is no longer maintained, has so far only a limited community behind it and seemingly low adoption. Please try again at a later stage.

PR #1881, opened 2021-02-22
Remove aaS entry from software listOD

RichiH:

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -51,7 +51,6 @@ data volumes.
   * [IRONdb](https://github.com/circonus-labs/irondb-prometheus-adapter): read and write
   * [Kafka](https://github.com/Telefonica/prometheus-kafka-adapter): write
   * [M3DB](https://m3db.github.io/m3/integrations/prometheus): read and write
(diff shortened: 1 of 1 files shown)


roidelapluie (changes requested):
I am blocking this to make sure we first notice them about this and explain them why we do this. I think it should also be part of the commit message.

roidelapluie:
What about newrelic?

RichiH:
I disagree that we are required to keep third parties informed about changes in all cases, but will do so. Do you have contact direct information? CC @ComradeAkko in any case.

New Relic is a distinct piece of software. That there is only one offering by a commercial vendor stands to reason. What would your suggestion be?

roidelapluie:
The split could be : software you install vs software you consume.

RichiH:
That change would be out of scope for the current PR, though. My main intention is to make the list correct as it is defined right now.

roidelapluie (approved):

roidelapluie:
LGTM, and we have now pinged them via github.

roidelapluie:
> That change would be out of scope for the current PR, though. My main intention is to make the list correct as it is defined right now.

From that perspective that pull request is unambihuous and non controversial. Merged.

PR #1883, opened 2021-02-23
Fix the M3DB integration link

lbordowitz:
Signed-off-by: Larry Bordowitz <larry@gomerits.com>

@RichiH  I think you're the right person to contact.


roidelapluie:
Thanks!

PR #1889, opened 2021-02-27
Fix typo in exporter name

lkiesow:
This patch fixes a typo in the name “Elasticsearch” which was written with an uppercase `S`, while the product [is written with a lowercase `s`](https://www.elastic.co/elasticsearch/).


roidelapluie:
Thanks!

PR #1892, opened 2021-03-03
Load all used weights and styles of the Open Sans font

dnrce:
Currently the stylesheets are using regular (400) and bold (700) weights of the Open Sans font, plus regular italic, but only the regular-weight normal style is being loaded from Google Fonts. In addition to missing intended emphasis, this can cause "bold" text to appear blurry as the browser attempts to make do without an actual bold font, as seen here in "Between two scalars":

blurry

This PR adds the `400italic` and `700` variants to the Google Fonts request.

Before:

before

After:

after

Note that the "after" is pretty dramatically different and won't necessarily be what everyone wants, but it is allowing the styles to come through as they're currently written. If it's then desirable to adjust from there, the styles themselves should be revised.

### DIFF
--- a/layouts/header.html
+++ b/layouts/header.html
@@ -57,7 +57,7 @@
 
     <!-- Custom Fonts -->
     <link href="/assets/font-awesome-4.7.0/css/font-awesome.min.css" rel="stylesheet" type="text/css">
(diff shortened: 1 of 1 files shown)


roidelapluie:
Thanks! I never noticed that because I have the font on my laptop...

roidelapluie:
Thanks! This looks great!

dnrce:
Thanks for the quick review!

PR #1894, opened 2021-03-04
Add speedtest-exporter

### LINKED ISSUE #1893: Add speedtest-exporter
Hey,

Me and @Doacola have created a [speedtest-exporter](https://github.com/MiguelNdeCarvalho/speedtest-exporter) written in `Python`, that uses the official `Speedtest CLI`, that shows accurate values. Most of the exporters, are using `speedtest-cli`, that is unofficial and was reported by some users that can show some wrong numbers sometimes.

Thanks,
MiguelNdeCarvalho

MiguelNdeCarvalho:
closes #1893


hdost (approved):
LGTM @juliusv

beorn7:
I think we should allow multiple entries for the same problem with some explanation attached about the different trade-offs.
However, that's clearly at odds with our explicitly stated practice in the past. That's why I think we should first discuss this topic at a dev summit before changing our practice.

RichiH:
Given current votes, it seems likely we'll get to discuss it at the next dev summit on 2022-07-28.

juliusv:
Sounds good to me! @MiguelNdeCarvalho sorry for the delay, but then for good measure I also think it's best to discuss it formally first, as it has been a contentious point for a while.

jan--f:
Hello from the bug scrub!
Sorry for the lag here. Seems like the project is no longer actively maintained. Please re-open this is I misread the situation.

PR #1898, opened 2021-03-09
Add the public Google calendar to the community page

beorn7:


beorn7:
@RichiH even the _new_ calendar event for the dev-summit is _not_ in this publicly visible calendar (in which all the other public events have ended up by now). Instead, it is in a mysterious calendar "Prometheus Community Calendar", which I seem to be unable to access. Could you please move the current new public dev-summit event from "Prometheus Community Calendar" to the [Public Prometheus Calendar](https://calendar.google.com/calendar/render?cid=cHJvbWV0aGV1cy5pb19iZGY5cWdtMDgxbnJkMGZlMzJnM29sc2xkMEBncm91cC5jYWxlbmRhci5nb29nbGUuY29t)? Thank you.

beorn7:
@RichiH and if that's in any way possible, could you check if there are any other events in the "Prometheus Community Calendar", move those as appropriate, and delete that calendar for good?

RichiH:
This took some digging...

What's in the PR is the internal calendar ID within the Prometheus org and according to various Google docucmentation, should not be used externally. Yet, that's what you get when clicking "Get shareable link".

The external calendar ID is prometheus.io_97r8b57tmmchdcl3527ge58vm0@group.calendar.google.com and should be used as such:
* Web UI: https://calendar.google.com/calendar/embed?src=prometheus.io_97r8b57tmmchdcl3527ge58vm0%40group.calendar.google.com -- this shows a webpage
* Embed https://calendar.google.com/calendar/ical/prometheus.io_97r8b57tmmchdcl3527ge58vm0%40group.calendar.google.com/public/basic.ics -- this DOWNLOADS the .ics unless you manually add it.

So, all in all, I think we should use the link by @beorn7 and add links to the others as well for people not using GCal, which I will put into the same PR.

roidelapluie:
https://calendar.google.com/calendar/embed?src=prometheus.io_bdf9qgm081nrd0fe32g3olsld0%40group.calendar.google.com&ctz=Europe%2FParis is the public calendar

RichiH:
You're right. I deleted the Community calendar to avoid future mistakes.

I figured out how to use the external ID of the Public calendar for automagic adding to GCal: https://calendar.google.com/calendar/render?cid=prometheus.io_bdf9qgm081nrd0fe32g3olsld0@group.calendar.google.com . To avoid leaking the internal ID, I will force push over @beorn7 's commit in a minute or two.

RichiH:
"Public" is in CE[S]T, "Community" was in UTC. I will try to get feedback on IRC or on -team if we can switch that to UTC...

RichiH:
@juliusv created that cal and was OK moving to UTC, so I did.

Getting rid of the old, internal invite is harder than it should be:
image
I want to scream.

roidelapluie:
Maybe I would make it a list, but LGTM if you think it's fine as is.

beorn7:
Wow, what a ride! Thanks, both of you.

PR #1899, opened 2021-03-09
Add GSuite to onboarding

beorn7:
_This is an informational change and therefore doesn't go through the
formal procedure of a governance change._

This also mentions the team's calendar. Note that those calendars are
notoriously difficult to find, so whoever onboards the new team member
should give them a direct link to it. (But I didn't want to get more
technical than I did in this PR.)

### DIFF
--- a/content/governance.md
+++ b/content/governance.md
@@ -169,6 +169,7 @@ The new member is
 * added to the [GitHub organization][gh] as _Owner_.
 * optionally added to the community, junkyard, and related organizations and repositories.
 * added to the [team mailing list][team].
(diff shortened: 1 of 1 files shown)


roidelapluie (approved):
The new member should be able to choose what comes before @

beorn7:
I tried to add the user name thing. Does that work for you?

roidelapluie (approved):
Yes, thanks :)

RichiH (approved):
It's called "Google Workspace" until the next VP takes over that part of Google and they rename it again. Not that I care about that name any more and GSuite is fine.

Agreed that it does not require the full dance for updating.

PR #1910, opened 2021-03-25
Describe work-around for negative observations

beorn7:
Interestingly, we had documented the problem here already (but at a
location where users of the instrumentation libraries probably won't
find it easily – I'll add links to here from the instrumentation
libraries later).

The only thing I want to add now is at least a hint towards a
work-around.


roidelapluie (approved):

matthiasr (approved):

RichiH:
Thanks!

PR #1911, opened 2021-03-26
Added GitLab to webhook integrations list

nagyv:
@simonpasquier please, check and merge.

The PR adds GitLab as a webhook integration. Prometheus Alerts can be surfaced into GitLab and a corresponding alert is created within GitLab as per the linked docs.


SuperQ (approved):
LGTM, Thanks.

PR #1916, opened 2021-03-30
Link to promcon

roidelapluie:


juliusv:
IMO the styling of this doesn't quite fit in and the page looks a bit broken now... we already had a newsbar style in the past, will send an update PR.

PR #1917, opened 2021-03-31
Bring back old newsbar styling for PromCon banner

juliusv:


juliusv:
I'm going to merge this already, since I think the current look makes our homepage look a bit too broken.

roidelapluie:
Thanks. I think May 3 instead of May 3rd would be even better

PR #1920, opened 2021-04-12
Update exporters to include Pure Storage exporters.md

genegr:
Added reference to Pure Storage exporter for FlashArray and FlashBlade.
Written using the Python Prometheus client, the Pure exporter is a quite consolidated project since three years and is widely used to monitor Pure appliances in several enterprise contexts.

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -114,6 +114,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Hadoop HDFS FSImage exporter](https://github.com/marcelmay/hadoop-hdfs-fsimage-exporter)
    * [Lustre exporter](https://github.com/HewlettPackard/lustre_exporter)
    * [NetApp E-Series exporter](https://github.com/treydock/eseries_exporter)
+   * [Pure Storage exporter](https://github.com/PureStorage-OpenConnect/pure-exporter)
    * [ScaleIO exporter](https://github.com/syepes/sio2prom)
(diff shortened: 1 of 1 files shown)


roidelapluie:
```
  - targets: [ mypureflashblade-01.lan ]
    labels:
      __pure_apitoken: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```

This is unsecure, since that token will be displayed in the UI. Did you consider configuring the token in the exporter itself, or as part of the `authorization` parameter of prometheus, then passing it again to the actual target?

RichiH:
Passing auth from Prometheus is also not recommended; what if you need
to auth to get to the exporter, and the auth behind it is different?

I recommend making an `auth` section in the exporter config, and
passing this as a URL parameter

genegr:
That is already the suggested method to use the exporter. The current
version still supports both the options: the old one using the GET request
parameter that was chosen mainly to condense multiple FlashArray/FlashBlade
targets into the same Prometheus job and the new preferred one that uses
the Authorization header. I am also thinking to deprecate the old method
and remove it in the 2.0 version which uses the Pure REST API 2.x.

RichiH:
Would it be an option to deprecate the old mechanism in code and documentation now, and remove it with the next major release? That would give your users both more time to migrate, and a strong signal to not use the insecure legacy path.

genegr:
> Would it be an option to deprecate the old mechanism in code and documentation now, and remove it with the next major release? That would give your users both more time to migrate, and a strong signal to not use the insecure legacy path.

Done. Now the README stated the old way of using the apitoken GET parameter is deprecated and will be removed from the next major release.

RichiH:
Thanks!

PR #1922, opened 2021-04-14
Add LabyrinthLabs to commercial support

martinhaus:
Add LabyrinthLabs to commercial support page.


martinhaus:
@juliusv

juliusv:
@martinhaus :+1: thanks!

Any chance you have a vector version of the logo, or a PNG with transparent background? It'll just make it look slightly nicer when hovering over it, since the button background turns grey on hover, but the image currently has a hardcoded white background. No requirement though.

Also, in case you have a more Prometheus-specific landing page to link to (now or in the future), that's encouraged too. We didn't require it for others so far and thus I don't want to require it for new vendors at the moment, but could happen at some point.

juliusv:
Btw. you can see the preview deploy at https://deploy-preview-1922--prometheus-docs.netlify.app/support-training/

martinhaus:
Hey @juliusv, thanks for your suggestions!
I've replaced the logo with an SVG version. It resolves the issue with grey background on hover.
For the Prometheus-specific landing page - we currently don't have one but will look into creating one in the future. If that's okay with you, I'd like this PR to be merged with the generic landing page and then once the new Prometheus landing page on our side is ready, I'll open another one with the corresponding change in the target URL.

Thank you.

juliusv:
Perfect, thanks!

Yeah, that's fine :)

PR #1925, opened 2021-04-17
update histograms.md

viktoryi:
Seems it's a descriptive error.

Signed-off-by: Viktor Yi <viktoryi@163.com>

### DIFF
--- a/content/docs/practices/histograms.md
+++ b/content/docs/practices/histograms.md
@@ -69,7 +69,7 @@ a histogram called `http_request_duration_seconds`.
 
 You can approximate the well-known [Apdex
 score](http://en.wikipedia.org/wiki/Apdex) in a similar way. Configure
-a bucket with the target request duration as the upper bound and
+a bucket with the target request duration as the target bound and
 another bucket with the tolerated request duration (usually 4 times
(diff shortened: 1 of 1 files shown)


RichiH:
Our buckets are `le`; thus "upper, inclusive". The phrasing seems to be correct, to me.

SuperQ:
The wording is correct as-is.

viktoryi:
Here is the document for now, and there are two `upper bound` in this sentence. That's why I'm confused.
> a bucket with the target request duration as the upper bound and another bucket with the tolerated request duration (usually 4 times the target request duration) as the upper bound

viktoryi:
Here is the document for now, and there are two `upper bound` in this sentence. That's why I'm confused.
> a bucket with the target request duration as the upper bound and another bucket with the tolerated request duration (usually 4 times the target request duration) as the upper bound

beorn7:
Oops. In fact, "upper bound" is mentioned twice in an attempt to be _less_ confusing and _more_ explicit. (o:

The "upper bound" is not coming from APDEX but from the Prometheus histogram, where an "upper bound" is what you configure for each bucket. If there's a better way to phrase it, suggestions are welcome.

PR #1928, opened 2021-04-24
Add Unit Modifier Section to Naming Best Practices

LeviHarrison:
This PR adds a section to the naming best practices for unit modifiers. Related to prometheus/prometheus#8754.

Fixes prometheus/prometheus#8718

### DIFF
--- a/content/docs/practices/naming.md
+++ b/content/docs/practices/naming.md
@@ -42,7 +42,10 @@ A metric name...
  * request duration
  * bytes of data transfer
  * instantaneous resource usage as a percentage
-
+* ...should add the unit modifier, like `failed`, before the metric name if needed.
+ * <code>foobar\_<b>failed</b>\_requests_total</code>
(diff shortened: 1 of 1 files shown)


roidelapluie:
I am converting this to draft as it requires more discussions.

RichiH:
Unless I am misunderstanding, this should be a label, no?

roidelapluie:
It should not, especially for failed stuff. A label would make sense if you can sum accross all labels, while keeping a consistent labelset.

If you have e.g. http_requests_failed_total, it can not have a `code`  label like a `http_requests_total` would have. 

In the prometheus code, we also have:

net_conntrack_dialer_conn_failed_total
net_conntrack_dialer_conn_established_total
net_conntrack_dialer_conn_closed_total
net_conntrack_dialer_conn_attempted_total

But labels are not appropriate because the same request would end up in multiple instances of the metric (e.g. attemped, established and closed), so a sum() across labels would be maningless.

PR #1939, opened 2021-04-27
Add InfiniBand exporter

treydock:
The main purpose is to collect InfiniBand counters from switches that may not support SNMP.  Our IB fabric switches are all "dumb" so they can only expose counters via `perfquery` or other methods that talk to them over IB protocols.  The opt-in collection of HCA metrics is to collect metrics from HCAs in the event a site doesn't want to install node_exporter on their IB hosts.

@RichiH


juliusv:
:+1: Thanks!

PR #1957, opened 2021-05-26
Delegate a matrix server

SuperQ:
Delegate matrix.org as the Matrix servie for prometheus.io.

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/content/.well-known/matrix/server
+++ b/content/.well-known/matrix/server
@@ -1,0 +1,5 @@
+---
+filename: server
+---
+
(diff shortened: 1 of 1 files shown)


RichiH:
Sgtm

SuperQ:
First I have to figure out how to Nanoc.

RichiH:
Any update here, or is this stale?

SuperQ:
I think the consensus was that we don't need/want to do this because it makes all matrix channels look like they're part of prometheus.io.

PR #1969, opened 2021-06-10
Test our blog

RichiH:


RichiH:
Test failed successfully.

RichiH:
@juliusv is this something we can make our blog do?

roidelapluie:
did you try `<a>`

roidelapluie:
not-as-easy-as it looks, does this propagate to RSS?

juliusv:
I mean YAML frontmatter is decidedly not Markdown. But we could either encode some HTML in there (no idea what currently happens if you do that, if it's escaped or not), or we could have separate fields for author name and author URL, and then combine those in the output somehow.

RichiH:
Works.

PR #1972, opened 2021-06-10
Marketing

RichiH:

### DIFF
--- a/content/blog/2018-08-09-prometheus-graduates-within-cncf.md
+++ b/content/blog/2018-08-09-prometheus-graduates-within-cncf.md
@@ -2,7 +2,7 @@
 title: Prometheus Graduates Within CNCF
 created_at: 2018-08-09
 kind: article
-author_name: Richard Hartmann
+author_name: '<a href="https://twitter.com/TwitchiH">Richard "RichiH" Hartmann</a>'
 ---
 
 We are happy to announce that as of today, Prometheus graduates within the [CNCF](https://www.cncf.io/).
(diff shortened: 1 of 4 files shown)


juliusv:
This puts the following into the feed at https://deploy-preview-1972--prometheus-docs.netlify.app/blog/feed.xml:

```xml
<author>
  <name><a href="https://twitter.com/TwitchiH">Richard "RichiH" Hartmann</a></name>
  <uri>https://prometheus.io/blog/</uri>
</author>
```

It would probably be better to set a separate field for the URL and put that into the `<uri>` XML tag for the author, rather than just pointing back to the blog.

RichiH:
Meh, I had hoped to bypass the need for any actual changed. Otoh, this
allows us to migrate to main at the same time.

juliusv:
Btw. I have no idea how this works, but my hunch is that https://github.com/prometheus/docs/blob/master/content/blog/feed.erb basically defines *defaults* for the author name and URL, so maybe if you just set the `uri` field in your YAML frontmatter (instead of putting the link into the `author` field, maybe it will just magically already do the right thing. And then in https://github.com/prometheus/docs/blob/master/content/blog/index.html we would just also output the URL as a link properly if it's set.

jan--f:
Hello from the bug scrub!
Is this still something we want to add/change?

jan--f:
Hello from the bug scrub!

Closing this for now as abandoned. Please feel free to reopen at any time.

PR #1975, opened 2021-06-14
Add HTTP SD recommendations

roidelapluie:

### DIFF
--- a/content/docs/instrumenting/http_sd.md
+++ b/content/docs/instrumenting/http_sd.md
@@ -1,0 +1,89 @@
+---
+title: HTTP SD
+sort_rank: 7
+---
+
+# Writing HTTP Service Discovery
+
+Prometheus provides a generic [HTTP Service Discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#http_sd_config),
+that enables to discover targets over an HTTP endpoint.
+
+The HTTP Service Discovery is complimentary to the supported service
+discoveries, and is an alternative to the [File-based Service Discovery](https://prometheus.io/docs/guides/file-sd/#use-file-based-service-discovery-to-discover-scrape-targets).
+
+## Comparison between File-Based SD and HTTP SD
+
+Here is a table comparing our two generic Service Discovery implementations.
+
+| Item | File SD | HTTP SD |
+| Event Based | Yes, via inotify | No |
+| Update frequency | Instant, thanks to inotify | Following refresh_interval |
+| Format | Yaml or JSON | JSON |
+| Transport | Local file | HTTP/HTTPS |
+| Security | File-Based security | TLS, Basic auth, Authorization header, OAuth2 |
+
+## Requirements of HTTP SD endpoints
+
+The HTTP SD endpoints should have been written specifically for Prometheus.
+The response is consumed as is, unmodified. On each refresh interval (default: 1
+minute), Prometheus will do a GET request to the HTTP SD endpoint. The GET
+request contains a `X-Prometheus-Refresh-Interval-Seconds` HTTP header with the
+refresh interval.
+
+The SD endpoint must answer with an HTTP 200 response, with the HTTP Header
+`Content-Type: application/json`. The answer must be UTF-8 formatted.
+In the case when no targets are returned, an HTTP 200 must also be emitted, with
+an empty list `[]`.
+
+It is expected that the whole list of targets is returned on every scrape. There
+is no client-side caching mechanism. A prometheus instance does not send its
+hostname and it is not possible for a SD endpoint to know if the SD requests
+is the first one after a restart or not.
+
+The URL to the HTTP SD is not considered secret. The authentication, and any API
+keys should be passed with the appropriate authentication mechanisms. Prometheus
+supports TLS authentication, basic authentication, oauth2, and authorization
+headers.
+
+## HTTP_SD format
+
+```json
+[
+  {
+    "targets": [ "<host>", ... ],
+    "labels": {
+      "<labelname>": "<labelvalue>", ...
+    }
+  },
+  ...
+]
+```
+
+
+Examples:
+
+```json
+[
+    {
+        "targets": ["10.0.10.2:9100", "10.0.10.3:9100", "10.0.10.4:9100", "10.0.10.5:9100"],
+        "labels": {
+            "__meta_datacenter": "london",
+            "__meta_prometheus_job": "node"
+        }
+    },
+    {
+        "targets": ["10.0.40.2:9100", "10.0.40.3:9100"],
+        "labels": {
+            "__meta_datacenter": "london",
+            "__meta_prometheus_job": "alertmanager"
+        }
+    },
+    {
+        "targets": ["10.0.40.2:9093", "10.0.40.3:9093"],
+        "labels": {
+            "__meta_datacenter": "newyork",
+            "__meta_prometheus_job": "alertmanager"
+        }
+    }
+]
+```
--- a/content/docs/instrumenting/http_sd.md
+++ b/content/docs/instrumenting/http_sd.md
@@ -36,7 +36,7 @@ refresh interval.
 The SD endpoint must answer with an HTTP 200 response, with the HTTP Header
 `Content-Type: application/json`. The answer must be UTF-8 formatted.
 If no targets should be transmitted, HTTP 200 must also be emitted, with
-an empty list `[]`.
+an empty list `[]`. The list of unordered.
 
 Prometheus caches target lists. If an error occurs while fetching an updated
 targets list, Prometheus keeps using the current targets list. The targets list
--- a/content/docs/instrumenting/http_sd.md
+++ b/content/docs/instrumenting/http_sd.md
@@ -6,10 +6,10 @@ sort_rank: 7
 # Writing HTTP Service Discovery
 
 Prometheus provides a generic [HTTP Service Discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#http_sd_config),
-that enables to discover targets over an HTTP endpoint.
+that enables it to discover targets over an HTTP endpoint.
 
 The HTTP Service Discovery is complimentary to the supported service
-discoveries, and is an alternative to the [File-based Service Discovery](https://prometheus.io/docs/guides/file-sd/#use-file-based-service-discovery-to-discover-scrape-targets).
+discovery mechanisms, and is an alternative to the [File-based Service Discovery](https://prometheus.io/docs/guides/file-sd/#use-file-based-service-discovery-to-discover-scrape-targets).
 
 ## Comparison between File-Based SD and HTTP SD
 
@@ -25,7 +25,9 @@ Here is a table comparing our two generic Service Discovery implementations.
 
 ## Requirements of HTTP SD endpoints
 
-The HTTP SD endpoints should have been written specifically for Prometheus.
+If you implement an HTTP SD endpoint, here is a few requirements you should be
+aware of.
+
 The response is consumed as is, unmodified. On each refresh interval (default: 1
 minute), Prometheus will do a GET request to the HTTP SD endpoint. The GET
 request contains a `X-Prometheus-Refresh-Interval-Seconds` HTTP header with the
@@ -33,11 +35,12 @@ refresh interval.
 
 The SD endpoint must answer with an HTTP 200 response, with the HTTP Header
 `Content-Type: application/json`. The answer must be UTF-8 formatted.
-In the case when no targets are returned, an HTTP 200 must also be emitted, with
+If no targets should be transmitted, HTTP 200 must also be emitted, with
 an empty list `[]`.
 
-If an error occurs while fetching the targets list, the current targets list is
-kept. The targets list is not saved across restart.
+Prometheus caches target lists. If an error occurs while fetching an updated
+targets list, Prometheus keeps using the current targets list. The targets list
+is not saved across restart.
 
 The whole list of targets must be returned on every scrape. There is no support
 for incremental update. A prometheus instance does not send its hostname and it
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,90 @@
...
+request contains a `X-Prometheus-Refresh-Interval-Seconds` HTTP header with the
+refresh interval.
+
+The SD endpoint must answer with an HTTP 200 response, with the HTTP Header
+`Content-Type: application/json`. The answer must be UTF-8 formatted.
+In the case when no targets are returned, an HTTP 200 must also be emitted, with
+an empty list `[]`.
+
+It is expected that the whole list of targets is returned on every scrape. There
+is no client-side caching mechanism. A prometheus instance does not send its
```
SuperQ, line 40: Does this imply that a failure (4xx, 5xx, TCP RST) will empty any existing discovery data? If any failure results in no targets, we should probably call this out.

roidelapluie: It should keep the previous targets set. I will update.

SuperQ: Rather than "no client-side cache", I would say "It does not support incremental updates".

SuperQ: Also "It does not store data between restarts"

roidelapluie:
@SuperQ  I have improved this based on your feedback

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,93 @@
+---
+title: HTTP SD
+sort_rank: 7
+---
+
+# Writing HTTP Service Discovery
+
+Prometheus provides a generic [HTTP Service Discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#http_sd_config),
+that enables to discover targets over an HTTP endpoint.
```
RichiH, line 9:
```suggestion
that enables it to discover targets over an HTTP endpoint.
```

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,93 @@
...
+
+The HTTP Service Discovery is complimentary to the supported service
+discoveries, and is an alternative to the [File-based Service Discovery](https://prometheus.io/docs/guides/file-sd/#use-file-based-service-discovery-to-discover-scrape-targets).
```
RichiH, line 12:
```suggestion
discovery mechanisms, and is an alternative to [File-based Service Discovery](https://prometheus.io/docs/guides/file-sd/#use-file-based-service-discovery-to-discover-scrape-targets).
```

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,93 @@
...
+| ---- | ------- | ------- |
+| Event Based | Yes, via inotify | No |
+| Update frequency | Instant, thanks to inotify | Following refresh_interval |
+| Format | Yaml or JSON | JSON |
+| Transport | Local file | HTTP/HTTPS |
+| Security | File-Based security | TLS, Basic auth, Authorization header, OAuth2 |
+
+## Requirements of HTTP SD endpoints
+
+The HTTP SD endpoints should have been written specifically for Prometheus.
```
RichiH, line 28: That's unclear.

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,93 @@
...
+The response is consumed as is, unmodified. On each refresh interval (default: 1
+minute), Prometheus will do a GET request to the HTTP SD endpoint. The GET
+request contains a `X-Prometheus-Refresh-Interval-Seconds` HTTP header with the
+refresh interval.
+
+The SD endpoint must answer with an HTTP 200 response, with the HTTP Header
+`Content-Type: application/json`. The answer must be UTF-8 formatted.
+In the case when no targets are returned, an HTTP 200 must also be emitted, with
```
RichiH, line 36:
```suggestion
If no targets should be transmitted, HTTP 200 must also be emitted, with
```

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,93 @@
...
+an empty list `[]`.
+
+If an error occurs while fetching the targets list, the current targets list is
+kept. The targets list is not saved across restart.
```
RichiH, line 40:
```suggestion
Prometheus caches target lists. If an error occurs while fetching an updated targets list, Prometheus keeps using the current targets list. The targets list is not saved across restart.
```

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,93 @@
...
+
+The whole list of targets must be returned on every scrape. There is no support
+for incremental update. A prometheus instance does not send its hostname and it
```
RichiH, line 43:
```suggestion
for incremental updates. A Prometheus instance does not send its hostname and it
```

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,93 @@
...
+is not possible for a SD endpoint to know if the SD requests is the first one
+after a restart or not.
+
+The URL to the HTTP SD is not considered secret. The authentication, and any API
+keys should be passed with the appropriate authentication mechanisms. Prometheus
+supports TLS authentication, basic authentication, oauth2, and authorization
```
RichiH, line 49:
```suggestion
supports TLS authentication, basic authentication, OAuth2, and authorization
```

RichiH (changes requested):

content/docs/instrumenting/http_sd.md
roidelapluie: What I try to say really is that we expect this to work only for endpoints crafted for this purpose. We don't aim to support e.g. file servers that would not provide appropriate content types etc...

roidelapluie: How can we better phrase it?

roidelapluie:
Thanks @RichiH , ready for a next round of reviews.

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,96 @@
...
+| Transport | Local file | HTTP/HTTPS |
+| Security | File-Based security | TLS, Basic auth, Authorization header, OAuth2 |
+
+## Requirements of HTTP SD endpoints
+
+If you implement an HTTP SD endpoint, here is a few requirements you should be
+aware of.
+
+The response is consumed as is, unmodified. On each refresh interval (default: 1
+minute), Prometheus will do a GET request to the HTTP SD endpoint. The GET
```
RichiH, line 32:
```suggestion
minute), Prometheus will perform a GET request to the HTTP SD endpoint. The GET
```

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,96 @@
...
+request contains a `X-Prometheus-Refresh-Interval-Seconds` HTTP header with the
```
RichiH, line 33: Why? What do you suggest implementers do with it?

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,96 @@
...
+
+| Item | File SD | HTTP SD |
+| ---- | ------- | ------- |
+| Event Based | Yes, via inotify | No |
+| Update frequency | Instant, thanks to inotify | Following refresh_interval |
+| Format | Yaml or JSON | JSON |
+| Transport | Local file | HTTP/HTTPS |
+| Security | File-Based security | TLS, Basic auth, Authorization header, OAuth2 |
+
+## Requirements of HTTP SD endpoints
```
RichiH, line 26: I wonder if we should call this v1 or 1.0.0 and use RFC 2119 language

RichiH (changes requested):

content/docs/instrumenting/http_sd.md
roidelapluie: X-Prometheus-Refresh-Interval-Seconds is 1, they might cache on their side. I don't want to call this explicitly, I just want to guarantee that we will keep this field in the future.

content/docs/instrumenting/http_sd.md
roidelapluie: We can version later. At this point, this is not even released. I would refrain from versioning or being too strict here. It is almost rfc2119, we could just turn lowercases into uppercases later on.

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,96 @@
...
+
+Prometheus caches target lists. If an error occurs while fetching an updated
+targets list, Prometheus keeps using the current targets list. The targets list
+is not saved across restart.
+
+The whole list of targets must be returned on every scrape. There is no support
+for incremental updates. A Prometheus instance does not send its hostname and it
+is not possible for a SD endpoint to know if the SD requests is the first one
+after a restart or not.
+
```
TimSimmons, line 49: random (probably irrelevant) question, need there be any consideration for sorting the results of the HTTP endpoint? would it cause unnecessary updates to the list of targets (potentially resetting scrape timers or something unintended) if the results "changed" by being in a different order?

content/docs/instrumenting/http_sd.md
roidelapluie: I propose we do RFC2119 and versioning in the compliance repository.

content/docs/instrumenting/http_sd.md
```diff
@@ -0,0 +1,96 @@
...
+The response is consumed as is, unmodified. On each refresh interval (default: 1
+minute), Prometheus will perform a GET request to the HTTP SD endpoint. The GET
+request contains a `X-Prometheus-Refresh-Interval-Seconds` HTTP header with the
+refresh interval.
+
+The SD endpoint must answer with an HTTP 200 response, with the HTTP Header
+`Content-Type: application/json`. The answer must be UTF-8 formatted.
+If no targets should be transmitted, HTTP 200 must also be emitted, with
+an empty list `[]`.
```
RichiH, line 39:
```suggestion
an empty list `[]`. The target list is unordered.
```

RichiH (changes requested):
@TimSimmons is right in https://github.com/prometheus/docs/pull/1975#discussion_r651926479 -- easy fix.

content/docs/instrumenting/http_sd.md
```diff
@@ -36,7 +36,7 @@ refresh interval.
 The SD endpoint must answer with an HTTP 200 response, with the HTTP Header
 `Content-Type: application/json`. The answer must be UTF-8 formatted.
 If no targets should be transmitted, HTTP 200 must also be emitted, with
-an empty list `[]`.
+an empty list `[]`. The list of unordered.
```
RichiH, line 39: I would call it target list, but nm.

`s/of/is/` and good to merge.

RichiH (changes requested):

content/docs/instrumenting/http_sd.md
roidelapluie: What about Lists are unordered ?

Because this is a list of  maps that contain a target list.

RichiH: We're on the long tail of the discussion; "Target lists are unordered" would be even more generic; I am fine with any.

RichiH:
Thanks!

juliusv:
Hi,

Two notes:

* Should this have been merged already, given that HTTP SD is not yet released?
* I'm unsure as to why this is placed under the "Instrumenting" section of our docs page? That section in my mind should only be about adding metrics to things / getting metrics out of things. Would it fit somehow under "Operating" or "Best Practices", or do we need to find another place to fit it in?

roidelapluie:
> Hi,
> 
> Two notes:
> 
> * Should this have been merged already, given that HTTP SD is not yet released?

Yes, several people are already working on HTTP SD instrumentation.

> * I'm unsure as to why this is placed under the "Instrumenting" section of our docs page? That section in my mind should only be about adding metrics to things / getting metrics out of things. Would it fit somehow under "Operating" or "Best Practices", or do we need to find another place to fit it in?

My reasoning:

I have put it into instrumenting because it is to instrument other software to offer SD capabilities. I see instrumenting as a section about development.

This is not best practices, as the rules are strict, and if you do not follow them, your SD would not work.

This is not operating, even if the list of software that supports it would go under operating. This HTTP SD development section is more about development, therefore not the same public.

juliusv:
> Yes, several people are already working on HTTP SD instrumentation.

At the moment it contains broken links though (the one to the configuration section on the config page, which doesn't contain the `http_sd_config` yet because 2.28 is not yet released.

> I have put it into instrumenting because it is to instrument other software to offer SD capabilities. I see instrumenting as a section about development.

Instrumenting is about adding metrics to things, and the section has always been about that only (see also https://en.wikipedia.org/wiki/Instrumentation). So I am almost 100% sure that it does not fit into that section.

Maybe we need to come up with a completely new section with information about building non-instrumentation integrations. Or for now, just have an "HTTP SD" page under the existing "Prometheus" point? That has already become kind of a catch-all for Prometheus-server-specific docs, and at least it would fit in nicely in that way.

roidelapluie:
We can't have in under Prometheus as I'd rather have this independant of the Prometheus version.

Best practice seems the best fit then for now?

juliusv:
> We can't have in under Prometheus as I'd rather have this independant of the Prometheus version.

How would it hurt to have it in the Prometheus docs though? If it's version-independent, all that means is that the content won't change between versions. But we do the same for "Federation" and some other points which are pretty version-independent, but thematically tied to the Prometheus server (like HTTP SD).

RichiH:
Versioned makes sense, either way. Putting it under the versioning of Prometheus is the quickest way.

roidelapluie:
Okay. @juliusv would you have the time to do this before the release?
Would be nice to also have an entry here: https://github.com/prometheus/docs/blob/master/content/_redirects

juliusv:
@roidelapluie Yeah. Since it's just docs, do you think it's ok to just put it into the full release right after rc.0, without doing an rc.1?

roidelapluie:
Yes ofc

juliusv:
Thought so. Done in https://github.com/prometheus/docs/pull/1976 and https://github.com/prometheus/prometheus/pull/8972.

PR #1977, opened 2021-06-23
Added Monitoring Studio X as a third-party software exposing Promethe…

iguitton:
@RichiH 

Added Monitoring Studio X as a third-party software exposing Prometheus metrics

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -282,7 +282,7 @@ separate exporters are needed:
    * [MidoNet](https://github.com/midonet/midonet)
    * [midonet-kubernetes](https://github.com/midonet/midonet-kubernetes) (**direct**)
    * [MinIO](https://docs.minio.io/docs/how-to-monitor-minio-using-prometheus.html)
-   * [Monitoring Studio X](https://www.sentrysoftware.com/library/swsyx/prometheus/exposing-patrol-parameters-in-prometheus.html)
+   * [PATROL with Monitoring Studio X](https://www.sentrysoftware.com/library/swsyx/prometheus/exposing-patrol-parameters-in-prometheus.html)
    * [Netdata](https://github.com/firehol/netdata)
    * [Pretix](https://pretix.eu/)
    * [Quobyte](https://www.quobyte.com/) (**direct**)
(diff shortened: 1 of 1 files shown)


RichiH:
Thanks @iguitton !

Taking a quick glance, `skm_volume_readrate_bytes_per_second` is an anti-pattern; is there any chance to expose the raw values like `skm_volume_read_bytes_total` so users can use `rate()` and such in PromQL?

bertysentry:
@RichiH: Unfortunately, the PATROL Agent does not expose the underlying raw values of its metrics. We tried to reconstruct the raw values from the rates, but it's a resource-intensive task for the agent when there are thousands of metrics to convert.

We're still doing a lot of automatic units conversions (e.g. the **FreeSpace** metric is originally expressed in MB, and we convert it to bytes, etc.) but this version of the exporter doesn't reconstruct the raw values from the rate metrics. It's still under consideration for a new version, but it will truly depend on the overall performance of the operation.

content/docs/instrumenting/exporters.md
```diff
@@ -282,6 +282,7 @@ separate exporters are needed:
    * [MidoNet](https://github.com/midonet/midonet)
    * [midonet-kubernetes](https://github.com/midonet/midonet-kubernetes) (**direct**)
    * [MinIO](https://docs.minio.io/docs/how-to-monitor-minio-using-prometheus.html)
+   * [Monitoring Studio X](https://www.sentrysoftware.com/library/swsyx/prometheus/exposing-patrol-parameters-in-prometheus.html)
```
bertysentry, line 285: Suggested title change:
```md
[PATROL with Monitoring Studio X](https://www.sentrysoftware.com/library/swsyx/prometheus/exposing-patrol-parameters-in-prometheus.html)
```

iguitton: @bertysentry, your comment has been taken into account.

iguitton:
@RichiH any update?

RichiH:
@bertysentry Rebuilding counters from rates is error-prone. If your upstream were to consider exposing raw values, that would be best.

@iguitton thanks for the poke & PR, merging.

PR #1981, opened 2021-07-01
Update basic auth guide

roidelapluie:


RichiH (approved):
I am wondering if we should retain the NGINX docs in case they are useful somehow, but I can't see much reason to actually retain them.

roidelapluie:
What I want now is our website to continue to say "we don't support basic auth & TLS".

RichiH:
Which is part of why I made the comment as part of approval. Thanks!

PR #1982, opened 2021-07-02
WIP: Add compaction analysis blog post

fpetkovski:


content/blog/2021-07-10-analyzing-compaction-efficiency.md
```diff
@@ -0,0 +1,62 @@
...
+
+An efficient compaction process would therefore create chunks with a number of samples as close as possible to the maximum count of 120. A chunk with less than 120 samples might, in absolute terms, use less space on disk than a chunk with exactly 120 samples. But in relative terms, a chunk with 120 samples will tend to have a better size per sample ratio than a chunk with less than 120 samples, leading to a reduction in the total amount of chunks as well as a reduction in the total disk space used by Prometheus.
+
+## Visualizing the compaction efficiency
+
+In order to help visualize how well Prometheus is compacting samples, we extended the `promtool tsdb analyze` command to plot a distribution of how full chunks are relative to their maximum capacity. Running the analysis on a TSDB block consuming 4.5 bytes per sample yields the following output.
```
fpetkovski, line 22: Do we need a reference to promtool? I am not sure how well the community is familiar with it.

content/blog/2021-07-10-analyzing-compaction-efficiency.md
```diff
@@ -0,0 +1,62 @@
+---
+title: "Analyzing Prometheus compaction efficiency"
```
bwplotka, line 2:
```suggestion
title: "Analyzing Prometheus Compaction Efficiency"
```
Title case might be better? 🤔

content/blog/2021-07-10-analyzing-compaction-efficiency.md
```diff
@@ -0,0 +1,62 @@
...
+created_at: 2021-07-10
```
bwplotka, line 3: deadline for review? =D

content/blog/2021-07-10-analyzing-compaction-efficiency.md
bwplotka: Yes, we need to - I think it's semi-known, but it would be no harm to reference it (even how to use it, how to build or that you can take it from release binaries)

bwplotka:
```suggestion
In order to help visualize how well Prometheus is compacting samples, we extended the `promtool tsdb analyze` command to plot a distribution of how full chunks are relative to their maximum capacity of 120 samples. Running the analysis on a TSDB block consuming the estimated 4.5 bytes per sample (instead of perfect 1.3 bytes) yields the following output.
```

content/blog/2021-07-10-analyzing-compaction-efficiency.md
```diff
@@ -0,0 +1,62 @@
...
+
+An efficient compaction process would therefore create chunks with a number of samples as close as possible to the maximum count of 120. A chunk with less than 120 samples might, in absolute terms, use less space on disk than a chunk with exactly 120 samples. But in relative terms, a chunk with 120 samples will tend to have a better size per sample ratio than a chunk with less than 120 samples, leading to a reduction in the total amount of chunks as well as a reduction in the total disk space used by Prometheus.
+
+## Visualizing the compaction efficiency
+
+In order to help visualize how well Prometheus is compacting samples, we extended the `promtool tsdb analyze` command to plot a distribution of how full chunks are relative to their maximum capacity. Running the analysis on a TSDB block consuming 4.5 bytes per sample yields the following output.
+
+```
+Fullness: Amount of chunks
```
bwplotka, line 25: https://github.com/prometheus/prometheus/pull/8940#pullrequestreview-697986227

content/blog/2021-07-10-analyzing-compaction-efficiency.md
```diff
@@ -0,0 +1,62 @@
...
+     40%: ####
+     50%: #
+     60%: ##
+     70%: 
+     80%: 
+     90%: 
+    100%: ##########################################################################################
+```
+
+We immediatelly notice that most blocks are at full capacity. The disk usage of this block is just 1.3 bytes per sample.
```
bwplotka, line 56:
```suggestion
We immediatelly notice that most blocks have chunks at full capacity. The disk usage of this block is just 1.3 bytes per sample.
```

content/blog/2021-07-10-analyzing-compaction-efficiency.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Potential reasons for sub-optimal compaction
+
+In the majority of cases, the compaction process is efficient and produces chunks with the ideal size of 1-2 bytes per sample. The efficiency can, however, become degraded when high CPU load is present while compaction is taking place. Prometheus assumes that samples are scraped at regular intervals, but a lack of available CPU can lead to drift in the samples' timestamps. This in turn would reduce the efficience of compacting those samples into chunks.
```
bwplotka, line 60: > The efficiency can, however, become degraded when high CPU load is present while compaction is taking place.

Do you mean in scrape time? I don't think compaction CPU load can cause any issues on how chunks are compacted. What can sometimes cause it is that during compaction we used to NOT compact chunks together (unless they overlap). See: https://github.com/prometheus/prometheus/blob/d95b0972505f61be0b23b4dfedb8b4816392491b/storage/merge.go#L628

So I am not sure what exactly behavior you are trying to explain

bwplotka (commented):
Nice blog post! Some comments. My overall feeling is that it's good to know, but user cannot do much with this situation other than reduce cardinality 🤔 right?

content/blog/2021-07-10-analyzing-compaction-efficiency.md
bwplotka: Oh, another reasons is miconfigured compaction or compaction not processed fast enough etc

fpetkovski: I thought that a drift in the timestamps (e.g. scrapes are not regular), this would cause compaction to become less efficient. Is this an actual problem?

bwplotka: Definitely no. Compaction at the current state does not touch samples unless it's overlap situation (vertical merge that has to be enabled explicitly). 

I am not sure how timestamp drift in scrape time can change this too. I suppose irregular timestamps can reduce gorilla compression ration, but probably marginally.

roidelapluie: Which version of prometheus created those blocks? This should not happen in Prometheus 2.22+

roidelapluie:
Thanks for your contribution.

I think there is a misunderstanding in this blog post. The "fullness" of the chunks is not related to the bytes/sample they take. This is a misleading shortcut.

I am wondering if we have to publish that kind of deep tsdb blockposts in our blog, considering the very high complexity involved.
Even the "reference" tsdb blog by codesome are not in prometheus.io.

It means that this blog post would be "the" reference when you would look for info about chunks etc, and it would be very very confusing for users.

roidelapluie:
To clarify my last comment: I strongly believe that:

1/ if there is such an issue in Prometheus, it is a bug that must be fixed. Users should never run the `promtool tsdb` commands unless they are TSDB developers.
2/ if we publish this on prometheus.io, users will start using those commands and come back with numbers / bug reports / etc that will be hard to tackle. It will also create  a lot of frustration because those commands are not optimized and not targeted at end users.

So I would rather have this published on a 3rd party platform than in our main website.

jan--f:
Hello from the bug scrub!
Given the age of this and the unresolved disagreement, I'll go ahead and close this for now.
Please feel free to reopen at any time.

Issue #1986, opened 2021-07-07
Table of contents conflicts with `NOTE` box


LeviHarrison:
This issue appears in:

* https://prometheus.io/docs/guides/go-application/
* https://prometheus.io/docs/guides/node-exporter/
* https://prometheus.io/docs/guides/basic-auth/
* https://prometheus.io/docs/guides/basic-auth/

(tested in Chrome and Firefox)

![image](https://user-images.githubusercontent.com/54278938/124687018-2878b680-dea2-11eb-961e-a1825c72577a.png)

RichiH:
Thanks for the report, reproduced locally in Chromium.

@grobie @juliusv this seems to be new?

juliusv:
It's because the NOTE block is a `div` which is a `display: block` element, which is 100% width by default and also doesn't seem to care about overlapping the float element.

Adding `display: inline-block; width: 100%;` to the note block seems to help with this particular situation:

![note-inline-block](https://user-images.githubusercontent.com/538008/124835000-e96d6280-df80-11eb-96d9-733b97fd5ed0.png)

I also tried to add a `clear: ...` with different values to the TOC div, but that also then moves the paragraph text out of the way in a way we don't want.

PR #1987, opened 2021-07-08
Fix NOTE block rendering

RichiH:
Fixes: https://github.com/prometheus/docs/issues/1986

### DIFF
--- a/content/css/docs.css
+++ b/content/css/docs.css
@@ -342,9 +342,16 @@ footer {
   font-size: 16px;
 }
 
-.doc-content p, .doc-content.ul, .doc-content .alert {
+.doc-content p, .doc-content.ul {
   margin: 15px 0 15px 0;
   line-height: 1.5;
+}
+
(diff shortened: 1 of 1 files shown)


RichiH:
@juliusv I was uncertain if I should split out `.alert` into its own section or not. Preferences?

content/css/docs.css
```diff
@@ -347,6 +347,12 @@ footer {
   line-height: 1.5;
 }
 
+.doc-content .alert {
```
juliusv, line 350: Could you add a comment line so we remember later why this is here and don't remove it:

```
// This is to prevent NOTE/admonition alert boxes from overlapping the table of contents at the top of a docs page.
```

juliusv:
:+1: Though there may be more block-style elements that could cause a similar problem if they occur at the top of the page.

content/css/docs.css
RichiH: done

PR #1996, opened 2021-07-21
Make a link more visible

axlrosen:
A link that is also code doesn't visually appear as a link. So I moved the link to some non-code text instead.


RichiH:
Thanks!

PR #2002, opened 2021-08-03
Add Swift (SwiftPrometheus) to client libraries

MrLotU:
This PR will add SwiftPrometheus (https://github.com/MrLotU/SwiftPrometheus) to the list of client libraries.

SwiftPrometheus is the current standard for Prometheus Client libraries in the ServerSide Swift ecosystem and is recognised as such by the SSWG. See [SSWG Projects](https://swift.org/sswg/#projects).

The library contains an example executable that can be used and altered to see what the output will look like.

Any questions and/or remarks are welcome.


ktoso:
We should close this one in favor of https://github.com/prometheus/docs/pull/2793

PR #2006, opened 2021-08-11
Adapt TLS guide to new Prometheus features

roidelapluie:

### DIFF
--- a/content/docs/guides/tls-encryption.md
+++ b/content/docs/guides/tls-encryption.md
@@ -5,27 +5,26 @@ sort_rank: 1
 
 # Securing Prometheus API and UI endpoints using TLS encryption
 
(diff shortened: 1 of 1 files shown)


content/docs/guides/tls-encryption.md
```diff
@@ -36,66 +35,46 @@ openssl req \
...
-```bash
-sudo nginx -c /usr/local/etc/nginx/nginx.conf
+```yaml
+tls_server_config:
+  cert_file: /home/prometheus/certs/example.com/example.com.crt
+  key_file: /home/prometheus/certs/example.com/example.com.key
 ```
 
-NOTE: This example uses `/usr/local/etc/nginx` as the location of the nginx configuration file, but this will vary based on the installation. Other [common nginx config directories](https://nginx.org/en/docs/beginners_guide.html) include `/usr/local/nginx/conf` and `/etc/nginx`.
+From now on, you would need to pass a new flag to Prometheus,
```
RichiH, line 48:
```suggestion
To make Prometheus use this config, you will need to call it with the flag
```

RichiH (approved):
One change, feel free to merge after.

PR #2011, opened 2021-08-23
Add link to docs about `up` metric

mkopinsky:
I read this section a few times, but was still confused about how I'd alert or graph uptime if the app or exporter doesn't export a 1/0 metric. The missing piece of information is that there's a built-in `up` metric. This PR adds a link so the next guy or gal doesn't get confused like I did. :-)

### DIFF
--- a/content/docs/instrumenting/writing_exporters.md
+++ b/content/docs/instrumenting/writing_exporters.md
@@ -497,7 +497,7 @@ that has a value of 0 or 1 depending on whether the scrape worked.
 The latter is better where there’s still some useful metrics you can get
 even with a failed scrape, such as the HAProxy exporter providing
 process stats. The former is a tad easier for users to deal with, as
-[`up` works in the usual way](https://prometheus.io/docs/concepts/jobs_instances/#automatically-generated-labels-and-time-series), although you can’t distinguish between the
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/writing_exporters.md
```diff
@@ -497,7 +497,7 @@ that has a value of 0 or 1 depending on whether the scrape worked.
 The latter is better where there’s still some useful metrics you can get
 even with a failed scrape, such as the HAProxy exporter providing
 process stats. The former is a tad easier for users to deal with, as
-`up` works in the usual way, although you can’t distinguish between the
+[`up` works in the usual way](https://prometheus.io/docs/concepts/jobs_instances/#automatically-generated-labels-and-time-series), although you can’t distinguish between the
```
SuperQ, line 500: This repo uses a static site generator, so in-repo links should be relative to the content dir.

```suggestion
[`up` works in the usual way](/docs/concepts/jobs_instances/#automatically-generated-labels-and-time-series), although you can’t distinguish between the
```

SuperQ (changes requested):
Thanks for the improvement, minor change to the formatting of the URL is needed.

SuperQ (approved):
LGTM

PR #2016, opened 2021-08-27
Correct bad link for Kuma integration

lahabana:
The link was wrong fixed it.


roidelapluie:
Thanks, we should actually remove that link now that kuma is natively supported.

lahabana:
While it's true it seems that prometheus docs are not versioned therefore we should probably keep both methods for a bit no?
On the Kuma side we're maintaining both methods for a bit.

roidelapluie:
Ack

roidelapluie:
Thanks!

PR #2030, opened 2021-09-27
Add link to Tasmota exporter

lfuelling:
Hi,
this PR adds a link to an exporter for Tasmota (IoT switches/sockets firmware).

Signed-off-by: Lukas Fülling <lukas@k40s.net>


lfuelling:
The DCO is failed because I used a name that is not the name I use on GitHub. 
When running the command listed by the DCO scan, my real name will be appended instead of the one I use on GitHub.

roidelapluie:
Not sure what happens here, but the DCO issue seems minor. I have manually marked it as passed.

PR #2037, opened 2021-10-12
Add acend to Support & Training 

Tobi-Tr:


RichiH:
Thanks for this PR. We currently only list English resources.

@juliusv thoughts?

roidelapluie:
> Thanks for this PR. We currently only list English resources.
> 
> @juliusv thoughts?

We list https://www.puzzle.ch/de/prometheus which is in German, this is fine to me.

RichiH:
@roidelapluie good point

@Tobi-Tr whichever you prefer then; I would make certain that it's really obvious that you offer both languages at that landing page.

Do I assume correctly that you will move to the Courses section?

PR #2039, opened 2021-10-12
adding scaphandre to softwares exposing prometheus metrics

bpetit:
[scaphandre](https://github.com/hubblo-org/scaphandre) is a monitoring agent dedicated to power consumption metrics

Signed-off-by: Benoit Petit <bpetit@nce.re>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -293,6 +293,7 @@ separate exporters are needed:
    * [Quobyte](https://www.quobyte.com/) (**direct**)
    * [RabbitMQ](https://rabbitmq.com/prometheus.html)
    * [RobustIRC](http://robustirc.net/)
+   * [Scaphandre](https://github.com/hubblo-org/scaphandre)
    * [ScyllaDB](http://github.com/scylladb/scylla)
    * [Skipper](https://github.com/zalando/skipper)
    * [SkyDNS](https://github.com/skynetservices/skydns) (**direct**)


bpetit:
> Thanks; some comments/questions:
> 
>     * The label cmdline can get into high cardinality. Do you know how many total values a user should normally expect?

This is the length of the commandline corresponding to the process as returned by procfs. There is no limit yet but we should add one in a near future.

>     * scaph_process_power_consumption_microwatts - is is possible to get counters in Joule? `rate()` turns those into Watt automagically. Alternatively, could you change to the base unit of Watt, please?

We use microwatts to manipulate more precise values as some processes may use only a small amount of the power consumed by the whole machine. This seems more practical to use round values than floats with a lot of decimals.

>     * scaph_processes_blocked_current - the _current seems superfluous

You are right, we could remove that to make the name simpler.

>     * scaph_socket_energy_microjoules - is it possible to change to Joule?

This is the same idea as before. That being said, this particular metric is redundant with the microwatts one, so one or the other should be removed soon.

>     * scaph_self_domain_records_nb - assuming _nb is number, it seems superfluous

Agreed, we could remove that suffix to.

>     * scaph_self_mem_shared_resident_size - should end in _bytes and memory written out

Okay, that makes sense.

>     * scaph_self_mem_shared_resident_size - can you do CPU seconds? That way it can be correlated with node_exporter. In this specific use case it might make sense to retain % (or ratio 0-n) as we can not be certain that node_exporter data exists.

We do get cpu time consumed by each process, in the process of computing the power consumption due to each process. But we don't expose those cpu time related metrics yet (that was not the purpose of the agent in the first place). Do you mean we should add a CPU usage % metric to the power consumption ones ?

>     * scaph_self_version - we commonly use _info metrics for this. Information in labels, the metric value is always 1

Okay, I didn't know that, my bad.

jan--f:
Hello from the bug scrub!
Thank you for your contribution. I'm afraid we need to decline this for now. We are happy to feature exporters that are actively developed and used. Unfortunately this one has seen no commits in two years. Please try again at a later stage.

PR #2046, opened 2021-10-15
Delete old file

RichiH:


roidelapluie:
#2047 has a redirect.

PR #2051, opened 2021-10-17
Add HART exporter

jszumigaj:
@RichiH Hi,

this PR adds a link to my implementation of exporter for devices with HART protocol (like modbus-exporter)


jan--f:
Hello from the bug scrub!

PR #2065, opened 2021-11-20
Added matcher syntax support for Routing tree editor.

### LINKED ISSUE #1638: Routing Tree Editor website not accurately simulating routes

### LINKED ISSUE #1965: Routing tree editor - support new matchers syntax
When testing with the latest example the editor is showing the wrong matched route

https://prometheus.io/docs/alerting/latest/configuration/#example

Evreke:
I added matchers syntax support to [alertmanager routing tree editor](https://www.prometheus.io/webtools/alerting/routing-tree-editor/)
PR closes #1638 closes #1965

### DIFF
--- a/static/routing-tree.js
+++ b/static/routing-tree.js
@@ -25,6 +25,8 @@ var tooltip = d3.select("body")
     .style("z-index", "10")
     .style("visibility", "hidden");
 
+const promQlRegExp = /\b(?<label>[a-z0-9_]\w*)(?<selector>=~?|![=~])"?(?<value>(?<=")(?:[^\\"]|\\.)*(?=")|\w+[^\s},]+)/gmi;
+
 function parseSearch(searchString) {
   var labels = searchString.replace(/{|}|\"|\'/g, "").split(",");
   var o = {};
@@ -164,6 +166,41 @@ function massage(root, receivers) {
       o.name = key;
       matchers.push(o);
     }
+  }
+
+  if (root.matchers) {
+    root.matchers.forEach((matcher) => {
+      let o = {};
+
+      let matchedExpressions;
+      while ((matchedExpressions = promQlRegExp.exec(matcher)) !== null) {
+        let [match, label, selector, value] = matchedExpressions;
+        o.name = label;
+
+        switch (selector) {
+          case "=~":
+            o.isRegex = true;
+            o.value = new RegExp("^(?:" + value + ")$");
+            matchers.push(o);
+            break;
+          case "!=":
+            o.isRegex = true;
+            o.value = new RegExp("^(?!" + value + "$)");
+            matchers.push(o);
+            break;
+          case "!~":
+            o.isRegex = true;
+            o.value = new RegExp("^(?!" + value + "$)");
+            matchers.push(o);
+            break;
+          case "=":
+            o.isRegex = false;
+            o.value = value
+            matchers.push(o);
+            break;
+        }
+      }
+    });
   }
 
   root.matchers = matchers;
--- a/static/routing-tree.js
+++ b/static/routing-tree.js
@@ -25,7 +25,7 @@ var tooltip = d3.select("body")
     .style("z-index", "10")
     .style("visibility", "hidden");
 
-const promQlRegExp = /\b(?<label>[a-z0-9_]\w*)(?<selector>=~?|![=~])"?(?<value>(?<=")(?:[^\\"]|\\.)*(?=")|\w+[^\s},]+)/gmi;
+const promQlRegExp = /\b(?<label>[a-z0-9_]\w*)\s?(?<selector>=~?|![=~])\s"?(?<value>(?<=")(?:[^\\"]|\\.)*(?=")|\w+[^\s},]+)/gmi;
 
 function parseSearch(searchString) {
   let o = {};
--- a/static/routing-tree.js
+++ b/static/routing-tree.js
@@ -25,7 +25,7 @@ var tooltip = d3.select("body")
     .style("z-index", "10")
     .style("visibility", "hidden");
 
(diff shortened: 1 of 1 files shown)


juliusv:
Hi, thanks!

I played around with it a bit at https://deploy-preview-2065--prometheus-docs.netlify.app/webtools/alerting/routing-tree-editor/ and it seems to only work when not putting any spaces around the match op, e.g. `foo=~bar` works, but `foo =~ bar` does not. But the docs at https://prometheus.io/docs/alerting/latest/configuration/#matcher mainly show the style including the spaces, which confused me initially... could we support whitespace here as well?

Also, should a regex matcher like `foo=~bar` be shown as `foo: /^(?:bar)$/` in the tree view? While technically not incorrect, should we maybe show the original string value instead? Also, I see `!=` being listed as `isRegex = true;` currently, but that shouldn't be a regex match, right?

Generally: For now going about it in this direction seems fine, but long term I think this piece of JS is not easy to maintain or get fully correct. At some point it would be great for this to have a Go backend that could do proper validation of everything and reuse the Go datatypes + code for all of this and just return a tree to the UI... kind of like PromLens (https://promlens.com/) is doing for PromQL.

Evreke:
I understood the comments. I'll see what can be done

You are right about maintainability of that solution and using existing code base would be preferable over reinventing the wheel in JS. 
I'd like to work on it in next iteration. 
Would you like to consider webassembly solution?

juliusv:
@Evreke Maybe webasm could be a solution, but I know very little about it so far... the other option would be just having a Go backend. Of course that would need to be hosted somewhere, which is annoying.

But yeah, the current piece of JS:

* has no tests
* is not typed (plain old single JS file)
* has to (imperfectly) re-create a lot of the logic from the Go code

...which is not great.

Evreke:
@juliusv 
Totally agree. Code should be refactored.
This is a big work and it should be done carefully. I would prefer to create one or better few issues to discuss and plan it there.

About comments. 
1. I add regexp fix to capture single or zero spaces around selector.
2. About `=~` matcher and non regexp value. This could be done but things will complicated even more as we need to check value is it regexp or not. Yeah it could be interpreted as regular equals matcher but it's just not worth it. I'll explain  further.
3. About using non regexp value for negate matcher `!=`. If we would like to avoid regexp for `!=` we will lose lucidity and make things more complicated. The most obvious solution from my perspective is to introduce new field and check it in `matchLabel` but it wouldn't be used in other cases.

I think 2,3 is just a waste of time in case if we want to rewrite this part of code to be evaluated by WASM\external API or something else. Correct me if i'm wrong.

In summary:
In my opinion, now it is better to refactor, so as not to get bogged down in fixing notoriously difficult decisions that will entail more complication.

Evreke:
@juliusv 
I've done small research and found that using WASM could be difficult from debugging perspective. We don't have mature debugging instruments yet. Also it could be difficult or impossible just to reuse existing codebase due to lots of package dependencies.

So we have to look at hosted backend for reusing existing code base.

juliusv:
@Evreke Thanks, yeah, makes sense! I'll be honest, I won't have too much time to put into something completely new with a hosted backend (especially since that needs to be run somewhere separately and maintained by the team). I think we have two options:

* For now, try to make as much as possible work with the current script, although it's not going to be perfect.
* If someone wants to really own and be responsible for a new and better version of this with a hosted backend, that would also be great. It could first be built standalone, and then we could eventually figure out where to run it (e.g. on Google Cloud Run).

Evreke:
@juliusv 

Sounds reasonable.

I'll watch what could be done on your second and third comment if you insist but from my point of view it is not necessary. 
Will try to finish it this week.

juliusv:
> I'll watch what could be done on your second and third comment if you insist but from my point of view it is not necessary.

With the third one, do you mean the one about `!=` not being a regex? As far as I can see that part actually needs to be fixed, because right now the non-regex value is just put into a regex, so if it contains any special regex characters (like `.`), they will be interpreted as such, although they're not supposed to be.

Evreke:
@juliusv
You are right! Appreciate, my bad.
I add mutable object `Matcher` to hold operators and I rewrite my part using it.
Now evaluating deprecated matchers tends to separate from my solution.

Evreke:
@juliusv 
Hi Julius! I congratulate all of the holidays have passed. Are we okay with the current solution?

static/routing-tree.js
```diff
@@ -166,6 +193,48 @@ function massage(root, receivers) {
...
+        let [match, label, selector, value] = matchedExpressions;
+        o.name = label;
+
+        switch (selector) {
+          case "=~":
+            if (/[*+?\\.^\[\]$&|]+/.test(value)) {
+              o.value = new RegExp("^(?:" + value + ")$");
+              o.op = Matcher.RE;
+            } else {
+              // If no special characters in RegexMatcher than we can consider it as Equal Matcher
```
juliusv, line 212: Is there a reason why we wouldn't just still treat it as a regex if the selector is `=~`? If it ends up matching just like an equality matcher, that's fine...

juliusv (commented):
Thanks and sorry for the extended pause on my side! Looks pretty good to me, one small question.

static/routing-tree.js
Evreke: This is probably done in terms of saving resources, but it looks strange and confusing, and there is almost no benefit in terms of saving. You're right

juliusv:
Great, thank you again for the work and the patience :) :+1:

PR #2073, opened 2021-12-01
Add RTPEngine to 3rd-party software that export Prometheus metrics

guillemj:
It can be configured to directly export Prometheus metrics.


jan--f:
Hello from the bug scrub!
Thanks for your contribution. Would you please rebase the PR? Happy to go ahead with this.

guillemj:
Done, thanks! (Adding an explicit comment, as I don't recall what triggers notifications and what not :) )

PR #2076, opened 2021-12-14
Remove English-centric phrasing for metric naming

mpontillo:
I think the phrasing around allowed characters in metric names should be changed, for a couple of reasons.

First, consider the metric name `合計httpリクエスト`. (If I believe Google Translate, that means "total HTTP requests".) The documentation seems to imply that, since the characters in this metric name are not allowed, one might instead call it `__http_____`. 🙂 Consider a less extreme example: if I have `jalapeños_purchased_total`, "sanitizing" it would result in `jalape_os_purchased_total`, which is similarly silly. I don't think this is a useful suggestion, so maybe it shouldn't be included in the documentation.

Second (and respectfully, because I think it's unlikely that this was the intention of the original author), the "sanitized" phrasing might imply that non-English character sets are somehow diseased and in need of cleansing. So I think this phrasing should just be removed.

Signed-off-by: Mike Pontillo <mpontillo@digitalocean.com>

### DIFF
--- a/content/docs/instrumenting/writing_exporters.md
+++ b/content/docs/instrumenting/writing_exporters.md
@@ -128,8 +128,7 @@ automatically doesn’t always produce nice results for things like
 Exposed metrics should not contain colons, these are reserved for user
 defined recording rules to use when aggregating.
 
-Only `[a-zA-Z0-9:_]` are valid in metric names, any other characters
-should be sanitized to an underscore.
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
Agreed, the extra wording is misleading.

RichiH:
Thanks!

PR #2086, opened 2022-01-21
Concepts: Remove confusing examples

mattiasb:
I have trouble understanding what saying that an instance is "usually
corresponding to a single process" is trying to convey here and how it
might help a new user trying to understand what an instance is. I
believe the point is made more clear by just removing this example.

The same is true for the explanation of what a job is, but here I
believe the example is actively harmful. It conveys the picture that a
job is a way to group servers together based on their purpose *as a
server*, when in reality it's often more the case that you want to group
up a set of instances that you want to scrape the same set of metrics
from.

Remove these examples and let the actual content stand on its own legs.

Signed-off-by: Mattias Bengtsson <mattias.jc.bengtsson@gmail.com>

### DIFF
--- a/content/docs/concepts/jobs_instances.md
+++ b/content/docs/concepts/jobs_instances.md
@@ -5,8 +5,8 @@ sort_rank: 3
 
 # Jobs and instances
 
-In Prometheus terms, an endpoint you can scrape is called an _instance_,
-usually corresponding to a single process. A collection of instances with the same purpose, a process replicated for scalability or reliability for example, is called a _job_.
+In Prometheus terms, an endpoint you can scrape is called an _instance_. A
+collection of instances is called a _job_.
 
 For example, an API server job with four replicated instances:
 


roidelapluie:
Thanks for your contribution. I disagree with this change, I think the current wording is approriate and helpful. It helped me when I started using Prometheus.

SuperQ:
I also think the wording is fine as-is.

mattiasb:
Would it be helpful if I explained the confusion a bit more? Or do you understand what I'm getting at but rather believe that most users won't fall into the thought traps I did? :)

RichiH (commented):
I also think it's clear as-is, but you are free to try to come up with other phrasing if you want.

mattiasb:
Hm, I don't think I have the energy to pursue this. The current wording made me misunderstand what a `job` is but I might very well be an outlier. Closing for now. Thanks for the input!

PR #2089, opened 2022-02-11
Add xamira networks to training and support page

fthorns:
Signed-off-by: Fabian Thorns <fabian@thorns.it>


fthorns:
@RichiH Sorry for bugging you about one of the less important PRs, I saw this one pending for a while and was wondering if there is anything I forgot or if there is anything I could do to help getting this one merged. Thank you :)

roidelapluie:
Thank you for this. I think the current rule is that the "courses" are limited to online trainings.

fthorns:
Hi @roidelapluie , thanks for the explanation and sorry for the late response. We do actually delivery training online, although it is instructor lead. I assume you're referring to self-paces training only, right? In that case, I could adjust the pull request to only add us to the commercial support section; although I think that the ability to collect instructor led training might be helpful, too. But that is probably another topic. Let me know how you'd like to proceed, if we have a chance to make it at least to the commercial support section I'll make the necessary changes. Thanks! :)

juliusv:
Yeah, Courses is just for self-paced online courses so far, but in general I support the idea of making this page more useful. Frequently people also ask me who could help them with some consulting aspect (since I'm doing less consulting with PromLabs now). If we knew what kinds of services each company offered (hosted service, consulting, self-paced training, live training, software engineering, commercial additional software, ...), then we could add tags and/or make things searchable / filterable somehow. Of course that would require reaching out to the various companies to get better information on what they offer, because it's often not 100% clear to us as well.

juliusv:
Yep, that's a larger project...

Thanks for the adjustment! :+1:

PR #2091, opened 2022-02-14
Add webscraper-exporter to the exporters list

cestef:
Signed-off-by: cstef <colin@petit-suisse.fr>

An exporter for website performance metrics built using puppeteer: https://github.com/cstefFlexin/webscraper-exporter

CLI preview:

image

`/` endpoint: 

image

Metrics exposed:

image

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -245,6 +245,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Teamspeak3 exporter](https://github.com/hikhvar/ts3exporter)
    * [Transmission exporter](https://github.com/metalmatze/transmission-exporter)
    * [Unbound exporter](https://github.com/kumina/unbound_exporter)
+   * [Webscraper exporter](https://github.com/cestef/webscraper-exporter)
    * [WireGuard exporter](https://github.com/MindFlavor/prometheus_wireguard_exporter)
    * [Xen exporter](https://github.com/lovoo/xenstats_exporter)
 


RichiH:
Did you close this on purpose?

Both the metric and the label names would need some work; I would be happy to help you polish them.

cestef:
I just realised I closed it by accident when adding the lastest commit, thank you so much for sending this message 😭

cestef:
I would be very grateful if you could help me rename the metrics etc. I'm pretty new to the prometheus ecosystem...

RichiH:
No problem! You should prefix them with a name, in your case I think it's `webscraper_` and suffix them with the unit, so:

Metrics should be in base units, so duration should be in seconds, and probably named `webscraper_test_duration_seconds`

`heap_usage` should probably be `webscraper_heap_usage_bytes`

CPU usage should ideally be in seconds and also allow users to calculate a total; that way they can calculate the % themselves as they need. If you can only do %, then `webscraper_cpu_usage_percent` would be OK.

`bytes_in` would be better as `webscraper_in_bytes`

cestef:
Thank you very much ! I'll update the metrics accordingly asap

cestef:
Sorry for all the useless logs from github about me force pushing commits or whatever, it was a complete mess so I just went over from scratch 😅
I renamed the metrics following what you said, so it should be all good now !

cestef:
These are now the metrics exposed:
image

cestef:
Any updates on this ? Would love to have my work featured :D

jan--f:

PR #2107, opened 2022-03-21
Replace 'build' with 'create'

youngrrrr:
At least for the Java client, it appears the 'build' method was renamed 'create'?

https://prometheus.github.io/client_java/io/prometheus/client/Summary.Builder.html

### DIFF
--- a/content/docs/instrumenting/writing_clientlibs.md
+++ b/content/docs/instrumenting/writing_clientlibs.md
@@ -116,14 +116,14 @@ For example in the Java Simpleclient we have:
 
 ```java
 class YourClass {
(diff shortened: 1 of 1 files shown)


youngrrrr:
Will figure out a way to properly redo this PR (e.g. include DCO)...

hdost:
Thank you for the PR, but this is incorrect a `Counter.build()` returns a `Counter.Builder` this has all the methods as shown.
https://prometheus.github.io/client_java/io/prometheus/client/Counter.html
the `.register()` at the end is inherited from Builder class which creates the class from the builder and then registers it to the default registry.

PR #2111, opened 2022-04-04
add a special note regarding the security scanners

Nexucis:
Signed-off-by: Augustin Husson <augustin.husson@amadeus.com>

### DIFF
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -15,16 +15,22 @@ attack vectors that some configurations may enable.
 As with any complex system, it is near certain that bugs will be found, some of
 them security-relevant. If you find a _security bug_ please report it
 privately to the maintainers listed in the MAINTAINERS of the relevant
-repository and CC prometheus-team@googlegroups.com. 
-
-Special note for security scanner users, please be careful with the report it is given.
-Most of the time the issues raised don't apply to the tool scanned. It is specially true
-regarding the vulnerabilities detected in the dependencies (Go or npm). 
-In this particular case we will kindly ask you to explain how it can affect the tool scanned.
-
-We will fix the issue as soon as possible and coordinate a release date with you. 
-You will be able to choose if you want public acknowledgement of your effort and if you want to be
+repository and CC prometheus-team@googlegroups.com.  We will fix the issue as soon
+as possible and coordinate a release date with you. You will be able to choose
+if you want public acknowledgement of your effort and if you want to be
 mentioned by name.
+
+### Automated security scanners
+
+Special note for security scanner users: Please be mindful with the reports produced.
+Most scanners are generic and produce lots of false positives. More and more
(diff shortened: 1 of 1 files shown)


content/docs/operating/security.md
```diff
@@ -15,9 +15,15 @@ attack vectors that some configurations may enable.
...
+repository and CC prometheus-team@googlegroups.com. 
+
+Special note for security scanner users, please be careful with the report it is given.
+Most of the time the issues raised don't apply to the tool scanned. It is specially true
+regarding the vulnerabilities detected in the dependencies (Go or npm). 
+In this particular case we will kindly ask you to explain how it can affect the tool scanned.
+
+We will fix the issue as soon as possible and coordinate a release date with you. 
+You will be able to choose if you want public acknowledgement of your effort and if you want to be
 mentioned by name.
```
RichiH, line 27:
```suggestion
repository and CC prometheus-team@googlegroups.com.  We will fix the issue as soon
as possible and coordinate a release date with you. You will be able to choose
if you want public acknowledgement of your effort and if you want to be
mentioned by name.

### Automated security scanners

Special note for security scanner users: Please be mindful with the reports produced.
Most scanners are generic and produce lots of false positives. More and more
reports are being sent to us, and it takes a significant amount of work to go
through all of them and reply with the care you expect. This problem is particularly
bad with Go and NPM dependency scanners.

As a courtesy to us and our time, we would ask you not to submit raw reports.
Instead, please submit them with an analysis outlining which specific results
are applicable to us and why.
```

RichiH:
I think this is better phrasing.

As an aside, I do believe that we should "fix" a lot of the dependencies even if we are not affected. It's busywork to some extent, but it's still cleaner for the wider ecosystem to simply "fix" them where it's simple.

Nexucis:
> I think this is better phrasing.

Ah yeah indeed. Thanks @RichiH !

> As an aside, I do believe that we should "fix" a lot of the dependencies even if we are not affected. It's busywork to some extent, but it's still cleaner for the wider ecosystem to simply "fix" them where it's simple.

Totally agree on that and that's we do before each release (i.e. update the dependencies).

content/docs/operating/security.md
```diff
@@ -15,11 +15,23 @@ attack vectors that some configurations may enable.
 As with any complex system, it is near certain that bugs will be found, some of
 them security-relevant. If you find a _security bug_ please report it
 privately to the maintainers listed in the MAINTAINERS of the relevant
-repository and CC prometheus-team@googlegroups.com. We will fix the issue as soon
+repository and CC prometheus-team@googlegroups.com.  We will fix the issue as soon
```
RichiH, line 18:
```suggestion
repository and CC prometheus-team@googlegroups.com. We will fix the issue as soon
```

RichiH (approved):
One whitespace too many, else LGTM.

RichiH:
Thanks!

PR #2113, opened 2022-04-05
Comparison with rather than "to"

tredondo:
https://writingexplained.org/compare-to-or-compare-with

### DIFF
--- a/content/docs/introduction/comparison.md
+++ b/content/docs/introduction/comparison.md
@@ -1,5 +1,5 @@
 ---
-title: Comparison to alternatives
+title: Comparison with alternatives
 sort_rank: 4
 ---
 


RichiH:
Thanks for this! This is not clear-cut, see e.g. https://jakubmarian.com/in-comparison-to-vs-in-comparison-with-in-english/ -- I believe that the current spelling is slightly more correct in this case.

tredondo:
> https://jakubmarian.com/in-comparison-to-vs-in-comparison-with-in-english/

Is that [source more reliable](https://en.wikipedia.org/wiki/Wikipedia:Reliable_sources) than [Collins](https://www.collinsdictionary.com/dictionary/english/compared-with-something-compared-to-something)? Its author is a [Czech](https://jakubmarian.com/about/) data scientist, and he doesn't cite any [primary sources](https://en.wikipedia.org/wiki/Primary_source).

Here are more arguments favoring "compare with" for highlighting differences:

https://english.stackexchange.com/questions/583/compared-with-vs-compared-to-which-is-used-when

PR #2115, opened 2022-04-05
Emphasize label names in Data Model

tredondo:


RichiH:
We're only emphasizing the well-defined terms within this doc. So while I will close this without merging, I still want to thank you for submitting :)

tredondo:
Thank you :) Would it help the new learner to format as `code` the labels, since they are formatted like that just a few lines below?

image

PR #2123, opened 2022-04-14
Exclude Web UI from the downloads

roidelapluie:
Prometheus now provides web UI tarballs to better integrate with
downstream distributions.

We need to explictly exclude them from our downloads.

Signed-off-by: Julien Pivotto <roidelapluie@o11y.eu>


SuperQ (approved):
LGTM

PR #2129, opened 2022-04-28
Introduce long term support

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@o11y.eu>


RichiH:
I like it.

I expect your new company to carry the workload, correct?

I would still suggest to discuss this during today's dev summit.

roidelapluie:
So, the discussion happened in prevous dev summit: https://youtu.be/CtnpN-_KssY?t=4277 and in this design document: https://docs.google.com/document/d/1wCXLnvh460UG10Mw019UbYMWDwSWWdvgwY61LHY0A0w/edit and

RichiH:
Wow, that's embarassing. I brainflushed on that.

RichiH:
Should we still have a quick discussion if anyone has concerns about choosing this specific version?

RichiH (approved):
We just touched on it during the dev summit. No one has any concerns about version & timing.

Yay!

PR #2130, opened 2022-04-28
integrations: add link to prom2tower

pja237:
Signed-off-by: Jager,Petar <petar.jager@gmail.com>

Hello @simonpasquier , @w0rm ,

please check this PR, it adds prom2tower link to the list of integrations.
Prometheus2tower is a service that integrates alertmanager and ansible tower (awx), more elaborate description is in the repo README :)

Best,
Petar

### DIFF
--- a/content/docs/operating/integrations.md
+++ b/content/docs/operating/integrations.md
@@ -91,6 +91,7 @@ For notification mechanisms not natively supported by the Alertmanager, the
   * [Matrix](https://github.com/matrix-org/go-neb)
   * [Phabricator / Maniphest](https://github.com/knyar/phalerts)
   * [prom2teams](https://github.com/idealista/prom2teams): forwards notifications to Microsoft Teams
+  * [Ansible Tower](https://github.com/pja237/prom2tower): call Ansible Tower (AWX) API on alerts (launch jobs etc.)
   * [Rocket.Chat](https://rocket.chat/docs/administrator-guides/integrations/prometheus/)
   * [ServiceNow](https://github.com/FXinnovation/alertmanager-webhook-servicenow)
   * [Signal](https://github.com/dgl/alertmanager-webhook-signald)


hdost (commented):
LGTM, @juliusv , but it's only 0.0.1 and very recently released. Is there any sort of criteria other than "under development"

content/docs/operating/integrations.md
```diff
@@ -91,6 +91,7 @@ For notification mechanisms not natively supported by the Alertmanager, the
   * [Matrix](https://github.com/matrix-org/go-neb)
   * [Phabricator / Maniphest](https://github.com/knyar/phalerts)
   * [prom2teams](https://github.com/idealista/prom2teams): forwards notifications to Microsoft Teams
+  * [prom2tower](https://github.com/pja237/prom2tower): call Ansible Tower (AWX) API on alerts (launch jobs etc.)
```
juliusv, line 94: Most items in this list are not named by the name of the tool, but by the thing they integrate with. The existing `prom2teams` right above is an exception and maybe should be changes to just `Microsoft Teams` as well. I think for this item, just `Ansible Tower` would be the expected item name?

pja237:
> LGTM, @juliusv , but it's only 0.0.1 and very recently released. Is there any sort of criteria other than "under development"

Hello @hdost, just to clarify it's current status, it's a fairly simple service, not too much to develop around this topic so even though i put "in development" in the readme and it's versioned as 0.0.1, it's actually quite done (for now) and it's been tested and rolled out to my companies monitoring stack (dual rail tde-prod setup doing ~1000+ scrape endpoints).
Would it make sense to make a 1.0.0 version out of it?

More development i guess might happen if bugs pop up or we come across something more to add to it (even though, i can't think of what that might be right now)

content/docs/operating/integrations.md
pja237: Sure, no problem with that, would you like me to push a change with "Ansible Tower" then?

juliusv: Yep please :)

juliusv:
@pja237 I think the version stuff is totally fine. People can still judge for themselves, and as long as there's nothing more mature out there for the same purpose, it's better to list it than not.

content/docs/operating/integrations.md
pja237: Done! :)

juliusv:
:+1:

Issue #2118, opened 2022-04-06
Having to sign the DCO discourages contributions


tredondo:
Background:

1. I'm a technical writer, and love to contribute to open source projects on GitHub, given its easy to use interface for editing Markdown and submitting a PR.
2. At the same time, I don't have the inclination to fork, clone, edit etc. a project just to fix a typo.
3. Prometheus is the first project I've seen in years that automatically rejects all such contributions because of _a thing called DCO_ (remember, I'm just a technical writer, and I'm unfamiliar with the reasons why this DCO is necessary here, but nowhere else I've been contributing).
4. I'm not the only one whose commits have been rejected for not signing the DCO. https://github.com/prometheus/docs/search?q=dco&type=issues shows 116 issues that have the word "DCO" in them. Just look at the first page of results.

I assume this DCO thing is worth rejecting contributions from the community, but there's got to be a better way.

[The PR template](https://github.com/prometheus/docs/edit/main/.github/PULL_REQUEST_TEMPLATE.md) is super cryptic for drive-by contributors: "Please sign CNCF's Developer Certificate of Origin". What's CNCF? Developer Certificate of Origin? Yeah, you can Google those, but the whole line made me gloss over it several times, and the vast majority won't take the time to Google these things. Also, this line suggests that the contributor would have to "sign" something like a CLA, which *can* be done after submitting the commit, and is far more likely to be done than someone forking the repo, clonig, and amending their commit.

Also, the PR template is displayed too late for GitHub users. At that point, the commit was already made. Again, look at the [hundreds of issues created by the DCO](https://github.com/prometheus/docs/search?q=dco&type=issues).

If you want to accept contributions from the community, please try to make it as easy as possible for the community to do so.

roidelapluie:
Thank you. How could we improve the PR template to make DCO easier to grasp?

DCO is mandated to us by CNCF, it is not a choice of Prometheus itself: https://github.com/cncf/foundation/blob/main/charter.md#11-ip-policy so we can not remove this requirement, but I am happy to see how we can make it easier.

tredondo:
I don't think updating the PR template would help contributions via GitHub, because it's displayed too late, after the commit is created.

What would help would be a comment block at the top of each source `.md` file, advising to sign off the DCO. Something like:

```
<!--
Please sign off your commits so that we can legally accept your contribution.

You can do this with `git commit -s` or, from the GitHub interface,
by adding the following line to the "extended description" of each commit:

[...]

PR #2138, opened 2022-05-17
docs: Fix links master -> main

hdost:
There are a few repos which have migrated from `master` to `main` this
reflects that change for the ones which have made the transition.

Additionally cleanup some End of Line whitespace.

Signed-off-by: Harold Dost <github@hdost.com>

PR #2140, opened 2022-05-17
Be more explicit about DCO

hdost:
* Add link about the browser extension.

Relates #2118

Signed-off-by: Harold Dost <github@hdost.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,10 @@
 
 Prometheus uses GitHub to manage reviews of pull requests.
 
+* Every commit *MUST* be signed off to agree to the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). [How does it work?](https://github.com/dcoapp/app#how-it-works)
+** There is a [browser extension](https://github.com/scottrigby/dco-gh-ui) if you're making edits
+   from the GitHub UI.
+
 * If you have a trivial fix or improvement, go ahead and create a pull request,
   addressing (with `@...`) the maintainer of this repository (see
   [MAINTAINERS.md](MAINTAINERS.md)) in the description of the pull request.
@@ -10,5 +14,3 @@ Prometheus uses GitHub to manage reviews of pull requests.
   on our [mailing list](https://groups.google.com/forum/?fromgroups#!forum/prometheus-developers).
   This will avoid unnecessary work and surely give you and us a good deal
   of inspiration.
- 
-* Be sure to sign off on the [DCO](https://github.com/dcoapp/app#how-it-works)


CONTRIBUTING.md
```diff
@@ -2,6 +2,10 @@
 
 Prometheus uses GitHub to manage reviews of pull requests.
 
+* Every commit *MUST* signed off on the [DCO](https://github.com/dcoapp/app#how-it-works)
```
SuperQ, line 5: Expand the acronym to make it a little more SEO, screen reader, and mobile screen friendly.

```suggestion
* Every commit *MUST* signed off on the [Developer Certificate of Origin](https://github.com/dcoapp/app#how-it-works) (DCO).
```

SuperQ (commented):
Thanks, this is a good improvement.

hdost:
@juliusv  Something appears to be going wrong with Netlify testing.

CONTRIBUTING.md
juliusv: Should it really be "on the" not "with the"?

juliusv:
@hdost Ah drat, I believe that's due to the Netlify setting to omit sensitive environment variables for forks:

variables

Not sure what to do there as a quick fix, because we can't just expose sensitive vars to the world, but the GitHub token is required for assembling the docs from the other repos.

juliusv:
Btw., I think this is the same CONTRIBUTING.md as in the other repos, so we probably want this change everywhere.

I thought it was part of the files that we automatically sync across repos, but it seems CONTRIBUTING.md is not included in that: https://github.com/prometheus/prometheus/blob/707600d84f5567354ef48d4593ddab84c4837811/scripts/sync_repo_files.sh#L40 - @SuperQ @roidelapluie should it be?

SuperQ:
@juliusv Yes, I think we should sync this file. But some repos have repo-specific instructions. We'll have to come up with pattern to handle this. Either move the repo-specific stuff to a separate file, or have some kind of "# Additional considerations` section that we can patch around.

CONTRIBUTING.md
hdost: Agreed

hdost: > Should it really be "on the" not "with the"?

So I believe "on the" does actually could make sense

But maybe
"Every commit *MUST* be signed off to agree to the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). [How does it work?](https://github.com/dcoapp/app#how-it-works)"

hdost:
> @hdost Ah drat, I believe that's due to the Netlify setting to omit sensitive environment variables for forks:
> 
> 
> 
> variables
> 
> 
> 
> Not sure what to do there as a quick fix, because we can't just expose sensitive vars to the world, but the GitHub token is required for assembling the docs from the other repos.

I'm not so sure because I just had this PR merged yesterday https://github.com/prometheus/docs/pull/2138

juliusv:
> I'm not so sure because I just had this PR merged yesterday https://github.com/prometheus/docs/pull/2138

Ah indeed, I retried the build and now it succeeded.

> @juliusv Yes, I think we should sync this file. But some repos have repo-specific instructions. We'll have to come up with pattern to handle this. Either move the repo-specific stuff to a separate file, or have some kind of "# Additional considerations` section that we can patch around.

Oh, I wasn't aware of some repos having more specific instructions. That complicates things... I mean for now I'd be ok just merging this one here then, if other files are diverging already as well.

CONTRIBUTING.md
juliusv: Yeah, that sounds even better!

hdost:
@juliusv Looks good to go, and I could always push PRs to some of the others with this same language. Just let me know 👍🏼

juliusv:
@hdost Definitely not gonna say no to that, but we do have a lot of repos and that would be creating a lot of work for you that seems like it should be automated in *some* way :)

SuperQ:
We need to resolve at least some of the wording diff with `prometheus/prometheus`. There appear to have been a number of small changes.

hdost:
> We need to resolve at least some of the wording diff with `prometheus/prometheus`. There appear to have been a number of small changes.

Yea `prometheus/prometheus` is significantly different. I'm not sure the others (more code based) would suffer as much from the same issues as docs are something I think more people are likely to not feel they need to try changes locally first.

RichiH:
On mobile.

There's also a "be" missing around the DCO link.

PR #2153, opened 2022-06-06
Unofficial Go client addition

pascaldekloe:
Output looks like the following.

```
< HTTP/1.1 200 OK
< Content-Type: text/plain;version=0.0.4;charset=utf-8
< Date: Sun, 07 Mar 2021 15:22:47 GMT
< Content-Length: 351
< 
# Prometheus Samples

# TYPE db_connects_total counter
# HELP db_connects_total Number of established initiations.
db_connects_total 4 1615130567389

# TYPE db_cache_bytes gauge
# HELP db_cache_bytes Size of collective responses.
db_cache_bytes 7600 1615130567389

# TYPE db_disk_usage_ratio gauge
db_disk_usage_ratio{device="sda"} 0.19 1615130563595
```

### DIFF
--- a/content/docs/instrumenting/clientlibs.md
+++ b/content/docs/instrumenting/clientlibs.md
@@ -28,6 +28,7 @@ Unofficial third-party client libraries:
 * [Dart](https://github.com/tentaclelabs/prometheus_client)
 * [Elixir](https://github.com/deadtrickster/prometheus.ex)
 * [Erlang](https://github.com/deadtrickster/prometheus.erl)
+* [Go](https://github.com/pascaldekloe/metrics)
 * [Haskell](https://github.com/fimad/prometheus-haskell)
 * [Lua](https://github.com/knyar/nginx-lua-prometheus) for Nginx
 * [Lua](https://github.com/tarantool/metrics) for Tarantool


roidelapluie:
I do not think we should document this unofficial client lib. I prefer users to stick to our main implementation.

pascaldekloe:
Makes sense in a way. Choice can be good though. Some Go developers are not using Prometheus because the standard library is too bloated and complicated. This implementation is quite a bit faster too.

beorn7:
Which, I guess, circles back to the question if we should link to different solutions for the same problem, with some comments about the different trade-offs taken.

The old stance used to be that we only list one solution for one problem to "unify the community behind it" or something. I'm not sure if we ever officially gave up on that.

beorn7:
FTR: I never believed that only listing one solution would unify the community. More the contrary. But we had this discussion elsewhere a while ago.

roidelapluie:
I don't strongly disagree. However, because it is in a different list than the official one, would it be possible to make it clearer that there is an official alternative?

roidelapluie:
official/unofficial is known in the prometheus ecosystem, I don't feel like changing the terminology here.

beorn7:
I think we could just list this with a short comment like "a more lightweight alternative to the official Go instrumentation library".

However, a the discussion at the team meeting today concluded that we do _not_ have made an explicit call that we want to list more than one solution for the same problem. I don't think we should change our past stance implicitly by sneaking in links. Let's put this question on the agenda of the dev-summit to come to an informed decision.

pascaldekloe:
You may have missed the concept here @SuperQ.

> “Counter, Integer, Real and Histogram are live representations of events. Value updates should be part of the respective implementation. Otherwise, use Sample for captures with a timestamp.”
> — https://pkg.go.dev/github.com/pascaldekloe/metrics#pkg-overview

Samples are a lightweight alternative to the (lazy-fetch) callback mechanisms provided by the original Go implementation. The mechanism also allows for more precise timing, which is critical to some use-cases.

pascaldekloe:
Metrics are unreliable without timestamps, especially when you do lazy fetches during the exposition formation, as the overview ends up with old and fresh values. Correctness by default is something I hold in high regard, even if most users won't notice the difference. We can be different. Choice is nice.

beorn7:
I should probably mention that I haven't really reviewed @pascaldekloe's instrumentation library in detail. I think the discussion if we should generally consider listing more than one solution for any one given problem on prometheus.io is separate from the discussion what our quality bar is to list a solution at all.

PR #2154, opened 2022-06-08
docs(naming): escape underscores in timestamp gauge metric name

naseemkullah:


RichiH:
This was already fixed in https://github.com/prometheus/docs/pull/2147 -- thanks, though!

Issue #2167, opened 2022-07-04
Contradictive recommendations on naming for application-specific metrics


Alexsey:
According to the [dosc](https://prometheus.io/docs/practices/naming/#metric-and-label-naming), the naming of application-specific metrics should be
> A metric name... <...> ...should have a (single-word) application prefix relevant to the domain the metric belongs to. <...> For metrics specific to an application, the prefix is usually the application name itself.

But at the same time below we have
> Do not put the label names in the metric name, as this introduces redundancy and will cause confusion if the respective labels are aggregated away.

So does it mean that it's recommended to omit the label with the name of the application for application-specific metrics, or application-specific metrics is an exception and it's recommended to duplicate the application name in both metric and label?

After clarifying the ambiguity I'm willing to make a PR for the docs for clarifying the actual recommendation

Alexsey:
@beorn7 thank you for the reply!

Could you please clarify
> If a metric is really application specific (e.g. prometheus_rule_evaluation_failures_total), then there is no need to duplicate the information

Do you mean that for application-specific metrics it's ok to put the application name as the prefix of the metric name, but in order not to duplicate the information we shouldn't add `job` label with application name?

beorn7:
The `job` label is typically a target label added by Prometheus. And yes, Prometheus should still do that. (There is also no obligation to keep the job name the same as the general application name, or to use a job label at all. It really depends on your deployment model. But should both your job label and the metric prefix coincide, that's fine. Just don't add a label with the application name during instrumentation.)

PR #2168, opened 2022-07-06
Solve contradictive metrics naming recomendation

Alexsey:
For metrics specific to an application it's recommended to put the application name as a prefix and for all metrics, it's recommended not to put labels into metrics names. After the discussion in the [issue](https://github.com/prometheus/docs/issues/2167) it appeared that for metrics specific to applications it's ok.

Besides that, the "Do not put the label **names** in the metric name" has been changed to "Do not put the label **values** in the metric name" that was probably just a typo.

Signed-off-by: Aleksey Bobyr <agat00@gmail.com>

### DIFF
--- a/content/docs/practices/naming.md
+++ b/content/docs/practices/naming.md
@@ -58,8 +58,10 @@ Use labels to differentiate the characteristics of the thing that is being measu
  * `api_http_requests_total` - differentiate request types: `operation="create|update|delete"`
  * `api_request_duration_seconds` - differentiate request stages: `stage="extract|transform|load"`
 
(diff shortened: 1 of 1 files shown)


RichiH:
I believe this is fine as-is. The "label name" is on purpose.

"job" is not a label name anyone would ever expose during normal operation; it would be only used in very few situations.

As such, unless I am misunderstanding the intention, I would lean towards closing this PR.

Alexsey:
@RichiH Well, my questions have been answered in the issue, so if you think docs are ok, I have nothing against closing the PR

PR #2170, opened 2022-07-07
New Relic Exporter new repository

klinux:
The actual exporter in the exporter page, was abbandoned and do not have chart to install the
exporter, and the exporter had a bug, that causes exporter terminated
by error in Kubernetes, we fixed this race condition and publish the
new exporter, with helm chart and image published at docker hub.

Signed-off-by: Kleber Rocha <klinux@gmail.com>

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -178,7 +178,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [JMX exporter](https://github.com/prometheus/jmx_exporter) (**official**)
    * [Munin exporter](https://github.com/pvdh/munin_exporter)
    * [Nagios / Naemon exporter](https://github.com/Griesbacher/Iapetos)
-   * [New Relic exporter](https://github.com/mrf/newrelic_exporter)
+   * [New Relic exporter](https://github.com/corabank/newrelic_exporter.git)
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/exporters.md
```diff
@@ -178,7 +178,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [JMX exporter](https://github.com/prometheus/jmx_exporter) (**official**)
    * [Munin exporter](https://github.com/pvdh/munin_exporter)
    * [Nagios / Naemon exporter](https://github.com/Griesbacher/Iapetos)
-   * [New Relic exporter](https://github.com/mrf/newrelic_exporter)
+   * [New Relic exporter](https://github.com/corabank/newrelic_exporter.git)
```
RichiH, line 181: This should be without the `.git` suffix.

RichiH (changes requested):

content/docs/instrumenting/exporters.md
klinux: Of course, in your time, in the meantime, we will be making other improvements to the chart and improving the documentation.

klinux: Hum Ok I will remove the .git and update the PR

jan--f:
Hello from the bug scrub!

It looks like the old link is more actively maintained then the replacement.

PR #2182, opened 2022-08-24
Update README: remove obsolete/unnecessary command

chalin:
There's no `docs` folder in this repo. Maybe the `cd` command was meant to refer to this repo itself, but there's no other context in the README implying that we wouldn't already be in this repo to execute the command.


juliusv:
:+1: Thanks!

PR #2184, opened 2022-08-24
Add GitPod config file

chalin:
/cc @nate-double-u


juliusv:
:+1: Thanks!

It'd even be possible to launch both a webserver to run the site (`bundle exec nanoc view`) and a watcher for changes in the sources (`bundle exec guard` IIRC), but this is good for now.

PR #2185, opened 2022-08-24
Gitignore core.* files

chalin:
Note sure what these files are, but they don't seem to be needed under version control. Local build causes a lot of them to be created:

```console
$ ls core.*
core.5096
...
core.5906
$ ls core.* | wc -l
103
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ prometheus_rsa
 # Ruby artifacts
 .bundle/
 vendor/
+
+/core.*


SuperQ:
You should probably put these in your global .gitconfig ignore.

```
[core]
	excludesfile = ~/.gitignore
```

SuperQ:
Also, core files means something is crashing. That's probably worth debugging.

chalin:
Ok, that works for me, thanks for the explanation.

PR #2187, opened 2022-09-05
Relace `ELK stack` with different solutions

Rohlik:
ELK is legacy name, nowadays we should use Elastic stack instead - https://www.elastic.co/what-is/elk-stack.

### DIFF
--- a/content/docs/introduction/faq.md
+++ b/content/docs/introduction/faq.md
@@ -71,7 +71,7 @@ For cases where you must push, we offer the [Pushgateway](/docs/instrumenting/pu
 
 ### How to feed logs into Prometheus?
 
-Short answer: Don't! Use something like the [ELK stack](https://www.elastic.co/products) instead.
+Short answer: Don't! Use something like the [Elastic stack](https://www.elastic.co/products) instead.
 
 Longer answer: Prometheus is a system to collect and process metrics, not an
 event logging system. The Raintank blog post
--- a/content/docs/introduction/faq.md
+++ b/content/docs/introduction/faq.md
@@ -71,7 +71,7 @@ For cases where you must push, we offer the [Pushgateway](/docs/instrumenting/pu
 
 ### How to feed logs into Prometheus?
 
-Short answer: Don't! Use something like [Loki](https://grafana.com/oss/loki/) or [OpenSearch](https://opensearch.org/) instead.
+Short answer: Don't! Use something like [Grafana Loki](https://grafana.com/oss/loki/) or [OpenSearch](https://opensearch.org/) instead.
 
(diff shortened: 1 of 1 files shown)


roidelapluie:
I would suggest we promote open source software instead (opensearch and
Loki)

Le lun. 5 sept. 2022, 15:30, Tomas Rohrer ***@***.***> a
écrit :

RichiH:
NB: I am doubly biased here: Not only do I work at Grafana, I have wanted and arguably worked towards what is now Loki since 2015.

I agree that we should be suggesting Open Source, that ELK/Elastic is listed seems more like an oversight. I would prefer to link to tools which have the same data model, but again, I am biased.

RichiH:
And thank you @Rohlik for opening this PR, even though it's currently not going the way you anticipated.

content/docs/introduction/faq.md
```diff
@@ -71,7 +71,7 @@ For cases where you must push, we offer the [Pushgateway](/docs/instrumenting/pu
 
 ### How to feed logs into Prometheus?
 
-Short answer: Don't! Use something like the [ELK stack](https://www.elastic.co/products) instead.
+Short answer: Don't! Use something like the [Elastic stack](https://www.elastic.co/products) instead.
```
roidelapluie, line 74:
```suggestion
Short answer: Don't! Use something like [Loki](https://grafana.com/oss/loki/) or [OpenSearch](https://opensearch.org/) instead.
```

RichiH: Nit: The official name is "Grafana Loki" for trademark reasons.

RichiH:
@Rohlik would you be ok merging @roidelapluie 's suggestion or would you prefer ir

Rohlik:
> @Rohlik would you be ok merging @roidelapluie 's suggestion or would you prefer if we made a new PR and closed this one?

Merged :)

content/docs/introduction/faq.md
```diff
@@ -71,7 +71,7 @@ For cases where you must push, we offer the [Pushgateway](/docs/instrumenting/pu
 
 ### How to feed logs into Prometheus?
 
-Short answer: Don't! Use something like the [ELK stack](https://www.elastic.co/products) instead.
+Short answer: Don't! Use something like [Loki](https://grafana.com/oss/loki/) or [OpenSearch](https://opensearch.org/) instead.
```
roidelapluie, line 74:
```suggestion
Short answer: Don't! Use something like [Grafana Loki](https://grafana.com/oss/loki/) or [OpenSearch](https://opensearch.org/) instead.
```

PR #2192, opened 2022-09-08
Add Udaan as Adopter of Prometheus

prajain12:
We use Prometheus in Production

Signed-off-by: Prakul Jain <prakul.jain@udaan.com>


roidelapluie:
Thank you for using Prometheus, as thousands of other companies. Unfortunately, we have decided to freeze the list of companies on prometheus.io and not accept new companies at the moment.

PR #2203, opened 2022-09-18
Fixing java client benchmarks broken URI

maximveksler:
Signed-off-by: Maxim Veksler <maxim@vekslers.org>


jan--f:
Hello from the bug scrub.
Seems like this was fixed elsewhere. https://github.com/prometheus/docs/blob/eae7ff3bc6da78650a8226d83084ee104bbd0bef/docs/introduction/faq.md?plain=1#L182 now has the correct link.

Issue #2181, opened 2022-08-24
Migrate to Google Analytics 4 (GA4)


chalin:
This is part of a CNCF-wide effort to upgrade project websites to GA4 since **Google has deprecated Universal Analytics** (UA). For more details, see:

- https://github.com/cncf/techdocs/issues/108, including reference to:
  - A suggested [migration process](https://github.com/cncf/techdocs/blob/main/docs/analytics.md#migrating-to-google-analytics-4), 
  which we've used to [migrate a few CNCF projects](https://docs.google.com/spreadsheets/d/1Mx4LhdI2Un-rvGMI73SlHxQH9D2HABAJclMB3dd6lnA) already.

### Tasks: stage 1 (no code changes are necessary)

- [x] Create a GA4 site tag under the **CNCF projects** account.
  The new GA4 stream measurement ID is: **`G-80ZM8LGB96`**
- [x] [Connect] to the GA4 site tag from the UA tag.

### Tasks: stage 1b - upgrade to use `gtag.js`

- [x] Upgrade website to use [gtag.js] library - #2207
- [x] ~Use GA4 setup assistant to "reconnect" the two tags.~ Can't use the GA4 setup assistant to connect site tags from different accounts it would seem.
- [x] Confirm that the GA4 site tag is receiving events
- [x] Confirm that the UA site tag, `UA-58468480-1`, is still receiving events

### Tasks: stages 2 & 3

- [x] [Connect] UA ID from GA4 site tag
- [x] Configure GA4 ID through [Netlify snippet injection], "Google Analytics 4 - only in production", just before `</head>`:
  ```liquid
  {% if CONTEXT == 'production' %}
  <script async src="https://www.googletagmanager.com/gtag/js?id=G-80ZM8LGB96"></script>
  <script>
    window.dataLayer = window.dataLayer || [];
    function gtag(){dataLayer.push(arguments);}
    gtag('js', new Date());
  
    gtag('config', 'G-80ZM8LGB96');
  </script>
  {% endif %}
  ```
- [x] Dis[connect] to the GA4 site tag from the UA tag
- [x] Confirm that the GA4 site tag is receiving events and that the UA site tag, `UA-58468480-1`, is still receiving events:
  > <img width="600" alt="image" src="https://user-images.githubusercontent.com/4140793/201354527-39073838-2d88-442b-b4eb-250439137133.png">

### Related

- https://github.com/promcon/website/issues/245
- Followup: #2234

### More:

- #2208
- #2209
[...]

PR #2207, opened 2022-09-26
Switch to use of gtag.js library for analytics

chalin:
- Contributes to #2181, a step towards using GA4
- Note that this PR only switches to using the [gtag.js] library.
  **The same UA property ID is used**. 
  By switching to use of [gtag.js], we enable [connected] site tags to work. In particular, the connection (forwarding) to the newly created GA4 site tag.

/cc @juliusv 

[connected]: https://support.google.com/analytics/answer/9973999
[gtag.js]: https://developers.google.com/analytics/devguides/collection/gtagjs

### DIFF
--- a/layouts/footer.html
+++ b/layouts/footer.html
@@ -6,15 +6,14 @@
     <script src="/assets/docs.js"></script>
     <!-- IE10 viewport hack for Surface/desktop Windows 8 bug -->
     <script src="/assets/ie10-viewport-bug-workaround.js"></script>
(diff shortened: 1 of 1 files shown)


chalin:
With this change, the GA4 tag starts receiving events:

> image

No events were being emitted to the GA4 stream before the preview was deployed.

Previews should _**not**_ generate analytics, nor be indexable, but that will be addressed separately.

juliusv:
:+1: Thanks!

Issue #2208, opened 2022-09-26
Analytics: resolve redundant-hostnames warning


chalin:
Google analytics reports the following issue:

> Property prometheus.io is receiving data from redundant hostnames. Some of the redundant hostnames are:
>
> - prometheus.io
> - www.prometheus.io
>
> Redundant hostnames are counted as separate rows in reports, so hits that are going to the same page on your site from different hostnames will be split into multiple rows. With data split across multiple rows, traffic to specific pages will appear lower than it actually is.

/cc @caniszczyk

chalin:
I can confirm the problem:

```console
$ curl -i https://prometheus.io 2>/dev/null | head -25 | grep '^[lH]'  # Drop the grep to see the full header
HTTP/2 200 
$ curl -i https://www.prometheus.io 2>/dev/null | head -25 | grep '^[lH]' 
HTTP/2 200 
```

The second response should be a 301. Note that this is an issue for both UA and GA4 analytics.

/cc @juliusv

chalin:
Following the merge of #2211, we're now getting a 301 as we should for `www`:

```console
$ curl -i https://www.prometheus.io 2>/dev/null | head -25 | grep '^[lH]' 
HTTP/2 301 
location: https://prometheus.io/
$ curl -i https://prometheus.io 2>/dev/null | head -25 | grep '^[lH]' 
HTTP/2 200 
```

I've triggered a re-check in Google Analytics redundant-hostname issue:

> <img width="400" alt="image" src="https://user-images.githubusercontent.com/4140793/192791571-c186ad22-eb96-44ca-8dc2-a30cccbd86a7.png">

chalin:
PR #2211 seems to have mitigated the issue but not fully resolved it. In any case, #2211 _should not_ have been required. I've inquired with Netlify support, and they suggested looking at the project's DNS settings.

@juliusv - do you have access to the DNS settings? DNS seems to be provided by Cloudflare for the Prometheus websites.

One option would be to drop Cloudflare DNS and go with Netlify DNS, which might make sense given that Netlify is hosting anyways. WDYT?

chalin:
Ah, I believe that the Linux Foundation is the registrant. I'll dig further on our side and confirm.

juliusv:
[...]

PR #2211, opened 2022-09-27
Redirect www to apex domain using Netlify config

chalin:
This is an attempt at resolving

- #2208

without any further Netlify configuration changes.

/cc @juliusv

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,4 @@
+[[redirects]]
+from = "https://www.prometheus.io/*"
+to = "https://prometheus.io/:splat"
+force = true


chalin:
@juliusv, I'm not sure why the build is failing in this way:

```
5:02:31 PM: $ make downloads compile
5:02:31 PM: curl -sf -H 'Accept: application/vnd.github.v3+json' <GITHUB_AUTHENTICATION> https://api.github.com/repos/prometheus/prometheus > downloads/prometheus/repo.json
5:02:31 PM: make: *** [Makefile:22: downloads/prometheus/repo.json] Error 22
```

Any idea? When I build locally, it works fine. The `curl` is failing maybe because the `GITHUB_AUTHENTICATION` can't be accessed due to the new `netlify.config` file?

juliusv:
Hi @chalin, yeah I think the reason would be the GitHub token, the Netlify build says:

```
11:02:14 PM: The following environment variables have been removed from this build to keep them secret:
11:02:14 PM: - GITHUB_AUTHENTICATION
```

Netlify omits secret env vars in builds triggered by non-team-members (because otherwise you could just extract them).

I just restarted the build myself in Netlify and now it seems to work. Of course we can't really test the apex domain redirect in the preview deployment anyway, so we just have to hope that it works.

However, we already have a bunch of Netlify redirects in https://github.com/prometheus/docs/blob/main/content/_redirects as well. Can we move this one in there as well, or is the `_redirects` file format not powerful enough for that?

chalin:
Thanks for looking info the issue @juliusv.

> Can we move this one in there as well, or is the `_redirects` file format not powerful enough for that?

The convention we've been using is to separate redirect rules based on their role:

- We use  `_redirects` exclusively for intra-site redirect rules, whose purpose is to avoid 404s. This file is subject to more frequent updates, and tends to grow considerably over time.
- We put (sub)domain redirect rules (and other HTTP-response header config) in `netlify.toml`. This is a file that contributors should _very_ rarely be making changes to.

This separation helps to reduce the chance of inadvertent changes to the domain redirect rules, which might easily go unnoticed. Does that make sense?

juliusv:
Thanks for the explanation, @chalin, that makes sense! I'll merge it and stick around to check that it's all working correctly once deployed.

PR #2235, opened 2022-11-11
Drop UA script

### LINKED ISSUE #2209: Enable analytics for the production deploys only
With the current site setup, _all_ deploys (this includes previews), generate analytics. They shouldn't -- analytics should _only_ be enabled for the production deploy.

Reference:

- [SEO, Analytics and site-local search](https://github.com/cncf/techdocs/blob/main/assessments/criteria.md#seo-analytics-and-site-local-search).

Related:

- #2181

Tasks:

- [ ] Fix this for UA - #2235
- [x] Fix this for GA4: snippet code added for #2181 

/cc @nate-double-u @caniszczyk

### LINKED ISSUE #2234: Drop Google Universal Analytics (UA) code
As is mentioned in #2181, the UA ID is [connected] to the new GA4 ID, and hence doesn't need to be explicitly registered. The UA ID will continue to receive events as long as it's [connected], until Google drops UA support.

[connected]: https://support.google.com/analytics/answer/9973999

chalin:
- Followup to #2181 
- Closes #2234
- Closes #2209 for UA

/cc @juliusv

### DIFF
--- a/layouts/footer.html
+++ b/layouts/footer.html
@@ -6,16 +6,6 @@
     <script src="/assets/docs.js"></script>
     <!-- IE10 viewport hack for Surface/desktop Windows 8 bug -->
     <script src="/assets/ie10-viewport-bug-workaround.js"></script>
-    <!-- Global site tag (gtag.js) - Google Analytics -->
-    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-58468480-1"></script>
-    <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-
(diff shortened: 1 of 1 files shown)


juliusv:
:+1:

chalin:
I confirm that:

- The UA ID is gone:

  ```console
  $ curl -i https://prometheus.io 2>/dev/null | grep UA-
      <meta http-equiv="X-UA-Compatible" content="IE=edge">
  ```

- The GA4 ID is still in place:

  ```console
  $ curl -i https://prometheus.io 2>/dev/null | grep 'www.google' | head -10
  <script async src="https://www.googletagmanager.com/gtag/js?id=G-80ZM8LGB96"></script>
  ```

- Both tags are still processing events:

  > image

PR #2238, opened 2022-11-17
Update metric_types.md

scottpantall:
If I'm reading the contributing guidelines correctly and this is a trivial improvement I think I'm supposed to tag @RichiH in this PR. Looking forward to using Prometheus more!


RichiH:
Thanks!

PR #2250, opened 2022-12-16
added the envite logo to the commercial support companies

ultimateJJ:
With this commit I would like to add the invite consulting GmbH to the commercial support site.

kind regards
Julian John
envite consulting GmbH
envite.de
Signed-off-by: Julian John <julian.john00j@gmail.com>


ultimateJJ:
Hi @juliusv @RichiH ,
Marius Brill told me I can ask you to take a look at this PR.
Would be nice if you could merge this request.

kind regards
Julian John

.gitignore
```diff
@@ -13,6 +13,9 @@ crash.log
 # macOS artifacts
 .DS_Store
 
+# IntelliJ artifacts
+.idea
```
juliusv, line 17: We don't add IDE-specific files to our .gitignores, but probably this unrelated change just slipped in by accident. Could you remove it?

content/support-training.html
```diff
@@ -76,6 +76,12 @@ <h2>Commercial support</h2>
               </a>
           </div>
 
+          <div class="col-md-3 col-sm-6 doc-content tc-col">
+            <a href="https://envite.de" class="btn btn-training btn-default btn-large" target="_blank">
```
juliusv, line 80: We don't require it (because others also don't always have it), but in case you have a Prometheus-specific page you can link to, it would be greatly preferred. That way, anyone looking for support and clicking through all the various support companies will understand much better how envite can help them specifically around Prometheus. Or if you don't have such a page yet, it's also a good idea to update this later with one.

juliusv (commented):
Thanks! Just two comments.

content/support-training.html
ultimateJJ: At the Moment we don't have one, but we will add it later.

.gitignore
ultimateJJ: yes I overlooked this change. It's reverted

juliusv:
:+1: Thanks!

PR #2261, opened 2023-01-17
Retire Russia/Ukraine banner after ~1 year

juliusv:
As discussed on prometheus-team in December.


SuperQ (approved):
Thanks, considering the continue support that is being given, I would say it was effective.

PR #2312, opened 2023-04-17
Update naming.md - fix data model hyperlink

ShaiDemri:
The link led to a 404 github page. Fixed it to the correct location.

### DIFF
--- a/content/docs/practices/naming.md
+++ b/content/docs/practices/naming.md
@@ -14,7 +14,7 @@ practices, e.g. naming conventions, differently.
 
 A metric name...
 
-* ...must comply with the [data model](/docs/concepts/data_model/#metric-names-and-labels) for valid characters.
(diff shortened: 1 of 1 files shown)


ShaiDemri:
@RichiH

RichiH:
This is working as expected already; https://prometheus.io/docs/practices/naming/ links to https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels correctly for me.

What browser are you using? Any special plugins?

I am closing this as it's the same link format we use throughout the website, but feel free to reopen if I got this wrong.

ShaiDemri:
broken data model link

I am using Chrome but checked in Firefox also, same behavior.
Take a look at the link that is about to open at the bottom of the attached screenshoot. I believe this is due to the fact that the used link is relative to the current URL. 
Also tried from incognito (not logged in to Github), same behavior.
The link does not direct to [https://prometheus.io](https://prometheus.io/) domain but rather stays in the [https://github.com/](https://github.com/) domain.

PR #2322, opened 2023-04-27
remote_write_spec: Update the receivers list.

hdost:
* Sort the list by Prometheus and Avalanche First, then sort the
  rest in alphabetical order.
* Add some relevant links.
* Add VictoriaMetrics as receiver.


juliusv:
:+1: Thanks! Coincidentally I didn't notice that this page had even been added. I don't think it belongs under "Concepts" in the top-level menu at all, those are meant to be the absolute fundamental concepts behind Prometheus, not detailed specs. But that's a different topic :)

PR #2325, opened 2023-04-27
Update remote_write_spec.md

mlunadia:
Adding Elastic Agent to the list of Compatible Senders and Receivers


juliusv:
:+1: Thanks!

PR #2349, opened 2023-07-11
Update instrumenting_http_server_in_go.md

nvictor:
This fixes a Markdown rendering issue. Hope I am tagging the right person.
@juliusv


juliusv:
:+1: Thanks! There's a couple more things in this doc that would need fixing (e.g. "Default Register" should be "default registry"), but that's a separate issue :)

PR #2352, opened 2023-07-14
Change DO logo from png to svg

RichiH:
As per https://github.com/prometheus/docs/pull/2351#issuecomment-1635619390


RichiH:
@juliusv same behaviour as before. Could you help with sizing the SVG, please?

RichiH:
@juliusv that works, thanks. Are you particular about 250px? I think the 200 looked better. Alternatively, the CNCF logo is 300 wide, so we can make them the same.

RichiH:
NB: The whole text width is a clickable link for both CNCF & DO, which is not HTML default behaviour -- but I don't think it's worth digging to find what causes this.

juliusv:
> Are you particular about 250px? I think the 200 looked better.

Either is fine :+1: 

> The whole text width is a clickable link for both CNCF & DO

That's because we have a CSS rule that sets doc content images to `display: block`, but fine with me.

:+1:

PR #2357, opened 2023-07-27
Adds Mixpanel to users

austinpray-mixpanel:
We recently decided to go all in on prom! ❤️

https://mixpanel.com/ / https://github.com/mixpanel

>image


juliusv:
:+1: Thanks! I think a while ago we were planning on not really extending this section any further and potentially even replacing it with something new / different (ADOPTERS.md file?), but I don't remember the specifics at the moment. Can't hurt too much to add some more for now I guess :)

PR #2359, opened 2023-07-29
Add recording rules to glossary

khallai:
Added recording rules to glossary. This will provide a quick definition to a user wanting to know what a recording rule is without searching the docs. Also adding a link to navigate the user to the configuration > recording rules section could be helpful.

sign off by: Lasea75(khallai) <lasea75@gmail.com>


juliusv:
:+1: Thanks! To cover both kinds of rules in Prometheus, adding alerting rules as an entry (separate from an alert itself) would also make sense. Ultimately though I'm not 100% sure how to define the inclusion criteria for the glossary. The current list of terms seems somewhat arbitrary, with some common ones missing and some less common ones present.

PR #2368, opened 2023-08-10
Correct what might be a possible grammar error

moijes12:
Changed `... scraping metrics HTTP endpoints..` to `...scraping metrics from HTTP endpoints...` as the latter might be more correct.

### DIFF
--- a/content/docs/introduction/first_steps.md
+++ b/content/docs/introduction/first_steps.md
@@ -5,7 +5,7 @@ sort_rank: 3
 
 # First steps with Prometheus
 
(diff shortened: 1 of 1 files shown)


juliusv:
Thanks! Hmm, I see where you're coming from. Both variants are fine, but I'm not sure which is better actually. It was indeed intended to say "metrics endpoints", as the scraping cannot happen from arbitrary endpoints, so that nuance could get lost with the change. And there's already another "from" just before that in the sentence.

moijes12:
@juliusv Thanks for clarifying this. I think I can close this PR now

PR #2372, opened 2023-08-23
Fix: Remove broken/incorrect links in docs/integrations.md

Hapshanko:
1. Removed servicenow integration link due to it resulting in a 404, as the repo does not seem to exist anymore.
2. Removed rocket.chat webhook integration link, as it leads to a link of monitoring rocket.chat with prometheus, and not integrating alertmanager with rocket.chat as a receiver.


juliusv:
:+1: Thanks!

PR #2378, opened 2023-08-31
Update exposition_formats.md

hoppedmalt:
Removed false instructions on how to use (unsupported) custom timestamps.


RichiH:
It's officially supported, just not encouraged.

PR #2379, opened 2023-08-31
Update first_steps.md

Pradeep-Kurapati:
Donwload link points to an empty page. So I updates the link to (https://prometheus.io/download/)

Please let me know if I'm not doing it right. Thank you.

@juliusv @SuperQ


juliusv:
@Pradeep-Kurapati Thanks! However, the relative link is correct as it is here - this docs repo is not meant to be consumed directly, it gets deployed as the Prometheus website. See https://prometheus.io/docs/introduction/first_steps/ - the link finds the right target there :)

PR #2382, opened 2023-09-03
Add Guide for sending Native OpenTelemetry Metrics to Prometheus

prathamesh-sonpatki:

### DIFF
--- a/content/docs/guides/opentelemetry.md
+++ b/content/docs/guides/opentelemetry.md
@@ -1,0 +1,208 @@
+---
+title: Ingesting OpenTelemetry Metrics
+---
+
+# Ingest Native OpenTelemetry Metrics to Prometheus
+
+Prometheus supports ingesting native OpenTelemetry Metrics from release 2.47 onwards. Before this change, Prometheus was not supporting ingesting the OpenTelemetry Metrics natively, there was a way to convert the OpenTelemetry Metrics into Prometheus Metrics format either via Prometheus Remote Write Exporter.
+
+In this guide, we will learn about
+
+- Configuration changes required in Prometheus setup to enable the ingestion of OpenTelemetry Metrics.
+- Setting up a Python application with OpenTelemetry SDK.
+- Emitting OpenTelemetry metrics via auto-instrumentation.
+- Ingesting those metrics in Prometheus natively.
+- Setup manual instrumentation of metrics.
+- Ingesting those metrics in Prometheus natively.
+
+> The guide expects that you have basic understanding of differences between Prometheus / OpenMetrics format and [OpenTelemetry Metrics standard](https://opentelemetry.io/docs/specs/otel/metrics/).
+
+## Prometheus Configuration
+
+There are no guarantees that OpenTelemetry data will arrive in order. However, Prometheus expects data to be in order by default. This is one of the nicer properties of a scrape-based system in which the timing is controlled centrally, and it allows substantial efficiency gains.
+
+Yet, because of this data may be dropped if sending OpenTelemetry data directly to Prometheus. The fix is easy: Simply enable Out of Order ingestion in Prometheus.
+
+
+``` yaml
+storage:
+  tsdb:
+    # A 10min time window is enough because it can easily absorb retries and network delays.
+    out_of_order_time_window: 10m
+```
+
+The support for OpenTelemetry Metrics ingestion is supported in Prometheus via a feature flag. Run Prometheus with the `otlp-write-receiver` flag enabled as follows:
+
+``` shell
+prometheus --enable-feature=otlp-write-receiver --config=prometheus.yml
+```
+
+Once the feature is successfully enabled, you can see the evidence in the logs.
+
+``` shell
+$ prometheus --enable-feature=otlp-write-receiver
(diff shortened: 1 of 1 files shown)


content/docs/guides/opentelemetry.md
```diff
@@ -0,0 +1,208 @@
...
+# Ingest Native OpenTelemetry Metrics to Prometheus
+
+Prometheus supports ingesting native OpenTelemetry Metrics from release 2.47 onwards. Before this change, Prometheus was not supporting ingesting the OpenTelemetry Metrics natively, there was a way to convert the OpenTelemetry Metrics into Prometheus Metrics format either via Prometheus Remote Write Exporter.
+
+In this guide, we will learn about
+
+- Configuration changes required in Prometheus setup to enable the ingestion of OpenTelemetry Metrics.
+- Setting up a Python application with OpenTelemetry SDK.
+- Emitting OpenTelemetry metrics via auto-instrumentation.
+- Ingesting those metrics in Prometheus natively.
```
hterik, line 14: There is a lot happening here. It's nice to have a complete example, but for the purpose of demonstrating Prometheus, isn't this a bit too much that is unrelated to the topic at hand? 

Setting up flask app and auto-instrumentations feels like out of scope for basic tutorial here? I believe a basic cli that pushes manually collected metrics once should be a more appropriate level. Then linking to opentelemetry tutorials for more advanced examples.

Especially considering this could be read by someone who doesn't come from Python background.

What do others think?

content/docs/guides/opentelemetry.md
```diff
@@ -0,0 +1,208 @@
...
+  "result": 105
+}
+```
+
+## Auto Instrumentation
+
+Let's add the OpenTelemetry instrumentation now. We will use Otel's Flask instrumentation library which auto-instruments all request handelers.
+
+``` shell
+pip opentelemetry-api opentelemetry-sdk opentelemetry-exporter-otlp opentelemetry-instrumentation-flask
```
hterik, line 125: Would be nice if example included creation of virtualenv. Installing to system python is against best practice. In newer versions of ubuntu, pip install fails and asks to provide `--break-system-packages` arg.

content/docs/guides/opentelemetry.md
```diff
@@ -0,0 +1,208 @@
...
+FlaskInstrumentor().instrument_app(app)
+```
+
+A lot is happening is here, let's break it down.
+
+- We created the OTLP Exporter with the Prometheus endpoint which accepts the OpenTelemetry metrics in native format. The endpoint path is `/api/v1/otlp/v1/metrics`.
+- We setup required `reader`, `resource` and `provider` for metrics instrumentation.
+- We wrapped the Flask app in `FlaskInstrumentor's` `instrument_app` method so that the auto instrumentation will kick in.
+
+When the application is run and few requests are sent to it, the auto instrumented metrics will be **pushed** from the application to Prometheus via the `FlaskInstrumentor`.
```
hterik, line 158: This can be misunderstood as its the FlaskInstrumentor that pushes the metrics. If i understand correctly, what happens is that FlaskInstrumentor only *collects* the metrics internally. PeriodicExportingMetricReader+OTLPMetricExporter does the push.

content/docs/guides/opentelemetry.md
```diff
@@ -0,0 +1,208 @@
...
+The Prometheus' OTLP write recevier endpoint takes care of coverting these OpenTelemetry metrics into Prometheus format seamlessly. Following metrics will be present in Prometheus.
+
+- `http_requests_duration_milliseconds_bucket`
+- `http_requests_duration_milliseconds_count`
+- `http_requests_duration_milliseconds_sum`
+- `http_server_active_requests`
+
+## Custom Instrumentation
+
+Let's add a custom metric in the Python application to count number of calculaor operations performed.
```
hterik, line 174: typo calculator

content/docs/guides/opentelemetry.md
prathamesh-sonpatki: @gouthamve Thoughts?

content/docs/guides/opentelemetry.md
```diff
@@ -0,0 +1,208 @@
...
+storage:
+  tsdb:
+    # A 10min time window is enough because it can easily absorb retries and network delays.
+    out_of_order_time_window: 10m
+```
+
+The support for OpenTelemetry Metrics ingestion is supported in Prometheus via a feature flag. Run Prometheus with the `otlp-write-receiver` flag enabled as follows:
+
+``` shell
+prometheus --enable-feature=otlp-write-receiver --config=prometheus.yml
```
krajorama, line 37: The feature flag is no longer needed since Prometheus 3.0

krajorama: Correction: need this flag instead: `--web.enable-otlp-receiver`.

krajorama:
Hello from the bug scrub. Thank you for starting the PR, we are unsure if the complexity of the example is too much. I (@krajorama ) think it's a bit too complicated to set up. @prathamesh-sonpatki will you continue working on this?

prathamesh-sonpatki:
@krajorama sorry, no. I don't have bandwidth right now to pick this up.

krajorama:
> @krajorama sorry, no. I don't have bandwidth right now to pick this up.

No worries, in that case we'll close it for now.

PR #2383, opened 2023-09-04
Patch 1

hoppedmalt:
Removed mention of user timestamps being supported, as they specifically are mentioned NOT to be, when use is attempted. If there is a way to use them, please document with example.

`
prometheus-node-exporter[1224]: ts=2023-09-04T07:14:27.714Z caller=textfile.go:219 level=error collector=textfile msg="failed to collect textfile data" file=test.prom err="textfile \"/var/lib/prometheus/node-exporter/junk.prom\" contains unsupported client-side timestamps, skipping entire file"
`

### DIFF
--- a/content/docs/instrumenting/exposition_formats.md
+++ b/content/docs/instrumenting/exposition_formats.md
@@ -81,7 +81,7 @@ The remaining lines describe samples (one per line) using the following syntax
 ```
 metric_name [
   "{" label_name "=" `"` label_value `"` { "," label_name "=" `"` label_value `"` } [ "," ] "}"
-] value [ timestamp ]
+] value
 ```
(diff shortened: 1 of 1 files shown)


hoppedmalt:
To me, this seems that client-side timestamps are unsupported:

`
prometheus-node-exporter[1224]: ts=2023-09-04T07:14:27.714Z caller=textfile.go:219 level=error collector=textfile msg="failed to collect textfile data" file=test.prom err="textfile \"/var/lib/prometheus/node-exporter/junk.prom\" contains unsupported client-side timestamps, skipping entire file"
`

SuperQ:
That is specific to the node exporter.

PR #2400, opened 2023-10-15
Add 404 page and redirect path

claudiusbr:
Relates to prometheus#1795

still TODO: 
- add link to latest docs. 

This is what the page looks like when I run it:

image

As this is my first contribution, I thought it would be a good idea to start the discussion by opening a PR. Hope this is ok, and look forward to some suggestions on how to improve.

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -49,3 +49,6 @@
 
 # Blog post duplicated
 /blog/2021/05/14/prometheus-conformance-results/ /blog/2021/10/14/prometheus-conformance-results/ 302!
+
+# Custom 404 page for all nonexistent assets
+/*                          /docs/404/                                  404
--- a/content/docs/404.md
+++ b/content/docs/404.md
@@ -1,0 +1,17 @@
+---
+title: 404
+is_hidden: true
+---
+
+# 404
+
+## Oops! The page you requested could not be found.
+
+Here are a few things you can try:
+
+* Check the URL to make sure you typed it correctly.
+* Try searching for the page you're looking for in the documentation search bar.
+* Browse the documentation table of contents to see if you can find the page you're looking for.
+* If you're still having trouble, please contact us for assistance.
+
+Thank you for your patience.


content/_redirects
```diff
@@ -39,6 +39,7 @@
 /docs/alerting/             /docs/alerting/latest/overview/             302!
 /docs/practices/            /docs/practices/naming/                     302!
 /docs/guides/               /docs/guides/basic-auth/                    302!
+/docs/*                     /docs/404/                                  302!
```
claudiusbr, line 42: I had a read through, and seems to me this is the way to do a redirect. But have not been able to test it on my local, so I'd be grateful for pointers on how to do it.

juliusv:
Yep that works, thanks :) I accepted the deploy preview on Netlify, which you can try here now: https://deploy-preview-2400--prometheus-docs.netlify.app/docs/404/

However it would be nice to have this for anything on prometheus.io, not just under /docs - so maybe we can just change the `/docs/*` to `/*`?

claudiusbr:
Thanks for the review @juliusv! 
I changed it to  `/*` now. Hope it still works.

juliusv:
I retriggered the preview deploy and it's now also redirecting for pages that should exist. I think I found out why: the Netlify documentation has this example: https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling and this explanation about shadowing and overrides: https://docs.netlify.com/routing/redirects/rewrites-proxies/#shadowing

I see two issues if I understand this correctly:

* The rule shouldn't return a 302 redirect code, but return a 404 (as in the Netlify example)
* We want to remove the exclamation mark at the end to not return the 404 page for pages that exist

claudiusbr:
Great. It makes semantic sense too.
I'm not that familiar with Netlify (yet), but hope it works now.
I'm assuming I can't re-trigger the preview deploys myself, so thanks for doing it @juliusv.

juliusv:
Noice! Looks good now, thanks :)

PR #2420, opened 2024-01-22
Removed remote-write-receiver allowed values in enable-feature flag

pawarpranav83:
The `--enable-feature=remote-write-receiver` flag is deprecated and is going to be removed from Prometheus 3.0 and instead users should use `--web.enable-remote-write-receiver`.

Refs - https://github.com/prometheus/prometheus/issues/13206 https://github.com/prometheus/prometheus/pull/13435

### DIFF
--- a/content/blog/2021-11-16-agent.md
+++ b/content/blog/2021-11-16-agent.md
@@ -138,7 +138,7 @@ Flags:
       --storage.agent.path="data-agent/"
                                  Base path for metrics storage. Use with agent mode only.
       (... other flags)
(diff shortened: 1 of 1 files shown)


bwplotka (changes requested):
Yes, but this is a blog post from 2021 and we cannot possibly maintain all old flags ever used in old blog posts (: In fact it might be surprising -- you expect 2021 version of Prometheus to have this feature flag etc.

Thus, I would NOT do this, but thank you for your contribution anyway! What do you think?

Tusenka (approved):

SuperQ:
Agreed, we also don't update old blog posts with all of the new flags as well. I don't think we should change this.

Tusenka (approved):

jan--f:
Hello from the bug scrub!
Closing as per comments above.

PR #2426, opened 2024-01-31
Adds Heii On-Call to alert manager webhook provider integrations

hevans66:
[Heii On-Call](https://heiioncall.com/) is a rotation management and alerting service. We implement the alert manager webhook receiver specification as a first class trigger type. This PR adds Heii On-Call to the list of supported integrations

image

image


hevans66:
@RichiH, tagging you as I think this is a "trivial fix or improvement". Let me know if I should tag someone over on alertmanager instead.

juliusv:
:+1: Thanks :)

PR #2427, opened 2024-02-01
Add Prefect2 exporter to list of miscellaneous exporters

pathfinder177:
Please add this new [exporter](https://github.com/pathfinder177/prefect2-prometheus-exporter) to exporter list. Please let me know if something missing, thanks.

Sample output:
```
# HELP prefect2_flows_runs_24h Prefect2 flows for 24h
# TYPE prefect2_flows_runs_24h gauge
prefect2_flows_runs_24h{state="FAILED"} 12.0
prefect2_flows_runs_24h{state="CANCELLED"} 0.0
prefect2_flows_runs_24h{state="CRASHED"} 0.0
```

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -240,6 +240,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [PHP-FPM exporter](https://github.com/bakins/php-fpm-exporter)
    * [PowerDNS exporter](https://github.com/ledgr/powerdns_exporter)
    * [Podman exporter](https://github.com/containers/prometheus-podman-exporter)
+   * [Prefect2 exporter](https://github.com/pathfinder177/prefect2-prometheus-exporter)
    * [Process exporter](https://github.com/ncabatoff/process-exporter)
    * [rTorrent exporter](https://github.com/mdlayher/rtorrent_exporter)
    * [Rundeck exporter](https://github.com/phsmith/rundeck_exporter)


juliusv:
@pathfinder177 Thanks! When it comes to counting things (the number of runs here), in Prometheus it's more common to use a continuous counter metric rather than exposing a precomputed rate (over 1h + 24h). But not sure if that might be challenging in your case if you have to define a start timestamp for the counter once, and then query over longer and longer time ranges, since there's no counter you can directly bridge, right?

juliusv:
@pathfinder177 Ping! :)

pathfinder177:
> @pathfinder177 Ping! :)

Hi Julius!
Sorry for delay, pls close the PR, I can reopen it.
Only March 12th I can start to implement changes

juliusv:
Oh we can just leave this open until later, no worries!

pathfinder177:
Starting tomorrow

pathfinder177:
@juliusv Hi!
I got the point about continuous counter and agree.
However there is indeed a requirement to measure such a metric for 1h and for 24h.
Why? Because prefect2 is a scheduler for "jobs" and we would like to track jobs across time periods to compare it.

If there are some improvements ideas I am very glad to improve the exporter and now I would like to merge it if you do not mind

juliusv:
@pathfinder177 Happy to merge it for now!

However, the way to track how many events per time period have happened in Prometheus is to just expose a raw counter that always increases, and then you can still do `increase(foo[24h])` or `increase(foo[1h])` over it as needed.

PR #2428, opened 2024-02-05
docs: add greptimedb as a compatible prw receiver

sunng87:
The patch adds [greptimedb](https://github.com/greptimeteam/greptimedb) as remote writer receiver.


juliusv:
@sunng87 Thanks! I have a feeling that this remote write spec doc (which is also somewhat out of place in the menu hierarchy and should be more of an internal hidden detail somewhere) is probably not the right place to maintain a full list of remote write integrations, especially since we already have that here: https://prometheus.io/docs/operating/integrations/#remote-endpoints-and-storage - so I would recommend adding yourself to that list instead (or in addition, I'm fine with both places being updated, I just think we should probably not maintain two in the long run).

sunng87:
@juliusv Nice. I was reading the spec and finding it might be helpful add greptimedb there. I just added on integration page as well.

juliusv:
:+1: Thanks! Yeah we might remove that other list sometime, but for now it's good :)

PR #2435, opened 2024-02-26
Replaced platformengineers logo

munir131:
Our current logo is not aligning with dimension of other logos. So, replaced that with appropriate dimension.

Current page:

image


juliusv:
Thanks! Yeah I guess the best would be if we changed the CSS to make logos look ok no matter their aspect ratio, but a more horizontal logo like your new one fits in better in general with the rest.

PR #2438, opened 2024-03-11
fix: CSS to make logos look same in support-training page

payalraviya:
- resolve https://github.com/prometheus/docs/pull/2435#issuecomment-1975094227

image


juliusv:
Thanks!

Old vs. new:
* This PR: https://deploy-preview-2438--prometheus-docs.netlify.app/support-training/
* Live: https://prometheus.io/support-training/

The difference is not too large, except for the Novatec logo, which now looks too small relative to the rest. But I think it's the fault of that logo PNG for including a lot of whitespace at the top and bottom (would be nice to fix that). Anyway, seems ok enough to merge :)

PR #2439, opened 2024-03-11
Use container labels instead of task labels in swarm

aryanbaghi:
`__meta_dockerswarm_container_label_prometheus_job` should be used instead of `__meta_dockerswarm_service_label_prometheus_job `


jan--f:
Hello from the bug scrub!

Why should user prefer the container label over the service label? What impact does this have on metric cardinality?

aryanbaghi:
Hi,
Honestly, I don’t clearly remember the original context since it’s been quite a while.
I close the pull request.

PR #2451, opened 2024-04-03
fix python links for Counter, Gauge, Summary and Histogram.

iamaliyousefi:
Python documentation links were not valid anymore, since they moved to prometheus.github.io/client_python.

They will be working with this commit.

Signed-off-by: Ali Yousefi iamaliyousefi@gmail.com


juliusv:
:+1: Thanks!

PR #2452, opened 2024-04-05
fix links in exposition_formats.md

Kratacoa:


juliusv:
@Kratacoa Hi, thanks! The links as they are are already correct though. They aren't meant to be used from within the Markdown files in the GitHub repo, but as they are deployed at prometheus.io. For example: https://prometheus.io/docs/instrumenting/clientlibs/

PR #2462, opened 2024-05-08
Added Remote-Write Specification 2.0.

bwplotka:
See proposal https://github.com/prometheus/proposals/pull/35 which explain rationales.

NOTE: The content of this spec is driven by the huge amount of work from @cstyan @alexgreenbank @npazosmendez side [on our public remote write 2.0 working group](https://docs.google.com/document/d/1TYKB_XqVgkEZlxosQ9BM0XR8JFi0pctlNQ-gdu1wuRw/edit).


bwplotka (commented):
Comments after quick sync

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+### Background
+
+The remote write protocol is designed to make it possible to reliably propagate samples in real-time from a sender to a receiver, without loss.
+
+The remote write protocol is designed to make stateless implementations of the server possible; as such there are little-to-no inter-message references.  As such the protocol is not considered "streaming." To achieve a streaming effect multiple messages should be sent over the same connection using e.g. HTTP/1.1 or HTTP/2. "Fancy" technologies such as gRPC were considered, but at the time were not widely adopted, and it was challenging to expose gRPC services to the internet behind load balancers such as an AWS EC2 ELB.
+
+The remote write protocol contains opportunities for batching, e.g. sending multiple samples for different series in a single request. It is not expected that multiple samples for the same series will be commonly sent in the same request, although there is support for this in the protocol.
+
+<!---
+TODO(bwplotka): Challenge last sentence (e.g. OTel coll usage)?
```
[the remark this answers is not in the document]
cstyan: Yeah, I think removing the current note but keeping some kind of note here would be preferable but I'm not sure about how to word it.

```
Note that while the remote write protocol technically allows for an applications to write metrics directly to Prometheus compatible recievers themselves remote write is not intended to be used this way. Prometheus compatible backends and some PromQL functions expect samples to be written to the storage at consistent intervals rather than on every event, like log lines, and as such writing metrics directly from your application could cause incompatibility problems.
```

I'm not sure if doing that :point_up: could actually cause any issues. I suppose it could also be easier to get out of order/duplicate issues sending metrics in this way.

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+-->
+The remote write protocol is not intended for use by applications to push metrics to Prometheus remote-write-compatible Receiver. It is intended that a Prometheus remote-write-compatible sender scrapes instrumented applications or exporters and sends remote write messages to a server.
+
+<!---
+TODO(bwplotka): Add support for 2.0 to those suites 
+-->
+A test suite can be found at https://github.com/prometheus/compliance/tree/main/remote_write_sender.
```
cstyan, line 36: should we just open an issue for this and include the link to the issue with `the 2.0 compatibility test suite is currently in progress here`?

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+
+### Glossary
+
+For the purposes of this document the following definitions MUST be followed:
+
+* a "Sender" is something that sends Prometheus Remote Write data.
+* a "Receiver" is something that receives Prometheus Remote Write data.
+* a "Sample" is a pair of (timestamp, value).
+* a "Histogram" is a pair of (timestamp, histogram value).
```
cstyan, line 45: should we/do we need to specify what a histogram value is here, by linking to the native histogram documentation?

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+
+### Protocol
+
+The Remote Write Protocol MUST consist of RPCs with the request body encoded using a Google Protobuf 3 message. The protobuf encoding MUST use either of the following schemas: 
+
+<!---
+TODO(bwplotka): Do we deprecate it?
+-->
+* [`prometheus.WriteRequest`](./remote_write_spec.md#protocol) introduced in the Remote Write 1.0 specification. As of 2.0 the `prometheus.WriteRequest` message is deprecated.
+* `io.prometheus.write.v2.Request` introduced in this specification and defined [below](#ioprometheuswritev2request-proto-schema). Senders SHOULD use `io.prometheus.write.v2.Request` when possible.
```
cstyan, line 59: :+1: if we're keeping info about the 2.0 proto format then I like this section `protobuf encoding MUST use either of the following schemas:`

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+
+* `Content-Encoding: <compression>`
+
+  Content encoding request header MUST follow [the RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-encoding). Sender MUST use the `snappy` value. More compression algorithms might come in 2.x or beyond.
+
+* `Content-Type: application/x-protobuf` or `Content-Type: application/x-protobuf;proto=<fully qualified name>`
+  
+  <!---
+  TODO(bwplotka): The framing here is a bit inconsistent with deprecation policy. We can either mention sender MUST use either 1.0 or 2.0 message. Or we can say sender MUST use 2.0 message and it MAY/SHOULD use 1.0 message against 1.x receiver. What do we prefer?
+  -->
```
cstyan, line 73: `MUST try the 2.0 format, and MAY fall back to using the 1.0 format if the receiver rejects 2.0 messages because it does not yet have support for the 2.0 format`?

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+* `X-Prometheus-Remote-Write-Version: <remote write specificiation version the sender follows>`
+
+  Sender SHOULD use `X-Prometheus-Remote-Write-Version: 0.1.0` for backward compatibility, when using 1.0 proto message.
+
+Sender MAY allow users to send custom HTTP headers; they MUST NOT allow users to configure them in such a way as to send reserved headers.
+
+<!---
+TODO(bwplotka): Kind of repeated statement, feedback welcome how to structure it.
+-->
+The remote write request in the body of the HTTP POST MUST be compressed with [Google’s Snappy](https://github.com/google/snappy). The block format MUST be used -- the framed format MUST NOT be used. The remote write request MUST be encoded using Google Protobuf 3, and MUST use either of the schemas defined above.
```
cstyan, line 94: we could include ` #### compression` sub section with notes about snappy (and other compressions in the future potentially) and how to use them if they have configuration options? `The block format MUST be used -- the framed format MUST NOT be used.` seems worth keeping at least, which would be relevant if within the Go community more and more people start using https://github.com/klauspost/compress/tree/master/s2#s2-compression instead of the default snappy library.

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+<!---
+TODO(bwplotka): Kind of repeated statement, feedback welcome how to structure it.
+-->
+The remote write request in the body of the HTTP POST MUST be compressed with [Google’s Snappy](https://github.com/google/snappy). The block format MUST be used -- the framed format MUST NOT be used. The remote write request MUST be encoded using Google Protobuf 3, and MUST use either of the schemas defined above. 
+
+#### `io.prometheus.write.v2.Request` Proto Schema
+
+<!---
+TODO(bwplotka): Move link to the one on Prometheus main or even buf. 
+-->
```
cstyan, line 100: :+1:

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+  repeated string symbols = 1;
+  // timeseries represents an array of distinct series with 0 or more samples.
+  repeated TimeSeries timeseries = 2;
+}
+
+// TimeSeries represents a single series.
+message TimeSeries {
+  // labels_refs is a list of label name-value pair references, encoded
+  // as indices to the Request.symbols array. This list's length is always
+  // a multiple of two, and the underlying labels should be sorted.
```
cstyan, line 122: if we're going to keep the note about sorting we should clarify how they should be sorted. At least within prometheus itself we're bound by the sorting that the TSDB does when it creates a series/SeriesRecord in the WAL, which I believe is lexicographically.

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+  //
+  // Note that there might be multiple TimeSeries objects in the same
+  // Requests with the same labels e.g. for different exemplars, metadata
+  // or created timestamp.
+  repeated uint32 labels_refs = 1;
+
+  // Timeseries messages can either specify samples or (native) histogram samples
+  // (histogram field), but not both. For typical clients (real-time metric
+  // streaming), in healthy cases, there will be only one sample or histogram.
```
cstyan, line 131: should we include a note `exporting both a regular metric sample and histogram sample for the exact same metric name and label set should not be possible, and if it occurs within a client library or remote write implementation it should be considered an error`?

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+
+  // exemplars represents an optional set of exemplars attached to this series' samples.
+  repeated Exemplar exemplars = 4;
+
+  // metadata represents the metadata associated with the given series' samples.
+  Metadata metadata = 5;
+
+  // created_timestamp represents an optional created timestamp associated with
+  // this series' samples in ms format, typically for counter or histogram type
+  // metrics. Note that some servers might require this and in return fail to
```
cstyan, line 145: servers -> receivers

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+  // help_ref is a reference to the Request.symbols array representing help
+  // text for the metric.
+  uint32 help_ref = 3;
+  // unit_ref is a reference to the Request.symbols array representing unit
+  // for the metric.
+  uint32 unit_ref = 4;
+}
+
+// A native histogram, also known as a sparse histogram.
+message Histogram { ... }
```
cstyan, line 200: should we include a link to the proto file or buf here as well?

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+All timestamps MUST be int64 counted as milliseconds since the Unix epoch. Sample's values MUST be float64.
+
+For every `TimeSeries` message:
+
+* Label references MUST be provided.
+* At least one element in Samples or in Histograms MUST be provided. For series which (rarely) would mix float and histogram samples, a separate `TimeSeries` message MUST be used.
+<!---
+TODO(bwplotka): We have some inconsistency here. In gdoc we talked this should be SHOULD. But proto has it as MUST (?). What do we want at the end? What's wrong with MUST here?
+-->
+* Metadata MUST be provided.
```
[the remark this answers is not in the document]
cstyan: I think should, and leave it up to sender implementations whether they want to allow users to configure this as required on their end. I can see there still being some systems that generate prometheus compatible metrics but not prometheus style metadata.

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+
+Symbols table containing deduplicated strings used in series and exemplar labels, metadata strings MUST be provided. References MUST point to the existing index in the Symbols string array.
+
+#### Series Labels
+
+The complete set of labels MUST be sent with each Sample or Histogram sample. Additionally, the label set associated with samples:
+
+- SHOULD contain a `__name__` label.
+- MUST NOT contain repeated label names.
+- MUST have label names sorted in lexicographical order.
```
cstyan, line 234: :+1: earlier where I made a comment about the sorting method we could just say `see #### series labels` with an internal markdown link?

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+- MUST NOT contain any empty label names or values.
+
+Sender MUST only send valid metric names, label names, and label values:
+
+<!---
+TODO(bwplotka): Add mention for UTF-8 support going forward, see https://docs.google.com/document/d/1PljkX3YLLT-4f7MqrLt7XCVPG3IsjRREzYrUzBxCPV0/edit?disco=AAAA4gSqQ7g
+-->
+- Metric names MUST adhere to the regex `[a-zA-Z_:]([a-zA-Z0-9_:])*`.
+- Label names MUST adhere to the regex `[a-zA-Z_]([a-zA-Z0-9_])*`.
+- Label values MAY be any sequence of UTF-8 characters .
```
cstyan, line 244: do we want to keep this section, or instead just link to some spec for prometheus metrics themselves?

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+Receiver MAY impose limits on the number and length of labels, but this will be receiver-specific and is out of scope for this document.
+
+Label names beginning with "__" are RESERVED for system usage and SHOULD NOT be used, see [Prometheus Data Model](https://prometheus.io/docs/concepts/data_model/).
+
+#### Samples
+
+<!---
+TODO(bwplotka): Does this apply to histograms?
+-->
+Sender MUST send samples for any given TimeSeries in timestamp order. Sender MAY send multiple requests for different series in parallel.
```
cstyan, line 255: Technically prometheus supports out of order samples now, and I think by proxy remote write ingestion via the write handler does as well? Should this be `SHOULD`? or `MUST, unless via some external system an out of order sample is generated such that a sample has a timestamp older than a previously sent sample`?

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+#### Metadata
+
+Metadata SHOULD follow the official guidelines for [TYPE](https://prometheus.io/docs/instrumenting/writing_exporters/#types) and [HELP](https://prometheus.io/docs/instrumenting/writing_exporters/#help-strings).
+
+#### Exemplars
+
+<!---
+TODO(bwplotka): Anything to say here?
+-->
+TBD
```
cstyan, line 277: exemplars contain labels that `MUST` also be interned in the symbols table

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+
+### Responses
+
+Receiver ingesting all samples successfully MUST return HTTP 200 status code. In such a successful case, the response body from the Receiver SHOULD be empty; Sender MUST ignore the response body. The response body is RESERVED for future use.
+
+The following subsections specify Sender and Receiver semantics around write errors.
+
+#### Partial Write
+
+Sender SHOULD use Prometheus Remote Write to request write of multiple samples, across multiple series. As a result, Receiver MAY ingest valid samples within a write request that contains invalid or otherwise unwritten samples, which represents a partial write case.
```
cstyan, line 294: > Sender SHOULD use Prometheus Remote Write to send samples for multiple series at a time. As a result, Receiver MAY ingest valid samples within a write request that contains invalid or otherwise unwritten samples, which represents a partial write case.

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+
+#### Unsupported Request Content
+
+Receiver MAY NOT support certain content types or encodings defined in [the Protocol section](#protocol). Receiver MUST return [415 HTTP Unsupported Media Type](https://www.rfc-editor.org/rfc/rfc9110.html#name-415-unsupported-media-type) status code if they don't support a given content type or encoding provided by the Sender.
+
+Sender SHOULD expect [400 HTTP Bad Request](https://www.rfc-editor.org/rfc/rfc9110.html#name-400-bad-request) for the above reasons from the 1.x Receiver, for backward compatibility.
+
+<!---
+TODO(bwplotka): Note sure if worth mentioning given we decided to not include auto negotiation logic in 2.0. I think I would delete it.
+-->
```
cstyan, line 306: :+1:

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+Sender MAY retry write requests on 415 HTTP status code, with different content type and compression settings.
+
+#### Invalid Samples
+
+<!---
+TODO(bwplotka): This wording assumes metadata is optional, which I think it should NOT be.
+-->
```
cstyan, line 313: unless we want to make extra rw2.0 implementation work for systems that don't already generate prometheus style metadata, I think we have to keep it optional 

(to be clear, I am not opposed to making it mandatory, but I think we should discuss further)

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+Receiver MAY NOT support certain metric types or samples (e.g. Receiver might reject sample without metadata or without created timestamp, while another Receiver might accept such sample.). It’s up to the Receiver what sample is invalid. Receiver MUST return a [400 HTTP Bad Request](https://www.rfc-editor.org/rfc/rfc9110.html#name-400-bad-request) status code for write requests that contain any invalid samples, unless the [partial retryable write](#retries-on-partial-writes) occurs.
```
cstyan, line 314: We should clarify, do we mean any invalid samples or all? I don't know for sure if any exists at the moment, but unless the spec clarifies that partial ingestion must be supported then it's possible that some systems are simply failing entire write requests if one sample is invalid?

In either case here, `the response SHOULD contain a useful human readable error about why samples were rejected`

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+
+Sender MUST NOT retry on 4xx HTTP (other than 429 and 415) status codes, which MUST be used by Receiver to indicate that the write will never be able to succeed and should not be retried.
+
+### Retries & Backoff
+
+Receiver MAY return a [429 HTTP Too Many Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) status code to indicate the overloaded server situation. Receiver MAY return [the Retry-After](https://www.rfc-editor.org/rfc/rfc9110.html#name-retry-after) header to indicate the time for the next write attempt. Receiver MAY return a 5xx HTTP status code to represent internal server errors.
+
+Sender MAY retry on 429 HTTP status code. Sender MUST retry write requests on 5xx HTTP. Sender MUST use a backoff algorithm to prevent overwhelming the server. Sender MAY handle [the Retry-After response header](https://www.rfc-editor.org/rfc/rfc9110.html#name-retry-after) to estimate the next retry time.
```
cstyan, line 322: > Sender MAY use the [Retry-After response header](https://www.rfc-editor.org/rfc/rfc9110.html#name-retry-after) to suggest/enforce the next retry time.

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+
+Receiver ingesting all samples successfully MUST return HTTP 200 status code. In such a successful case, the response body from the Receiver SHOULD be empty; Sender MUST ignore the response body. The response body is RESERVED for future use.
+
+The following subsections specify Sender and Receiver semantics around write errors.
+
+#### Partial Write
+
+Sender SHOULD use Prometheus Remote Write to request write of multiple samples, across multiple series. As a result, Receiver MAY ingest valid samples within a write request that contains invalid or otherwise unwritten samples, which represents a partial write case.
+
+In a partial write case, Receiver MUST NOT return HTTP 200 status code. Receiver MUST provide a human-readable error message in the response body. Sender MUST NOT try and interpret the error message, and SHOULD log it as is.
```
rfratto, line 296: > Sender MUST NOT try and interpret the error message, and SHOULD log it as is.

Can we standardize a way for senders to interpret the error message? 

Today while debugging failed sample errors with Remote-Write 1.0, I was wishing I could visualize (from the sender's perspective) the types of errors being returned from the server and the number of samples discarded per error type the server generated. 

I don't think this information has to be in the response body, but it could be a header field or another place where machines can read it.

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,394 @@
...
+
+# Prometheus Remote-Write Specification
+
+* Version: 2.0
+* Status: Proposed
+* Date: May 2024
+
+The remote write specification, in general, is intended to document the standard for how Prometheus and Prometheus remote-write-compatible agents send data to a Prometheus or Prometheus remote-write compatible receivers.
+
+This document is intended to define a second version of the [Prometheus Remote Write](./remote_write_spec.md) API with minor changes to protocol and semantics. This second version also adds a new wire format with new features enabling more use cases and wider adoption on top of performance and cost savings. Finally, this spec outlines how to implement backward compatible senders and receivers (even under a single endpoint) using existing basic content negotiation request headers. More advanced, automatic content negotiation mechanisms might come in future versions, if needed. For the rationales behind the 2.0 specification, see [the formal proposal](https://github.com/prometheus/proposals/pull/35).
```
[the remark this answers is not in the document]
bwplotka: Good point. I think yes, because in this version we also publish a new wire format. If another version will NOT publish anything substantial, it could be skipped. Equally if another version adds 2, both could be mentioned in this doc.

Counter argument is that proto is spec, but this needs more human explanation as we do with old proto, so perhaps here makes sense?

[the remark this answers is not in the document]
bwplotka: Assuming we agree we put proto explanation here in this spec.

content/docs/concepts/remote_write_spec_2_0.md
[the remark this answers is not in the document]
bwplotka: Yea, I think we are then not really sure what we try to share here. The fact that scraping is be preferred for Prometheus (which I agree) should be likely advertised in Prometheus, not in spec 🤔 

The only reason, I think, to have such statement, is to give intend that we won't add more spec features for pure push metric from monitored application use case,  but I can't imagine any example feature ATM. Should we remove this piece and add it back once we have a clear example of feature this spec will not implement?

> I suppose it could also be easier to get out of order/duplicate issues sending metrics in this way.

Hm, we don't specify anything about out of order - it's up to receiver. So perhaps not specific enough reason to add such intention paragraph?

Proposing removing for now 🤗

content/docs/concepts/remote_write_spec_2_0.md
bwplotka: 👍🏽 https://github.com/prometheus/compliance/issues/101

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,363 @@
...
+#### Series Labels
+
+The complete set of labels MUST be sent with each Sample or Histogram sample. Additionally, the label set associated with samples:
+
+* SHOULD contain a `__name__` label.
+* MUST NOT contain repeated label names.
+* MUST have label names sorted in lexicographical order.
+* MUST NOT contain any empty label names or values.
+
+Metric names, label names, and label values MAY be any sequence of UTF-8 characters. Receiver MAY reject some series with metric names and label names characters that does not follow [previous patterns](https://prometheus.io/docs/concepts/remote_write_spec/#:~:text=Metric%20names%20MUST,UTF%2D8%20characters%20) given [the UTF-8 support is still in progress](https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md).
```
bwplotka, line 224:
```suggestion
Metric names, label names, and label values MAY be any sequence of UTF-8 characters. Receiver MAY reject series with metric names and label names characters that does not follow [previous patterns](https://prometheus.io/docs/concepts/remote_write_spec/#:~:text=Metric%20names%20MUST,UTF%2D8%20characters%20) given [the UTF-8 support is still in progress](https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md).
```

Notes:
* Why not support UTF-8 as a MUST? Prometheus 3.0 will scrape it. Why it cannot send it through PRW 2.0? Proposal -> remove "MAY".
   * Open question: What's the progress? There is a risk that PRW 2.0 will be not adoptable by Prometheus itself.
   * Gather more feedback

bwplotka: Started discussion: https://cloud-native.slack.com/archives/C01AUBA4PFE/p1716394693414599

content/docs/concepts/remote_write_spec_2_0.md
bwplotka: Thanks! 

We discussed this with @cstyan a lot and decided to skip the standirazation of error message for now. As mentioned on Slack our recommendation would be organically grow this feature - so propose it to Prometheus (both sender and receiver) and see how it will be actually used/implemented. Then we can judge it and decide if we need this in 2.x or 3.0. This feature feels useful, but we might need data if "programmatic" use will be there. There is some more adoption work and rabbit holes (what error types etc?) to discuss if we ever propose it, so more data first would be nice. So far it feels error log should suffice. 

We extended the sentence to:

```
... Receiver MUST provide a human-readable error message in the response body. The Receiver's error SHOULD contain information about the amount of the samples being rejected and for what reasons.
```

Hope that's ok! And thanks for amazing idea!

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,367 @@
...
+
+## Introduction
+
+### Background
+
+The remote write protocol is designed to make it possible to reliably propagate samples in real-time from a sender to a receiver, without loss.
+
+The remote write protocol is designed to make stateless implementations of the server possible; as such there are little-to-no inter-message references.  As such the protocol is not considered "streaming." To achieve a streaming effect multiple messages should be sent over the same connection using e.g. HTTP/1.1 or HTTP/2. "Fancy" technologies such as gRPC were considered, but at the time were not widely adopted, and it was challenging to expose gRPC services to the internet behind load balancers such as an AWS EC2 ELB.
+
+The remote write protocol contains opportunities for batching, e.g. sending multiple samples for different series in a single request. It is not expected that multiple samples for the same series will be commonly sent in the same request, although there is support for this in the protocol.
```
cstyan, line 26: > although there is support for this in the protocol

should this refer to the protobuf format rather than the protocol? that's technically what allows for this though I'm not sure specifying either way is important

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,367 @@
...
+    * `Content-Type: application/x-protobuf;proto=prometheus.WriteRequest`
+  For the message introduced in PRW 2.0, identified by `io.prometheus.write.v2.Request`:
+    * `Content-Type: application/x-protobuf;proto=io.prometheus.write.v2.Request`
+  
+  When talking to 1.x Receiver, the Sender SHOULD use `Content-Type: application/x-protobuf` for backward compatibility. Otherwise, Sender SHOULD use `Content-Type: application/x-protobuf;proto=io.prometheus.write.v2.Request`. More proto messages might come in 2.x or beyond.
+
+* `User-Agent: <name & version of the sender>`
+* `X-Prometheus-Remote-Write-Version: <remote write spec major and minor version>`
+
+  When talking to 1.x Receiver, the Sender MUST use `X-Prometheus-Remote-Write-Version: 0.1.0` for backward compatibility. Otherwise, Sender SHOULD use the newest remote write version it is compatible with e.g. `X-Prometheus-Remote-Write-Version: 2.0.0`.
```
cstyan, line 77: did we not allow for `X-Prometheus-Remote-Write-Version: 1.0` in the 1.0 spec as well?

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,367 @@
...
+The 2.x protocol is breaking compatibility with 1.x by introducing a new `io.prometheus.write.v2.Request` content type (wire format) and deprecating the `prometheus.WriteRequest`.
+
+2.x Senders MAY support 1.x Receivers by allowing users to configure what content type sender should use. 2.x Senders also MAY automatically fall back to different content types, if the Receiver returns 415 HTTP status code.
+
+#### `io.prometheus.write.v2.Request` Proto Schema
+
+<!---
+TODO(bwplotka): Move link to the one on Prometheus main or even buf. 
+-->
+The source of truth is [here](https://github.com/prometheus/prometheus/blob/remote-write-2.0/prompb/io/prometheus/write/v2/types.proto#L32). The `gogo` dependency and options CAN be ignored. They are not part of the specification as they don't impact the serialized format.
```
cstyan, line 141: you could mention the ongoing gsoc project to replace gogoproto if you want?

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,367 @@
...
+
+The simplified version of the new `io.prometheus.write.v2.Request` is presented below.
+
+```
+// Request represents a request to write the given timeseries to a remote destination.
+message Request {
+  // symbols contains a de-duplicated array of string elements used for various
+  // items in a Request message, like labels and metadata items. For the sender convenience
```
cstyan, line 149: nit: sender -> senders

content/docs/concepts/remote_write_spec_2_0.md
cstyan: nevermind, confirmed myself that the 1.0 spec says `0.1.0` must be used

cstyan (commented):
few minor nits and then I think we're good to go

bwplotka:
This spec is ready for wider, final review 🤗 🎉

bwplotka:
The proposal for moving forward is:

* Merge NOW the actual [spec](https://github.com/prometheus/docs/pull/2462) as "2.0-rc.0" into our docs and announce it as experimental. 
* [The Proposal PR](https://github.com/prometheus/proposals/pull/35) will be kept open until protocol is stable.
* Continue to gather organic feedback (on proposals and issues) and data, based on early adoption. Add 2.0-rc.X for potential changes (ideally small, major ones ONLY if absolutely necessary) and eventually e.g. if we feel so, around PromCon time do the stable 2.0.

content/docs/concepts/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,422 @@
...
+  // symbols contains a de-duplicated array of string elements used for various
+  // items in a Request message, like labels and metadata items. For the sender's convenience
+  // around empty values for optional fields like unit_ref, symbols array MUST start with
+  // empty string.
+  //
+  // To decode each of the symbolized strings, referenced, by "ref(s)" suffix, you
+  // need to lookup the actual string by index from symbols array. The order of
+  // strings is up to the sender. The receiver should not assume any particular encoding.
+  repeated string symbols = 1;
+  repeated string symbols = 1;
```
tomwilkie, line 181: Is this a typo? Is this field supposed to appear twice?  Thats illegal, surely?

bwplotka: copy-pasta, referenced proto does not have it, thanks for finding 🙈

bwplotka:
One idea around custom native histograms -> we could add an information if it was auto-translated or not 🤔 cc @beorn7 @krajorama

content/docs/specs/remote_write_spec.md
```diff
@@ -19,8 +20,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 The remote write protocol is designed to make it possible to reliably propagate samples in real-time from a sender to a receiver, without loss.
 
-The remote write protocol is designed to make stateless implementations of the server possible; as such there are little-to-no inter-message references.  As such the protocol is not considered "streaming." To achieve a streaming effect multiple messages should be sent over the same connection using e.g. HTTP/1.1 or HTTP/2.  "Fancy" technologies such as gRPC were considered, but at the time were not widely adopted, and it was challenging to expose gRPC services to the internet behind load balancers such as an AWS EC2 ELB.
-
+The Remote-Write protocol is designed to be stateless; there is strictly no inter-message communication. As such the protocol is not considered "streaming. To achieve a streaming effect multiple messages should be sent over the same connection using e.g. HTTP/1.1 or HTTP/2.  "Fancy" technologies such as gRPC were considered, but at the time were not widely adopted, and it was challenging to expose gRPC services to the internet behind load balancers such as an AWS EC2 ELB.
```
ArthurSens, line 23:
```suggestion
The Remote-Write protocol is designed to be stateless; there is strictly no inter-message communication. As such the protocol is not considered "streaming". To achieve a streaming effect multiple messages should be sent over the same connection using e.g. HTTP/1.1 or HTTP/2. "Fancy" technologies such as gRPC were considered, but at the time were not widely adopted, and it was challenging to expose gRPC services to the internet behind load balancers such as an AWS EC2 ELB.
```

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+## Introduction
+
+### Background
+
+The Remote-Write protocol is designed to make it possible to reliably propagate samples in real-time from a sender to a receiver, without loss.
+
+<!---
+For the detailed rationales behind each 2.0 Remote-Write decision, see: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-proposal/proposals/2024-04-09_remote-write-20.md
+-->
+The Remote-Write protocol is designed to be stateless; there is strictly no inter-message communication. As such the protocol is not considered "streaming." To achieve a streaming effect multiple messages should be sent over the same connection using e.g. HTTP/1.1 or HTTP/2. "Fancy" technologies such as gRPC were considered, but at the time were not widely adopted, and it was challenging to expose gRPC services to the internet behind load balancers such as an AWS EC2 ELB.
```
ArthurSens, line 29:
```suggestion
The Remote-Write protocol is designed to be stateless; there is strictly no inter-message communication. As such the protocol is not considered "streaming". To achieve a streaming effect multiple messages should be sent over the same connection using e.g. HTTP/1.1 or HTTP/2. "Fancy" technologies such as gRPC were considered, but at the time were not widely adopted, and it was challenging to expose gRPC services to the internet behind load balancers such as an AWS EC2 ELB.
```

ArthurSens: Is the claim about gRPC not being widely adopted and hard to expose still true to this day?

Legit just curious, I don't mean to doubt what is written

ArthurSens (approved):

beorn7:
> One idea around custom native histograms -> we could add an information if it was auto-translated or not 🤔 cc @beorn7 @krajorama

If anyone needs this information, sure. But who actually needs this information? I would expect _all_ classic histograms to be converted to NHCBs in the not too far future.

bwplotka:
> If anyone needs this information, sure. But who actually needs this information? I would expect all classic histograms to be converted to NHCBs in the not too far future.

Yea, this depends on migration strategies. Let's talk on separate thread maybe @beorn7, will ping on Slack - we can add this info later on too.

content/docs/specs/remote_write_spec_2_0.md
bwplotka: Fair question!

I think it's adopted now, but don't challenge it in this wording. We just said it was not adopted back then, right?

If you ask "if it's worth to re-eval now given adoption", the answer is in the FAQ below -- 

> **Why did you not use gRPC?**
> Because 1.0 protocol does not use gRPC, breaking it would increase friction in the adoption. See 1.0 [reason](./remote_write_spec.md#faq).

Does it make sense? 🤗

ArthurSens: Yes, a lot! Thanks

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+
+* **Alternative wire formats**. The OpenTelemetry community has shown the validity of Apache Arrow (and potentially other columnar formats) for over-wire data transfer with their OTLP protocol. We would like to do experiments to confirm the compatibility of a similar format with Prometheus’ data model and include benchmarks of any resource usage changes. We would potentially maintain both a protobuf and columnar format long term for compatibility reasons and use our content negotiation to add different proto messages for this purpose.
+
+* **Global symbols**. Pre-defined string dictionary for interning The protocol could pre-define a static dictionary of ref->symbol that includes strings that are considered common, e.g. “namespace”, “le”, “job”, “seconds”, “bytes”, etc. Sender and refer to these without the need to include them in the request’s symbols table. This dictionary could incrementally grow with a minor version releases of this protocol.
```
npazosmendez, line 422:
```suggestion
* **Global symbols**. Pre-defined string dictionary for interning The protocol could pre-define a static dictionary of ref->symbol that includes strings that are considered common, e.g. “namespace”, “le”, “job”, “seconds”, “bytes”, etc. Sender could refer to these without the need to include them in the request’s symbols table. This dictionary could incrementally grow with a minor version releases of this protocol.
```

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+Because 1.0 protocol does not use gRPC, breaking it would increase friction in the adoption. See 1.0 [reason](./remote_write_spec.md#faq).
+
+**Why not stream protobuf messages?**
+If you use persistent HTTP/1.1 connections, they are pretty close to streaming. Of course headers have to be re-sent, but yes that is less expensive than a new TCP set up.
+
+**Why do we send samples in order?**
+The in-order constraint comes from the encoding we use for time series data in Prometheus, the implementation of which is append-only. It is possible to remove this constraint, for instance by buffering samples and reordering them before encoding.
+
+**How can we parallelise requests with the in-order constraint?**
+Samples must be in-order _for a given series_. Remote-Write requests can be sent in parallel as long as they are for different series. In Prometheus, we shard the samples by their labels into separate queues, and then writes happen sequentially in each queue. This guarantees samples for the same series are delivered in order, but samples for different series are sent in parallel - and potentially "out of order" between different series.
```
npazosmendez, line 438: Is this up to date given that Prometheus supports out of order now? I know we still want in-order within a single request, I'm not saying we change the spec, but maybe these answers in the FAQ should be updated.

Is it for performance reasons maybe?

npazosmendez (approved):
LGTM!

A few non-blocking nits and questions, nothing about the spec per se.

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+
+<!---
+For the detailed rationales behind each 2.0 Remote-Write decision, see: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-proposal/proposals/2024-04-09_remote-write-20.md
+-->
+The Remote-Write protocol is designed to be stateless; there is strictly no inter-message communication. As such the protocol is not considered "streaming." To achieve a streaming effect multiple messages should be sent over the same connection using e.g. HTTP/1.1 or HTTP/2. "Fancy" technologies such as gRPC were considered, but at the time were not widely adopted, and it was challenging to expose gRPC services to the internet behind load balancers such as an AWS EC2 ELB.
+
+The Remote-Write protocol contains opportunities for batching, e.g. sending multiple samples for different series in a single request. It is not expected that multiple samples for the same series will be commonly sent in the same request, although there is support for this in the Proto Message.
+
+A test suite can be found at https://github.com/prometheus/compliance/tree/main/remote_write_sender. The test's 2.0 compatibility [is in progress](https://github.com/prometheus/compliance/issues/101).
```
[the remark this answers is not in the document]
bwplotka: It a shorthand of "The 2.0 compatibility of compliance tests will be done soon", will rephrase!

content/docs/specs/remote_write_spec_2_0.md
bwplotka: Feel free to propose changes. I personally would avoid using OOO in Prometheus because of efficiency reasons.. and my lack of experience of using it (: ..but it's better and better and viable for "rare" cases.

Still I think the intention of this question is to "how do I send LONG back log of samples faster". Current wording (probably too much Prometheus details here which can get obsolete easily) suggests you NEVER send not-in-order samples (not allowed), but also NEVER send not-in-order samples in different TimeSeries for same series (spec does NOT prohibits this, but also don't recommends). You are right that technically the latter is an option for some systems e.g Prometheus with OOO enabled, if not common I guess. We could mention that "ensure Receiver supports OOO as it's not guaranteed/nor blocked by spec and then you can do it e.g. in Prometheus do XYZ"

Feel free to propose changes here, we could share something like above. 🤗

npazosmendez: How about... ?
```suggestion
**Why do we send samples in order?**
The in-order constraint comes from the encoding we use for time series data in Prometheus, the implementation of which is optimized for append-only workloads. Receiver may support out-of-order ingestion, though it is not guaranteed nor blocked by the specification.

**How can we parallelise requests with the in-order constraint?**
Samples must be in-order _for a given series_. Even if Receiver does not support out-of-order ingestion, Remote-Write requests can be sent in parallel as long as they are for different series. In Prometheus, we shard the samples by their labels into separate queues, and then writes happen sequentially in each queue. This guarantees samples for the same series are delivered in order, but samples for different series are sent in parallel - and potentially "out of order" between different series.
```

LMK what you think!

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+
+### Retries & Backoff
+
+Receiver MAY return a [429 HTTP Too Many Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) status code to indicate the overloaded server situation. Receiver MAY return [the Retry-After](https://www.rfc-editor.org/rfc/rfc9110.html#name-retry-after) header to indicate the time for the next write attempt. Receiver MAY return a 5xx HTTP status code to represent internal server errors, that should be retried.
+
+Sender MAY retry on a 429 HTTP status code. Sender MUST retry write requests on 5xx HTTP. Sender MUST use a backoff algorithm to prevent overwhelming the server. Sender MAY handle [the Retry-After response header](https://www.rfc-editor.org/rfc/rfc9110.html#name-retry-after) to estimate the next retry time.
+
+The difference between 429 vs 5xx handling is due to a potential situation for Sender “falling behind” if the Receiver cannot keep up. As a result, the ability to NOT retry on 429 allows progress is made when there are Sender side errors (e.g. too much traffic), while the data is not lost when there are Receiver side errors.
```
cstyan, line 155:
```suggestion
The difference between 429 vs 5xx handling is due to the potential situation of the Sender “falling behind” when the Receiver cannot keep up with the request volume, or the Receiver choosing to rate limit the Sender to protect it's own availability. As a result, the Sender has the option to NOT retry on 429, which allows progress to be made when there are Sender side errors (e.g. too much traffic), while the data is not lost when there are Receiver side errors (5xx).
```

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+
+<!---
+Rationales: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-proposal/proposals/2024-04-09_remote-write-20.md#partial-writes#samples-vs-native-histogram-samples
+-->
+* At least one element in `samples` or in `histograms` MUST be provided. A `TimeSeries` MUST NOT include both `samples` and `histograms`. For series which (rarely) would mix float and histogram samples, a separate `TimeSeries` message MUST be used.
+
+<!---
+Rationales: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-proposal/proposals/2024-04-09_remote-write-20.md#always-on-metadata
+-->
+* Metadata fields SHOULD be provided. Receiver MAY reject series with unspecified type.
```
cstyan, line 323:
```suggestion
* Metadata fields SHOULD be provided. Receiver MAY reject series with unspecified Type.
```

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+* Exemplars SHOULD be provided if they exist for a series.
+* Created timestamp SHOULD be provided for metrics that follow counter semantics (e.g. counters and histograms). Receiver MAY reject those series without the created timestamp being set.
```
cstyan, line 325:
```suggestion
* Created Timestamp SHOULD be provided for metrics that follow counter semantics (e.g. counters and histograms). Receiver MAY reject those series without the Created Timestamp being set.
```

content/docs/specs/remote_write_spec_2_0.md
cstyan: I think capitalizing the metadata field names here makes it clearer

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+The following subsections define some schema elements in detail.
+
+#### Symbols
+
+<!---
+Rationales: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-proposal/proposals/2024-04-09_remote-write-20.md#partial-writes#string-interning
+-->
+The `io.prometheus.write.v2.Request` Proto Message is designed to [intern all strings](https://en.wikipedia.org/wiki/String_interning) for the proven additional compression and memory efficiency gains on top of the standard compressions.
+
+Symbols table MUST be provided and it MUST contain deduplicated strings used in series, exemplar labels and metadata strings. The first element of the symbols table MUST be an empty string. References MUST point to the existing index in the Symbols string array.
```
cstyan, line 336:
```suggestion
Symbols table MUST be provided and it MUST contain deduplicated strings used in series, exemplar labels, and metadata strings. The first element of the symbols table MUST be an empty string, which is used to represent empty or unspecified values such as when Unit or Help metadata are not provided. References MUST point to the existing index in the Symbols string array.
```

cstyan (commented):
I don't think my comment about `Proto Message` -> `Protobuf Message` is really necessary anymore but I will let you decide @bwplotka

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+
+Sender MUST NOT retry on a 4xx HTTP status codes (other than [429](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429)), which MUST be used by Receiver to indicate that the write operation will never be able to succeed and should not be retried. Sender MAY retry on the 415 HTTP status code with a different content type or encoding to see if Receiver supports it.
+
+### Retries & Backoff
+
+Receiver MAY return a [429 HTTP Too Many Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) status code to indicate the overloaded server situation. Receiver MAY return [the Retry-After](https://www.rfc-editor.org/rfc/rfc9110.html#name-retry-after) header to indicate the time for the next write attempt. Receiver MAY return a 5xx HTTP status code to represent internal server errors, that should be retried.
```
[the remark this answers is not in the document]
bwplotka: Nice! I removed the suffix "that wants retried." etc as it's clearly explained below. Let's be concise and avoid dups.

content/docs/specs/remote_write_spec_2_0.md
bwplotka: Thanks! I added something as you proposed, with minor changes for clarity PTAL:

```
**Why do we send samples in order?**
The in-order constraint comes from the encoding we use for time series data in Prometheus, the implementation of which is optimized for append-only workloads. However, this requirement is also shared across many other databases and vendors in the ecosystem. In fact, [Prometheus with OOO feature enabled](https://youtu.be/qYsycK3nTSQ?t=1321), allows out-of-order writes, but with the performance penalty, thus reserved for rare events. To sum up, Receiver may support out-of-order ingestion, though it is not permitted by the specification. In the future e.g. 2.x spec versions, we could extend content type to negotiate the out-of-order writes, if needed.

**How can we parallelise requests with the in-order constraint?**
Samples must be in-order _for a given series_. However, even if Receiver does not support out-of-order ingestion, the Remote-Write requests can be sent in parallel as long as they are for different series. Prometheus shards the samples by their labels into separate queues, and then writes happen sequentially in each queue. This guarantees samples for the same series are delivered in order, but samples for different series are sent in parallel - and potentially "out of order" between different series.
```

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+
+# Prometheus Remote-Write Specification
+
+* Version: 2.0-rc.0
+* Status: **Experimental**
+* Date: May 2024
+
+The Remote-Write specification, in general, is intended to document the standard for how Prometheus and Prometheus Remote-Write compatible senders send data to Prometheus or Prometheus Remote-Write compatible receivers.
+
+This document is intended to define a second version of the [Prometheus Remote-Write](./remote_write_spec.md) API with minor changes to protocol and semantics. This second version adds a new Proto Message with new features enabling more use cases and wider adoption on top of performance and cost savings. The second version also deprecates the previous Proto Message from a [1.0 Remote-Write specification](./remote_write_spec.md#protocol). Finally, this spec outlines how to implement backwards-compatible senders and receivers (even under a single endpoint) using existing basic content negotiation request headers. More advanced, automatic content negotiation mechanisms might come in a future minor version, if needed. For the rationales behind the 2.0 specification, see [the formal proposal](https://github.com/prometheus/proposals/pull/35).
```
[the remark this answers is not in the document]
bwplotka: Hm, so we define this in glossary below, plus both Proto and Protobuf are both shortcuts of "Protocol buffers" 🙈 


I agree https://github.com/protocolbuffers/protobuf#:~:text=Protocol%20Buffers%20(a.k.a.%2C%20protobuf)%20are `protobuf` might be more known than "proto"? But then we could move all "Proto Message" to "Protobuf Message"? Happy with that, but I would move all Proto -> Protobuf or none (:

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+
+A test suite can be found at https://github.com/prometheus/compliance/tree/main/remote_write_sender. The test's 2.0 compatibility [is in progress](https://github.com/prometheus/compliance/issues/101).
+
+### Glossary
+
+In this document, the following definitions are followed:
+
+* a `Remote-Write` is the name of this Prometheus protocol.
+* a `Protocol` is a communication specification that enables the client and server to transfer metrics.
+* a `Proto Message` refers to the [content type](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-type) definition of the data structure for this Protocol. Since the specification uses [Google Protocol Buffers ("protobuf")](https://protobuf.dev/) exclusively, the schema is defined in a ["proto" file](https://protobuf.dev/programming-guides/proto3/) and represented by a single Protobuf ["message"](https://protobuf.dev/programming-guides/proto3/#simple).
```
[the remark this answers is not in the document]
bwplotka: I wouldn't be too strict here, it's just tl'dr part,  it's fine to use some shortcuts. 

But we could change all to `Protobuf Message` if you want.

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+* a `Wire Format` is the format of the data as it travels on the wire (i.e. in a network). In the case of Remote-Write, this is always the compressed binary protobuf format.
+* a `Sender` is something that sends Remote-Write data.
+* a `Receiver` is something that receives Remote-Write data.
+* a `Sample` is a pair of (timestamp, value).
+* a `Histogram` is a pair of (timestamp, [histogram value](https://github.com/prometheus/docs/blob/b9657b5f5b264b81add39f6db2f1df36faf03efe/content/docs/concepts/native_histograms.md)).
+* a `Label` is a pair of (key, value).
```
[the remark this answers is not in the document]
bwplotka: Not here, it's explained well in Series Label section, also consistent with other items?

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+
+<!---
+Rationales: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-proposal/proposals/2024-04-09_remote-write-20.md#a-new-protobuf-message-identified-by-fully-qualified-name-old-one-deprecated
+-->
+The protobuf serialization MUST use either of the following Proto Messages:
+
+* The `prometheus.WriteRequest` introduced in [the Remote-Write 1.0 specification](./remote_write_spec.md#protocol). As of 2.0, this message is deprecated. It SHOULD be used only for compatibility reasons. Sender and Receiver MAY NOT support the `prometheus.WriteRequest`.
+* The `io.prometheus.write.v2.Request` introduced in this specification and defined [below](#proto-message). Sender and Receiver SHOULD use this message when possible. Sender and Receiver MUST support the `io.prometheus.write.v2.Request`.
+
+The Proto Message MUST use binary Wire Format. Then, MUST be compressed with [Google’s Snappy](https://github.com/google/snappy). The block format MUST be used -- the framed format MUST NOT be used.
```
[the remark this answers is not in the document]
bwplotka: added link too -- should be this https://github.com/google/snappy/blob/2c94e11145f0b7b184b831577c93e5a41c4c0346/format_description.txt

content/docs/specs/remote_write_spec_2_0.md
bwplotka: I will merge but will remove `the` before Receiver and Sender since those are now names (simplicity and consistency). If we add `the` here we have to add everywhere.

content/docs/specs/remote_write_spec_2_0.md
bwplotka: I think we started to put those in code quotes https://github.com/prometheus/docs/pull/2462#discussion_r1636508998 will change

content/docs/specs/remote_write_spec_2_0.md
bwplotka: Same as https://github.com/prometheus/docs/pull/2462#discussion_r1636508998 will change to code quotes

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -0,0 +1,443 @@
...
+All timestamps MUST be int64 counted as milliseconds since the Unix epoch. Sample's values MUST be float64.
+
+For every `TimeSeries` message:
+
+* Label references MUST be provided.
+
+<!---
+Rationales: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-proposal/proposals/2024-04-09_remote-write-20.md#partial-writes#samples-vs-native-histogram-samples
+-->
+* At least one element in Samples or in Histograms MUST be provided. For series which (rarely) would mix float and histogram samples, a separate `TimeSeries` message MUST be used.
```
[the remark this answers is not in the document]
bwplotka: Actually we might want to revert this and use capital case instead, probably more consistent with the rest e.g. https://github.com/prometheus/docs/pull/2462#discussion_r1643209084

content/docs/specs/remote_write_spec_2_0.md
bwplotka: Actually let's do this https://github.com/prometheus/docs/pull/2462#discussion_r1645729977

content/docs/specs/remote_write_spec_2_0.md
bwplotka: Actually let's do this https://github.com/prometheus/docs/pull/2462#discussion_r1645729977A

content/docs/specs/remote_write_spec_2_0.md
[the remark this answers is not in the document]
bwplotka: In the code `ProtoMsg` would make sense, but in spec... in our mtg we decided `ProtobufMessage`

cstyan (approved):

bwplotka:
This experimental spec was approved here, but also internally over -team and specific member emails. Let's goo! 💪🏽 🤗 🎉 

This spec is ready to get implemented (it's already the case for Prometheus on https://github.com/prometheus/prometheus/tree/remote-write-2.0). Official comms/announcements will come once we have impl branch merged and blog post.

[Proposal](https://github.com/prometheus/proposals/pull/35) will be merged on stable release. Up to date list of follow up task are mentioned [here](https://github.com/prometheus/prometheus/issues/13105#issue-1982617705).

PR #2472, opened 2024-06-06
Update support page

bboreham:
I suspected that many of the companies listed no longer provide Prometheus support, so I went through and clicked on every button. In each case I read through the page, searched for "Prometheus" and did a Google search of <company name> Prometheus.

In this PR I updated a couple, and removed those where I could detect no trace of Prometheus support.

I left a few which are arguable.

### DIFF
--- a/content/support-training.html
+++ b/content/support-training.html
@@ -72,12 +72,6 @@ layout: jumbotron
           </div>
 
           <div class="col-md-3 col-sm-6 doc-content tc-col">
-              <a href="https://www.asyncdatado.com" class="btn btn-training btn-default btn-large" target="_blank">
-                <img src="../assets/commercial-support-logos/asyncdatado.png" alt="asyncdata do dominicana">
-              </a>
-          </div>
-            
-          <div class="col-md-3 col-sm-6 doc-content tc-col">
               <a href="https://www.credativ.com/" class="btn btn-training btn-default btn-large" target="_blank">
                 <img src="../assets/commercial-support-logos/credativ.png" alt="Credativ">
               </a>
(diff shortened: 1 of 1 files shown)


beorn7:
:thinking: Maybe we could ask the removed companies, or even those that are arguable, if they are still committed to support Prometheus? I mean, we don't have to hunt down contacts, but we could @-mention the people that originally added the entry…

As a different line of thought: What do we even mean by "commercial support"? A lot of companies will have _some_ kind of Prometheus support these days. With a very wide definition of commercial support, we could also completely remove the whole section, declaring Prometheus mainstream… (nobody would list companies providing "Linux support", either…). A narrower definition could be "company will send you a consultant to answer your Prometheus questions and/or help you with setting up Prometheus in your organization", i.e. it would be more a fit to consulting companies that to cloud providers offering Prometheus as a service.

bboreham:
I interpret “support” as meaning “I can pay someone to help me out with running Prometheus”.

beorn7:
Makes sense. But is that even true for Grafana Labs?

bboreham:
If “yes we can run it for you” is acceptable. 

> I left a few which are arguable.

SuperQ:
IIRC, Grafana does some amount of direct Prometheus support for customers. It's just not a publicly listed offering.

I think it's fine to keep Grafana if they want to stay listed.

SuperQ (approved):

simonpasquier (approved):
LGTM

roidelapluie (approved):

beorn7:
With all the approvals, let's just merge and not do the extra effort of explicitly ping'ing the companies involved.

If you belong to one of the removed companies, and you feel your company should still be listed here, please let us know.

PR #2473, opened 2024-06-07
Rename "Prometheus" docs section to "Prometheus Server"

juliusv:
As per Björn's comment on the team mailing list:

"I think the current "Prometheus" section is not very well named. It is essentially the part coming from the prometheus/prometheus repo (not prometheus/docs, as the rest fo the site), and it means "The Prometheus Server". I think it should be named that way because just "Prometheus" could mean a lot of things. On prometheus.io, "Prometheus" mostly refers to the overarching project."


beorn7 (approved):

bwplotka:
❤️

PR #2474, opened 2024-06-10
Added "specification" section for specs as agreed in -team discussion.

bwplotka:
This section will contain specs used by more projects than just Prometheus server e.g. RW protocols. 

Exposition formats are for now in https://prometheus.io/docs/introduction/overview/ but one idea is to mention those there but have full specs in this new section.

Used `unlock` icon - I like the intention of not-locking the ecosystem with OSS specs and standards. Also considered: `bullhorn` icon (important? announced loud?), `certificate` (kind of official, but icon does not look like cert really), `globe` (aka connecting the world) and `users` (aka connecting ppl).

I ensured https://prometheus.io/docs/concepts/remote_write_spec/ link still will works (check https://deploy-preview-2474--prometheus-docs.netlify.app/docs/concepts/remote_write_spec/)

image

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -20,6 +20,9 @@
 /docs/alerting/notification_examples/       /docs/alerting/latest/notification_examples/
 /docs/alerting/management_api/              /docs/alerting/latest/management_api/
 
+# Moved RW spec from Concepts to a new section.
+/docs/concepts/remote_write_spec /docs/specs/remote_write_spec
+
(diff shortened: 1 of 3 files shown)


juliusv:
:+1: Thanks :)

bwplotka:
Sounds fair to me, adjusted. I used `file-text-o` as it looks better (to me) next to the darker book icon for tutorials.

PR #2481, opened 2024-06-21
[PRW 2.0] (rc.1) Breaking change of io.prometheus.write.v2.Request for Receiver convenience.

bwplotka:
See rationales in https://github.com/prometheus/prometheus/pull/14330

Sucks to make breaking change, but I just merged the rc.0 a few hours ago, so I think it's fine. If it would be weeks we would probably NOT do this change.


RichiH:
You increased the version, so all good

npazosmendez (approved):

cstyan (approved):

PR #2484, opened 2024-06-25
remote-write: Add hint to check for CPU and network saturation

beorn7:
This is very general advice, but we don't have a rule of thumb available right now (plus, PRW 2.0 will change a lot here anyway).

This "fixes" #8890.

@bboreham @krajorama as discussed.


krajorama (approved):
ideally we include what metrics to look at for rw lag, but this is already an improvement

PR #2485, opened 2024-07-01
blog: State of Service Discoveries in Prometheus

ArthurSens:


ArthurSens:
After discussing more with the team, we're happy to not move forward in this direction :)

PR #2486, opened 2024-07-05
[PRW 2.0] rc.2 (breaking change): Added required response headers for durability.

bwplotka:
Fixes: https://github.com/prometheus/prometheus/issues/14359

### DIFF
--- a/content/docs/specs/remote_write_spec_2_0.md
+++ b/content/docs/specs/remote_write_spec_2_0.md
@@ -11,7 +11,7 @@ sort_rank: 4
 
 The Remote-Write specification, in general, is intended to document the standard for how Prometheus and Prometheus Remote-Write compatible senders send data to Prometheus or Prometheus Remote-Write compatible receivers.
 
-This document is intended to define a second version of the [Prometheus Remote-Write](./remote_write_spec.md) API with minor changes to protocol and semantics. This second version adds a new Protobuf Message with new features enabling more use cases and wider adoption on top of performance and cost savings. The second version also deprecates the previous Protobuf Message from a [1.0 Remote-Write specification](./remote_write_spec.md#protocol) and adds mandatory [`X-Prometheus-Remote-Write-Written-*` HTTP response headers](#required-x-prometheus-remote-write-written--response-headers) for reliability purposes. Finally, this spec outlines how to implement backwards-compatible senders and receivers (even under a single endpoint) using existing basic content negotiation request headers. More advanced, automatic content negotiation mechanisms might come in a future minor version if needed. For the rationales behind the 2.0 specification, see [the formal proposal](https://github.com/prometheus/proposals/pull/35).
+This document is intended to define a second version of the [Prometheus Remote-Write](./remote_write_spec.md) API with minor changes to protocol and semantics. This second version adds a new Protobuf Message with new features enabling more use cases and wider adoption on top of performance and cost savings. The second version also deprecates the previous Protobuf Message from a [1.0 Remote-Write specification](./remote_write_spec.md#protocol) and adds mandatory [`X-Prometheus-Remote-Write-*-Written` HTTP response headers](#required-written-response-headers)for reliability purposes. Finally, this spec outlines how to implement backwards-compatible senders and receivers (even under a single endpoint) using existing basic content negotiation request headers. More advanced, automatic content negotiation mechanisms might come in a future minor version if needed. For the rationales behind the 2.0 specification, see [the formal proposal](https://github.com/prometheus/proposals/pull/35).
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
 
@@ -41,7 +41,8 @@ In this document, the following definitions are followed:
 * a `Protobuf Message` (or Proto Message) refers to the [content type](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-type) definition of the data structure for this Protocol. Since the specification uses [Google Protocol Buffers ("protobuf")](https://protobuf.dev/) exclusively, the schema is defined in a ["proto" file](https://protobuf.dev/programming-guides/proto3/) and represented by a single Protobuf ["message"](https://protobuf.dev/programming-guides/proto3/#simple).
 * a `Wire Format` is the format of the data as it travels on the wire (i.e. in a network). In the case of Remote-Write, this is always the compressed binary protobuf format.
 * a `Sender` is something that sends Remote-Write data.
-* a `Receiver` is something that receives (writes) Remote-Write data. The meaning of `writing` is up to the Receiver e.g. usually it means storing received data in a database, but also just validating, splitting or enhancing it.
+* a `Receiver` is something that receives (writes) Remote-Write data. The meaning of `Written` is up to the Receiver e.g. usually it means storing received data in a database, but also just validating, splitting or enhancing it.
+* `Written` refers to data the `Receiver` has received and is accepting. Whether or not it has ingested this data to persistent storage, written it to a WAL, etc. is up to the `Receiver`. The only distinction is that the `Receiver` has accepted this data rather than explicitly rejecting it with an error response.
 * a `Sample` is a pair of (timestamp, value).
 * a `Histogram` is a pair of (timestamp, [histogram value](https://github.com/prometheus/docs/blob/b9657b5f5b264b81add39f6db2f1df36faf03efe/content/docs/concepts/native_histograms.md)).
 * a `Label` is a pair of (key, value).
@@ -136,25 +137,25 @@ Receivers MUST NOT return a 2xx HTTP status code if any of the pieces of sent da
 
 The following subsections specify Sender and Receiver semantics around headers and different write error cases.
 
-#### Required `X-Prometheus-Remote-Write-Written-*` Response Headers
+#### Required `Written` Response Headers
 
 <!---
 Rationales: https://github.com/prometheus/prometheus/issues/14359
 -->
-Upon a successful content negotiation, Receivers process (write) the received batch of data. Once completed (with success or failure) for each important piece of data (currently Samples, Histograms and Exemplars) Receivers MUST send a dedicated HTTP `X-Prometheus-Remote-Write-Written-*` response header with the precise number of successfully written elements. 
+Upon a successful content negotiation, Receivers process (write) the received batch of data. Once completed (with success or failure) for each important piece of data (currently Samples, Histograms and Exemplars) Receivers MUST send a dedicated HTTP `X-Prometheus-Remote-Write-*-Written` response header with the precise number of successfully written elements. 
 
 Each header value MUST be a single 64-bit integer. The header names MUST be as follows:
 
-* `X-Prometheus-Remote-Write-Written-Samples <integer; count of all successfully written Samples from this request>`
-* `X-Prometheus-Remote-Write-Written-Histograms <integer; count of all successfully Histogram samples from this request>`
-* `X-Prometheus-Remote-Write-Written-Exemplars <integer; count of all successfully Exemplars from this request>`
+* `X-Prometheus-Remote-Write-Samples-Written <integer; count of all successfully written Samples from this request>`
+* `X-Prometheus-Remote-Write-Histograms-Written <integer; count of all successfully Histogram samples from this request>`
+* `X-Prometheus-Remote-Write-Exemplars-Written <integer; count of all successfully Exemplars from this request>`
 
-Upon receiving a 2xx or a 4xx status code, Senders CAN assume that any missing `X-Prometheus-Remote-Write-Written-*` response header means no element from this category (e.g. Sample) was written by the Receiver (count of `0`). Senders MUST NOT assume the same when using the deprecated `prometheus.WriteRequest` Protobuf Message due to the risk of hitting 1.0 Receiver without this feature.
+Upon receiving a 2xx or a 4xx status code, Senders CAN assume that any missing `X-Prometheus-Remote-Write-*-Written` response header means no element from this category (e.g. Sample) was written by the Receiver (count of `0`). Senders MUST NOT assume the same when using the deprecated `prometheus.WriteRequest` Protobuf Message due to the risk of hitting 1.0 Receiver without this feature.
 
 Senders MAY use those headers to confirm which parts of data were successfully written by the Receiver. Common use cases:
 
 * Better handling of the [Partial Write](#partial-write) failure situations: Senders MAY use those headers for more accurate client instrumentation and error handling.
-* Detecting broken 1.0 Receiver implementations: Senders SHOULD assume [415 HTTP Unsupported Media Type](https://www.rfc-editor.org/rfc/rfc9110.html#name-415-unsupported-media-type) status code when sending the data using `io.prometheus.write.v2.Request` request and receiving 2xx HTTP status code, but none of the `X-Prometheus-Remote-Write-Written-*` response headers from the Receiver. This is a common issue for the 1.0 Receivers that do not check the `Content-Type` request header; accidental decoding of the `io.prometheus.write.v2.Request` payload with `prometheus.WriteRequest` schema results in empty result and no decoding errors.
+* Detecting broken 1.0 Receiver implementations: Senders SHOULD assume [415 HTTP Unsupported Media Type](https://www.rfc-editor.org/rfc/rfc9110.html#name-415-unsupported-media-type) status code when sending the data using `io.prometheus.write.v2.Request` request and receiving 2xx HTTP status code, but none of the `X-Prometheus-Remote-Write-*-Written` response headers from the Receiver. This is a common issue for the 1.0 Receivers that do not check the `Content-Type` request header; accidental decoding of the `io.prometheus.write.v2.Request` payload with `prometheus.WriteRequest` schema results in empty result and no decoding errors.
 * Detecting other broken implementations or issues: Senders MAY use those headers to detect broken Sender and Receiver implementations or other problems.
 
 Senders MUST NOT assume what Remote Write specification version the Receiver implements from the remote write response headers.
--- a/content/docs/specs/remote_write_spec_2_0.md
+++ b/content/docs/specs/remote_write_spec_2_0.md
@@ -143,7 +143,7 @@ Rationales: https://github.com/prometheus/prometheus/issues/14359
 -->
 Upon a successful content negotiation, Receivers process (write) the received batch of data. Once completed (with success or failure) for each important piece of data (currently Samples, Histograms and Exemplars) Receivers MUST send a dedicated HTTP `X-Prometheus-Remote-Write-Written-*` response header with the precise number of successfully written elements. 
 
-Each header value MUST be a single 64-byte integer. The header names MUST be as follows:
+Each header value MUST be a single 64-bit integer. The header names MUST be as follows:
 
 * `X-Prometheus-Remote-Write-Written-Samples <integer; count of all successfully written Samples from this request>`
 * `X-Prometheus-Remote-Write-Written-Histograms <integer; count of all successfully Histogram samples from this request>`
--- a/content/docs/specs/remote_write_spec_2_0.md
+++ b/content/docs/specs/remote_write_spec_2_0.md
@@ -130,9 +130,9 @@ Senders MUST include a user agent header that SHOULD follow [the RFC 9110 User-A
 
 ### Response
 
-Receivers ingesting all samples successfully MUST return a [success 2xx HTTP status code](https://www.rfc-editor.org/rfc/rfc9110.html#name-successful-2xx). In such a successful case, the response body from the Receiver SHOULD be empty and the status code SHOULD be [204 HTTP No Content](https://www.rfc-editor.org/rfc/rfc9110.html#name-204-no-content); Senders MUST ignore the response body. The response body is RESERVED for future use.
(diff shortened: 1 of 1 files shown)


content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -134,7 +134,30 @@ Receivers ingesting all samples successfully MUST return a 200 HTTP status code.
...
+
+Receivers MUST send those response headers, even if they use asynchronous or eventual write mechanisms e.g. when Receivers can't confirm that downstream storage actually written them in durable way. The headers are only to confirm how many Samples, Histograms and Exemplars were accepted by the Receiver, especially useful on [Partial Write](#partial-write) situations.
+
+Each header value MUST be a single 64 bytes integer. The header names MUST be as follows:
+
+```
+X-Prometheus-Remote-Write-Received-Samples <integer; count of all written Samples from this request>
+X-Prometheus-Remote-Write-Received-Histograms <integer; count of all written Histogram samples from this request>
+X-Prometheus-Remote-Write-Received-Exemplars <integer; count of all written Exemplars from this request>
+```
```
cstyan, line 154:
```suggestion
For all HTTP status codes, Receivers MUST send back the HTTP response `X-Prometheus-Remote-Write-Accepted-` type of the header indicating write statistics, i.e. how many Samples, Histograms and Exemplars were successfully accepted/ingested by the Receiver. Receivers MAY omit each header, if its value is zero.

Receivers MUST send those response headers, even if they use asynchronous or eventual write mechanisms e.g. when Receivers can't confirm that downstream storage actually written them in durable way. The headers are only to confirm how many Samples, Histograms and Exemplars were accepted by the Receiver, especially useful on [Partial Write](#partial-write) situations.

Each header value MUST be a single 64 bytes integer. The header names MUST be as follows:

```
X-Prometheus-Remote-Write-Accepted-Samples <integer; count of all written Samples from this request>
X-Prometheus-Remote-Write-Accepted-Histograms <integer; count of all written Histogram samples from this request>
X-Prometheus-Remote-Write-Accepted-Exemplars <integer; count of all written Exemplars from this request>
```

Comment (cstyan) [reply content/docs/specs/remote_write_spec_2_0.md:154]: I think we should change `Received` to `Accepted` to make it clearer that the values in the header are supposed to be the # of each data type successfully ingested rather than just the # of each that were received in the request.

Comment (bwplotka) [reply content/docs/specs/remote_write_spec_2_0.md:154]: Hm, good point. We can.

I wonder if naming it "accepted" won't conflict with the statement added to the spec in this PR about eventual storage.

What if there is Receiver by design stores things in some queue and cannot confirm durability?

I guess "accept" does not mean "durablyStoredForverWeDoubleChecked" so it's fine too 🙈

Comment (cstyan) [reply content/docs/specs/remote_write_spec_2_0.md:154]: That's a whole different can of worms, and I think outside the scope for RW spec. We can really only specify that they should tell us how many of each samples they are going to accept, or reject. The 200 status code or retryable error is only able to tell the Sender "I can accept these" or "some of these are not accepted but retryable". Whether or not that data has been committed to persistent storage is outside RW's concerns IMO. Every RW Receiver implementer is going to have different guarantees about how they accept data and the timeline on which accepted data is committed to persistent storage/available for reads.

Comment (bwplotka) [reply content/docs/specs/remote_write_spec_2_0.md:154]: Let's do accept!

**Comment by bwplotka** [comment]
Ok, with new wording suggested by @cstyan I was able to simplify the section plus add clear example. Also updated 200 section (we often use 204 No content HTTP status instead, which makes more sense).

Changes:
* Added summary on top.
* Be precise on 2xx / 204 status codes.
* Simplified response header section
* Removed eventual durability case, with "accepted" wording that's obvious.
* Grammarly

PTAL @cstyan @npazosmendez @alexgreenbank

**Comment by bwplotka** [comment]
See Receiver https://github.com/prometheus/prometheus/pull/14427 and Sender https://github.com/prometheus/prometheus/pull/14444 implementation with those.

Generally it allows Sender to catch common bug AND calculate correct partial error metrics.

File: content/docs/specs/remote_write_spec_2_0.md [hunk 2]
```diff
@@ -130,18 +130,42 @@ Senders MUST include a user agent header that SHOULD follow [the RFC 9110 User-A
...
 
-The following subsections specify Sender and Receiver semantics around different write error cases.
+The following subsections specify Sender and Receiver semantics around headers and different write error cases.
+
+#### Response Headers
+
+<!---
+Rationales: https://github.com/prometheus/prometheus/issues/14359
+-->
+Upon a successful content negotiation, Receivers ingest the received Samples, Histograms and Exemplars. In such cases, Receivers MUST track the total number of Samples, Histograms or Exemplars that were accepted or ingested. Receivers MUST share those statistics with the Sender as the HTTP response `X-Prometheus-Remote-Write-Accepted-*` headers specified below. Receivers MAY omit those when the value is zero.
```
cstyan, line 144: > Receivers MAY omit those when the value is zero.

I don't think we should allow this. It's yet another way to confuse the end user. Did I get no values because nothing was accepted, or because I sent 2.0 and the Receiver only speaks 1.0?

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -130,18 +130,42 @@ Senders MUST include a user agent header that SHOULD follow [the RFC 9110 User-A
 
 ### Response
 
-Receivers ingesting all samples successfully MUST return a 200 HTTP status code. In such a successful case, the response body from the Receiver SHOULD be empty; Senders MUST ignore the response body. The response body is RESERVED for future use.
+Receivers ingesting all Samples and Histograms successfully MUST return a [success 2xx HTTP status code](https://www.rfc-editor.org/rfc/rfc9110.html#name-successful-2xx). In such a successful case, the response body from the Receiver SHOULD be empty and the status code SHOULD be [204 HTTP No Content](https://www.rfc-editor.org/rfc/rfc9110.html#name-204-no-content); Senders MUST ignore the response body. The response body is RESERVED for future use.
```
cstyan, line 133: > In such a successful case, the response body from the Receiver SHOULD be empty and the status code SHOULD be [204 HTTP No Content](https://www.rfc-editor.org/rfc/rfc9110.html#name-204-no-content);

What's the reasoning for changing to 204?

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -141,17 +141,17 @@ The following subsections specify Sender and Receiver semantics around headers a
...
 
-For example, Senders MAY assume [415 HTTP Unsupported Media Type](https://www.rfc-editor.org/rfc/rfc9110.html#name-415-unsupported-media-type) status code, when `io.prometheus.write.v2.Request` request with samples, results in 2xx HTTP status code, but no response headers from the Receiver (a common issue for the 1.0 Reciever without content-type header check).
+Senders MUST treat the omission of any `X-Prometheus-Remote-Write-Accepted-*` header as either no Samples, Histograms or Exemplars were accepted by the Receiver or the Receiver only supports Remote Write 1.0 protocol and `prometheus.WriteRequest` Protobuf Message. For example, Senders MAY assume [415 HTTP Unsupported Media Type](https://www.rfc-editor.org/rfc/rfc9110.html#name-415-unsupported-media-type) status code, when the `io.prometheus.write.v2.Request` request with samples, results in 2xx HTTP status code, but no `X-Prometheus-Remote-Write-Accepted-*` response headers from the Receiver (a common issue for the 1.0 Receiver without content-type header check). Senders MUST NOT assume the same for `prometheus.WriteRequest` (for compatibility reasons) or when receiving zero counts (Receiver might )
```
cstyan: this would be simpler if we didn't allow for omitting the header if the value was 0

content/docs/specs/remote_write_spec_2_0.md
bwplotka: Great question! 

First of all this was previously wrong as we technically NOT allowed 204 ;p so I moved this to 2xx as in 1.0 https://prometheus.io/docs/specs/remote_write_spec/#:~:text=Prometheus%20remote%20Write%20compatible%20receivers%20MUST%20respond%20with%20a%20HTTP%202xx%20status%20code%20when%20the%20write%20is%20successful.

Secondly I noticed Prometheus handler and others use [204](https://www.rfc-editor.org/rfc/rfc9110.html#name-204-no-content), not 200. I dived deeper and the reason is simple - we won't write any response, so correct response is 204. This helps clients to be more efficient and assume no body (in theory).

Thanks for challenging, because now I might say we don't need this SHOULD. RFC 9110 states that no response means 204, so no need to specify this in our spec. However one benefit of specifying this here is that not many might now about 204, so worth highlighting? I think generally I would keep it 🙈 No harm in highlighting, it's just a SHOULD and takes a few words really. 🤔

bwplotka: Actually since you are here I noticed another change we did in the past vs 1.0.

1.0 states:

> Prometheus remote Write compatible receivers MUST respond with a HTTP 2xx status code when the write is successful

We state:

> Receivers ingesting all Samples and Histograms successfully MUST return a [success 2xx HTTP status code](https://www.rfc-editor.org/rfc/rfc9110.html#name-successful-2xx).

In some way we specify **the success condition**. In 1.0 they were up to Receiver. More open definition allow variety of Receive use cases perhaps? Not sure yet if this good or bad, but it matters for Response Header Stats. Let's iterate with some examples (helps me too - thinking loudly):

A) Like a validation receiver that only checks stuff and NOT ingest (like literally our compliance test) - it might not ingest (as in-store anywhere) anything and still return true, so even "ingesting" is dubious here. 

B) What if Receiver does not support native histograms yet? If we kept open definition from 1.0 it could either return 400 on those samples or 200 (silent ignore). In current 2.0 it HAS to return 400 partial error (which is kind of nice, because receivers are more predictable form sender perspective)?

C) What if somebody creates a Receiver that ONLY wants native histograms but nothing else. Then from this Receiver perspective success means ingested histograms, but samples would cause 400 effectively blocking this kind of receiver from being implemented (likely no senders would be happy to have mode of "skip samples"). 

D) What if somebody creates a Receiver that ONLY wants exemplars but nothing else (like a separate store). This sucks a lot now because it will always have to return 400 kind of, so can't silently drop other stuff (maybe more predictable again?)

E) What if future Sender wants to ensure exemplars are part of success (ultra durability of everything). Sender can't rely on 2xx anymore but has to check Stats. I guess that's fine, with 1.0 undefined success they cannot technically rely on 2xx even unless they know implementation of Receiver.

So generally.. so far it feels to me, that binding the success definition to Remote Write and not Receiver implementation is helping because Sender can rely on spec more, less on different kinds of implementations 🤔  I remember we discussed a bit about same in the past, but can't find it now 🙈 

Now, for the response headers it matters, because of that assumption that if you got 2xx AND (accepted samples == 0 and/or accepted histograms == 0) AND you sent some samples/histograms == you talk to 1.0 receiver accidently. If success was not defined.. we couldn't assume so. Maybe there are simpler ways to achieve so? I ruled out `X-Remote-Write-Version` in Response HTTP header, because it would still not guarantee bug didn't happened in the content negotiation. @npazosmendez proved alternative of deterministic unmarshal fail is also out of question. 

Anyway, will add those reasoning to proposal for future readers. Wanted to rechallenge and confirm we are ok here (:

bwplotka: And finally I defined `samples` in previous statement meant Samples and Histograms... but maybe we wanted to include Exemplars too? In some way "successful write" of Sample with Exemplar is only successful when exemplar is also ingested? Maybe not? Maybe it's fine to not specify fully and leave that to Receiver?

content/docs/specs/remote_write_spec_2_0.md
bwplotka: Great question and I'm glad you challenged this! Was removing and adding this back like 5 times. It's not clear what will confuse "user" (I assume you mean Sender) here (: 

TL;DR I see the Pros of this, but not sure yet, let's chat on our sync.

Let's unpack:

**What if we say it's MUST for Receiver to put those headers even if it's 0.**

**Pros:**
* For 2xx it's indeed clear if we talk to Receiver 1.0 or 2.0 (potentially 2.1 in future too). This helps a bit on 2 cases below:
  a. It simplifies the known problem detection for 2xx a lot e.g. used to be 

    > you got 2xx AND (accepted samples == 0 and/or accepted histograms == 0) AND you sent some samples/histograms == you talk to 1.0 receiver with known bug of not checking content type.

    Now, it's more like "you got 2xx and none of the required headers in 2.0 == that case". 

  b. Allows us actually to check if 2.0 Receiver correctly uses 1.0 payload we sent. Had to ignore this case in https://github.com/prometheus/prometheus/pull/14444 as you don't know what Receiver version it is.

I don't see any other benefits yet. Particularly on any non 2xx case Sender cannot assume the Receiver spec version (see cons why).

**Cons:**
* Sender implementation complexity. Treating empty and zero value as different cases is generally asking for complexity, so wanted to avoid this. In Go it sucks to represent this, you are force to use pointers or flags (`hasSpecified`). It's the reason why it sucks sometimes to work with proto or yamls or any data format and we end up with those `aws.String` or `proto.Int` that just pointerize for little reason. In other languages it requires more bytes to store and complex data structures too. If we can avoid it, it would be amazing.
* Sender confusion due to inconsistency with other failure use cases. Zero default is consistent with thousands of cases where you will NOT have those headers because e.g. HTTP didn't even hit Receiver (e.g. proxy errors, initial HTTP validation before parsing proto). This means we can't rely on this behaviour for non 2xx.
* Wanted to avoid relying on Sender doing special things for certain spec version as long as possible. Let's list why it's bad/not ideal, especially for failure cases:
  * Similar to above, we will have cases of non-2xx AND No headers EVEN for 2.0 Receivers. So now Senders can't fully assume Receiver spec version on ANY error case anyway, but they might be tempted too.
  * Receiver Implementation now need to ask themselves "do I need to add those headers the moment I start processing the request or only after I decode the proto or should my forward proxy do that? How far should I unsure this". It will pain to even define this, I tried with "when processing decoded proto" mention.
* Inconsistency with forward compatibility of 2.0 Receiver. If we add new header in 2.1 (perhaps a new type of sample) the 2.0 receivers would NOT add them. However such new sample does not exists, so it's definitely means zero, which offers a clean Sender implementation. With required 0 header, Senders now require more friction to implement as `X-Prometheus-Remote-Write-Accepted-SomethingAddedIn2_1` now holds two meaning, but this information is useless it all means just zero after all.
  * Counter argument: This only works if we can guarantee future headers omission will mean zero. One tricky example is e.g. metadata header in future, so can we guarantee that?

bwplotka: Another question - what Sender should do with formatted response header (e.g. `X-Prometheus-Remote-Write-Accepted-Samples 12.44-yolo`)?

bwplotka: We found some consensus in https://cloud-native.slack.com/archives/C06B8KJUU2Y/p1720628447889389 will try to put that on the PR.

content/docs/specs/remote_write_spec_2_0.md
bwplotka: We found some consensus in https://cloud-native.slack.com/archives/C06B8KJUU2Y/p1720628447889389 will try to put that on the PR.

content/docs/specs/remote_write_spec_2_0.md
bwplotka: Yes and no - we discussed this heavily with @cstyan  in https://cloud-native.slack.com/archives/C06B8KJUU2Y/p1720628447889389 will try to put the consensu on the PR. 🤗

bwplotka:
Rdy for final call!

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -130,18 +130,43 @@ Senders MUST include a user agent header that SHOULD follow [the RFC 9110 User-A
...
+* Broken 1.0 Receiver implementations: Senders SHOULD assume [415 HTTP Unsupported Media Type](https://www.rfc-editor.org/rfc/rfc9110.html#name-415-unsupported-media-type) status code when sending the data using `io.prometheus.write.v2.Request` request and receiving 2xx HTTP status code, but none of the `X-Prometheus-Remote-Write-Written-*` response headers from the Receiver. This is a common issue for the 1.0 Receivers that do not check the `Content-Type` request header; accidental decoding of the `io.prometheus.write.v2.Request` payload with `prometheus.WriteRequest` schema results in empty result and no decoding errors.
+* Unknown broken implementations or issues: Senders MAY use those headers to detect broken Sender and Receiver implementations or other problems.
+
+On the 2xx HTTP status code, Senders CAN assume that a missing write response header means that no such a piece of data was written by the Receiver (count of `0`). value. One exception is that Senders MUST NOT assume the same (empty header means zero) when using the deprecated `prometheus.WriteRequest` Protobuf Message in the request.
```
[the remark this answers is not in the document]
thampiotr: Maybe we can rephrase this for clarity. Something like:

> Upon receiving a 2xx HTTP status code with missing write response headers:
> - If the Sender used the `io.prometheus.write.v2.Request` schema to encode the request, it CAN assume that no data was written by the Receiver (count of `0`).
> - If the Sender used a deprecated `prometheus.WriteRequest` Protobuf Message, it MUST NOT assume that no data was written by the Receiver (count of `0`).

In the second case, should Sender assume the data was written correctly? (I might be missing some context here).

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -130,18 +130,43 @@ Senders MUST include a user agent header that SHOULD follow [the RFC 9110 User-A
...
+
+Senders MUST NOT assume what Remote Write specification version the Receiver implements from the remote write response headers.
```
thampiotr, line 160:
```suggestion
Senders MUST NOT assume what Remote Write specification version the Receiver implements the remote write response headers.
```

content/docs/specs/remote_write_spec_2_0.md
```diff
@@ -130,18 +130,43 @@ Senders MUST include a user agent header that SHOULD follow [the RFC 9110 User-A
...
+The following subsections specify Sender and Receiver semantics around headers and different write error cases.
+
+#### Required `X-Prometheus-Remote-Write-Written-*` Response Headers
+
+<!---
+Rationales: https://github.com/prometheus/prometheus/issues/14359
+-->
+Upon a successful content negotiation, Receivers process (write) the received batch of data. Once completed (with success or failure) for each important piece of data (currently Samples, Histograms and Exemplars) Receivers MUST send a dedicated HTTP `X-Prometheus-Remote-Write-Written-*` response header with the precise number of successfully written elements. 
+
+Each header value MUST be a single 64-byte integer. The header names MUST be as follows:
```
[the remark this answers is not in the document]
bwplotka: lol

content/docs/specs/remote_write_spec_2_0.md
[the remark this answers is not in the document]
bwplotka: Thanks!

So first case is broader - as there might be more messages in future, but I will check how we can improve.

> In the second case, should Sender assume the data was written correctly? (I might be missing some context here).

Correct, there is no way to tell UNLESS you see the header with values for 1.0 message (even if 2.0 receiver was used, we don't know that). I will add example perhaps - might help

content/docs/specs/remote_write_spec_2_0.md
bwplotka: actually that's wrong suggestion no? (: I merged this but will revert, it should be "from"?

content/docs/specs/remote_write_spec_2_0.md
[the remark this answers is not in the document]
bwplotka: Done, PTAL @thampiotr

thampiotr (approved):
LGTM

cstyan (approved):
LGTM. I still think the `accepted` wording was okay, it was really `ingest/ingested` we needed to remove imo since that implied some kind of persistence which rules out some receivers or at least significantly increases the response time on writes.

cstyan:
> I want to make sure you are happy with final wording decision too (: We could switch back to `Accepted` but:

I'm not unhappy, but I do think `Accepted` reads nicer than `X-Prometheus-Remote-Write-Written-X`. But the point that we have to define `Accepted` is a valid one

> * We don't use it anywhere, it has unknown definition. What definition of it you would put? Would it be the same or different to `Write`?

What about:
> Accepted, and a 200 response, does not strictly require that data has already been persisted to long term storage and is queryable, but rather implies that the data has been accepted in some form and _will_ be persisted. This could mean the data has been written to a WAL or queuing system with delivery guarantees such as Kafka. Receiver implementations should be careful, users will expect that any piece of data that is "Accepted" will be persisted to durable storage and not lost.


And then...
> * We ruled out `Receive` word [[PRW 2.0] rc.2 (breaking change): Added required response headers for durability. #2486 (comment)](https://github.com/prometheus/docs/pull/2486#discussion_r1666965320) from the opposite reason to `ingest`: We don't want to rule out ingesting receivers and with `ingest` we would rule only validating receivers. It feels `accept` has the same collocation as `receive` here - for the receivers that actually ingest it does not give the feeling you should bump this number ONLY if ingested perhaps?

Damn, now I'm just contradicting myself over the course of multiple comments, including this one I am writing right now.

I think `Ingested` implies too much that we can't enforce, and `Write-Written` doesn't read nicely, `Received` or `Accepted` is good with some version of the description from earlier in this comment.

bwplotka:
Thanks! I agree Received == Accepted, but the reason we rule out receive fits accepted IMO, so it implies it's NEVER durable write (in practice, it depends on the receiver). Write is universal, it's whatever remote write is for, so best. Ingest is too in the other extreme.

Agree it does not read well. 

What about we imply "write" and have 'X-Prometheus-Remote-Written-Samples'? (:

bwplotka:
The more I think about it, the more I like accepted/received like we had and really focus on this NEVER being durable write count, and always about initial validation/how many samples not rejected.. but then we need rewrite this section a bit for that and perhaps never send them on internal errors (non recoverable?). Tricky.

bwplotka:
We switched to `X-Prometheus-Remote-Write-*-Written` form. Naming is hard, but this should work great. Thanks @cstyan for last updates 💪🏽 Let's go!

PR #2487, opened 2024-07-11
Add anynines to the list of commercial support providers

asmagulov:
Please add anynines to the list of commercial support providers of Prometheus.


jan--f:
Hello from the bug scrub!

@asmagulov I would like to repeat bwplotka's question. We happily list support providers for upstream Prometheus, less so for prometheus-like SaaS services.

jan--f:
Hello from the bug scrub!
I'll close this for now, but please feel free to re-open in the future.

PR #2491, opened 2024-07-18
Remote Write defaults have increased

jamessanford:
Note that I multiplied the old "500 kB" figure by 4, keeping the same "200 bytes per sample" -- but I'm not sure where that number comes from.

The increased defaults are from prometheus/prometheus@889ef998f4f73d0d1fbd961e59494584e3616e42

Added breadcrumb to `prometheus_remote_storage_samples_pending` metric.


jamessanford:
Noticed the docs were out of date while creating https://github.com/jamessanford/remote-tsdb-clickhouse/pull/12

juliusv:
:+1: Thanks!

PR #2494, opened 2024-07-26
fix(docs): amend exposition_formats.md broken relative links

javiergarea:


juliusv:
Hi, and thanks! In this case, the existing links are actually already correct, since the content is meant to be viewed from the Prometheus website at https://prometheus.io/docs/instrumenting/exposition_formats/ (where you can see that they work). This repo here is just the content for the website. So closing this PR.

javiergarea:
> Hi, and thanks! In this case, the existing links are actually already correct, since the content is meant to be viewed from the Prometheus website at https://prometheus.io/docs/instrumenting/exposition_formats/ (where you can see that they work). This repo here is just the content for the website. So closing this PR.

Oops, my bad! 😄

PR #2495, opened 2024-07-30
fix(instrumenting): Fix link to python pushgateway docs

RodneyMKay:
The Link to the Python docs in the Pushgateway Section of the docs seems to be outdated. It used to point to a specific heading in the readme of the python client, but it seems that project has since moved to a dedicated documentation website. This PR addresses this by using the new Link.

Note: I'm not sure whether this is the exact location that it used to point to, but I think the connection makes sense this way.


RodneyMKay:
@RichiH I saw in CONTRIBUTING.md, that the maintainer should be mentioned for PRs for trivial fixes... sooo, I'm hoping I'm doing this correctly (:

SuperQ (approved):
Thanks!

CC @csmarchbanks

PR #2505, opened 2024-09-01
Update exporters.md - Fix Repojacking in Prometheus Exporters

yakirk:
Repojacking is a security vulnerability that occurs when an attacker takes control of a repository name that was previously owned by another user or organization. This typically happens when the original owner deletes or renames their repository. If this repository is still referenced by other projects, an attacker can create a new repository with the same name, introducing malicious code that could be executed by those relying on the original repository link.
In Prometheus's case, several exporters listed on Prometheus's official documentation could be claimed by an attacker. The attacker could create a repository with the old username and project name, potentially hosting a malicious exporter. 

For example, currently, when a user clicks on the "aws-cost-exporter" link, GitHub redirects them to https://github.com/electrolux-oss/aws-cost-exporter, even though the original link points to https://github.com/opensourceelectrolux/aws-cost-exporter. Since the username "opensourceelectrolux" is available for claim, an attacker could take over this username and host a malicious exporter under https://github.com/opensourceelectrolux/aws-cost-exporter. 

This PR updates the links to ensure they point to trusted sources.

Thanks,
Yakir Kadkoda @yakirk 
Ofek Itach @Ofekitach

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -129,10 +129,10 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Apache exporter](https://github.com/Lusitaniae/apache_exporter)
    * [HAProxy exporter](https://github.com/prometheus/haproxy_exporter) (**official**)
    * [Nginx metric library](https://github.com/knyar/nginx-lua-prometheus)
-   * [Nginx VTS exporter](https://github.com/hnlq715/nginx-vts-exporter)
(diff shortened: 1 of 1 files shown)


beorn7 (approved):
Thank you very much.

I was wondering how to ensure that those repos weren't already repojacked. I checked all the original orgs, and they are all not taken at the moment, which strongly implies that the redirects are legit. But would be interesting to hear if you have applied any other check for that.

beorn7:
I merge this now as things won't get worse than the current redirect situation, even if the repos have already been repojacked. But if anyone has insights one way or the other, please share.

PR #2509, opened 2024-09-08
Added GraphQL exporter

ricardbejarano:

### DIFF
--- a/content/docs/instrumenting/exporters.md
+++ b/content/docs/instrumenting/exporters.md
@@ -151,6 +151,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Fastly exporter](https://github.com/peterbourgon/fastly-exporter)
    * [GitHub exporter](https://github.com/githubexporter/github-exporter)
    * [Gmail exporter](https://github.com/jamesread/prometheus-gmail-exporter/)
+   * [GraphQL exporter](https://github.com/ricardbejarano/graphql_exporter)
    * [InstaClustr exporter](https://github.com/fcgravalos/instaclustr_exporter)
    * [Mozilla Observatory exporter](https://github.com/Jimdo/observatory-exporter)
    * [OpenWeatherMap exporter](https://github.com/RichiH/openweathermap_exporter)


juliusv:
Thanks! Do you have an example /metrics output to look at (I realize this would vary based on the GraphQL endpoint and query)?

When linking to an exporter, we just want to ensure that metric names etc. roughly stick to our guidelines (https://prometheus.io/docs/practices/naming/ and https://prometheus.io/docs/instrumenting/writing_exporters/).

ricardbejarano:
They're generated based on the response schema of the GraphQL API you're querying.
This is to make it generic and allow querying any GraphQL API with any schema.

A query like this:
```yaml
scrape_configs:
  - job_name: "cloudflare-analytics"
    metrics_path: "/query"
    bearer_token: "<YOUR CLOUDFLARE API TOKEN>"
    params:
      endpoint: ["https://api.cloudflare.com/client/v4/graphql"]
      query:
        - |
          {
            viewer {
              zones(filter: {zoneTag: $zoneTag}) {
                httpRequests1hGroups(limit: 1, filter: {datetime_geq: "{{ Now "-1h" }}"}) {
                  dimensions {
                    datetime
                  }
                  sum {
                    bytes
                    cachedBytes
                    cachedRequests
                    requests
                  }
                }
              }
            }
          }
      zoneTag: ["<YOUR CLOUDFLARE ZONE ID>"]
    static_configs:
      - targets: ["127.0.0.1:9199"]
```

Would produce metrics (all gauges) such as:
```
query_viewer_zones_httpRequests1hGroups_sum_bytes{...} 10000
query_viewer_zones_httpRequests1hGroups_sum_cachedBytes{...} 9000
query_viewer_zones_httpRequests1hGroups_sum_cachedRequests{...} 900
query_viewer_zones_httpRequests1hGroups_sum_requests{...} 1000
```

Where the name of the metrics is "query_" plus the keys of JSON path to get to the value, joined by underscores.

Again, this all depends on the API you're querying and its schema.

juliusv:
I see, thanks! Yeah, I guess it's hard to get metrics that are compliant with our usual naming and labeling best conventions in that case.

:+1:

PR #2512, opened 2024-09-19
OTel guide: Add section on `target_info`

aknuds1:
Add to the OTel guide a section on the `target_info` metric, and its usage.

### DIFF
--- a/content/docs/guides/opentelemetry.md
+++ b/content/docs/guides/opentelemetry.md
@@ -82,7 +82,7 @@ target_info
 
 What happens in this query is that the time series resulting from `rate(http_server_request_duration_seconds_count[2m])` are augmented with the `k8s_cluster_name` label from the `target_info` series that share the same `job` and `instance` labels.
 In other words, the `job` and `instance` labels are shared between `http_server_request_duration_seconds_count` and `target_info`, akin to SQL foreign keys.
-The `k8s_cluster_name` label, OTOH, corresponds to the OTel resource attribute `k8s.cluster.name` (Prometheus converts dots to underscores).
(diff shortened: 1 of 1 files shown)


content/docs/guides/opentelemetry.md
```diff
@@ -67,6 +67,31 @@ otlp:
     - ...
 ```
 
+## Including resource attributes at query time
+
+An alternative to promoting resource attributes, as described in the previous section, is to add labels from the `target_info` metric when querying.
+This is conceptually known as a "join" query.
```
jesusvazquez, line 73: Maybe add a newline here.
```suggestion

This is conceptually known as a "join" query.
```

content/docs/guides/opentelemetry.md
```diff
@@ -67,6 +67,31 @@ otlp:
...
+
+```promql
+rate(http_server_request_duration_seconds_count[2m])
+* on (job, instance) group_left (k8s_cluster_name)
+target_info
+```
+
+What happens in this query is that the time series resulting from `rate(http_server_request_duration_seconds_count[2m])` are augmented with the `k8s_cluster_name` label from the `target_info` series that share the same `job` and `instance` labels.
+In other words, the `job` and `instance` labels are shared between `http_server_request_duration_seconds_count` and `target_info`, akin to SQL foreign keys.
+The `k8s_cluster_name` label, OTOH, corresponds to the OTel resource attribute `k8s.cluster.name` (Prometheus converts dots to underscores).
```
jesusvazquez, line 84: Let's try to use less acronyms. I think OTel is valid due to the context.
```suggestion
The `k8s_cluster_name` label, On the other hand, corresponds to the OTel resource attribute `k8s.cluster.name` (Prometheus converts dots to underscores).
```

jesusvazquez (approved):
LGTM, left two nits.

Issue #2514, opened 2024-09-20
Docs point to version 1.8 as latest


zeroshift:
Currently accessing the docs for the Prometheus config shows docs for `1.8` by default.

juliusv:
Uh oh, that's not good. Thanks for reporting!

What I've found out so far, digging in the Ruby [repo docs download script](https://github.com/prometheus/docs/blob/main/lib/data_sources/repo_docs.rb):

The `versions` method scans the `release-*` branches and extracts the latest supposed version from each of those branches, then limiting it down to one version per major version. This ends up returning `["3.0", "2.55", "1.8"]`, which is a problem because:

* For finding the latest version in `latest_version`, we go through this array and try to find the first of the versions that has a matching tag that does not include a dash (`-`), because that would usually be an RC or beta.
* Neither 3.0 nor 2.55 have a matching tag yet (3.0 is still beta, 2.55 is just a branch to prepare a release, but isn't cut yet), so they are not selected. So that only leaves 1.8 and that gets selected as the `latest` release.

Probably the `versions` method should be changed such that it only considers branches for which there is already a matching stable release tag.

PR #2516, opened 2024-09-21
Only select branches with released stable versions for version dropdown

juliusv:
Fixes https://github.com/prometheus/docs/issues/2514

Before:

image

After:

image


juliusv:
I'm going to allow myself to already merge this before a review since it seems fairly urgent to fix this, as we are currently showing docs from 7 years ago instead of the current ones.

kakkoyun (commented):
LGTM

PR #2517, opened 2024-09-21
Only consider non-draft releases for downloads page

juliusv:
Currently the Makefile uses authenticated GitHub accesses to download a JSON containing *all* releases (since we're authenticated), including draft ones.

The script that then assembles the downloads page tries to also download the checksums for those draft releases, which fails with a 404 if you are not authenticated (and we probably don't want to show these anyway).

Interestingly the build works on Netlify anyway. Probably the GitHub API token used there doesn't actually have the rights to see draft releases on the Prometheus org. But it still produces build errors for people building the site locally with their own GitHub access token.

### DIFF
--- a/lib/helpers/download.rb
+++ b/lib/helpers/download.rb
@@ -99,7 +99,7 @@ module Downloads
 
     def self.load(dir)
       repo = JSON.parse(File.read(File.join(dir, 'repo.json')))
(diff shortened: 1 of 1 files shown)


juliusv:
@roidelapluie plz :)

PR #2520, opened 2024-09-24
Added blog for "Setting Up Prometheus Alerts with Webex Integration"

AkshayAwate:


AkshayAwate:
@RichiH would you review this PR ?

jan--f:
Hello from the bug scrub!

This looks like a more appropriate post for a Webex blog, as this assumes the reader has an account.
@AkshayAwate Any thoughts on why this should be a Prometheus blog post?

AkshayAwate:
Hi @jan--f , 

Thanks for the review!

My intention with this post was to demonstrate how Prometheus Alertmanager can be extended using webhooks for custom integrations. Webex is used as a concrete example, similar to how Slack or PagerDuty integrations are documented.

The focus of the post is on:
	•	Configuring Alertmanager receivers
	•	Using webhooks for custom notifications
	•	Grouping and formatting alerts

Webex is simply the notification medium in this example, and the webhook pattern can be adapted for other systems as well.

I’d be happy to adjust the introduction to clarify that the main goal is showcasing Prometheus and Alertmanager extensibility rather than Webex specifically.

jan--f:
Hey, we have a basic tutorial with alerts send to a webhook under https://prometheus.io/docs/tutorials/alerting_based_on_metrics/

For a blog post we typically aim for more technical depth or development insight.

Do you think there is something missing from the alerting tutorial that we should add?

AkshayAwate:
@jan--f 

Thanks for pointing that out.

I agree the basic webhook flow is already covered in the tutorial.

My blog shows specific to Webex webhooks, I think we can add Webex part in the original doc, instead of creating new one. What do you think ?

jan--f:
Iḿ afraid I don´t see what the webex part would add to Prometheus documentation. As I suggested before, this might be better of hosted in a webex blog post.
Thanks a lot for your contributions, but Iĺl have to decline this addition. If you find anything is missing from our current doc that all Prometheus users should know (as opposed to the subset that wants to send alerts to Webex) please feel free to open another PR at any time.

PR #2521, opened 2024-09-29
Fix links to java client documentation

ChrisMcD1:
I noticed some of these links were not working since the java client has moved their documentation from GitHub to their new docs site.

There are still 2 broken links that I did not fix in this PR because I could not find the corresponding new file.

https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/writing_exporters.md?plain=1#L350

and

https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/writing_exporters.md?plain=1#L387

But I figured it wouldn't hurt to make a PR for the ones that were trivial to fix!

I believe the `CONTRIBUTING.md` is asking me to ping the maintainers of the java client since the docs relate to their project, so @fstab @dhoard @tomwilkie

### DIFF
--- a/content/docs/concepts/metric_types.md
+++ b/content/docs/concepts/metric_types.md
@@ -25,7 +25,7 @@ use a counter for the number of currently running processes; instead use a gauge
 Client library usage documentation for counters:
 
    * [Go](http://godoc.org/github.com/prometheus/client_golang/prometheus#Counter)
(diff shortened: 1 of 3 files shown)


ChrisMcD1:
@RichiH This seems to fall under "everything else", so you seem appropriate to reach out to according to `MAINTAINERS.md`

PR #2522, opened 2024-09-30
Document last 10 releases, with at least 1 per major release

roidelapluie:

### DIFF
--- a/lib/data_sources/repo_docs.rb
+++ b/lib/data_sources/repo_docs.rb
@@ -17,7 +17,7 @@ class RepoDocsDataSource < ::Nanoc::DataSource
 
   DOCS_DIRECTORY = 'docs'.freeze
   BRANCH_PATTERN = 'release-*'.freeze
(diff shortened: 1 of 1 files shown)


lib/data_sources/repo_docs.rb
```diff
@@ -136,19 +136,24 @@ def sync_repository
...
-      .sort_by { |v| v.split('.').map(&:to_i) }
-      .reverse
-
-    # Number of versions is reduced to speed up site compilation time.
+            .select { |v| v.match(VERSION_REGEXP) }
+            .map { |v| v.delete_prefix('v').split('.')[0, 2].join('.') }
+            .uniq
+            .select { |v| branches.include?('release-' + v) }
+            .sort_by { |v| v.split('.').map(&:to_i) }
+            .reverse
```
juliusv, line 144: AFAICS nothing changed in these lines, maybe keep the same indent of 2 as before and everywhere else?

juliusv (approved):
Ah thanks, I was also just working on this, but your version is better :)

PR #2525, opened 2024-10-02
Fix compliance test suite link in remote write specs

karsov:
While going through the remote write specs page I noticed that the compliance test suite link is broken.
This PR should fix it.

@roidelapluie I guess :slightly_smiling_face:

### DIFF
--- a/content/docs/specs/remote_write_spec.md
+++ b/content/docs/specs/remote_write_spec.md
@@ -26,7 +26,7 @@ The remote write protocol contains opportunities for batching, e.g. sending mult
 
 The remote write protocol is not intended for use by applications to push metrics to Prometheus remote-write-compatible receivers.  It is intended that a Prometheus remote-write-compatible sender scrapes instrumented applications or exporters and sends remote write messages to a server.
 
-A test suite can be found at https://github.com/prometheus/compliance/tree/main/remote_write_sender.
+A test suite can be found at <https://github.com/prometheus/compliance/tree/main/remotewrite/sender>.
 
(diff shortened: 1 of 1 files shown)


content/docs/specs/remote_write_spec.md
```diff
@@ -26,7 +26,7 @@ The remote write protocol contains opportunities for batching, e.g. sending mult
 
 The remote write protocol is not intended for use by applications to push metrics to Prometheus remote-write-compatible receivers.  It is intended that a Prometheus remote-write-compatible sender scrapes instrumented applications or exporters and sends remote write messages to a server.
 
-A test suite can be found at https://github.com/prometheus/compliance/tree/main/remote_write_sender.
+A test suite can be found at <https://github.com/prometheus/compliance/tree/main/remotewrite/sender>.
```
alexgreenbank, line 29: Good spot, we don't need the < or > though. I would suggest:

```suggestion
A test suite can be found at https://github.com/prometheus/compliance/tree/main/remotewrite/sender.
```

alexgreenbank (changes requested):

content/docs/specs/remote_write_spec.md
karsov: Done.
I added the <> because my markdown linter complains if I don't have them around links :)

alexgreenbank: Thank you!. It's not wrong but usually it's best to follow the convention in the rest of the file.

PR #2531, opened 2024-10-14
Add documentation on how to send metrics to the OTLP endpoint

fstab:
We didn't have any documentation what protocol (`http/protobuf`) and what endpoint (`/api/v1/otlp/`) to use when sending OpenTelemetry metrics to a Prometheus server.

This PR adds the complete set of OpenTelemetry environment variables needed to send OpenTelemetry metrics to a Prometheus server.

### DIFF
--- a/content/docs/guides/opentelemetry.md
+++ b/content/docs/guides/opentelemetry.md
@@ -16,6 +16,37 @@ To enable the receiver you need to toggle the flag `--web.enable-otlp-receiver`.
 $ prometheus --enable-feature=otlp-write-receiver
 ```
 
+## Send OpenTelemetry Metrics to the Prometheus Server
+
+OpenTelemetry SDKs and instrumentation libraries can be configured via [standard environment variables](https://opentelemetry.io/docs/languages/sdk-configuration/). The following are the OpenTelemetry variables needed to send OpenTelemetry metrics to a Prometheus server on localhost:
+
+```shell
+export OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
+export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:9090/api/v1/otlp/
+```
+
+Turn off traces and logs, because the Prometheus server supports metrics only:
+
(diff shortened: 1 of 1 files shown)


content/docs/guides/opentelemetry.md
```diff
@@ -16,6 +16,37 @@ To enable the receiver you need to toggle the flag `--web.enable-otlp-receiver`.
...
+export OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
+export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:9090/api/v1/otlp/
+```
+
+Turn off traces and logs, because the Prometheus server supports metrics only:
+
+```shell
+export OTEL_TRACES_EXPORTER=none
+export OTEL_LOGS_EXPORTER=none
+```
```
ArthurSens, line 33: Should we tell people to use [OTEL_EXPORTER_OTLP_METRICS_ENDPOINT](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_metrics_endpoint) instead? That way you don't need to disable traces/logs I believe

ArthurSens (approved):
Looks great, just one question :)

jmichalek132:
Do we want to also mention other approaches for this such as:
* Configuring otel within the app to use prometheus exporter and scrape it
* Configure otel collector to send metrics via OTLP or remote write into prometheus

Do we also want to get into the copying of resources attributes / target_info / dots translation or is the goal to keep it short?

fstab:
> Do we want to also mention other approaches for this such as:
> 
>     * Configuring otel within the app to use prometheus exporter and scrape it
> 
>     * Configure otel collector to send metrics via OTLP or remote write into prometheus
> 
> 
> Do we also want to get into the copying of resources attributes / target_info / dots translation or is the goal to keep it short?

I think that would be nice, but let's better do this in a different PR so that we don't block this one for too long.

jmichalek132 (approved):

PR #2539, opened 2024-11-04
Add native histogram specification

beorn7:
Here is finally the opus magnum that was in the making for far too long. Partially addressing https://github.com/prometheus/prometheus/issues/11561 .

It is partially more documentation than specification, but I assume the specification section is still the least bad place for it.

This is intentionally not yet complete. TODOs mark the missing parts, most of which cannot be resolved at the moment because implementation work is still missing and/or decisions still have to be made.

The intention here is to finally get something out here, even if it is pretty much still a "living document". It will probably take us another year or so to resolve all TODOs.

Dear reviewers, primary goal of the review is to detect what I got wrong here and what I have forgotten. Language improvements are of course welcome, but of secondary importance. And as said above, the TODOs should mostly stay for this first version. Of course, if you have an easy way of resolving one, be my guest.

Tagging potential reviewers here (simply focus on your areas of expertise):
@bwplotka @juliusv @krajorama @zenador @fatsheep9146 @NeerajGartia21 @Maniktherana @fpetkovski @vesari @carrieedwards @fionaliao @codesome @ArthurSens @csmarchbanks @fstab

### DIFF
--- a/content/docs/specs/native_histograms.md
+++ b/content/docs/specs/native_histograms.md
@@ -1,0 +1,2353 @@
+---
+title: Native Histograms [EXPERIMENTAL]
+sort_rank: 6
+---
+
+# Native Histograms
+
+Native histograms were introduced as an experimental feature in November 2022.
+They are a concept that touches almost every part of the Prometheus stack. The
+first version of the Prometheus server supporting native histograms was
+v2.40.0. The support had to be enabled via a feature flag
+`--enable-feature=native-histograms`. (TODO: This is still the case with the
+current release v2.55 and v3.00. Update this section with the stable release,
+once it has happened.)
+
+Due to the pervasive nature of the native-histogram related changes, the
+documentation of those changes and explanation of the underlying concepts are
+widely distributed over various channels (like the documentation of affected
+Prometheus components, doc comments in source code, sometimes the source code
+itself, design docs, conference talks, …). This document intends to gather all
+these pieces of information and present them concisely in a unified context.
+This document prefers to link existing detailed documentation rather than
+restating it, but it contains enough information to be comprehensible without
+referring to other sources. With all that said, it should be noted that this
+document is neither suitable as an introduction for beginners nor does it focus
+on the needs of developers. For the former, the plan is to provide an updated
+version of the [Best Practices article on histograms and
+summaries](../../practices/histograms/). (TODO: And a blog post or maybe even a
+series of them.) For the latter, there is Carrie
+Edward's [Developer’s Guide to Prometheus Native
+Histograms](https://docs.google.com/document/d/1VhtB_cGnuO2q_zqEMgtoaLDvJ_kFSXRXoE0Wo74JlSY/edit).
+
+While formal specifications are supposed to happen in their respective context
+(e.g. OpenMetrics changes will be specified in the general OpenMetrics
+specification), some parts of this document take the shape of a specification.
+In those parts, the key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL
+NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” are used as
+described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+This document still contains a lot of TODOs. In most cases, they are not just
+referring to incompleteness of this doc but more importantly to incomplete
+implementation or open questions. For now, this is essentially a living
+document that will receive updates as implementations and specifications catch
+up.
+
+## Introduction
+
+The core idea of native histograms is to treat histograms as first class
+citizens in the Prometheus data model. Elevating histograms to a “native”
+sample type is the fundamental prerequisite for the key properties listed
+below, which explains the choice of the name _native histograms_.
+
+Prior to the introduction of native histograms, all Prometheus sample values
+have been 64-bit floating point values (short _float64_ or just _float_). These
+floats can directly represent _gauges_ or _counters_. The Prometheus metric
+types _summary_ and (the classic version of) _histogram_, as they exist in
+exposition formats, are broken down into float components upon ingestion: A
+_sum_ and a _count_ component for both types, a number of _quantile_ samples
+for a summary and a number of _bucket_ samples for a (classic) histogram.
+
+With native histograms, a new structured sample type is introduced. A single
+sample represents the previously known _sum_ and _count_ plus a dynamic set of
+buckets.
+
+Native histograms have the following key properties:
+1. A sparse bucket representation, allowing (near) zero cost for empty buckets.
+2. Coverage of the full float64 range of values.
+3. No configuration of bucket boundaries during instrumentation.
+4. Dynamic resolution picked according to simple configuration parameters.
+5. Sophisticated exponential bucketing schemas, ensuring mergeability between
+   all histograms using those schemas.
+6. An efficient data representation for both exposition and storage.
+
+Compared to the previously existing “classic” histograms, native histograms
+allow a higher bucket resolution across arbitrary ranges of observed values at
+a lower storage and query cost with very little to no configuration required.
+Even partitioning histograms by labels is now much more affordable.
+
+Because the sparse representation (property 1 in the list above) is so crucial
+for many of the other benefits of native histograms, _sparse histograms_ was
+a common name for _native histograms_ early during the design process. However,
+other key properties like the exponential bucketing schema or the dynamic
+nature of the buckets are also very important, but not caught at all in the
+term _sparse histograms_.
+
+### Design docs
+
+These are the design docs that guided the development of native histograms.
+Some details are obsolete now, but they describe rather well the underlying
+concepts and how they evolved.
+
+- [Sparse high-resolution histograms for
+  Prometheus](https://docs.google.com/document/d/1cLNv3aufPZb3fNfaJgdaRBZsInZKKIHo9E6HinJVbpM/edit),
+  the original design doc.
+- [Prometheus Sparse Histograms and
+  PromQL](https://docs.google.com/document/d/1ch6ru8GKg03N02jRjYriurt-CZqUVY09evPg6yKTA1s/edit),
+  more an exploratory document than a proper design doc about the handling of
+  native histograms in PromQL.
+
+### Conference talks
+
+A more approachable way of learning about native histograms is to watch
+conference talks, of which a selection is presented below. As an introduction,
+it might make sense to watch these talks and then return to this document to
+learn about all the details and technicalities.
+
+- [Secret History of Prometheus
+  Histograms](https://fosdem.org/2020/schedule/event/histograms/) about the
+  classic histograms and why Prometheus kept them for so long.
+- [Prometheus Histograms – Past, Present, and
+  Future](https://promcon.io/2019-munich/talks/prometheus-histograms-past-present-and-future/)
+  is the inaugural talk about the new approach that lead to native histograms.
+- [Better Histograms for
+  Prometheus](https://www.youtube.com/watch?v=HG7uzON-IDM) explains why the
+  concepts work out in practice.
+- [Native Histograms in
+  Prometheus](https://promcon.io/2022-munich/talks/native-histograms-in-prometheus/)
+  presents and explains native histograms after the actual implementation.
+- [PromQL for Native
+  Histograms](https://promcon.io/2022-munich/talks/promql-for-native-histograms/)
+  explains the usage of native histograms in PromQL.
+- [Prometheus Native Histograms in
+  Production](https://www.youtube.com/watch?v=TgINvIK9SYc) provides an analysis
+  of performance and resource consumption.
+- [Using OpenTelemetry’s Exponential Histograms in
+  Prometheus](https://www.youtube.com/watch?v=W2_TpDcess8) covers the
+  interoperability with the OpenTelemetry.
+
+## Glossary
+
+- A __native histogram__ is the new complex sample type representing a full
+  histogram that this document is about. Where the context is sufficiently
+  clear, it is often just called a _histogram_ below.
+- A __classic histogram__ is the older sample type representing a histogram
+  with fixed buckets, formerly just called a _histogram_. It exists as such in
+  the exposition formats, but is broken into a number of float samples upon
+  ingestion into Prometheus.
+- __Sparse histogram__ is an older, now deprecated name for _native
+  histogram_. This name might still be found occasionally in older
+  documentation. __Sparse buckets__ remains a meaningful term for the buckets
+  of a native histogram.
+
+## Data model
+
+This section describes the data model of native histograms in general. It
+avoids implementation specifics as far as possible. This includes terminology.
+For example, a _list_ described in this section will become a _repeated
+message_ in a protobuf implementation and (most likely) a _slice_ in a Go
+implementation.
+
+### General structure
+
+Similar to a classic histogram, a native histogram has a field for the _count_
+of observations and a field for the _sum_ of observations. In addition, it
+contains the following components, which are described in detail in dedicated
+sections below:
+
+- A _schema_ to identify the method of determining the boundaries of any given
+  bucket with an index _i_.
+- A sparse representation of indexed buckets, mirrored for positive and
+  negative observations.
+- A _zero bucket_ to count observations close to zero.
+- A (possibly empty) list of _custom values_.
+- _Exemplars_.
+
+### Flavors
+
+Any native histogram has a specific flavor along each of two independent
+dimensions:
+
+1. Counter vs. gauge: Usually, a histogram is “counter like”, i.e. each of its
+   buckets acts as a counter of observations. However, there are also “gauge
+   like” histograms where each bucket is a gauge, representing arbitrary
+   distributions at a point in time. The concept of a gauge histogram was
+   previously introduced for classic histograms by
+   [OpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#gaugehistogram).
+2. Integer vs. floating point (short: float): The obvious use case of
+   histograms is to count observations, resulting in integer numbers of
+   observations ≥ 0 within each bucket, including the _zero bucket_, and for
+   the total _count_ of observations, represented as unsigned 64-bit integers
+   (short: uint64). However, there are specific use cases leading to a
+   “weighted” or “scaled” histogram, where all of these values are represented
+   as 64-bit floating point numbers (short: float64). Note that the _sum_ of
+   observations is a float64 in either case.
+
+Float histograms are occasionally used in direct instrumentation for “weighted”
+observations, for example to count the number of seconds an observed value was
+falling into different buckets of a histogram. The far more common use case for
+float histograms is within PromQL, though. PromQL generally only acts on float
+values, so the PromQL engine converts every histogram retrieved from the TSDB
+to a float histogram first, and any histogram stored back into TSDB via
+recording rules is a float histogram. If such a histogram is effectively an
+integer histogram (because the value of all non-_sum_ fields can be represented
+precisely as uint64), a TSDB implementation MAY convert them back to integer
+histograms to increase storage efficiency. (As of Prometheus v3.00, the TSDB
+implementation within Prometheus is not utilizing this option.) Note, however,
+that the most common PromQL function applied to a counter histogram is `rate`,
+which generally produces non-integer numbers, so that results of recording
+rules will commonly be float histograms with non-integer values anyway.
+
+Treating native histograms explicitly as integer histograms vs. float histogram
+is a notable deviation from the treatment of conventional simple numeric
+samples, which are always treated as floats throughout the whole stack for the
+sake of simplicity.
+
+The main reason for the more involved treatment of histograms is the easy
+efficiency gains in protobuf-based exposition formats. Protobuf uses varint
+encoding for integers, which reduces the data size for small integer values
+without requiring an additional compression layer. This benefit is amplified by
+the [delta encoding of integer buckets](#buckets), which generally results in
+smaller integer values. Floats, in contrast, always require 8 bytes in
+protobuf. In practice, many integers in an integer histogram will fit in 1
+byte, and most will fit in 2 bytes, so that the explicit presence of integer
+histogram in a protobuf-exposition format results directly in a data size
+reduction approaching 8x for histograms with many buckets. This is particularly
+relevant as the overwhelming majority of histograms exposed by instrumented
+targets are integer histograms.
+
+For similar reasons, the representation of integer histograms in RAM and on
+disk is generally more efficient than that of float histograms. This is less
+relevant than the benefits in the exposition format, though. For one,
+Prometheus uses Gorilla-style XOR encoding for floats, which reduces their
+size, albeit not as much as the double-delta encoding used for integers. More
+importantly, an implementation could always decide to internally use an integer
+representation for histogram fields that are effectively integer values (see
+above). (Historical note: Prometheus v1 used exactly this approach to improve
+the compression of float samples, and Prometheus v3 might very well adopt this
+approach again in the future.)
+
+In a counter histogram, the total _count_ of observation and the counts in the
+buckets individually behave like Prometheus counters, i.e. they only go down
+upon a counter reset. However, the _sum_ of observation may decrease as a
+consequence of the observation of negative values. PromQL implementations MUST
+detect counter resets based on the whole histogram (see the [counter reset
+considerations section](#counter-reset-considerations) below for details).
+(Note that this always has been a problem for the _sum_ component of classic
+histograms and summaries, too. The approach so far was to accept that counter
+reset detection silently breaks for _sum_ in those cases. Fortunately, negative
+observations are a very rare use case for Prometheus histograms and summaries.)
+
+### Schema
+
+The _schema_ is a signed integer value with a size of 8 bits (short: int8). It
+defines the way bucket boundaries are calculated. The currently valid values
+are -53 and the range between and including -4 and +8. More schemas may be
+added in the future. -53 is a schema for so-called _custom bucket boundaries_
+or short _custom buckets_, while the other schema numbers represent the
+different standard exponential schemas (short: _standard schemas_).
+
+The standard schemas are mergeable with each other and are RECOMMENDED for
+general use cases. Larger schema numbers correspond to higher resolutions.
+Schema _n_ has half the resolution of schema _n_+1, which implies that a
+histogram with schema _n_+1 can be converted into a histogram with schema _n_
+by merging neighboring buckets.
+
+For any standard schema _n_, the boundaries of a bucket with index _i_
+calculated as follows (using Python syntax):
+
+- The upper inclusive limit of a positive bucket:
+  `(2**2**-n)**i`
+- The lower exclusive limit of a positive bucket:
+  `(2**2**-n)**(i-1)`
+- The lower inclusive limit of a negative bucket:
+  `-((2**2**-n)**i)`
+- The upper exclusive limit of a negative bucket:
+  `-((2**2**-n)**(i-1))`
+
+_i_ is an integer number that may be negative.
+
+There are exceptions to the rules above concerning the largest and smallest
+finite values representable as a float64 (called `MaxFloat64` and `MinFloat64`
+in the following) and the positive and negative infinity values (`+Inf` and
+`-Inf`):
+
+- The positive bucket that contains `MaxFloat64` (according to the boundary
+  formulas above) has an upper inclusive limit of `MaxFloat64` (rather than the
+  limit calculated by the formulas above, which would overflow float64).
+- The next positive bucket (index _i_+1 relative to the bucket from the
+  previous item) has a lower exclusive limit of `MaxFloat64` and an upper
+  inclusive limit of `+Inf`. (It could be called a _positive overflow bucket_.)
+- The negative bucket that contains `MinFloat64` (according to the boundary
+  formulas above) has a lower inclusive limit of `MinFloat64` (rather than the
+  limit calculated by the formulas above, which would underflow float64).
+- The next negative bucket (index _i_+1 relative to the bucket from the
+  previous item) has an upper exclusive limit of `MinFloat64` and an lower
+  inclusive limit of `-Inf`. (It could be called a _negative overflow bucket_.)
+- Buckets beyond the `+Inf` and `-Inf` buckets described above MUST NOT be used.
+  
+There are more exceptions for values close to zero, see the [zero bucket
+section](#zero-bucket) below.
+
+The current limits of -4 for the lowest resolution and 8 for the highest
+resolution have been chosen based on practical usefulness. Should a practical
+need arise for even lower or higher resolution, an extension of the range will
+be considered. However, a schema greater than 52 does not make sense as the
+growth factor from one bucket to the next would then be smaller than the
+difference between representable float64 numbers. Likewise, a schema smaller
+than -9 does not make sense either, as the growth factor would then exceed the
+largest float representable as float64. Therefore, the schema numbers between
+(and including) -9 and +52 are reserved for future standard schemas (following
+the formulas for bucket boundaries above) and MUST NOT be used for any other
+schemas.
+
+For schema -53, the bucket boundaries are set explicitly via _custom values_,
+described in detail in the [custom values section](#custom-values) below. This
+results in a native histogram with custom bucket boundaries (or short _custom
+buckets_, often further abbreviated to NHCB). Such a histogram can be used to
+represent a classic histogram as a native histogram. It can also be used if
+the exponential bucketing featured by the standard schemas is a bad match for
+the distribution to be represented by the histogram. Histograms with different
+custom bucket boundaries are generally not mergeable with each other.
+Therefore, schema -53 SHOULD only be used as an informed decision in specific
+use cases. (TODO: NHCB aren't fully merged into main as of now (2024-11-03).
+They are worked into this document as far as possible already. This information
+might not yet be relevant for released Prometheus versions.)
+
+### Buckets
+
+For standard schemas, buckets are represented as two lists, one for positive
+buckets and one for negative buckets. For custom buckets (schema -53), only the
+positive bucket list is used, but repurposed for all buckets.
+
+Any unpopulated buckets MAY be excluded from the lists. (Which is the reason
+why the buckets are often called _sparse buckets_.)
+
+For float histograms, the elements of the lists are float64 and
+represent the bucket population directly.
+
+For integer histograms, the elements of the lists are signed 64-bit integers
+(short: int64), and each element represents the bucket population as a delta to
+the previous bucket in the list. The first bucket in each list contains an
+absolute population (which can also be seen as a delta relative to zero).
+
+To map buckets in the lists to the indices as defined in the previous section,
+there are two lists of so-called _spans_, one for the positive buckets and one
+for the negative buckets.
+
+Each span consists of a pair of numbers, a signed 32-bit integer (short: int32)
+called _offset_ and an unsigned 32-bit integer (short: uint32) called _length_.
+Only the first span in each list can have a negative offset. It defines the
+index of the first bucket in its corresponding bucket list. The length defines
+the number of consecutive buckets the bucket list starts with. The offsets of
+the following spans define the number of excluded (and thus unpopulated
+buckets). The lengths define the number of consecutive buckets in the list
+following the excluded buckets.
+
+The sum of all length values in each span list MUST be equal to the length of
+the corresponding bucket list.
+
+Empty spans (with a length of zero) are valid and MAY be used, although they
+are generally not useful and they SHOULD be eliminated by adding their offset
+to the offset of the following span. Similarly, spans that are not the first
+span in a list MAY have an offset of zero, although those offsets SHOULD be
+eliminated by adding their length to the previous span. Both cases are allowed
+so that producers of native histograms MAY pick whatever representation has the
+best resource trade-offs at that moment. For example, if a histogram is
+processed through various stages, it might be most efficient to only eliminate
+redundant spans after the last processing stage.
+
+In a similar spirit, there are situation where excluding every unpopulated
+bucket from the bucket list is most efficient, but in other situations, it
+might be better to reduce the number of spans by representing small numbers of
+unpopulated buckets explicitly.
+
+Note that future high resolution schemas might require offsets that are too
+large to be represented with an int32. An extension of the data model will be
+required in that case. (The current standard schema with the highest resolution
+is schema 8, for which the bucket that contains `MaxFloat64` has index 262144,
+and thus the `+Inf` overflow bucket has index 262145, while the largest number
+representable with int32 is 2147483647. The highest standard schema that would
+still work with int32 offsets would be schema 20, corresponding to a growth
+factor from bucket to bucket of only ~1.000000661.)
+
+#### Examples
+
+An integer histogram has the following positive buckets (index→population):
+
+`-2→3, -1→5, 0→0, 1→0, 2→1, 3→0, 4→3, 5→2`
+
+They could be represented in this way:
+
+- Positive bucket list: `[3, 2, -4, 2, -1]`
+- Positive span list: `[[-2, 2], [2,1], [1,2]]`
+
+The second and third span could be merged into one if the single unpopulated
+bucket with index 3 is represented explicitly, leading to the following result:
+
+- Positive bucket list: `[3, 2, -4, -1, 3, -1]`
+- Positive span list: `[[-2, 2], [2,4]]`
+
+Or merge all the spans into one by representing all unpopulated buckets above
+explicitly:
+
+- Positive bucket list: `[3, 2, -5, 0, 1, -1, 3, -1]`
+- Positive span list: `[[-2, 8]]`
+
+### Zero bucket
+
+Observations of exactly zero do not fit into any bucket as defined by the
+standard schemas above. They are counted in a dedicated bucket called the _zero
+bucket_.
+
+The number of observations in the zero bucket is tracked by a single uint64
+(for integer histograms) or float64 (for float histograms).
+
+The zero bucket has an additional parameter called the _zero threshold_, which
+is a float64 ≥ 0. If the threshold is set to zero, only observations of exactly
+zero go into the zero bucket, which is the case described above. If the
+threshold has a positive value, all observations within the closed interval
+[-threshold, +threshold] go to the zero bucket rather than a regular bucket.
+This has two use cases:
+
+- Noisy observations close to zero tend to populate a high number of buckets.
+  Those observations might happen due to numerical inaccuracies or if the
+  source of the observations are actual physical measurements. A zero bucket
+  with a relatively small threshold redirects those observations into a single
+  bucket.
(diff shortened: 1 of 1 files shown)


beorn7:
https://deploy-preview-2539--prometheus-docs.netlify.app/docs/specs/native_histograms/ is the preview, very convenient to see the rendered layout.

krajorama:
cc @jhesketh  @charleskorn have also dealt with native histograms due to their query engine work.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+the previous bucket in the list. The first bucket in each list contains an
+absolute population (which can also be seen as a delta relative to zero).
+
+To map buckets in the lists to the indices as defined in the previous section,
+there are two lists of so-called _spans_, one for the positive buckets and one
+for the negative buckets.
+
+Each span consists of a pair of numbers, a signed 32-bit integer (short: int32)
+called _offset_ and an unsigned 32-bit integer (short: uint32) called _length_.
+Only the first span in each list can have a negative offset. It defines the
```
krajorama, line 340: For NHCB not even the first offset can be negative as the offset is interpreted as an offset into the custom values.

krajorama (commented):
Couple of comments. Haven't gone through yet even once, on line 738 Instrumentation libraries.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+
+This is the corresponding query for a native histograms:
+```
+histogram_quantile(0.9, sum by (job) (rate(http_request_duration_seconds[10m])))
+```
+
+As with classic histograms, an estimation of the maximum and minimum
+observation in a histogram can be performed using 1 and 0, respectively, as the
+first parameter of `histogram_quantile`. This estimation is much more useful
+with native histograms, though, not only because of the usually higher
```
krajorama, line 1879: native histograms with standard schemas

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+- `histogram_sum()`
+- `histogram_stddev()`
+- `histogram_stdvar()`
+
+All these functions silently ignore float samples as input. Each function
+returns a vector of float samples.
+
+`histogram_count()` and `histogram_sum()` return the count of observations or the sum of
+observations, respectively, that are contained in a native histogram. As they are normal
+function, their result cannot be used in a range selector. Instead of using
```
krajorama, line 1903:
```suggestion
functions, their result cannot be used in a range selector. Instead of using
```

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+ability to handle conflicting metadata between different targets). There are
+plans, though, to improve metadata handling in Prometheus in general. Those
+efforts will also take into account how to support native histograms properly.
+(TODO: Update as progress is made.)
+
+## Prometheus UI
+
+This section describes the rendering of histograms by Prometheus's own UI. This
+MAY be used as a guideline for external graphing frontends (like Perses or
+Grafana).
```
krajorama, line 2123:
```suggestion
MAY be used as a guideline for third party graphing frontends.
```

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+remote write (by setting the `send_native_histograms` remote write config
+setting to true).
+
+In [remote write v2](https://prometheus.io/docs/specs/remote_write_spec_2_0/),
+native histograms are a stable feature.
+
+TODO: Is there more to say? I'm not a remote write & read specialist. Does
+remote read really work?
+
+TODO: What are the plans to auto-convert classic histograms to NHCBs in
```
krajorama, line 2196: No plan at the moment because we cannot guarantee that the classic histogram came from scrape where we for sure know we have all the buckets present. This is true for both receiving and sending remote write. In fact one of the reason for having NHCB is to avoid scenarios where classic histograms are not consistent (missing/late due to independent series). Therefor currently the only plans are to have NHCB when we know we have all data : scrape - being implemented, receiving OTEL histogram - issue there, expose from instrumentation directly.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+for native histograms as an experimental feature. Receivers not capable of
+processing native histograms will simply ignore the newly added fields.
+Nevertheless, Prometheus has to be configured to send native histograms via
+remote write (by setting the `send_native_histograms` remote write config
+setting to true).
+
+In [remote write v2](https://prometheus.io/docs/specs/remote_write_spec_2_0/),
+native histograms are a stable feature.
+
+TODO: Is there more to say? I'm not a remote write & read specialist. Does
```
krajorama, line 2193: Remote read should work , either it sends chunks or converts the result: https://github.com/prometheus/prometheus/blob/fc3d19725d83e13363ce40d56672c6f97e2955e7/storage/remote/codec.go#L155

Remove write works for sure, it's in use.

The one TODO I think is to verify/fix the situation when exemplars of a native histogram are sent in two requests instead of one - we might consider the first batch out-of-order. But it has been a while since I worked on this so not 100% sure it's a problem.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+## OTLP
+
+The OTLP receiver built into Prometheus will convert incoming OTel exponential
+histograms to Prometheus native histograms utilizing the compatibility
+described [above](#opentelemetry-interoperability). The resolution of a
+histogram using a schema (“scale” in OTel lingo) greater than 8 will be reduced
+to match schema 8. (In the unlikely case that a schema smaller than -4 is used,
+the ingestion will fail.)
+
+TODO: Is the OTLP receiver documented anywhere? Link the documentation here.
```
krajorama, line 2218: + TODO convert to NHCB explicit bucket histograms. https://github.com/prometheus/prometheus/issues/15022

krajorama: @aknuds1  about documentation

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+instrumentation the same but convert classic histograms to NHCBs upon
+ingestion. This leverages the increased storage performance of native
+histograms, but still requires to address (1) and (2) in the same way as for a
+full migration to native histograms (see next paragraphs).
+
+The conservative way of addressing (1) and (2) is to allow a long transition
+period, which comes at the cost of collecting and storing classic and native
+histograms in parallel for a while.
+
+The first step is to update the instrumentation to expose classic and native
```
krajorama, line 2320: or start scraping NHCB on top of classic histograms

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+moment. While that is in principle possible, it is tricky. If feasible, the
+transition period during which classic and native histograms are collected in
+parallel, can be quite long to minimize the necessity to implement tricky
+switch-overs. For example, once classic and native histograms have been
+collected in parallel for a month, any dashboard not looking farther back than
+a month can simply be switched over from a classic histogram query to a native
+histogram query without any consideration about the right switch-over.
+
+Once there is confidence that all queries have been migrated correctly,
+configure Prometheus to only scrape native histograms (which is the “normal”
```
krajorama, line 2347: or start dropping some metrics with relabel rules to make this incremental

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+setting). If everything still works, it is time to remove classic histograms
+from the instrumentation.
+
+The Grafana Mimir documentation contains [a detailed migration
```
krajorama, line 2351: thanks!

krajorama (commented):
done with first read through

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+content negotiation to prefer the classic protobuf-based exposition format over
+the OpenMetrics text format. (TODO: This behavior will change once native
+histograms are a stable feature.)
+
+### Finetuning content negotiation
+
+With Prometheus v2.49 and later, it is possible to finetune the scrape protocol
+negotiation globally or per scrape config via the `scrape_protocols` config
+setting. It is a list defining the content negotiation priorities. Its default
+value depends on the `--enable-feature=native-histograms` flag. If the flage is
```
fionaliao, line 930:
```suggestion
value depends on the `--enable-feature=native-histograms` flag. If the flag is
```

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+
+To enable the Prometheus server to scrape native histograms, the feature flag
+`--enable-feature=native-histograms` is required. This flag also changes the
+content negotiation to prefer the classic protobuf-based exposition format over
+the OpenMetrics text format. (TODO: This behavior will change once native
+histograms are a stable feature.)
+
+### Finetuning content negotiation
+
+With Prometheus v2.49 and later, it is possible to finetune the scrape protocol
```
fionaliao, line 927:
```suggestion
With Prometheus v2.49 and later, it is possible to fine-tune the scrape protocol
```

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+negotiation globally or per scrape config via the `scrape_protocols` config
+setting. It is a list defining the content negotiation priorities. Its default
+value depends on the `--enable-feature=native-histograms` flag. If the flage is
+set, it is `[ PrometheusProto, OpenMetricsText1.0.0, OpenMetricsText0.0.1,
+PrometheusText0.0.4 ]`, otherwise `[ OpenMetricsText1.0.0,
+OpenMetricsText0.0.1, PrometheusText0.0.4 ]`. This results in the behavior
```
fionaliao, line 933: Should the diff between the lists be explicitly pointed out - i.e. point out the difference is `PrometheusProto`. Just so you don't need to jump between the lists to figure out the change.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+parentheses):
+
+- A change of sample type between integer histogram and float histogram
+  (because both require different chunk encodings altogether).
+- A change of sample type between gauge histogram and counter histogram
+  (because the leading byte has to denote the different type).
+- A counter reset for a counter histogram (to be stored in the leading byte as
+  counter reset information, see details below).
+- A schema change (which means we need a new chunk layout, and one and the same
+  chunk can only have one chunk layout).
```
fionaliao, line 1223: (Opinion) I think this reads better:

```suggestion
- A schema change (which means we need a new chunk layout, as a 
  chunk can only have one chunk layout).
```

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+### Staleness markers
+
+Staleness markers used to be simply a special float value among the many that
+can be used to represent the `NaN` value. The same float value could be used to
+mark a histogram time series as stale. However, this would require cutting a
+new chunk, just for the purpose of marking the series as stale, because a float
+value following a histogram value has to be stored in a different chunk (see
+above). Therefore, there is also a histogram version of a stale marker where
+the field for the sum of observations is set to the special stale `NaN` value.
+In this case, all other fields are ignored, so they can be set to values
```
fionaliao, line 1257: Having NaN observations could also result in the sum of observations being NaN according to
https://github.com/prometheus/docs/blob/2d3624ced474a0eeb70ec9a63d1d4ef7c96283a3/content/docs/specs/native_histograms.md?plain=1#L492-L496

How are the two cases (staleness markers vs NaN observations) distinguished? For the NaN observations case we wouldn't want to ignore the other fields.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+The “trim” operators `>/` and `</` were introduced specifically for native
+histograms. They only work for a histogram on the left hand side and a float
+sample or a scalar on the right hand side. (They do not work for float samples
+or scalars on _both_ sides. An info-level annotation is returned in this case.)
+These operators remove observations from the histogram that are greater or
+smaller than the float value on the right side, respectively, and return the
+resulting histogram. The removal is only precise if the threshold coincides
+with a bucket boundary. Otherwise, interpolation within the affected buckets
+has to be used, as described [above](#interpolation-within-a-bucket). The
+counter vs. gauge flavor of the histogram is preserved. (TODO: These operaters
```
fionaliao, line 1779:
```suggestion
counter vs. gauge flavor of the histogram is preserved. (TODO: These operators
```

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+
+- The timestamp, encoded as varbit-int, with an absolute value in the 1st
+  sample, a delta between the 1st and 2nd sample for the 2nd sample, and a
+  “delta of deltas” for any further samples (i.e. the same “double delta”
+  encoding used for timestamps in conventional float chunks, just with a
+  different bit bucketing for the varbit-int encoding).
+- The count of observations, encoded as varbit-uint for the 1st sample and as
+  varbit-int for any further samples, using the same “delta of deltas” approach
+  as for timestamps.
+- The zero bucket population, enccoded as varbit-uint for the 1st sample and as
```
carrieedwards, line 1188:
```suggestion
- The zero bucket population, encoded as varbit-uint for the 1st sample and as
```

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+creating a histogram, it is considered a counter reset and the detection
+procedure is concluded.
+
+If both conditions are met, the first histogram has to be converted so that its
+schema and zero bucket width matches those of the second histogram. This
+happens in the same way as [previously described](#limit-the-bucket-count):
+Neighboring buckets are merged to reduce the schema, and regular buckets are
+merged with the zero bucket to increase the width of the zero bucket.
+
+At this point in the procedure, both histograms have the schema and zero bucket
```
carrieedwards, line 1326:
```suggestion
At this point in the procedure, both histograms have the same schema and zero bucket
```

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+- For the 1st histogram in a counter histogram chunk, the counter reset
+  information from the chunk is generally used as the `CounterResetHint`
+  (either `CounterReset`, `NotCounterReset`, or `UnknownCounterReset`).
+  However, overlapping blocks and out-of-order ingestion can break this rule,
+  see below.
+  
+TODO: Currently, Prometheus returns a `CounterResetHint` of
+`UnknownCounterReset` for the 1st histogram of a chunk marked as
+`NotCounterReset` (which is essentially situation (4) from above). Once [this
+issue](https://github.com/prometheus/prometheus/issues/15247) is resolves, we
```
carrieedwards, line 1435:
```suggestion
issue](https://github.com/prometheus/prometheus/issues/15247) is resolved, we
```

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+
+Cutting a new chunk after a counter reset aims to improve the compression
+ratio. A counter reset sets all bucket populations to zero, so there are fewer
+buckets to represent. A chunk, however, has to represent the superset of all
+buckets of all histograms in the chunk, so cutting a new chunk enables a
+simpler set of buckets for the new chunk.
+
+This in turn implies that there will never be a counter reset within a chunk.
+The only histogram for which counter reset information has to be persisted is
+therefore the 1st histogram in a chunk. This happens in the so-called
```
carrieedwards, line 1358: I think this phrasing is a little bit confusing. Maybe something like this?

```suggestion
This in turn implies that there will never be a counter reset after the first sample in a chunk. Therefore, the only counter reset information that has to be persisted is is that of the first histogram in the chunk. This happens in the so-called
```

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+
+The sum of observations is calculated as usual by adding the observation to the
+sum of observations, following normal floating point arithmetic. (For example,
+an observation of `NaN` will set the sum to `NaN`. An observation of `+Inf`
+will set the sum to `+Inf`, unless it is already `NaN` or `-Inf`, in which case
+the sum is set to `NaN`.)
+
+An observation of `NaN` goes into no bucket, but increments the count of
+observations. This implies that the count of observations can be greater than
+the sum of all buckets, and the difference is the number of `NaN` observations.
```
carrieedwards, line 500: Should it mention that for histograms that haven't had any NaN observations/have a sum of NaN, the count has to be equal to the sum of overall counts in the negative buckets, positive buckets, and zero count?

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+Prom has reserved more schema numbers to extend its range, should it ever by
+required in practice.
+
+The index is offset by one, i.e. a Prom bucket with index _n_ has index _n-1_
+for OTel.
+
+OTel has a dense rather than a sparse representation of buckets. One might see
+OTel as “Prom with only one span”.
+
+The Prom _zero bucket_ is called _zero count_ for OTel, but works the same,
```
carrieedwards, line 530: I believe the attribute is called ZeroCount in both the Histogram/Float Histogram struct in Prometheus and the OTel Exponential Histogram

beorn7:
Thanks everyone for the review so far.

In the commit I have just pushed I have addressed all the feedback and resolved the conversation where I assumed that my response is uncontroversial. The conversations still open might require more discussion.

I have _not yet_ reworded the counter reset handling according to our recent discussions. I will do that tomorrow with a refreshed brain.

beorn7:
OK, counter reset handling should be up to date now.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+in the TSDB: `chunkenc.EncHistogram` (string representation `histogram`,
+numerical value 2) for integer histograms, and `chunkenc.EncFloatHistogram`
+(string representation `floathistogram`, numerical value 3) for float
+histograms.
+
+Similarly, there are two new record types for the WAL and the in-memory
+snapshot (Go type `record.Type`): `record.HistogramSamples` (string
+representation `histogram_samples`, numerical value 7) for integer histograms,
+and `record.FloatHistogramSamples` (string representation
+`float_histogram_samples`, numerical value 8) for float histograms.
```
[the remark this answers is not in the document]
krajorama: https://github.com/prometheus/prometheus/pull/15467

content/docs/specs/native_histograms.md
krajorama: I would guess count==0 is the distinguishing feature, so not all fields are ignored

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+
+The most common case is a counter histogram being followed by another counter
+histogram. In this case, a possible counter reset is detected by the following
+procedure:
+
+If the two histograms differ in schema or in the zero bucket width, these
+changes could be part of a compatible resolution reduction (which happens
+regularly to [reduce the bucket count of a
+histogram](#limit-the-bucket-count)). Both of the following is true for a
+compatible resolution reduction:
```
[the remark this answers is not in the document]
krajorama: on second read it's ok to have here, but I don't see it in the code

content/docs/specs/native_histograms.md
bboreham: There are lots of NaN values; we pick [one of them](https://github.com/prometheus/prometheus/blob/cd1f8ac129a289be8e7d98b6de57a9ba5814c406/model/value/value.go#L24-L28) to mean "stale".
It is not expected that the same value can be generated by accident.

beorn7:
Note: Apparently, I forgot to send all the 52 pending replies I wrote several weeks ago.
They are submitted now, so now it makes way more sense where I resolved a conversation and where not.

content/docs/specs/native_histograms.md
beorn7: The count value shouldn't matter. If the sum is the specific bit pattern that defines a staleness marker, the whole histogram must be treated as a staleness marker.

content/docs/specs/native_histograms.md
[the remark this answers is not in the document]
beorn7: It's kind of distributed in various parts of the appender (and the "ok to append" checks) in the tsdb code. I also believe that the current ingestion code doesn't cover all edge cases. I have noted down a TODO on my list to verify and fix that. (The `DetectReset` method MUST result in the exact same counter reset detection as the one we do on the fly during ingestion. The reason why we don't just call `DetectReset` on each ingested histogram is performance.)

content/docs/specs/native_histograms.md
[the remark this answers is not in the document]
beorn7: Thank you. I added the new record types.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+
+NOTE: Switching the used exposition format between text-based and
+protobuf-based might have surprising effects on the formatting of label values
+for `quantile` labels (used in summaries) and `le` labels (used in classic
+histograms). This is not a problem for native histograms, but might show up in
+the same context because enabling native histograms requires the protobuf
+exposition format. See details in the [documentation for the
+`native-histograms` feature
+flag](https://prometheus.io/docs/prometheus/latest/feature_flags/#native-histograms).
+(TODO: Normalization will become the default in v3. Update this as appropriate.)
```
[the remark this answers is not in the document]
beorn7: I have inverted the wording now, so that everything is from the perspective of v3 being the reference point.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+
+Native histograms were introduced as an experimental feature in November 2022.
+They are a concept that touches almost every part of the Prometheus stack. The
+first version of the Prometheus server supporting native histograms was
+v2.40.0. The support had to be enabled via a feature flag
+`--enable-feature=native-histograms`. (TODO: This is still the case with the
+current release v2.55 and v3.00. Update this section with the stable release,
+once it has happened.)
+
+Due to the pervasive nature of the native-histogram related changes, the
```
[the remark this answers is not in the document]
krajorama: lgtm

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+desirable because there are many situations where the generation of protobuf is
+infeasible. A text format has to make a trade-off between readability for
+humans and efficient handling by machines (encoding, transport, decoding). Work
+on it is in progress. See the [design
+doc](https://github.com/prometheus/proposals/blob/main/proposals/2024-01-29_native_histograms_text_format.md)
+for more details.
+
+(TODO: Update section as progress is made.)
+
+## Instrumentation libraries
```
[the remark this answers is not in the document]
krajorama: yes please, I think one sentence about not forgetting exemplars would be appreciated

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+
+Buckets of native histograms are created dynamically when they are populated
+for the first time. An unexpectedly broad distribution of observed values can
+lead to an unexpectedly high number of buckets, requiring more memory than
+anticipated. If the distribution of observed values can be manipulated from the
+outside, this could even be used as a DoS attack vector via exhausting all the
+memory available to the program. Therefore, an instrumentation library SHOULD
+offer a bucket limitation strategy. It MAY set one by default, depending on the
+typical use cases the library is used for. (TODO: Maybe we should say that a
+strategy SHOULD be set by default. The Go library is currently not limiting the
```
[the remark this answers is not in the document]
krajorama: ack, let's remember this

content/docs/specs/native_histograms.md
krajorama: This is super confusing :( Not sure how we can improve the text, but it seems like having `NaN` written here is the source of trouble. Let me suggest something in a new comment to try and clarify.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2385 @@
...
+
+### Staleness markers
+
+Staleness markers used to be simply a special float value among the many that
+can be used to represent the `NaN` value. The same float value could be used to
+mark a histogram time series as stale. However, this would require cutting a
+new chunk, just for the purpose of marking the series as stale, because a float
+value following a histogram value has to be stored in a different chunk (see
+above). Therefore, there is also a histogram version of a stale marker where
+the field for the sum of observations is set to the special stale `NaN` value.
```
krajorama, line 1282:
```suggestion
the field for the sum of observations is set to the special stale `NaN` value. Note: this special `NaN` value is a float value different to other representations of `NaN`, in particular other `NaN` values that would arise from arithmetic operations, thus it is different from observing a "natural" `NaN` value as described in [Special cases of observed values](#special-cases-of-observed-values).
```

beorn7: I have tried to explain this in the first sentence of this section (albeit in a more condensed form). The confusion encountered here proves that we need a better explanation. However, I think it should be at the beginning of the section, not in the middle. See what I have done (in the next commit).

content/docs/specs/native_histograms.md
beorn7: I'll resolve this conversation, because we have a new one below with @krajorama 's suggestion.

content/docs/specs/native_histograms.md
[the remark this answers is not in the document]
beorn7: I even added two paragraphs now, also extending a bit about how instrumentation libraries can track classic and native histograms in parallel and that the API (`Observe(...)`) doesn't change between them.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2353 @@
...
+
+Addition (`+`) and subtraction (`-`) work between two histograms. These
+operators add or subtract all matching bucket populations and the count and the
+sum of observations. Missing buckets are assumed to be empty and treated
+accordingly. Subtraction might result in negative histograms, see [notes
+above](#unary-minus-and-negative-histograms). Generally, both operands should
+be gauges. Adding and subtracting counter histograms requires caution, but
+PromQL allows it. Adding a gauge histogram and a counter histogram results in a
+gauge histogram. Adding two counter histograms with contradicting counter reset
+hints triggers a warn-level annotation. (TODO: The latter not yet implemented.
```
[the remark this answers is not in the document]
beorn7: I assume the explanation above is satisfactory.

beorn7:
OK, I think we are done. @krajorama please let me know if I have missed anything.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+relevant than the benefits in the exposition format, though. For one,
+Prometheus uses Gorilla-style XOR encoding for floats, which reduces their
+size, albeit not as much as the double-delta encoding used for integers. More
+importantly, an implementation could always decide to internally use an integer
+representation for histogram fields that are effectively integer values (see
+above). (Historical note: Prometheus v1 used exactly this approach to improve
+the compression of float samples, and Prometheus v3 might very well adopt this
+approach again in the future.)
+
+In a counter histogram, the total _count_ of observation and the counts in the
```
bwplotka, line 232: I would expect this section to tell us what to do for gauge histograms. Are there representable by native histograms? If yes, then how?

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+reset detection silently breaks for _sum_ in those cases. Fortunately, negative
+observations are a very rare use case for Prometheus histograms and summaries.)
+
+### Schema
+
+The _schema_ is a signed integer value with a size of 8 bits (short: int8). It
+defines the way bucket boundaries are calculated. The currently valid values
+are -53 and the range between and including -4 and +8. More schemas may be
+added in the future. -53 is a schema for so-called _custom bucket boundaries_
+or short _custom buckets_, while the other schema numbers represent the
```
bwplotka, line 249: Let's add those names to glossary 🤗

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+
+- A _schema_ to identify the method of determining the boundaries of any given
+  bucket with an index _i_.
+- A sparse representation of indexed buckets, mirrored for positive and
+  negative observations.
+- A _zero bucket_ to count observations close to zero.
+- A (possibly empty) list of _custom values_.
+- _Exemplars_.
+
+### Flavors
```
bwplotka, line 168: Perhaps standard schema and non-standard schemas is another dimension worth calling out?

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+// from the Spans).
+message BucketSpan {
+  sint32 offset = 1; // Gap to previous span, or starting point for 1st span (which can be negative).
+  uint32 length = 2; // Length of consecutive buckets.
+}
+
+// [...]
+```
+
+(TODO: The above does not yet contain the custom values needed for NHCBs. Update
```
bwplotka, line 685: Oh nice, so it will be possible to do nhcb on scrape? That's would be nice and somehow along the https://github.com/prometheus/OpenMetrics/issues/283 ideas.

bwplotka: Ok, explained in NHCB section, makes sense!

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+could not be enforced at all.) With very few histograms per chunk, the
+compression ratio becomes worse. Therefore, a minimum number of 10 histograms
+per chunks has to be reached before the size limit of 1024 bytes kicks in. This
+implies that histogram chunks can be much larger than 1024 bytes.
+
+Requiring a minimum of 10 histograms per chunk is an initial, very simplistic
+approach, which might be improved in the future to find a better trade-off
+between chunk size and compression ratio.
+
+### Counter reset considerations
```
bwplotka, line 1338: This is quite complex. What do you think about switching to Created Timestamp feature for it? Would it simplify it? We will have it for all counters soon.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+histogram has no populated buckets, and the sum of observations, the count of
+observations, and the zero bucket population are all zero. Schema, zero bucket
+width, custom values, and the float vs. integer flavor of the histogram SHOULD
+match the sample that directly follows the synthetic zero sample (to not
+trigger the detection of a spurious counter reset).
+
+The counter reset information of the synthetic zero sample is always set to
+`CounterReset`. (TODO: Currently, Prometheus probably sets it to
+`UnknownCounterReset` for the first sample of a series, which is not wrong, but
+I think setting it to `CounterReset` makes more sense.)
```
bwplotka, line 1561: Yea, I think we want to switch to native CT handling, not synthetic zeros. So switching counter reset to CT would be one way of doing this, going forward 🤔

bwplotka (commented):
Nice! Quite an effort and level of the details. Thanks for this work!

Added a few comments, I wonder if we could prepare the reader for nhcb dimension in the glossary and flavors 🤔 Also should we version this spec (like we do with PRW) 

I also have a general questions to counter reset hints. We could simplify by using the CT feature on native histograms instead of reset hints. Those feature do the same, just CT allows a bit more information to be passed (when exactly reset happened). I understand the reset hints are implemented and "baking" for stability, so perhaps it should be improved in some 2.0 🤔

content/docs/specs/native_histograms.md
krajorama: thank you

krajorama (approved):
LGTM as starting point, awesome work .

beorn7:
Thanks @krajorama & @bwplotka 

@bwplotka : I'll address the individual comments soon. Just a few general things in advance:

> I wonder if we could prepare the reader for nhcb dimension in the glossary and flavors 

I tried to include NHCB wherever it was already possible. NHCB is a schema, though, and thus an already defined dimension of the data model. I wouldn't put it under "flavor" (which is anyway a bit of a makeshift solution to have a word for the counter/gauge and integer/float dichotomies).

WRT glossary: My intention was to nail down the words that the reader might get wrong (because of historical baggage ("sparse histogram") or confusion with previously existing concepts (classic vs. native histograms)). Adding all the terms newly introduced by this document would be quite repetitive, as one of the main purposes of this document is to introduce and explain those terms.

> Also should we version this spec (like we do with PRW)

Maybe. But not while it is not really done. It is kind-of 0.x right now. But even once it is finalized, I'm not sure if versioning makes sense because this is more a concept document than strictly a spec. Let's discuss once we have declared native histograms a stable feature.

> I also have a general questions to counter reset hints. We could simplify by using the CT feature on native histograms instead of reset hints. Those feature do the same, just CT allows a bit more information to be passed (when exactly reset happened). I understand the reset hints are implemented and "baking" for stability, so perhaps it should be improved in some 2.0

Counter resets are more deeply baked into NH. For example, it is impossible to have a counter reset in the middle of a chunk, based on the way the encoding works. Thus, we cannot rely on an external feature, that might or might not be present, to mark counter resets. Or even if we did rely on it, we would still store the counter reset implicitly. In any case, we have to see how CT shakes out in practice, how clock skew is handled and all of that.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
+---
+title: Native Histograms [EXPERIMENTAL]
+sort_rank: 6
+---
+
+# Native Histograms
+
+Native histograms were introduced as an experimental feature in November 2022.
```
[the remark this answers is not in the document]
beorn7: It feels more like 0.1 (or 0.0.1 ;). Once we have addressed all (or at least most of) the TODOs here, and declared NH a stable feature, we might think about versioning. Right now, things are just too much in flux. (There are the git hashes, in any case.)

But I'm not even sure if there should be one version for all of NH. There is the exposition format (with its own versioning), there is RW (with its own versioning), there are chunks (with their own versioning), etc.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+for a summary and a number of _bucket_ samples for a (classic) histogram.
+
+With native histograms, a new structured sample type is introduced. A single
+sample represents the previously known _sum_ and _count_ plus a dynamic set of
+buckets. This is not limited to ingestion, but PromQL expressions may also
+return the new sample type where previously it was only possible to return
+float samples.
+
+Native histograms have the following key properties:
+1. A sparse bucket representation, allowing (near) zero cost for empty buckets.
```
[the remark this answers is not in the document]
beorn7: Yeah, weird. I'll try with the blank line.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+2. Coverage of the full float64 range of values.
+3. No configuration of bucket boundaries during instrumentation.
```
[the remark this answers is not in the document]
beorn7: True. I'll add a clarifying sentence.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+4. Dynamic resolution picked according to simple configuration parameters.
+5. Sophisticated exponential bucketing schemas, ensuring mergeability between
+   all histograms using those schemas.
+6. An efficient data representation for both exposition and storage.
+
+Compared to the previously existing “classic” histograms, native histograms
```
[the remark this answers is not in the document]
beorn7: will also add a note for clarification.

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+- [Prometheus Native Histograms in
+  Production](https://www.youtube.com/watch?v=TgINvIK9SYc) provides an analysis
+  of performance and resource consumption.
+- [Using OpenTelemetry’s Exponential Histograms in
+  Prometheus](https://www.youtube.com/watch?v=W2_TpDcess8) covers the
+  interoperability with the OpenTelemetry.
+
+## Glossary
+
+- A __native histogram__ is an instance of the new complex sample type
```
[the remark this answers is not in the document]
beorn7: I think a native histogram is a native histogram of any schema. We should rather clarify at other locations (like above) rather than somehow suggesting that an NHCB is not a native histogram. (I tried to do that already, but as you have found, I wasn't entirely consistent.)

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2416 @@
...
+implementation.
+
+### General structure
+
+Similar to a classic histogram, a native histogram has a field for the _count_
+of observations and a field for the _sum_ of observations. In addition, it
+contains the following components, which are described in detail in dedicated
+sections below:
+
+- A _schema_ to identify the method of determining the boundaries of any given
```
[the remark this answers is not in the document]
beorn7: I wouldn't call it a part of the data model. It is, on the one hand, a result of the way we encode it. On the other hand, it is an implementation detail to not do the expensive counter reset detection too often.

content/docs/specs/native_histograms.md
beorn7: The concept of a gauge histogram is not linked to NH. It was explained (and referenced) above. There, it is explicitly said that a NH can have gauge flavor. Not sure what else there is to explain for the representation. (Should we say somewhere that in a gauge histogram, all the counts are gauges? That's kind of the idea of a gauge histogram and not specific to native histograms, so maybe we shouldn't create the impression it is. Note that this whole paragraph here exists because classic histograms were treating _sum_ wrongly from the beginning, and it got finally fixed with NH.)

content/docs/specs/native_histograms.md
beorn7: Maybe we can consider extending the glossary to all the new terms introduced by this document. However, there would be a lot. If this was a printed document, it might still be helpful for cross reference, but since this is electronic, the reader will simply search for any term they don't remember. (The few terms in the glossary so far are those that might easily be misunderstood or confused.)

content/docs/specs/native_histograms.md
beorn7: CT is in itself a new and experimental feature. I wouldn't conflate them.

Even if CT were already established, I don't think it can be relied upon for NH ingestion. We need to detect counter reset as part of the encoding anyway. CT is more an additional source of information: We can mark (and store) a counter reset even if it is not detectable by conventional means, see below.

content/docs/specs/native_histograms.md
beorn7: See above.

beorn7:
One more thought about the terms I put into the glossary: I think I understand my reluctance to put all the new terms into it better now. What I want is that the reader can just read through the whole document in order, including the glossary. This means it should only contain terms the reader is already somewhat familiar with. A "conventional" glossary is more something you would put at the end of the document where the reader can refer to if they don't remember a certain term in the middle of the text. But the glossary I'm using here is more along the lines of "let's get some common misunderstandings out of the way first".

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2422 @@
...
+- The next positive bucket (index _i_+1 relative to the bucket from the
+  previous item) has a lower exclusive limit of `MaxFloat64` and an upper
+  inclusive limit of `+Inf`. (It could be called a _positive overflow bucket_.)
+- The negative bucket that contains `MinFloat64` (according to the boundary
+  formulas above) has a lower inclusive limit of `MinFloat64` (rather than the
+  limit calculated by the formulas above, which would underflow float64).
+- The next negative bucket (index _i_+1 relative to the bucket from the
+  previous item) has an upper exclusive limit of `MinFloat64` and an lower
+  inclusive limit of `-Inf`. (It could be called a _negative overflow bucket_.)
+- Buckets beyond the `+Inf` and `-Inf` buckets described above MUST NOT be used.
```
fatsheep9146, line 295: Using the `positive overflow bucket` and `negative overflow bucket` might be more accurate than `+Inf` and `-Inf` buckets?

beorn7: Those terms are only introduced in parentheses with "they could be called". I'd rather refer to the limits as that's firmly established before. It's also very accurate because the only observed value that will go into the +Inf bucket is, well, +Inf, and likewise for -Inf.

content/docs/specs/native_histograms.md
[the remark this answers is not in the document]
bwplotka: I think we are on the same page then, thanks!

content/docs/specs/native_histograms.md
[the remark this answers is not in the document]
bwplotka: fair, thanks!

content/docs/specs/native_histograms.md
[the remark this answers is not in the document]
bwplotka: Perhaps make it 1.0-rc.0 then, like we iterate over PRW 2.0? https://prometheus.io/docs/specs/remote_write_spec_2_0/ 🤔

content/docs/specs/native_histograms.md
bwplotka: fair!

content/docs/specs/native_histograms.md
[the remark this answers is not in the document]
beorn7: I would argue that version numbers aren't the right thing for this pseudo spec. It's way too large of a surface to version. As said above, we should version formats and protocols, not this entire beast.

(I'm also somewhat hesitant to add this to the "specification" section. It could as well be in the "concepts" section. Not saying I want to change it now, just to stress the point that this is only partially a spec.)

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2422 @@
...
+eliminated by adding their length to the previous span. Both cases are allowed
+so that producers of native histograms MAY pick whatever representation has the
+best resource trade-offs at that moment. For example, if a histogram is
+processed through various stages, it might be most efficient to only eliminate
+redundant spans after the last processing stage.
+
+In a similar spirit, there are situation where excluding every unpopulated
+bucket from the bucket list is most efficient, but in other situations, it
+might be better to reduce the number of spans by representing small numbers of
+unpopulated buckets explicitly.
```
fatsheep9146, line 373: I don't fully understand here. How could we reduce the number of spans by representing small numbers of unpopulated buckets explicitly? Do you mean combine two successive spans into a new span with some unpopulated buckets in it?

beorn7:
@bwplotka is there anything left to address right now? Can I merge this?

bwplotka (approved):
> @bwplotka is there anything left to address right now? Can I merge this?

All good, thanks for an amazing and deep work here! 💪🏽

beorn7:
There we go. And this is just the beginning. 43 TODOs to address… :sweat_smile:

content/docs/specs/native_histograms.md
```diff
@@ -0,0 +1,2422 @@
...
+- Noisy observations close to zero tend to populate a high number of buckets.
+  Those observations might happen due to numerical inaccuracies or if the
+  source of the observations are actual physical measurements. A zero bucket
+  with a relatively small threshold redirects those observations into a single
+  bucket.
+- If the user is more interested in the long tail of a distribution, far away
+  from zero, a relatively large threshold of the zero bucket helps to avoid
+  many high resolution buckets for a range that is not of interest.
+
+The threshold of the zero bucket SHOULD coincide with a boundary of a regular
```
fatsheep9146, line 432: Do this should be "SHOULD NOT" instead of "SHOULD"? I take "coincide with" as "same or overlapped".

PR #2543, opened 2024-11-12
Downloads: Add "popular" arch to show darwin arm64

### LINKED ISSUE #2542: Add a "popular" architecture entry on the downloads page, which includes arm64
### Proposal

https://prometheus.io/download/ has darwin amd64, but not darwin arm64 (M1 Mac). I think nowadays, the latter is just as popular as the former and we should also include arm64 images.

roidelapluie:
Fixes #2542


bwplotka:
Thanks!

PR #2546, opened 2024-11-13
Add guide for UTF8 usage

ywwg:
I thought it would be useful to have one place that covers UTF-8 from end-to-end -- scrape configuration, instrumentation, and querying.

### DIFF
--- a/content/docs/guides/utf8.md
+++ b/content/docs/guides/utf8.md
@@ -1,0 +1,122 @@
+---
+title: UTF-8 in Prometheus
+---
+
+# Introduction
+
+Versions of Prometheus before 3.0 required that metric and label names adhere to
+a strict set of character requirements. With Prometheus 3.0, all UTF-8 strings
+are valid names, but there are some manual changes needed for other parts of the ecosystem to introduce names with any UTF-8 characters.
+
+There may also be circumstances where users want to enforce the legacy character
+set, perhaps for compatibility with an older system or one that does not yet
+support UTF-8.
+
+This document guides you through the UTF-8 transition details.
+
+# Go Instrumentation 
+
+Currently, metrics created by the official Prometheus [client_golang library](github.com/prometheus/client_golang) will reject UTF-8 names
+by default. It is necessary to change the default validation scheme to allow
+UTF-8. The requirement to set this value will be removed in a future version of
+the common library.
+
+```golang
+import "github.com/prometheus/common/model"
+
+func init() {
+	model.NameValidationScheme = model.UTF8Validation
+}
+```
+
+If users want to enforce the legacy character set, they can set the validation
+scheme to `LegacyValidation`.
+
+Setting the validation scheme must be done before the instantiation of metrics
+and can be set on the fly if desired.
+
+## Instrumenting in other languages
+
+Other client libraries may have similar requirements to set the validation
+scheme. Check the documentation for the library you are using.
+
+# Configuring Name Validation during Scraping
+
+By default, Prometheus 3.0 accepts all UTF-8 strings as valid metric and label
+names. It is possible to override this behavior for scraped targets and reject
+names that do not conform to the legacy character set.
+
+This option can be set in the Prometheus YAML file on a global basis:
+
+```yaml
+global:
+  metric_name_validation_scheme: legacy
+```
+
+or on a per-scrape config basis:
+
+```yaml
+scrape_configs:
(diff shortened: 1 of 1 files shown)


content/docs/guides/utf8.md
```diff
@@ -0,0 +1,107 @@
...
+# Introduction
+
+Versions of Prometheus before 3.0 required that metric and label names adhere to
+a strict set of character requirements. With Prometheus 3.0, all UTF-8 strings
+are valid names, but due to technical limitations it was not possible to create
+a totally seamless transition to the new character set.
+
+There may also be circumstances where users want to enforce the legacy character
+set, perhaps for compatibility with an older system or one that does not yet
+support UTF-8.
```
bwplotka, line 14:
```suggestion
are valid names, but there are some manual changes needed for other parts of the ecosystem to introduce names with any UTF-8 characters.

There may also be circumstances where users want to enforce the legacy character
set, perhaps for compatibility with an older system or one that does not yet
support UTF-8.

This document guides you through the UTF-8 transition details.
```

bwplotka: Just to put it a bit softly more? I think transition should be smooth for majority of users, no?

content/docs/guides/utf8.md
```diff
@@ -0,0 +1,107 @@
...
+
+# Instrumentation
```
bwplotka, line 16:
```suggestion
# Go Instrumentation 
```

content/docs/guides/utf8.md
```diff
@@ -0,0 +1,107 @@
...
+
+Currently, metrics created by the client_golang library will reject UTF-8 names
```
bwplotka, line 18:
```suggestion
Currently, metrics created by the official Prometheus [client_golang library](github.com/prometheus/client_golang) will reject UTF-8 names
```

content/docs/guides/utf8.md
```diff
@@ -0,0 +1,107 @@
...
+by default. It is necessary to change the default validation scheme to allow
+UTF-8. The requirement to set this value will be removed in a future version of
+the common library.
+
+```golang
+import "github.com/prometheus/common/model"
+
+func init() {
+	model.NameValidationScheme = model.UTF8Validation
```
bwplotka, line 27: Side idea: Should we also allow this per registry? 🤔

content/docs/guides/utf8.md
```diff
@@ -0,0 +1,107 @@
...
+  characters with the unicode value, surrounded by underscores. Single
+  underscores are replaced with double underscores. This mode allows for full
+  round-tripping of UTF-8 names with a legacy system.
+
+## Remote Write 2.0
+
+Remote Write 2.0 automatically accepts all UTF-8 names in Prometheus 3.0. There
+is no way to enforce the legacy character set validation with Remote Write 2.0.
+
+# Querying
```
bwplotka, line 89: Can we add "OTLP metrics" with something like 

```suggestion
# OTLP Metrics

OTLP receiver in Prometheus 3.0 still normalizes all names to Prometheus format by default. You can change this in `otlp` section of the Prometheus configuration as follows:


    otlp:
      # Ingest OTLP data keeping UTF-8 characters in metric/label names.
      translation_strategy: NoUTF8EscapingWithSuffixes


See [OpenTelemetry guide](./opentelemetry) for more details.


# Querying

```

content/docs/guides/utf8.md
```diff
@@ -0,0 +1,107 @@
...
+
+```yaml
+scrape_configs:
+  - job_name: prometheus
+    metric_name_validation_scheme: legacy
+```
+
+Scrape config settings override the global setting.
+
+## Content Negotiation
```
bwplotka, line 65:
```suggestion
## Scrape Content Negotiation for UTF-8 escaping
```

bwplotka (approved):
LGTM, thanks! Just some suggestions, thanks!

content/docs/guides/utf8.md
ywwg: Perhaps! This could be done as part of the solution to https://github.com/prometheus/client_golang/issues/1638

content/docs/guides/utf8.md
ywwg: Sure -- should we flesh this out more now or is this enough?

ywwg:
squashed down and added coauthor line

PR #2548, opened 2024-11-14
Document known bug in /otlp endpoint

ArthurSens:
Although I plan to work on this issue very very soon, for at least one release it will be present in the Prometheus codebase.

The bug is rare to occur, but let's document it just in case we get someone by surprise.


jan--f (approved):
Was gonna point out the typo but you beat me to it.
Small optional improvement would be a quick example?

> Currently there's a known limitation in the OTLP translation package where characters get removed from metric/label names if multiple UTF-8 characters are concatenated between words. E.g. `my___metric` is translated to `my_metric`. Please see https://github.com/prometheus/prometheus/issues/15362 for more details.

jan--f (approved):

bwplotka:
This is for all strategies or only one? 🤔  worth noting?

Issue #2469, opened 2024-06-04
Naming conventions: Refine the recommendations.


beorn7:
https://github.com/prometheus/prometheus/issues/8718 discusses "misnamed" metrics and comes to the conclusion that their names are actually fine and we should improve the recommendations for naming metrics to match the actually existing "fine" metric names.

So the task here is to turn the discussion in https://github.com/prometheus/prometheus/issues/8718 into changes of the [metric naming best practices](https://prometheus.io/docs/practices/naming/#metric-names) page.

Gopi-eng2202:
Hi @beorn7 , I have a couple of questions here,
1. Is this naming issue only for failed units? (Thats what i understood from the #8718 issue)
2. If so, would it be a good idea to add a point in the docs page saying that,
A metric name that deals with failed units must have the keyword "_failed" right after the name of the unit.
Examples:

- prometheus_tsdb_head_truncations_failed_total
- net_conntrack_dialer_conn_failed_total
- prometheus_target_scrape_pools_failed_total

If you think my understanding is right here, i can work on this and create a PR.

beorn7:
I'm not sure about the precise answers to your questions. I guess finding the answers is part of the task here. @juliusv and @SuperQ might be better suited to loop in here.

All I can say is that we want the best practices worded in a way that metric names that are in fact fine should be covered by the best practices. From my limited understanding, this is not at all about metrics for "failed" things or the `_failed` suffix (ir infix) in particular. I think this is more about defining what a "unit" is in the Prometheus context. Once it is clarified that "truncations" is not seen as a unit in `prometheus_tsdb_head_truncations_failed_total`, maybe the problem is solved already. The aspect of sorting related metrics together by moving their difference to a position in the name as late as possible might be something to mention in the best practices, but again, this is not specific to `_failed`.

In different news, I think we still want to keep "real" units going last even for metrics that have `failed` in their name. For example, if there is  a metric called `request_size_bytes_total`, and we want the size of failed requests in a separate counter, I could see different ways of calling it:
- `failed_request_size_bytes_total` because it reads most naturally
- `request_size_failed_bytes_total` to still have the "unit" last (`bytes` is explicitly called out as a unit in the best practices).
- `request_size_bytes_failed_total` for the best lexicographical sorting experience.
- Maybe even `request_failed_size_bytes_total`?!?

[...]

PR #2549, opened 2024-11-15
Updating "Best practices: Naming lables" #2469

Gopi-eng2202:
Hi @SuperQ  and @juliusv, 
cc @beorn7 ,
This pr is about the issue #8718.  
I have updated the documentation on Best practices: Metric naming, based on the discussion from the mentioned issue#. I request you to have a look at it and let me know what you think.

Signed-off-by: Gopi-eng2202 <gopi.singaravelan.k@gmail.com>


content/docs/practices/naming.md
```diff
@@ -39,6 +39,26 @@ A metric name...
    (for a pseudo-metric that provides [metadata](https://www.robustperception.io/exposing-the-software-version-to-prometheus) about the running binary)
  * <code>data\_pipeline\_last\_record\_processed\_<b>timestamp_seconds</b></code>
   (for a timestamp that tracks the time of the latest record processed in a data processing pipeline)
+* ...that has a `unit` in its name, must always have that `unit` as the last word. Note that an accumulating count such as `total` will be the last word, in addition to the unit if applicable. Examples,
```
SuperQ, line 42: "Must" is probably too strong.

```suggestion
* ...that has a `unit` in its name, should have that `unit` as the last word. Note that an accumulating count such as `total` will be the last word, in addition to the unit if applicable. Examples,
```

content/docs/practices/naming.md
```diff
@@ -39,6 +39,26 @@ A metric name...
...
+  * <code>request\_size\_<b>bytes</b></code>
+  * <code>request\_size\_<b>bytes</b>\_<b>total</b></code>
+  * <code>process\_cpu\_<b>seconds</b></code>
+  * <code>process\_cpu\_<b>seconds</b>\_<b>total</b></code>
+* ...that has a accumulating count such as `total`, can have two different naming scenarios. 
+  * If the metric name has a unit, then the unit must be the last word before the accumulating count. Examples,
+    * <code>process\_cpu\_<b>seconds</b>\_<b>total</b></code>
+    * <code>request\_size\_<b>bytes</b>\_<b>total</b></code>
+  * If the metric name does'nt have a real unit, then the naming can follow any sorting order in a way that it fits your use case, such as `grouping similar metric names`. Examples,
```
SuperQ, line 51: I'm not sure I agree with this section. The example `connections` is a real countable unit.

However, we have agreed in the past that unit modifiers are prefered to be before, but are still allowed to be after. For example, both `process_thing_failed_bytes_total` and `process_thing_bytes_failed_total` are both valid.

Gopi-eng2202: I understand, i can try to rephrase that. 

But with regards to your overall comment, I understood that only standalone units which are measurable, like "seconds", "bytes" are considered to be a real unit. In a different context "connections" could be a unit, but it does'nt represent a unit of measurement though.

Correct me if i am wrong, ill read the discussion for this issue again to see if i can rephrase these lines.

Thanks for the feedback, will get back with this issue soon.

beorn7: I understood the previous discussions such that we had reached the consensus that we won't consider things like `connections` a "real" unit for the sake of metric naming. (I'm not sure, though, if we should use the wording "real unit" in the doc here.) If my understanding is flawed, maybe @juliusv and @SuperQ could go back to the drawing board and find out what the consensus is.

Very generally, this document is just a recommendation anyway, so we aren't defining a spec here and maybe don't have to be super precise.

On the other hand, OpenMetrics in its current state specifies about the UNIT metadata field:

> Unit specifies MetricFamily units. If non-empty, it MUST be a suffix of the MetricFamily name separated by an underscore. Be aware that further generation rules might make it an infix in the text format.

I'm not sure if we are ever going to enforce this in Prometheus (and the strict requirement might even be lifted in OM v2), but since the OM standard was evolved from existing best practices, it gives us some hints. I would believe the following to be true:
- A "real" unit is one that would show up in the UNIT metadata field. (E.g. we would probably not have a line like `# UNIT go_sql_stats_connections_idle connections` in the OM exposition.)
- Such a "real" unit really should be the last thing before any other suffix like `_total` (in OM, it's a MUST even).

So maybe that's what should go in here: Somehow explain what "real" units are (using a better wording than just saying "real unit"), and then have a very strong recommendation to put them last.

content/docs/practices/naming.md
```diff
@@ -39,6 +39,26 @@ A metric name...
...
+* ...that has a `unit` in its name, should have that `unit` as the last word. Note that an accumulating count such as `total` will be the last word, in addition to the unit if applicable. Examples,
+  * <code>request\_size\_<b>bytes</b></code>
+  * <code>request\_size\_<b>bytes</b>\_<b>total</b></code>
+  * <code>process\_cpu\_<b>seconds</b></code>
+  * <code>process\_cpu\_<b>seconds</b>\_<b>total</b></code>
+* ...that has a accumulating count such as `total`, can have two different naming scenarios. 
+  * If the metric name has a unit, then the unit must be the last word before the accumulating count. Examples,
+    * <code>process\_cpu\_<b>seconds</b>\_<b>total</b></code>
+    * <code>request\_size\_<b>bytes</b>\_<b>total</b></code>
+  * If the metric name does'nt have a real unit, then the naming can follow any sorting order in a way that it fits your use case, such as `grouping similar metric names`. Examples,
```
beorn7, line 51: Typo: does'nt → doesn't

content/docs/practices/naming.md
```diff
@@ -39,6 +39,26 @@ A metric name...
...
+    * If you decided to name something as "<code>net\_conntrack\_dialer\_conn</code>", then it can have the following group names. (Note that this is only an example and you can follow your own order, if it improves your work flow)
+      * <code>net\_conntrack\_dialer\_conn\_total</code>
+      * <code>net\_conntrack\_dialer\_conn\_failed\_total</code>
+      * <code>net\_conntrack\_dialer\_conn\_established\_total</code>
+      * <code>net\_conntrack\_dialer\_conn\_closed\_total</code>
+      or
+      * <code>net\_conntrack\_dialer\_conn\_total</code>
+      * <code>net\_conntrack\_dialer\_conn\_total\_failed</code>
+      * <code>net\_conntrack\_dialer\_conn\_total\_established</code>
+      * <code>net\_conntrack\_dialer\_conn\_total\_closed</code>
```
beorn7, line 61: I'm pretty sure that we _always_ want the `_total` last in a counter. So these examples should not be here.

beorn7 (changes requested):
To take a step back:

The changes as proposed right now are in part trying to recommend even more strictly that the unit has to go last. But that's already in the guidelines, even without a change. Remember that the whole discussion started because people felt the rule should be expressed in a _less_ strict way, because they saw "units" like "connections" not being last in widely used metric names. The discussion then turned more the other way by concluding that things like "connections" are not "real" units.

To summarize:

I think we need far fewer changes here. What I want to see is a good rule-of-thumb what a unit is in the narrow sense (maybe referring to the "base units" section further down, which essentially lists many relevant examples, notably excluding things like "connections") and then we don't really have to change much else. _Maybe_ we could add a new paragraph about ordering the metric name components in a way that the lexicographic sorting makes most sense, and that then you should still put the "real" units last, but say that this doesn't apply to things like "connections".

content/docs/practices/naming.md
SuperQ, line 61: Yes, this example is invalid.

```suggestion
```

Gopi-eng2202: Got it

Gopi-eng2202: ok, I'll work on this

content/docs/practices/naming.md
```diff
@@ -39,6 +39,16 @@ A metric name...
...
+* ...that has a accumulating count such as `total`, but doesn't have a base unit in the name like, "<code>prometheus\_tsdb\_head\_truncations</code>", then it can have the following naming scenarios. Note that the former examples would be suitable if you want to follow lexographic sorting order for the names.
+ * <code>prometheus\_tsdb\_head\_truncations\_total</code>
+ * <code>prometheus\_tsdb\_head\_truncations\_failed\_total</code>
+ * <code>prometheus\_tsdb\_head\_truncations\_established\_total</code>
+ * <code>prometheus\_tsdb\_head\_truncations\_closed\_total</code>
+  or
+ * <code>prometheus\_tsdb\_head\_truncations\_total</code>
+ * <code>prometheus\_tsdb\_head\_failed\_truncations\_total</code>
+ * <code>prometheus\_tsdb\_head\_established\_truncations\_total</code>
+ * <code>prometheus\_tsdb\_head\_closed\_truncations\_total</code>
```
beorn7, line 51: I would word this a bit differently. Also, the indentation of the lists has to be different to work properly. Also, the example metrics should actually be sorted lexicographically to prove the point. See suggestion below.

```suggestion
* ...may order its name components in a way that leads to convenient grouping when a list of metric names is sorted lexicographically, as long as all the other rules are followed. The following examples have their the common name components first so that all the related metrics are sorted together:
  * <code>prometheus\_tsdb\_head\_truncations\_closed\_total</code>
  * <code>prometheus\_tsdb\_head\_truncations\_established\_total</code>
  * <code>prometheus\_tsdb\_head\_truncations\_failed\_total</code>
  * <code>prometheus\_tsdb\_head\_truncations\_total</code>
  The following examples are also valid, but are following a different trade-off. They are easier to read individually, but unrelated metrics like <code>prometheus\_tsdb\_head\_series</code> might get sorted in between.
  * <code>prometheus\_tsdb\_head\_closed\_truncations\_total</code>
  * <code>prometheus\_tsdb\_head\_established\_truncations\_total</code>
  * <code>prometheus\_tsdb\_head\_failed\_truncations\_total</code>
  * <code>prometheus\_tsdb\_head\_truncations\_total</code>
```

content/docs/practices/naming.md
```diff
@@ -27,7 +27,7 @@ A metric name...
  * <code><b>http</b>\_request\_duration\_seconds</code>
    (for all HTTP requests)
 * ...must have a single unit (i.e. do not mix seconds with milliseconds, or seconds with bytes).
-* ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). See below for a list of base units.
+* ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). Note that things like connections, notifications are not considered as base units.See below for a list of base units.
 * ...should have a suffix describing the unit, in plural form. Note that an accumulating count has `total` as a suffix, in addition to the unit if applicable.
```
beorn7, line 31: This is where the note about "real" units can happen. You could append something like:

Also note that this applies to units in the narrow sense (like the units in the table below), but not to countable things in general. For example, `connections` or `notifications` are not considered units for this rule and do not have to be at the end of the metric name. (See also examples in the next paragraph.)

content/docs/practices/naming.md
```diff
@@ -39,6 +39,16 @@ A metric name...
...
+ * <code>prometheus\_tsdb\_head\_truncations\_established\_total</code>
+ * <code>prometheus\_tsdb\_head\_truncations\_closed\_total</code>
+  or
+ * <code>prometheus\_tsdb\_head\_truncations\_total</code>
+ * <code>prometheus\_tsdb\_head\_failed\_truncations\_total</code>
+ * <code>prometheus\_tsdb\_head\_established\_truncations\_total</code>
+ * <code>prometheus\_tsdb\_head\_closed\_truncations\_total</code>
 * ...should represent the same logical thing-being-measured across all label
   dimensions.
  * request duration
```
beorn7, line 54: Line 54 to 56 need to be indented by one more blank.

beorn7 (changes requested):

Gopi-eng2202:
Hi @beorn7,

Resolved all the requested changes. Do have a look and merge this pr if no further changes are required. 

Thank you

content/docs/practices/naming.md
```diff
@@ -20,30 +20,40 @@ A metric name...
...
- * <code>http\_requests\_<b>total</b></code>
-   (for a unit-less accumulating count)
- * <code>process\_cpu\_<b>seconds\_total</b></code>
-   (for an accumulating count with unit)
- * <code>foobar_build<b>\_info</b></code>
-   (for a pseudo-metric that provides [metadata](https://www.robustperception.io/exposing-the-software-version-to-prometheus) about the running binary)
- * <code>data\_pipeline\_last\_record\_processed\_<b>timestamp_seconds</b></code>
-  (for a timestamp that tracks the time of the latest record processed in a data processing pipeline)
+* ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers).See below for a list of base units.
+* ...should have a suffix describing the unit, in plural form. Note that an accumulating count has `total` as a suffix, in addition to the unit if applicable.Also note that this applies to units in the narrow sense (like the units in the table below), but not to countable things in general. For example, <code>connections</code> or <code>notifications</code> are not considered units for this rule and do not have to be at the end of the metric name. (See also examples in the next paragraph.)
```
beorn7, line 31:
```suggestion
* ...should have a suffix describing the unit, in plural form. Note that an accumulating count has `total` as a suffix, in addition to the unit if applicable. Also note that this applies to units in the narrow sense (like the units in the table below), but not to countable things in general. For example, <code>connections</code> or <code>notifications</code> are not considered units for this rule and do not have to be at the end of the metric name. (See also examples in the next paragraph.)
```

(Blank is missing. I'll commit this directly, no need to update the PR.)

content/docs/practices/naming.md
```diff
@@ -20,30 +20,40 @@ A metric name...
...
+  * <code>foobar_build<b>\_info</b></code>
+    (for a pseudo-metric that provides [metadata](https://www.robustperception.io/exposing-the-software-version-to-prometheus) about the running binary)
+  * <code>data\_pipeline\_last\_record\_processed\_<b>timestamp_seconds</b></code>
+    (for a timestamp that tracks the time of the latest record processed in a data processing pipeline)
+* ...may order its name components in a way that leads to convenient grouping when a list of metric names is sorted lexicographically, as long as all the other rules are followed. The following examples have their the common name components first so that all the related metrics are sorted together:
+  * <code>prometheus\_tsdb\_head\_truncations\_closed\_total</code>
+  * <code>prometheus\_tsdb\_head\_truncations\_established\_total</code>
+  * <code>prometheus\_tsdb\_head\_truncations\_failed\_total</code>
+  * <code>prometheus\_tsdb\_head\_truncations\_total</code>
+  The following examples are also valid, but are following a different trade-off. They are easier to read individually, but unrelated metrics like <code>prometheus\_tsdb\_head\_series</code> might get sorted in between.
```
beorn7, line 47:
```suggestion

    The following examples are also valid, but are following a different trade-off. They are easier to read individually, but unrelated metrics like <code>prometheus\_tsdb\_head\_series</code> might get sorted in between.
```

I think we need a blank line here to convince markdown to end the previous list item. (I'll commit this directly, no need to change the PR.)

beorn7 (approved):
I'll merge this now to get things rolling. It's "just" documentation after all. @SuperQ please follow up if you have any concerns.

PR #2557, opened 2024-12-12
Small amendment to the native histogram spec

beorn7:
This completes the list of functions acting on range vectors that filter out native histograms.

/cc @NeerajGartia21


krajorama (approved):
stamp, I don't immediately see where in the code we do this , but we can fix the code later if needed.

beorn7:
This is not implemented yet, see https://github.com/prometheus/prometheus/issues/13934.

Issue #2201, opened 2022-09-13
Update and clarify Security Model page


beorn7:
https://prometheus.io/docs/operating/security/#security-model is somewhat out of date WRT TLS and basic auth support.

I also think that the assumptions about how to handle `/metrics` endpoints should be clarified. It is by design that metrics are exposed, and that needs to pay attention about who has access to those endpoints and what information and DoS potential is exposed. Issues like https://github.com/prometheus/prometheus/issues/11290 create the impression that this exposure is surprising or a security flaw. See also https://kccnceu2022.sched.com/event/ytmB/how-attackers-use-exposed-prometheus-server-to-exploit-kubernetes-clusters-david-de-torres-huerta-miguel-hernandez-sysdig , which is less naive than the report above but also creates the impression that the design is flawed and that a "secure by default" setup is the expectation.

The Security Model page should not only make it very clear, even for a naive reader, that the `/metrics` endpoint of monitored targets and the query endpoints for the Prometheus server must not be accessible to the public internet in general, but it should also explain why this setup is a deliberate decision and a "secure by default" setup wouldn't really work.

beorn7:
We discussed this again at the dev summit today. An important outcome was that we have confirmed that we do not aim for a "secure by default" setup, not only because it is contrary to the operational model of Prometheus, but also because it would raise a claim that we cannot ever fulfill. It's therefore even more important to dispel any "secure by default" expectation.

As discussed, @jan--f volunteered as the assignee for this issue, but more to coordinate the effort, not so much to do all the work on his own. Any help is welcome.

PR #2558, opened 2024-12-13
Add a "tl;dr:" style note at the top of the security page

beorn7:
The security model is written in a very technical form that is hard to understand for naive users. Naive users are, sadly, also most prone to run Prometheus components exposed to the internet. I have little hope that those users will read this page at all, but if they do, they should get the message. Furthermore, we have had multiple incidents where 3rd parties "uncover" "vulnerabilities and security flaws". There is nothing to uncover, and the flagged behaviors are by design. We have to make that more clear.

This partially addresses #2201, but I think we need to do more. In particular, we should add a section about /pprof (and why we have it on by default), be clearer about DoS'ing ("It is more likely that a component will be accidentally taken out by a trusted user than by malicious action." can easily be understood in an unintended way), and provide more context about the usage of the word "untrusted".

### DIFF
--- a/content/docs/operating/security.md
+++ b/content/docs/operating/security.md
@@ -3,6 +3,17 @@ title: Security
 sort_rank: 4
 ---
 
+NOTE: Before we dive into the technical details below, we would like to
+emphasize that Prometheus, as a monitoring system, collects and serves
+information about the systems it is monitoring. Therefore, the HTTP endpoints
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
LGTM

bwplotka (approved):
LGTM!

Issue #2556, opened 2024-12-12
Replace(?) deprecated installation instructions in "Instrumenting a Go application for Prometheus" docs


atimeofday:
Hello,

As a new user of both Go and Prometheus, I ran into some issues and tentatively believe the instructions for installing the `prometheus/client_golang` libraries are outdated as of Go 1.17 in... August 2021. It appears `go get` was deprecated outside of Go Modules, and requires the creation of a Go Module first. Please let me know if this impression is incorrect, or if a small PR to adjust the instructions might be welcome.

Issue location: https://github.com/prometheus/docs/blob/main/content/docs/guides/go-application.md#installation
Issue contents:
```
go get github.com/prometheus/client_golang/prometheus
go get github.com/prometheus/client_golang/prometheus/promauto
go get github.com/prometheus/client_golang/prometheus/promhttp
```

Potential solution: preface `go get` instructions with `go mod init <module_name>` and a link to Go Module documentation

beorn7:
I guess you are right, these are outdated instruction to get Go packages. A PR fixing this would be very welcome.

lhajouji:
any one did take a look please ?

beorn7:
@lhajouji IIUC #2563 has been waiting for your response for quite some time.

lhajouji:
Sorry for my late response. the bellow PR fixed the issue.
We may close this issue.
Thanks !

PR #2563, opened 2024-12-20
fixing outdated instruction to get Go packages

lhajouji:
This PR main purpose is to resolve this issue [fixing deprecated installation instructions in "Instrumenting a Go application for Prometheus" docs](https://github.com/prometheus/docs/issues/2556)

### DIFF
--- a/content/docs/guides/go-application.md
+++ b/content/docs/guides/go-application.md
@@ -10,7 +10,21 @@ NOTE: For comprehensive API documentation, see the [GoDoc](https://godoc.org/git
 
 ## Installation
 
-You can install the `prometheus`, `promauto`, and `promhttp` libraries necessary for the guide using [`go get`](https://golang.org/doc/articles/go_command.html):
+To integrate Prometheus with your Go application, you'll need the `prometheus`, `promauto`, and `promhttp` libraries. Starting with Go 1.17, Go Modules are required for managing dependencies.
+
+#### Using Go Modules
+
+If your project does not already use Go Modules, initialize a new module in your project directory:
+
+```bash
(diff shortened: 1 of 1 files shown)


content/docs/guides/go-application.md
```diff
@@ -10,7 +10,21 @@ NOTE: For comprehensive API documentation, see the [GoDoc](https://godoc.org/git
 
 ## Installation
 
-You can install the `prometheus`, `promauto`, and `promhttp` libraries necessary for the guide using [`go get`](https://golang.org/doc/articles/go_command.html):
+To integrate Prometheus with your Go application, you'll need the `prometheus`, `promauto`, and `promhttp` libraries. Starting with Go 1.17, Go Modules are required for managing dependencies.
```
machine424, line 13: Thanks for this, but I'm afraid the more we add, the more maintenance it will require.
Let's just replace the existing link https://golang.org/doc/articles/go_command.html with with https://go.dev/ref/mod#go-get which is more module oriented.

And to be honest the warning when using `go get` outside a module is clear enough:

```
go: go.mod file not found in current directory or any parent directory.
	'go get' is no longer supported outside a module.
	To build and install a command, use 'go install' with a version,
	like 'go install example.com/cmd@latest'
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
```

sultanonwork:
Hello @machine424 just a friendly ping, is this issue resolved?

lhajouji:
[docs: update Go instrumentation installation instructions Username46786/docs_#1](https://github.com/Username46786/docs_/pull/1)

PR #2572, opened 2025-01-10
highlight le and quantile label values normalization in Prom 3.0

machine424:
As discussed in https://github.com/prometheus/prometheus/pull/15164/files#r1806414216

### DIFF
--- a/content/docs/concepts/metric_types.md
+++ b/content/docs/concepts/metric_types.md
@@ -79,6 +79,10 @@ observations. Native histograms allow much higher resolution at a fraction of
 the cost. Detailed documentation will follow once native histograms are closer
 to becoming a stable feature.
 
+NOTE: Beginning with Prometheus v3.0, the values of the `le` label of classic
+histograms are normalized during ingestion to follow the format of
+[OpenMetrics Canonical Numbers](https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md#considerations-canonical-numbers).
+
 Client library usage documentation for histograms:
 
    * [Go](http://godoc.org/github.com/prometheus/client_golang/prometheus#Histogram)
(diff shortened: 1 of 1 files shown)


content/docs/concepts/data_model.md
```diff
@@ -33,6 +33,9 @@ Note: The colons are reserved for user defined recording rules. They should not
  * Label values may contain any Unicode characters.
  * Labels with an empty label value are considered equivalent to labels that do not exist.
 
+NOTE: Beginning with Prometheus v3.0, the values of the `le` label of classic
+histograms and the `quantile` label in summaries are consistently normalized to
```
machine424, line 37: are invariably? always?

machine424: maybe link the migration guide? https://prometheus.io/docs/prometheus/latest/migration/#le-and-quantile-label-values

beorn7: Not sure if this is a good location for this factoid. The normalization is just an ingestion concern. Non-normalized `le` and `quantile` labels still fit the data model (and they even work with PromQL, it's just inconvenient if you have a mix of differently formatted `le` and `quantile` labels).

content/docs/concepts/metric_types.md
```diff
@@ -79,6 +79,9 @@ observations. Native histograms allow much higher resolution at a fraction of
 the cost. Detailed documentation will follow once native histograms are closer
 to becoming a stable feature.
 
+NOTE: Beginning with Prometheus v3.0, the values of the `le` label of classic
+histograms is consistently normalized to a float-like representation upon ingestion.
```
beorn7, line 83: "Float-like representation" is pretty vague. How about this:

```suggestion
NOTE: Beginning with Prometheus v3.0, the values of the `le` label of classic
histograms are normalized during ingestion to follow the format of [OpenMetrics Canonical Numbers](https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md#considerations-canonical-numbers).
```

content/docs/concepts/metric_types.md
```diff
@@ -106,6 +109,9 @@ See [histograms and summaries](/docs/practices/histograms) for
 detailed explanations of φ-quantiles, summary usage, and differences
 to [histograms](#histogram).
 
+NOTE: Beginning with Prometheus v3.0, the values of the `quantile` label is consistently
+normalized to a float-like representation upon ingestion.
```
beorn7, line 113: I would propose the same change as above.

PR #2573, opened 2025-01-14
Fix documentation links

ilovelinux:
Fix broken links inside documentation


bwplotka (changes requested):
Hm, can you double check? I can see the current website links you are changing works fine https://prometheus.io/docs/instrumenting/exposition_formats 🤔

SuperQ:
The paths in this codebase are processed by Nanoc. They are not meant to work in-repo without processing

This change actually makes the links broken.

PR #2576, opened 2025-01-29
Update native histogram spec

beorn7:
Add the newly established behavior of rate, irate & friends.

Remove some obsolete TODOs.

@NeerajGartia21 @krajorama

### DIFF
--- a/content/docs/specs/native_histograms.md
+++ b/content/docs/specs/native_histograms.md
@@ -318,9 +318,7 @@ the exponential bucketing featured by the standard schemas is a bad match for
 the distribution to be represented by the histogram. Histograms with different
 custom bucket boundaries are generally not mergeable with each other.
 Therefore, schema -53 SHOULD only be used as an informed decision in specific
-use cases. (TODO: NHCB aren't fully merged into main as of now (2024-11-03).
-They are worked into this document as far as possible already. This information
-might not yet be relevant for released Prometheus versions.)
+use cases.
 
 ### Buckets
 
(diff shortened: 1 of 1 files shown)


content/docs/specs/native_histograms.md
```diff
@@ -1861,19 +1858,26 @@ native histogram:
...
 These functions SHOULD be applied to either gauge histograms or counter
 histograms as noted above. However, they all work with both flavors, but if at
 least one histogram of an unsuitable flavor is contained in the range vector, a
 warn-level annotation is added to the result.
 
 All these functions return no result for series that contain a mix of float
 samples and histogram samples within the range. A warn-level annotation is
-added for each output element missing for that reason.
+added for each output element missing for that reason. `idelta()` and `irate()`
+only consider the last two samples in the range for this check.
```
NeerajGartia21, line 1869: Athough the last line "`idelta()` and `irate()` only consider the last two samples in the range for this check." clarifies that we are only considering the last two samples, still I think it would be better to have this conveyed in the initial sentence.

NeerajGartia21 (commented):
Looks good to me!
Just a small suggestion.

content/docs/specs/native_histograms.md
beorn7: The problem here is that the initial sentence is the general one that is valid for all these function. The sentence about idelta and irate is the exception to it for these two special functions. Maybe I can munge the last and first sentence together…

beorn7: How do you like it now?

NeerajGartia21: Now it looks good to me!
Thanks for the change.

beorn7:
With @NeerajGartia21 :+1: , I'll merge this now. It's "just documentation". @krajorama if you have any concerns, let me know and I'll address them in a separate PR.

PR #2585, opened 2025-02-15
move prometheus-smartraid-exporter to the “Storage” section

calestyo:


SuperQ:
Hmm, the storage section generally is about software-based storage.

calestyo:
https://github.com/prometheus/docs/blob/ef8460e2dc132a7d7fbd7b2daefecc192c6617df/content/docs/instrumenting/exporters.md?plain=1#L122-127

Hmm, at least:
- HPE storage array exporter
- NetApp E-Series exporter

seem to be hardware... and for some others (e.g. ScaleIO, Tivoli), one could argue so, too.

PR #2608, opened 2025-03-17
Welcome George Robinson!

gotjosh:
After a vote pass, George Robinson is now part of the Prometheus Team.


beorn7 (approved):
@grobinson-grafana should also approve this before merging.

PR #2609, opened 2025-03-17
Add Scraparr to API Exporters

TheGameProfi:
Hey there,
I created a Exporter for the *Arr Suite, and would like to include it into the Exporters List.
It would be nice if somebody could have a look at it :smile:

### DIFF
--- a/docs/instrumenting/exporters.md
+++ b/docs/instrumenting/exporters.md
@@ -259,6 +259,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Rundeck exporter](https://github.com/phsmith/rundeck_exporter)
    * [SABnzbd exporter](https://github.com/msroest/sabnzbd_exporter)
    * [SAML exporter](https://github.com/DoodleScheduling/saml-exporter)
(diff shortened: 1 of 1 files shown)


content/docs/instrumenting/exporters.md
```diff
@@ -160,6 +160,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Rancher exporter](https://github.com/infinityworks/prometheus-rancher-exporter)
    * [Speedtest exporter](https://github.com/nlamirault/speedtest_exporter)
    * [Tankerkönig API Exporter](https://github.com/lukasmalkmus/tankerkoenig_exporter)
+   * [Scraparr](https://github.com/thecfu/scraparr)
```
SuperQ, line 163: This should be in alphabetical order.

SuperQ (changes requested):

content/docs/instrumenting/exporters.md
TheGameProfi: Sorry didn't noticed thought the Program would do it, but i fixed it

PR #2623, opened 2025-04-10
Update exposition_formats.md

Propfend:
fixing broken links in documentation.

>library you can [create your own](/content/docs/instrumenting/writing_clientlibs.md/).
exposition format. There are various [client libraries](/content/docs/instrumenting/clientlibs.md/)

these links from these first two lines of this documentation section were broken.


SuperQ:
These files are processed by Nanoc, which renders the actual site. This would break the site.

PR #2626, opened 2025-04-17
naming: Update metric and label name restrictions and recommendations with the latest context

bwplotka:
### Notable changes

* Cleaned up should/must/may to upper case for clarity (as per RFC style we do in specs).
* Mentioned UTF-8 support.
* Lifted limited character set from MUST to SHOULD with a clear about consequences and downsides.
* Mentioned alternative notations.
* Removed experimental mentions about native histograms, made it part of the main sample definition.
* Added section on WHY Prometheus metric SHOULD contain type and unit suffix, even with our type and unit labels one day.

### Motivation

* [OpenMetrics WG 2.0 Action item](https://docs.google.com/document/d/1FCD-38Xz1-9b3ExgHOeDTQUKUatzgj5KbCND9t-abZY/edit?tab=t.lvx6fags1fga#heading=h.3xjg8jcpctk4 ) and blocker to relax MUST to SHOULD type and unit suffix in OpenMetrics 2.0.
* Similarly, to have something clear to reference to highlight huge downsides of skipping suffixes for OTLP (https://github.com/prometheus/prometheus/pull/16441)
* To help with explaining this problem to Prometheus users in general (on recent KubeCons we had to explain this hundrends of times).

### DIFF
--- a/content/docs/concepts/data_model.md
+++ b/content/docs/concepts/data_model.md
@@ -17,42 +17,39 @@ Every time series is uniquely identified by its metric name and optional key-val
 
 ***Metric names:***
 
- * Specify the general feature of a system that is measured (e.g. `http_requests_total` - the total number of HTTP requests received). 
- * Metric names may contain ASCII letters, digits, underscores, and colons. It must match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*`.
-   
-Note: The colons are reserved for user defined recording rules. They should not be used by exporters or direct instrumentation.
+* Metric names SHOULD specify the general feature of a system that is measured (e.g. `http_requests_total` - the total number of HTTP requests received). 
+* Metric names MAY use any UTF-8 characters.
+* Metric names SHOULD match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*` for the best experience and compatibility (see the warning below). Metric names outside of that set will require escaping e.g. when used in PromQL (see the [UTF-8 guide](../guides/utf8.md#querying)).
 
-
+NOTE: The colons are reserved for user defined recording rules. They should not be used by exporters or direct instrumentation.
 
 ***Metric labels:***
 
- * Enable Prometheus's dimensional data model to identify any given combination of labels for the same metric name. It identifies a particular dimensional instantiation of that metric (for example: all HTTP requests that used the method `POST` to the `/api/tracks` handler). The query language allows filtering and aggregation based on these dimensions. 
- * The change of any label's value, including adding or removing labels, will create a new time series.
- * Labels may contain ASCII letters, numbers, as well as underscores. They must match the regex `[a-zA-Z_][a-zA-Z0-9_]*`. 
- * Label names beginning with `__` (two "_") are reserved for internal use.
- * Label values may contain any Unicode characters.
- * Labels with an empty label value are considered equivalent to labels that do not exist.
+Labels enable Prometheus's dimensional data model to identify any given combination of labels for the same metric name. It identifies a particular dimensional instantiation of that metric (for example: all HTTP requests that used the method `POST` to the `/api/tracks` handler). The query language allows filtering and aggregation based on these dimensions. The change of any label's value, including adding or removing labels, will create a new time series.
 
+* Label names MAY use any UTF-8 characters. 
+* Label names beginning with `__` (two "_") MUST be reserved for internal Prometheus use.
+* Label names SHOULD match the regex `[a-zA-Z_][a-zA-Z0-9_]*` for the best experience and compatibility (see the warning below). Label names outside of that regex will require escaping e.g. when used in PromQL (see the [UTF-8 guide](../guides/utf8.md#querying)).
+* Label values MAY contain any UTF-8 characters.
+* Labels with an empty label value are considered equivalent to labels that do not exist.
 
-See also the [best practices for naming metrics and labels](/docs/practices/naming/).
+WARNING: The [UTF-8](../guides/utf8.md) support for metric and label names was added relatively recently in Prometheus v3.0.0. It might take time for the wider ecosystem (downstream PromQL compatible projects and vendors, tooling, 3P instrumentation, collectors, etc.) to adopt new escaping mechanisms, relaxed validation etc. For the best compatibility it's recommended to stick to the recommended ("SHOULD") character set.
+
+INFO: See also the [best practices for naming metrics and labels](/docs/practices/naming/).
 
 ## Samples
 
 Samples form the actual time series data. Each sample consists of:
 
-   * a float64 value
-   * a millisecond-precision timestamp
-
-NOTE: Beginning with Prometheus v2.40, there is experimental support for native
-histograms. Instead of a simple float64, the sample value may now take the form
-of a full histogram.
+* a float64 or [native histogram](https://prometheus.io/docs/specs/native_histograms/) value
+* a millisecond-precision timestamp
 
 ## Notation
 
 Given a metric name and a set of labels, time series are frequently identified
 using this notation:
 
-    <metric name>{<label name>=<label value>, ...}
+    <metric name>{<label name>="<label value>", ...}
 
 For example, a time series with the metric name `api_http_requests_total` and
 the labels `method="POST"` and `handler="/messages"` could be written like
@@ -61,3 +58,14 @@ this:
     api_http_requests_total{method="POST", handler="/messages"}
 
 This is the same notation that [OpenTSDB](http://opentsdb.net/) uses.
+
+Names with the UTF-8 characters outside the recommended set must be escaped, using
+this notation:
+
+    {"<metric name>", <label name>="<label value>", ...}
+
+Since an internal representation of a metric name is in a form of a label pair
+with a special label name (`__name__="<metric name>"`) one could also use the following notation:
+
+    {__name__="<metric name>", <label name>="<label value>", ...}
+
--- a/content/docs/concepts/data_model.md
+++ b/content/docs/concepts/data_model.md
@@ -21,14 +21,14 @@ Every time series is uniquely identified by its metric name and optional key-val
 * Metric names MAY use any UTF-8 characters.
 * Metric names SHOULD match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*` for the best experience and compatibility (see the warning below). Metric names outside of that set will require escaping e.g. when used in PromQL (see the [UTF-8 guide](../guides/utf8.md#querying)).
 
-NOTE: The colons are reserved for user defined recording rules. They should not be used by exporters or direct instrumentation.
+NOTE: Colons (':') are reserved for user-defined recording rules. They should not be used by exporters or direct instrumentation.
 
 ***Metric labels:***
 
-Labels enable Prometheus's dimensional data model to identify any given combination of labels for the same metric name. It identifies a particular dimensional instantiation of that metric (for example: all HTTP requests that used the method `POST` to the `/api/tracks` handler). The query language allows filtering and aggregation based on these dimensions. The change of any label's value, including adding or removing labels, will create a new time series.
+Labels let you capture different instances of the same metric name. For example: all HTTP requests that used the method `POST` to the `/api/tracks` handler. We refer to this as Prometheus's "dimensional data model". The query language allows filtering and aggregation based on these dimensions. The change of any label's value, including adding or removing labels, will create a new time series.
 
 * Label names MAY use any UTF-8 characters. 
-* Label names beginning with `__` (two "_") MUST be reserved for internal Prometheus use.
+* Label names beginning with `__` (two underscores) MUST be reserved for internal Prometheus use.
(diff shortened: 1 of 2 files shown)


content/docs/concepts/data_model.md
```diff
@@ -17,42 +17,39 @@ Every time series is uniquely identified by its metric name and optional key-val
...
- * The change of any label's value, including adding or removing labels, will create a new time series.
- * Labels may contain ASCII letters, numbers, as well as underscores. They must match the regex `[a-zA-Z_][a-zA-Z0-9_]*`. 
- * Label names beginning with `__` (two "_") are reserved for internal use.
- * Label values may contain any Unicode characters.
- * Labels with an empty label value are considered equivalent to labels that do not exist.
+Labels enable Prometheus's dimensional data model to identify any given combination of labels for the same metric name. It identifies a particular dimensional instantiation of that metric (for example: all HTTP requests that used the method `POST` to the `/api/tracks` handler). The query language allows filtering and aggregation based on these dimensions. The change of any label's value, including adding or removing labels, will create a new time series.
```
bboreham, line 28: Long words, hard to read.
```suggestion
Labels let you capture different instances of the same metric name. For example: all HTTP requests that used the method `POST` to the `/api/tracks` handler. We refer to this as Prometheus's "dimensional data model". The query language allows filtering and aggregation based on these dimensions. The change of any label's value, including adding or removing labels, will create a new time series.
```

content/docs/practices/naming.md
```diff
@@ -26,9 +26,9 @@ A metric name...
...
+* ...MUST have a single unit (i.e. do not mix seconds with milliseconds, or seconds with bytes).
+* ...SHOULD use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). See [below](#base-units) for a list of base units.
+* ...SHOULD have a suffix describing the unit, in plural form. Note that an accumulating count has `total` as a suffix, in addition to the unit if applicable. Also note that this applies to units in the narrow sense (like the units in the table below), but not to countable things in general. For example, <code>connections</code> or <code>notifications</code> are not considered units for this rule and do not have to be at the end of the metric name. (See also examples in the next paragraph.)
```
bboreham, line 31: I find the "units ... countable" part hard to follow.  As far as I can see there are no examples with `connections` or `notifications`, so referring to examples is perverse.  
Is this emphasis because we don't put `requests` in the UNIT metadata?

Given the long section you have added below, maybe remove this part and refer to that section?

content/docs/practices/naming.md
```diff
@@ -61,6 +61,33 @@ meaningful, split the data up into multiple metrics. For example, having the
 capacity of various queues in one metric is good, while mixing the capacity of a
 queue with the current number of elements in the queue is not.
 
+### Why including unit (and _total for counter) suffixes in metric names?
```
bboreham, line 64:
```suggestion
### Why include unit (and _total for counter) suffixes in metric names?
```

bboreham (commented):
Generally fine.  Some thoughts below.

content/docs/concepts/data_model.md
```diff
@@ -17,42 +17,39 @@ Every time series is uniquely identified by its metric name and optional key-val
...
 ***Metric names:***
 
- * Specify the general feature of a system that is measured (e.g. `http_requests_total` - the total number of HTTP requests received). 
- * Metric names may contain ASCII letters, digits, underscores, and colons. It must match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*`.
-   
-Note: The colons are reserved for user defined recording rules. They should not be used by exporters or direct instrumentation.
-
+* Metric names SHOULD specify the general feature of a system that is measured (e.g. `http_requests_total` - the total number of HTTP requests received). 
+* Metric names MAY use any UTF-8 characters.
+* Metric names SHOULD match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*` for the best experience and compatibility (see the warning below). Metric names outside of that set will require escaping e.g. when used in PromQL (see the [UTF-8 guide](../guides/utf8.md#querying)).
```
beorn7, line 22: escaping → quoting

(I would say escaping is technically incorrect.)

content/docs/concepts/data_model.md
```diff
@@ -17,42 +17,39 @@ Every time series is uniquely identified by its metric name and optional key-val
...
- * The change of any label's value, including adding or removing labels, will create a new time series.
- * Labels may contain ASCII letters, numbers, as well as underscores. They must match the regex `[a-zA-Z_][a-zA-Z0-9_]*`. 
- * Label names beginning with `__` (two "_") are reserved for internal use.
- * Label values may contain any Unicode characters.
- * Labels with an empty label value are considered equivalent to labels that do not exist.
+Labels enable Prometheus's dimensional data model to identify any given combination of labels for the same metric name. It identifies a particular dimensional instantiation of that metric (for example: all HTTP requests that used the method `POST` to the `/api/tracks` handler). The query language allows filtering and aggregation based on these dimensions. The change of any label's value, including adding or removing labels, will create a new time series.
+
+* Label names MAY use any UTF-8 characters. 
+* Label names beginning with `__` (two "_") MUST be reserved for internal Prometheus use.
```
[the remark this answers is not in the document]
beorn7: Or maybe even

```suggestion
* Label names beginning with `__` (two underscores) MUST be reserved for internal Prometheus use.
```

content/docs/concepts/data_model.md
```diff
@@ -17,42 +17,39 @@ Every time series is uniquely identified by its metric name and optional key-val
...
+* Label names SHOULD match the regex `[a-zA-Z_][a-zA-Z0-9_]*` for the best experience and compatibility (see the warning below). Label names outside of that regex will require escaping e.g. when used in PromQL (see the [UTF-8 guide](../guides/utf8.md#querying)).
```
beorn7, line 32: escaping → quoting, as above.

content/docs/concepts/data_model.md
```diff
@@ -17,42 +17,39 @@ Every time series is uniquely identified by its metric name and optional key-val
...
+* Label values MAY contain any UTF-8 characters.
+* Labels with an empty label value are considered equivalent to labels that do not exist.
 
+WARNING: The [UTF-8](../guides/utf8.md) support for metric and label names was added relatively recently in Prometheus v3.0.0. It might take time for the wider ecosystem (downstream PromQL compatible projects and vendors, tooling, 3P instrumentation, collectors, etc.) to adopt new escaping mechanisms, relaxed validation etc. For the best compatibility it's recommended to stick to the recommended ("SHOULD") character set.
```
beorn7, line 36: 3P → third-party

(Assuming that's what you mean…)

content/docs/concepts/data_model.md
```diff
@@ -61,3 +58,14 @@ this:
     api_http_requests_total{method="POST", handler="/messages"}
 
 This is the same notation that [OpenTSDB](http://opentsdb.net/) uses.
+
+Names with the UTF-8 characters outside the recommended set must be escaped, using
```
beorn7, line 62:
```suggestion
Names with UTF-8 characters outside the recommended set must be quoted, using
```

For completion you could also mention that quotes within the quotes must be escaped (e.g. `{"my metric name \"with quotes\""}`).

This also demonstrates the difference between escaping and quoting.

content/docs/concepts/data_model.md
```diff
@@ -61,3 +58,14 @@ this:
...
+this notation:
+
+    {"<metric name>", <label name>="<label value>", ...}
+
+Since an internal representation of a metric name is in a form of a label pair
```
beorn7, line 67:
```suggestion
Since metric name are internally represented as a label pair
```

content/docs/practices/naming.md
beorn7: These words have been added as a reaction to misunderstandings in the past. We can move it elsewhere, but it should stay prominent enough to avoid repeating the misunderstanding.

There _is_ an example below: 

```
* <code>http\_requests\_<b>total</b></code>
    (for a unit-less accumulating count)
```

content/docs/practices/naming.md
```diff
@@ -61,6 +61,33 @@ meaningful, split the data up into multiple metrics. For example, having the
...
+argument is that those pieces of information are already defined somewhere else (e.g. schema, 
+metadata, other labels, etc.).
+
+Prometheus strongly recommends including unit and type in a metric name, even if you store that
+information elsewhere, because of the two main reasons:
+
+* **Metric consumption reliability and UX**: When interacting with a modern UI to 
+use such a metric in PromQL, it's possible to display rich information about the metric's type and unit
+(autocompletion, overlays, pop-ups). Unfortunately, interactive, adhoc querying in a powerful UI is only
+a tiny fraction of the metric consumption use journeys (CUJs). Metric consumption ecosystem is vast. Majority
```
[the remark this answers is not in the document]
beorn7: Metric consumption ecosystem is vast. → The metric consumption ecosystem is vast.

content/docs/practices/naming.md
```diff
@@ -61,6 +61,33 @@ meaningful, split the data up into multiple metrics. For example, having the
...
+of the consumption comes in a form of the plain YAML configuration for variety of observability tools like
+alerting, recording, autoscaling, dashboard, analysis, processing, etc. It's **critical**, especially 
+during monitoring/SRE incident practices to look on PromQL expressions in the plain YAML and understand
```
beorn7, line 80:
```suggestion
during monitoring/SRE incident practices to look on PromQL expressions in plain YAML and understand
```

But I wouldn't even restrict this to YAML. You might encounter the naked metric name outside of YAML, too.

content/docs/practices/naming.md
beorn7: Maybe even

```suggestion
### Why include unit and type suffixes in metric names?
```

And then later explain that there is `_total` for counters and `_info` for info metrics.
I would also suggest to put a literal suffix with the `_` always in back quotes for better readability.

beorn7 (changes requested):
Thanks for doing this.

content/docs/concepts/data_model.md
```diff
@@ -17,42 +17,39 @@ Every time series is uniquely identified by its metric name and optional key-val
...
- * Specify the general feature of a system that is measured (e.g. `http_requests_total` - the total number of HTTP requests received). 
- * Metric names may contain ASCII letters, digits, underscores, and colons. It must match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*`.
-   
-Note: The colons are reserved for user defined recording rules. They should not be used by exporters or direct instrumentation.
-
+* Metric names SHOULD specify the general feature of a system that is measured (e.g. `http_requests_total` - the total number of HTTP requests received). 
+* Metric names MAY use any UTF-8 characters.
+* Metric names SHOULD match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*` for the best experience and compatibility (see the warning below). Metric names outside of that set will require escaping e.g. when used in PromQL (see the [UTF-8 guide](../guides/utf8.md#querying)).
 
+NOTE: The colons are reserved for user defined recording rules. They should not be used by exporters or direct instrumentation.
```
ArthurSens, line 24:
```suggestion
NOTE: The colons are reserved for user defined recording rules. They SHOULD not be used by exporters or direct instrumentation.
```

ArthurSens (commented):
besides all that has already been said, found one other thing

content/docs/concepts/data_model.md
beorn7: In fact SHOULD NOT.

content/docs/concepts/data_model.md
bwplotka: sorry, Google jargon 👍🏽

content/docs/concepts/data_model.md
bwplotka: Nice! Thanks, applying, but I would vote to keep it simple and not mention escaping quoting - hopefully not very common problem.

content/docs/practices/naming.md
bwplotka: "narrow sense units" (: 

While odd it makes sense for names and it's something we recommend for years. I will keep this form for now, we can change later.

content/docs/practices/naming.md
```diff
@@ -61,6 +61,33 @@ meaningful, split the data up into multiple metrics. For example, having the
...
+the underlying metric type and unit you work with.
+* **Metric collisions**: With the enormous adoption and metric changes over time, there are cases where lack
+of unit and type information in the metric name will cause certain series to collide (e.g. `process_cpu` for seconds and milliseconds).
+
+NOTE: While Prometheus community experiments with ways to reduce the above negative effects
```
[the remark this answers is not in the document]
bwplotka: True. However I don't want to give impression there is a solution planned that will make Prometheus to NOT recommend type and unit in names.

Maybe to NOT give this impression and NOT go into details I can skip this NOTE completely (:

ArthurSens (approved):
thank you!

beorn7 (approved):

PR #2637, opened 2025-04-30
Remove some obsolete TODOs from histogram spec

beorn7:
- Counter reset detection in subqueries fixed in https://github.com/prometheus/prometheus/pull/15987
- Optional translation of OTel fixed histograms to NHCB in https://github.com/prometheus/prometheus/issues/15022

The OTLP receiver is documented by now, but it is integrated into the respective sections (feature flags, config, general OTel guidelines) that linking to a particular part doesn't make sense.


beorn7:
@NeerajGartia21 you might want to review this, too.

NeerajGartia21:
Looks good to me!

beorn7:
Thanks.

This is just removing a TODO from the spec, so I merge now. @krajorama if you find any inaccuracies, let me know and I can fix them separately.

PR #2641, opened 2025-05-02
Fix various aspects of OTel survey banner styling / display

juliusv:
* Remove spurious margins to vertically center the banner text
* Use less in-your-face background color for banner
* Use icon instead of literal "X" for close button
* Don't show the banner on every docs page, only on the homepage
* Keep banner hidden via localStorage once it's dismissed once
* Remove italic style from link (italic usually just makes things harder to read)

Before:

image

After:

image

### DIFF
--- a/content/css/docs.css
+++ b/content/css/docs.css
@@ -79,8 +79,7 @@ body {
 }
 
 .banner {
-  background-color: #ffc107;
-  text-align: center;
+  background-color: rgb(252, 232, 170);
   width: 100%;
   display: flex;
   gap: 32px;
@@ -100,12 +99,11 @@ body {
 
 .banner a {
   text-decoration: underline;
-  font-style: italic;
 }
 
 .banner h3 {
+  margin: 1.2em 0;
   font-size: 16px;
-  font-weight: 600;
   font-family: Lato, sans-serif;
 }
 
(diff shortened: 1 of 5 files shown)


juliusv:
@ArthurSens @Wineshuga Heya, I just accidentally discovered the new banner on prometheus.io and trying to improve a few things here that were off styling-wise and in the way it shows up again and again on every docs page (instead of only on the front page, and only when it hasn't been dismissed once already).

On a general note, could you please mention one of the maintainers of the docs repo (usually me) when doing a change like this to the front page?

ArthurSens:
> On a general note, could you please mention one of the maintainers of the docs repo (usually me) when doing a change like this to the front page?

Thank you and we'll definitely ping you next time. I wasn't aware you paid close attention here, sorry

ArthurSens:
> @ArthurSens @Wineshuga Heya, I just accidentally discovered the new banner on prometheus.io and trying to improve a few things here that were off styling-wise and in the way it shows up again and again on every docs page (instead of only on the front page, and only when it hasn't been dismissed once already).


Could we keep it on all pages instead of only the front page, though? I believe the front page doesn't get much traffic, so showing it only there wouldn't help much on recruiting 😬 

I totally agree that once the banner is closed, it shouldn't show up anywhere else. Is that a possibility?

ArthurSens (approved):
Thank you!

PR #2645, opened 2025-05-06
Add Neo4j/Apoc exporter

stscoundrel:


juliusv:
Thanks, and sorry for nobody responding. Could you rebase the PR on the latest `main`? With the new site, the docs directory has now moved from `content/docs` to just `docs`.

Regarding the exporter itself, I noticed that https://github.com/stscoundrel/neo4j-apoc-exporter/blob/main/src/main/java/io/github/stscoundrel/MetricsRegistry.java goes through the pain of doing a manual mapping to Prometheus metrics. In case such a manual mapping is maintained anyway, it's usually a good idea to then also generate Prometheus metrics that stick to the metrics and label naming best practices (see https://prometheus.io/docs/practices/naming/), so for example, `neo4j_monitor_store_logSize` would rather be called something like `neo4j_monitor_store_log_size_bytes` (or whatever the unit is). And for example for a metric like `neo4j_monitor_tx_rolledBackTx` I am not 100% sure if it represents a counter, so if it was called `neo4j_monitor_tx_rollbacks_total` (as per naming best practices), I would have more confidence in interpreting the metric name.

jan--f:

PR #2653, opened 2025-05-21
docs[open-telemetry]: Add specifications for content negotation and escaping, and docs for translation strategies for Otel metric name translation

ywwg:
This updates the docs to represent the current state of otel configuration options. It also creates specification files to document scraping content negotiation and escaping schemes.


ArthurSens (approved):
LGTM

ArthurSens (approved):

content/docs/concepts/content_negotiation.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Supported Protocols
+
+The following protocols are be supported by Prometheus:
+
+1. `PrometheusProto` - Binary protobuf format
+2. `PrometheusText0.0.4` - Prometheus text format version 0.0.4
+3. `PrometheusText1.0.0` - Prometheus text format version 1.0.0
+4. `OpenMetricsText0.0.1` - OpenMetrics text format version 0.0.1
+5. `OpenMetricsText1.0.0` - OpenMetrics text format version 1.0.0
```
ywwg, line 31: do we need this at all? I do find it useful to cover the currently-implemented options

content/docs/concepts/escaping_schemes.md
```diff
@@ -0,0 +1,83 @@
...
+- Other non-legacy characters MUST be replaced with underscores
+- Example: `metric.name.with.dots` becomes `metric_dot_name_dot_with_dot_dots`
+
+### Value Encoding Escaping (values)
+
+**Header Value**: `escaping=values`
+
+**Behavior**:
+- The name MUST be prefixed with `U__`
+- Each non-legacy character MUST be replaced with its Unicode code point in
```
ArthurSens, line 64:
```suggestion
- Each character that is not part of the legacy character set (a-zA-Z0-9_:) MUST be replaced with its Unicode code point in
```

I think it would be good to always mention the regex when we say "legacy". Legacy is clear to us maintainers, but probably don't mean anything to readers that don't have the historical context

ArthurSens (commented):
Thank you for the effort writing this :) 

I have some suggestions 👇

content/docs/guides/utf8.md
```diff
@@ -82,6 +82,14 @@ setting the `escaping` header to a different value.
   underscores are replaced with double underscores. This mode allows for full
   round-tripping of UTF-8 names with a legacy system.
 
+Enabling UTF-8 support in Prometheus indicates that the system is *capable* of
```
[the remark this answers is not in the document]
ywwg: yeah I can tweak that.  There are cases where the spec isn't necessarily referencing prometheus, it could be referencing an endpoint like the otel SDK.  So there are places where we want to be more generic.  But I think this one can be switched over.

content/docs/concepts/escaping_schemes.md
```diff
@@ -0,0 +1,83 @@
...
+  hexadecimal, surrounded by underscores
+- Single underscores MUST be replaced with double underscores
+- Example: `metric.name` becomes `U__metric_2E_name` (where 2E is the hex Unicode code point for '.')
+
+## Default Behavior
+
+If no escaping scheme is specified:
+1. For legacy validation (metric_name_validation_scheme=legacy):
```
dashpole, line 72: Is this a reference to the constant in prometheus/common?  If so, we should probably specify what the default value of that constant is. But generally, It seems better to just clearly write what the default overall escaping scheme is if there is no `escaping` header provided.

content/docs/guides/opentelemetry.md
```diff
@@ -147,16 +147,18 @@ For each of a resource's OTel metrics, Prometheus converts it to a corresponding
 
 ## UTF-8
 
-From the 3.x version, Prometheus supports UTF-8 for metric names and labels, so [Prometheus normalization translator package from OpenTelemetry](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/translator/prometheus) can be omitted.
+From the 3.x version, Prometheus supports UTF-8 for metric names and labels, so [Prometheus normalization translator package from OpenTelemetry](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/translator/prometheus) can be omitted. Note that when Prometheus announces through content negotiation that it allows UTF-8 characters, this does not require that metric names contain previously-unsupported characters. The OTLP metrics may be converted in several different ways, depending on the configuration of the endpoint. So while UTF-8 is enabled by default in Prometheus storage and UI, you need to set the `translation_strategy` for OTLP metrics receiver, which by default is set to old normalization `UnderscoreEscapingWithSuffixes`.
```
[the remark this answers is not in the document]
ywwg: applied manually

content/docs/concepts/escaping_schemes.md
```diff
@@ -0,0 +1,83 @@
...
+
+**Header Value**: `escaping=allow-utf-8`
+
+**Behavior**:
+- Metric and label names MUST be valid UTF-8 strings
+- No character escaping is performed
+- Because names are quoted in the exposition text formats, UTF-8 strings MUST
+  be escaped using Go-style backslash escape codes for special characters,
+  particularly quote and whitespace characters
+- Names MAY contain any valid UTF-8 character
```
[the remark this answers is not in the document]
ywwg: cursor dropped this in, yeah it does seem superfluous.  applied manually.

content/docs/concepts/escaping_schemes.md
ywwg: yes this section is talking about  the Prometheus defaults.  I'm not sure if we need this block at all, it seems outside what a spec needs to define and is more a choice in Prometheus.

dashpole (approved):

content/docs/concepts/escaping_schemes.md
dashpole: I think we should define client behavior if no escaping header is provided somewhere

ywwg: done

juliusv:
Hey hey, I just saw this as I was launching the new prometheus.io homepage at the Berlin Prometheus meetup and had to merge it into my PR branch for that. I noticed a few things:

* The `sort_rank` frontmatter fields need to be adjusted relative to their existing sibling documents in the same folder, I took the liberty to fix those in https://github.com/prometheus/docs/commit/60c4d8863548f22edcc0cc232eed96ed6a14ffd8
* The documents are currently under "Concepts", but that top-level menu point is meant for basic Prometheus concepts  that every Prometheus beginner should familiarize themselves with. The new docs are more detailed specifications that only some experts need to know, and not something a new Prometheus user should be confronted with. I talked with @beorn7 about it at the meetup as well - how about moving these to "Instrumenting", since they are both about that, and there are similar docs like that already under there? It almost sounds like the two docs are related to the existing "Exposition Formats" doc under there. I wonder if opening a new subsection under "Instrumenting" for all three would make sense, or otherwise combining them somehow?
* The titles of the documents are quite long - that's fine for the page itself, but nav and prev/next pagination buttons weren't really made with such long titles in mind. You can set an alternate title to be shown in the side nav using the `nav_title` frontmatter field, which could be shorter.

juliusv:
I'll make a PR with a beginning of a move to the "Instrumenting" section. It does seem like both new sections should be referenced from (if not merged with) https://prometheus.io/docs/instrumenting/exposition_formats/.

juliusv:
See: https://github.com/prometheus/docs/pull/2657

PR #2657, opened 2025-05-27
Move some new docs from "Concepts" to "Instrumenting", fix up titles

juliusv:
See https://github.com/prometheus/docs/pull/2653#issuecomment-2913952990

Also change the titles:

* "Prometheus" is redundant in the context of Prometheus, let's make it more about UTF-8
* Make the nav titles shorter so they fit nicer into the nav (title on the page itself can still be long)
* We don't use title casing (at least in the "Instrumenting" section), so lower-casing the titles

### DIFF
--- a/docs/instrumenting/content_negotiation.md
+++ b/docs/instrumenting/content_negotiation.md
@@ -1,9 +1,9 @@
 ---
-title: Prometheus Scraping Protocol Content Negotiation
-sort_rank: 5
+title: Content negotiation
+sort_rank: 8
 ---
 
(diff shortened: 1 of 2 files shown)


juliusv:
As mentioned in https://github.com/prometheus/docs/pull/2653#issuecomment-2913952990, there needs to be more work done to link these from the existing "Exposition formats" sibling page and/or potentially merge some content with it.

ArthurSens (approved):
The move makes sense to me, thanks for the suggestion!

What do you think @ywwg ?

PR #2658, opened 2025-05-27
Wrap search results properly in narrow window

juliusv:


Nexucis (approved):
a single line that changes the life of so many ❤️

PR #2662, opened 2025-05-28
Standardize H1 title use in local Markdown docs

juliusv:
See https://groups.google.com/g/prometheus-developers/c/cwL3cM66Em8

In addition, I took the editorial liberty to make slight other modifications to some of the titles, where it seemed necessary. Also had to adjust deeper header levels further into the doc when the page started off with the wrong one.

And I fixed one completely broken list that was close to a title :P

The PR also fixes up the React code to already handle the new titling scheme for local docs outside of the Markdown content.

### DIFF
--- a/docs/concepts/data_model.md
+++ b/docs/concepts/data_model.md
@@ -3,8 +3,6 @@ title: Data model
 sort_rank: 1
 ---
 
-# Data model
-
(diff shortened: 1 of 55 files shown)


kakkoyun (approved):
LGTM

PR #2656, opened 2025-05-27
Merge new Next.js docs site and completely wipe old one

juliusv:

PR #2666, opened 2025-06-04
Update blog RSS feed with rendered content

sysadmind:
After migrating to Next.js in #2656, the blog RSS feed was missing the rendered markdown content for the blog. This uses the unified package, similar to how we render the blog entires, to render the entries to HTML inside the feed.


juliusv:
👍 Thanks, looks good :)

PR #2669, opened 2025-06-05
Add NetBird exporter to the exporters list in documentation

matanbaruch:
This pull request includes a small addition to the `docs/instrumenting/exporters.md` file. The change adds a new entry for the NetBird exporter to the list of available exporters.

### DIFF
--- a/docs/instrumenting/exporters.md
+++ b/docs/instrumenting/exporters.md
@@ -242,6 +242,7 @@ wide variety of JVM-based applications, for example [Kafka](http://kafka.apache.
    * [Meteor JS web framework exporter](https://atmospherejs.com/sevki/prometheus-exporter)
    * [Minecraft exporter module](https://github.com/Baughn/PrometheusIntegration)
    * [Minecraft exporter](https://github.com/dirien/minecraft-prometheus-exporter)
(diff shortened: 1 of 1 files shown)


juliusv:
Thanks! Just a question, before adding to the list: There's various metrics in the exporter that end with a `_total` suffix that according to the metric naming conventions (https://prometheus.io/docs/practices/naming/) should only be used for counter metrics. But it's not clear to me e.g. if `netbird_peers_total` or `netbird_groups_total` are really cumulative counts over time, or whether they represent a current count (in which case they would be a gauge metric and should not have a `_total` suffix)? Could you clarify that?

matanbaruch:
> Thanks! Just a question, before adding to the list: There's various metrics in the exporter that end with a `_total` suffix that according to the metric naming conventions (https://prometheus.io/docs/practices/naming/) should only be used for counter metrics. But it's not clear to me e.g. if `netbird_peers_total` or `netbird_groups_total` are really cumulative counts over time, or whether they represent a current count (in which case they would be a gauge metric and should not have a `_total` suffix)? Could you clarify that?

@juliusv  You are absolutely right. I released a fix for this.
https://github.com/matanbaruch/netbird-api-exporter/pull/10

juliusv:
:+1: Cool, thanks :)

PR #2672, opened 2025-06-11
test(om2): test that examples follow the spec and fix if not

krajorama:
I did not want to clutter the examples with adding "# EOF" to all, so I made a new marker to explicitly add it on request.

There was only one faulty example where we had an extra space, see the Exemplars section.


docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -781,7 +781,7 @@ foo_bucket{le="10"} 17 # {trace_id="oHg5SJYRHA0"} 9.8 1520879607.789
 foo_bucket{le="+Inf"} 17
 foo_count 17
 foo_sum 324789.3
-foo_created  1520430000.123
```
krajorama, line 784: Note to reviewers: this was the one mistake in the examples.

PR #2677, opened 2025-06-13
feat(om2): relax the rules around unit and total suffixes

krajorama:
Fixes: https://github.com/prometheus/OpenMetrics/issues/286
Fixes: https://github.com/prometheus/OpenMetrics/issues/282

### DIFF
--- a/docs/specs/om/open_metrics_spec_2_0.md
+++ b/docs/specs/om/open_metrics_spec_2_0.md
@@ -32,6 +32,8 @@ author:
 
 Created in 2012, Prometheus has been the default for cloud-native observability since 2015. A central part of Prometheus' design is its text metric exposition format, called the Prometheus exposition format 0.0.4, stable since 2014. In this format, special care has been taken to make it easy to generate, to ingest, and to understand by humans. As of 2020, there are more than 700 publicly listed exporters, an unknown number of unlisted exporters, and thousands of native library integrations using this format. Dozens of ingestors from various projects and companies support consuming it.
 
+Since 2019, OpenTelemetry has emerged as an alternative way to define a metrics data model. There is a need to be able to expose that data model within OpenMetrics, such that the transport of OpenTelemetry metrics signal is feasable and doesn't violate the OpenMetrics specification.
+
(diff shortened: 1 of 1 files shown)


krajorama:
I agree with [this comment](https://github.com/prometheus/OpenMetrics/issues/282#issuecomment-2532185128) from @beorn7 that the metric family name should be the same in the metadata and what we write in the metric(point)s. 
```
# TYPE foo_total counter
# HELP foo_total A help string.
foo_total 42
```
or
```
# TYPE foo counter
# HELP foo A help string.
foo 42
```

but to be able to make a generic statement, we need these to be fixed:
https://github.com/prometheus/OpenMetrics/issues/285 Created timestamp not exposed as "foo_create" anymore.
https://github.com/prometheus/OpenMetrics/issues/283 Complex types for classic histograms and summaries.

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -116,15 +118,23 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
 
 ##### Name
 
-MetricFamily names are a string and MUST be unique within a MetricSet. Names SHOULD be in snake_case. Metric names MUST follow the restrictions in the ABNF section.
+MetricFamily names are a string and combined with the TYPE and UNIT MUST be unique within a MetricSet.
+
+MetricFamily names SHOULD be uniq within a MetricSet. This rule relaxes the requirement in previous versions of OpenMetrics and is intended to facilitate exposing OpenTelemetry data model. For other use cases it is highly recommended to continue to adhere to this rule.
```
dashpole:
```suggestion
MetricFamily names SHOULD be unique within a MetricSet. This rule relaxes the requirement in previous versions of OpenMetrics and is intended to facilitate exposing OpenTelemetry data model. For other use cases it is highly recommended to continue to adhere to this rule.
```

dashpole (commented):
FWIW, https://github.com/prometheus/OpenMetrics/issues/286 and https://github.com/prometheus/OpenMetrics/issues/282 only ask for it to be allowed to not use suffixes. IMO the primary use-case for allowing the same name, but different unit and type is for federation-like endpoints (e.g. the prometheusexporter in the collector) where metric data from multiple applications is combined. We could require OTel SDK users to resolve naming collisions using views.  I'm still OK with this change, but want to make sure we are clear on what the requirements from OTel are.

krajorama:
This PR is mixing up relaxed suffix rules and federation. I'll open a PR based on #2750 instead.

PR #2679, opened 2025-06-16
Classic histogram and summary as complex types

krajorama:
Ref: https://github.com/prometheus/OpenMetrics/issues/283


docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -774,13 +806,7 @@ The "0.01" bucket has no Exemplar. The 0.1 bucket has an Exemplar with no Labels
...
 ```openmetrics-add-eof
 # TYPE foo histogram
-foo_bucket{le="0.01"} 0
-foo_bucket{le="0.1"} 8 # {} 0.054
-foo_bucket{le="1"} 11 # {trace_id="KOO5S4vxi0o"} 0.67
-foo_bucket{le="10"} 17 # {trace_id="oHg5SJYRHA0"} 9.8 1520879607.789
-foo_bucket{le="+Inf"} 17
-foo_count 17
-foo_sum 324789.3
+foo {count:17,sum:324789.3,bucket:[0.01:0,0.1:8,1.0:11,10.0:17,+Inf:17]} # {} 0.054 1520879607.7 # {trace_id="KOO5S4vxi0o"} 0.67 1520879602.890 # {trace_id="oHg5SJYRHA0"} 9.8 1520879607.789
```
krajorama: The problem with exemplars is that if this line is converted into native histogram with custom buckets during scrape, then we require exemplars with time stamp. However if this line is converted into classic histogram during scrape, then there is no such requirement. See https://github.com/prometheus/proposals/blob/main/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md#exemplars.

Maybe a way around is to not require the timestamp in exemplars for complex types after all, but state that any exemplar without timestamp may be discarded.

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -62,7 +62,15 @@ This section MUST be read together with the ABNF section. In case of disagreemen
...
 
-Metric values in OpenMetrics MUST be either floating points or integers. Note that ingestors of the format MAY only support float64. The non-real values NaN, +Inf and -Inf MUST be supported. NaN MUST NOT be considered a missing value, but it MAY be used to signal a division by zero.
+Metric values in OpenMetrics MUST be either numbers or complex data types.
+
+Numbers MUST be either floating points or integers. Note that ingestors of the format MAY only support float64. The non-real values NaN, +Inf and -Inf MUST be supported. NaN MUST NOT be considered a missing value, but it MAY be used to signal a division by zero.
+
+Complex data types MUST contain all information necessary to recreate a sample of a Metric Type, with the exception of Created Value and Exemplars.
+
+List of complex data types:
+ - Integer counter classic histogram for the Metric Type Histogram.
```
dashpole: nit: are these float histograms, rather than integer histograms?

dashpole (dismissed):

bwplotka (dismissed):
I would love that, but as we discussed on our sync - we need a bigger decision if we want to follow this path.

If we do, we likely need some rough `uncomplex` `inlined` format explained that is not meant to be used for scraping and can be surfaced with promtool etc.

docs/specs/om/open_metrics_spec_2_0.md
krajorama: Nope, histograms count observations, so they have integers.

Only if you expose result of calculations (recording rules) or aggregate histograms (federate/reduce/etc) do you need floats. Currently we're only targeting exposing metrics for instrumentation. We need to decide what other use cases to support.

dashpole (approved):

bwplotka:
Results are (expectedly) impressive: https://github.com/prometheus/prometheus/pull/17293

One observation is that buckets are a bit dense (readability). However we don't have spaces in any other place (e.g. labels)

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -1054,33 +1040,23 @@ The "0.01" bucket has no Exemplar. The 0.1 bucket has an Exemplar with no Labels
...
-foo {count:10,sum:1.0,schema:0,zero_threshold:1e-4,zero_count:0,positive_spans:[0:2],positive_buckets:[5,5]} st@1520430000.123 # {trace_id="shaZ8oxi"} 0.67 1520879607.789 # {trace_id="ookahn0M"} 1.2 1520879608.589
-foo_bucket{le="0.01"} 0 st@1520430000.123
-foo_bucket{le="0.1"} 8 st@1520430000.123 # {} 0.054
-foo_bucket{le="1"} 11 st@1520430000.123 # {trace_id="KOO5S4vxi0o"} 0.67
-foo_bucket{le="10"} 17 st@1520430000.123 # {trace_id="oHg5SJYRHA0"} 9.8 1520879607.789
-foo_bucket{le="+Inf"} 17 st@1520430000.123
-foo_count 17 st@1520430000.123
-foo_sum 324789.3 st@1520430000.123
+foo {count:17,sum:324789.3,schema:0,zero_threshold:1e-4,zero_count:0,positive_spans:[0:2],positive_buckets:[5,12]} st@1520430000.123 # {trace_id="shaZ8oxi"} 0.67 1520879607.789 # {trace_id="ookahn0M"} 1.2 1520879608.589
+foo {count:17,sum:324789.3,bucket:[0.01:0,0.1:8,1.0:11,10.0:17,+Inf:17]} st@1520430000.123 # {} 0.054 1520879607.7 # {trace_id="KOO5S4vxi0o"} 0.67 1520879602.890 # {trace_id="oHg5SJYRHA0"} 9.8 1520879607.789
```
bwplotka, line 1054: WG discussions:  what to do with float vs int histogram

Likely storage optimization, easy in the format.

For syntax:

A)Should we accept the float
B)If we do how:
*. B1) What if we accept floats e.g. and optimize on storage e.g. `count:17/17.0` `count:17.5` and buckets.
* B2) Strict mode - count decides if it's float or not.

Perhaps B1 or B2 if we want more efficient implementation
Perhaps for RC we do relax mode and try to implement?

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -754,38 +752,42 @@ MetricPoints MUST NOT be interleaved.
...
-foo_seconds_count{a="bb"} 0 123
-foo_seconds_sum{a="bb"} 0 123
-foo_seconds_count{a="bb"} 0 456
-foo_seconds_sum{a="bb"} 0 456
-foo_seconds_count{a="ccc"} 0 123
-foo_seconds_sum{a="ccc"} 0 123
-foo_seconds_count{a="ccc"} 0 456
-foo_seconds_sum{a="ccc"} 0 456
+# TYPE foo stateset
+foo{entity="controller",foo="a"} 1.0
```
bwplotka, line 766: what if we do stateset too:

```
# TYPE foo stateset
foo{entity="controller"} {states:["a":1.0, "bb":0.0, "ccc":0.0]}
...
```

bwplotka: One argument against moving to complex type here is the PromQL. We don't have plans to change PromQL for those so far (while we have plans for histograms). We don't have clear plans for summaries, so maybe that's fine 🙃

dashpole: It doesn't seem too far-fetched to me to add summary_XXX functions, similar to histogram if we went this path. In general, it seems like this is a good thing to do, just main question is whether it is worth the effort.

Another crazy idea is to add precomputed quantiles to histogram, remove the summary type entirely, and use the histogram functions. Summary users could then more easily migrate to histograms/exponential histograms without breaking.

dashpole:
This looks good to me overall. I think we can save float histograms for a follow-up.

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -498,9 +498,9 @@ normal-char = %x00-09 / %x0B-21 / %x23-5B / %x5D-D7FF / %xE000-10FFFF
 start-timestamp = %d115.116 "@" timestamp
 
 ; Complex values
-complex-value = nativehistogram
+complex-value = native-histogram / classic-histogram / classic-summary
```
bwplotka: From WG: Compound or Composite to consider

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -754,38 +752,42 @@ MetricPoints MUST NOT be interleaved.
...
-foo_seconds_sum{a="bb"} 0 456
-foo_seconds_count{a="ccc"} 0 123
-foo_seconds_sum{a="ccc"} 0 123
-foo_seconds_count{a="ccc"} 0 456
-foo_seconds_sum{a="ccc"} 0 456
+# TYPE foo stateset
+foo{entity="controller",foo="a"} 1.0
+foo{entity="controller",foo="bb"} 0.0
+foo{entity="controller",foo="ccc"} 0.0
+foo{entity="replica",foo="a"} 1.0
```
bwplotka, line 769: Why there are 2 1.0?

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -754,38 +752,42 @@ MetricPoints MUST NOT be interleaved.
...
-foo_seconds_count{a="ccc"} 0 456
+```openmetrics-add-eof
+# TYPE foo stateset
+foo{entity="controller",foo="a"} 1.0
+foo{entity="controller",foo="bb"} 0.0
+foo{entity="controller",foo="ccc"} 0.0
+foo{entity="replica",foo="a"} 1.0
+foo{entity="replica",foo="bb"} 0.0
+foo{entity="replica",foo="ccc"} 1.0
+foo{entity="controller",foo="a"} 1.0
```
bwplotka: Maybe you meant 

```suggestion
foo{entity="controller",foo="ddd"} 1.0
```

Otherwise it's a duplicate?

docs/specs/om/open_metrics_spec_2_0.md
bwplotka: WG Sync discussing: 

* It's tempting, but to discuss in separate PR.
* Otel does not have it, but some prev discussions happened.
* ENUM is subcase, it's 

* We don't know much about UX and usage to make clear decisions
* Ensure integer not float
* ST not relevant/useful here as those are gauges

**Cons for single:**
* Readability is impaired with single
* What you see what you query is violated unless we plan to change it. We don't know how it will look like.

**Cons for keeping multi-line:**
* Efficiency/reliability of parsing (label matching to realize if it's same "sample"). Efficiency of stateset is actually a main blocker for using them 
* Inconsistency 
* We need to have extra rule around interleaving

TODO: Add issue

bwplotka: For first RC we could keep it as-is

bwplotka: Issue to track this https://github.com/prometheus/OpenMetrics/issues/310

docs/specs/om/open_metrics_spec_2_0.md
krajorama: I've change it in a different way, by including another label.

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -955,54 +957,42 @@ An example of a Metric with no labels and a MetricPoint with Sum, Count and Star
...
-foo_bucket{le="1.0"} 10 st@1522430000.123
-foo_bucket{le="10.0"} 11 st@1520430123.123
-foo_bucket{le="100000.0"} 11 st@1521430010.123
-foo_bucket{le="1e+06"} 15 st@1520430301.123
-foo_bucket{le="1e+23"} 16 st@1521430001.123
-foo_bucket{le="1.1e+23"} 17 st@1522430220.123
-foo_bucket{le="+Inf"} 17 st@1520430000.123
-foo_count 17 st@1520430000.123
-foo_sum 324789.3 st@1520430000.123
+foo {count:17,sum:324789.3,bucket:[0.0:0,1e-05:0,0.0001:5,0.1:8,1.0:10,10.0:11,100000.0:11,1e+06:15,1e+23:16,1.1e+23:17,+Inf:17]} st@1520430000.123
```
bwplotka, line 990: Not for this PR, but one last thing to check is the readability/mode or accepting whitespaces (or at least  `[ \n\t].*`) inside `{ }`. We might want to compare pros and cons, but eating all those chars when in `{` is not too difficult (it might be prone to errors though).

This format asks for it (:

bwplotka:
```
foo {
  count: 17,
  sum: 324789.3,
  bucket: [ 
     0.0: 0,
     1e05: 0,
     0.0001: 5,
     ....
   ],
} st@1520430000.123
```

Pros:
* Looks nicer
Cons:
* Not very practical (grep)

Header or tool (CLI)


I like the CLI idea

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -764,38 +762,42 @@ MetricPoints MUST NOT be interleaved.
 A correct example where there were multiple MetricPoints and Samples within a MetricFamily would be:
```
bwplotka:
```suggestion
A correct example where there were multiple MetricPoints and Samples within a MetricFamily would be with StateSet:
```

docs/specs/om/open_metrics_spec_2_0.md
bwplotka: Idea from @krajorama allow whitespaces, but SHOULD without?

Consensus among me and @krajorama : Keep it simple and deterministic and do not allow whitespaces for now, have a CLI tool for it.

bwplotka: From WG: Machine readability = human readability is a nice aspect (reliability, confusion).

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -955,54 +957,42 @@ An example of a Metric with no labels and a MetricPoint with Sum, Count and Star
...
 An example of a Metric with no labels and a MetricPoint with two quantiles and Start Timestamp values:
 
 ```openmetrics-add-eof
 # TYPE foo summary
-foo{quantile="0.95"} 123.7 st@1520430000.123
-foo{quantile="0.99"} 150.0 st@1520430000.123
+foo {count:0,sum:0.0,quantile:[0.95:123.7,0.99:150]} st@1520430000.123
 ```
 
 Quantiles MAY be in any order.
```
dashpole: Can we require them to be ordered now?

krajorama: yes, updated. Also some other missing pieces. (github is down currently, hence the ci fails)

bwplotka:
The failure relates to AM release... 🤔 

7:00:32 AM: Checking out release-0.31 of prometheus/alertmanager from ./generated/repos/prometheus/alertmanager.git into /opt/build/repo/generated/repo-docs/prometheus/alertmanager/0.31...
7:00:32 AM: Creating new worktree...
7:00:32 AM: fatal: invalid reference: release-0.31
7:00:32 AM: node:internal/errors:983
7:00:32 AM:   const err = new Error(message);

bwplotka (approved):
Let's Go!

PR #2680, opened 2025-06-16
fix(abnf-checker): return error on example error

krajorama:
We just printed the error before but not set exit code.


krajorama:
See testrun in last but one commit: https://github.com/prometheus/docs/actions/runs/15679291421/job/44166744492

PR #2684, opened 2025-06-28
Update basic-auth.md

erselbey:


jan--f:
Hello from the bug scrub!
Using `getpass4` over the stdlib `getpass` seems unnecessary.

PR #2685, opened 2025-07-01
refactor(om2): rename, align Created Value to Created Timestamp

krajorama:
I think it's a bit confusing when we want to refer to the value of a Created Value, for example when we list values like Sum, Count, Created Value values.

Also why isn't the token for the Created Value cv@ ? Create Timestamp seems cleaner to me.

For your consideration :)


ArthurSens (approved):
Yeah, that's a lot better. LGTM

jesusvazquez (approved):

PR #2691, opened 2025-07-15
Add Guidance wrt Labelling to Naming and Rules Best Practices

### LINKED ISSUE #2690: Best Practice Docs Don't Call Out the Importance of Job Label
https://prometheus.io/docs/practices/naming/ and https://prometheus.io/docs/practices/rules/ should call out the importance of the `job` label, particularly in a multi-tenant prometheus installation.

https://prometheus.io/docs/practices/naming/ does not mention the job label at all.

https://prometheus.io/docs/practices/rules does mention

```
Always specify a without clause with the labels you are aggregating away. This is to preserve all the other labels such as job, which will avoid conflicts and give you more useful metrics and alerts.
```

and all it's examples include an explicit job label

But that's about as explicit as it gets.

conallob:
Add Guidance wrt Labelling to Naming and Rules Best Practices to docs/practices/naming.md and docs/practices/rules.md, specifically:

* The primary purposes of `job` and `instance`
* Include WARNINGS about accidentally stripping the `job` label, especially in multi-tenant systems

This Fixes https://github.com/prometheus/docs/issues/2690


conallob:
Friendly ping @SuperQ @beorn7 ?

conallob:
Obligatory post-it note reminder: https://photos.app.goo.gl/Bkfir4wRiLtNVG4W8

beorn7:
With my current patchy availability, there is little chance I get to this anytime soon. Maybe @juliusv has a qualified opinion here?

conallob:
Friendly ping @SuperQ @juliusv

andrechalella:
Hey @conallob, congrats for the nice PR!

Just one thing: maybe you could fix the `eaach` typo in

>    * The `job` label is a primary key to differentiate metrics from eaach other.

There is a small confusion I would love to see fixed, in a paragraph just below one of your edits. It is:

> To keep the operations clean, `_sum` is omitted if there are other operations,
> as `sum()`.

I don't understand the `as sum()` part. Like, "`x` is omitted if there are other operations such as `x`"? It doesn't make sense to me, in a very basic way. I know it's out of the scope of this PR, but maybe you could touch it to clarify.

conallob:
> Hey @conallob, congrats for the nice PR!
> 
> Just one thing: maybe you could fix the `eaach` typo in
> 
> > * The `job` label is a primary key to differentiate metrics from eaach other.
> 
> There is a small confusion I would love to see fixed, in a paragraph just below one of your edits. It is:

Fixed the typo.
 
> > To keep the operations clean, `_sum` is omitted if there are other operations,
> > as `sum()`.
> 
> I don't understand the `as sum()` part. Like, "`x` is omitted if there are other operations such as `x`"? It doesn't make sense to me, in a very basic way. I know it's out of the scope of this PR, but maybe you could touch it to clarify.

I'm afraid that best practice is unrelated.

It also makes sense as written, once you've written enough rules. It's weighing up the trade-off between tracking the chain of operations across a pipeline of rules vs the rule name growing unwieldy. Many of these best practices trace back to specific philosophies from Prometheus' predecessor.

If you still think it needs a polish, please a separate doc bug.

conallob:
Ping @juliusv , since @SuperQ is currently unavailable for life reasons

docs/practices/naming.md
```diff
@@ -80,7 +80,18 @@ the underlying metric type and unit you work with.
 * **Metric collisions**: With growing adoption and metric changes over time, there are cases where lack
 of unit and type information in the metric name will cause certain series to collide (e.g. `process_cpu` for seconds and milliseconds).
 
-## Labels
+### Important Labels
```
SuperQ: Let's not rename this section, so that we don't break inbound relative links.

```suggestion
### Labels
```

conallob: Oops, my bad. Fixed.

docs/practices/naming.md
```diff
@@ -80,7 +80,18 @@ the underlying metric type and unit you work with.
 * **Metric collisions**: With growing adoption and metric changes over time, there are cases where lack
 of unit and type information in the metric name will cause certain series to collide (e.g. `process_cpu` for seconds and milliseconds).
 
-## Labels
+### Labels
```
SuperQ:
```suggestion
## Labels
```

docs/practices/rules.md
```diff
@@ -19,6 +19,8 @@ This page documents proper naming conventions and aggregation for recording rule
 Keeping the metric name unchanged makes it easy to know what a metric is and
 easy to find in the codebase.
 
+IMPORTANT: `job` label acts as a primary key. It is **strongly** recommended that you use it to scope your PromQL expressions to the system you are monitoring.
```
SuperQ: This is misleading. Prometheus doesn't have the concept of "primary key". Not even metric names are a "primary key".

conallob: Fair, especially since folks used to SQL DBs will jump to the conclusion that it's a SQL DB, which it isn't.

Iterated on the language to avoid creating ambiguity

conallob:
PTAL

conallob:
Friendly ping?

conallob:
Friendly, you're not at SRECon EMEA this week, ping?

docs/practices/naming.md
```diff
@@ -82,6 +82,17 @@ of unit and type information in the metric name will cause certain series to col
...
 ## Labels
 
+* `job`
+   * The `job` label is one of the few ubiquitious labels, set at scrape time, and is used to identify metrics scraped from the same target/exporter.
+   * If not specified in PromQL expressions, they will match unrelated metrics with the same name. This is especially true in a multi system or multi tenant installation
+
+WARNING: When using `without`, be careful not to strip out the `job` label accidentally.
+
+* `instance`
+   * The `instance` label will include the `ip:port` what was scraped, providing a crucial breadcrumb for debugging scrape time issues
```
SuperQ:
```suggestion
   * The `instance` label by default will include the `ip:port` what was scraped.
```

docs/practices/naming.md
```diff
@@ -82,6 +82,17 @@ of unit and type information in the metric name will cause certain series to col
 
 ## Labels
 
+* `job`
```
SuperQ: I think this section needs a bit of rework. This section is about best practices for label naming.

In order to explain this we should probably talk about and reference where various labels come from. Just explaining the label(s) doesn't give the reader enough context.

```suggestion
Prometheus labels can come from both the target and from [relabeling in discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) as well as from the target itself.

By default Prometheus configures two primary discovery target labels.

* `job`
```

docs/practices/naming.md
```diff
@@ -82,6 +82,17 @@ of unit and type information in the metric name will cause certain series to col
...
+   * The `job` label is one of the few ubiquitious labels, set at scrape time, and is used to identify metrics scraped from the same target/exporter.
```
SuperQ:
```suggestion
   * The `job`  is a default target label set by the scrape configs and is used to identify metrics scraped from the same target/exporter.
```

docs/practices/naming.md
```diff
@@ -82,6 +82,17 @@ of unit and type information in the metric name will cause certain series to col
...
+   * If not specified in PromQL expressions, they will match unrelated metrics with the same name. This is especially true in a multi system or multi tenant installation
```
SuperQ: I'm not sure this is really a useful note here, as this applies to all label matching.

```suggestion
```

docs/practices/naming.md
```diff
@@ -82,6 +82,17 @@ of unit and type information in the metric name will cause certain series to col
...
+
+WARNING: When using `without`, be careful not to strip out the `job` label accidentally.
+
```
SuperQ: This warning doesn't make a lot of sense to me. It has a high probability of being quoted as copy-pasta without being understood. Let's just drop it.

```suggestion
```

SuperQ (changes requested):

docs/practices/naming.md
conallob: It applies to all labels. But `job` and `instance` are two uniform labels found on every metric, including ubiquitous synthetic metrics such as `up`

docs/practices/naming.md
conallob: I don't want to encourage copy-pasta, but this is an important point.

If using alerting expressions like `up{job=bla} > 0 for 3m` , you need to be careful not to accidentally strip the `job` label. If you do, your alert no longer works as intended.

I'll polish the wording here

conallob:
For perspective, one of the motivations behind this PR is the anti-patterm of writing alert expressions intended for a single tenant system, which has evolved into a multi-tenant system.

e.g `up{} for 5m` without defining a job label works for one job.

Once you start adding additional jobs that match on the same labels (e.g Daemonsets, fleet-wide node_exporter, etc), teams start getting paged for systems they don't own or care about

docs/practices/naming.md
SuperQ: This doesn't seem related to naming practices, which is what this guide is about.

docs/practices/naming.md
SuperQ: Yes, but it's not related to job, but related to "target labels" and discovery. That is a different thing and related to querying, not creating labels.

jan--f:
Hello from the bug scrub!
@conallob Looks like there some feedback to address still, are you still working on this?

conallob:
Coming back to this long neglected PR, afaik I can tell, the main issue here is whether to put label best practices in the variable naming best practices or not. I only did that since there is no label best practices section currently.


Would creating a new section for label best practices help to unstick this?

docs/practices/naming.md
conallob: I've kept this warning, but moved it to the new labels.md file

docs/practices/naming.md
conallob: Applied

docs/practices/rules.md
conallob: After noodling on this further, I think the message is still important, but to avoid any ambiguous terms, let's just go with "scoping", since that is what the `job` label is for.

docs/practices/naming.md
conallob: Integrated this suggestion, but this section is now in the new labels.md

conallob (commented):
Reviving this PR, PTAL @SuperQ 


* I've split the label sections of naming.md out into a new labels.md, as discussed in OOB chat with @SuperQ 
* I've also reworded the line mentioning "primary key" to be about defining scope in PromQL expressions, which is the ultimate goal.
* It looks like I need a maintainer to approve the Netlify CI checks though

conallob:
@SuperQ Friendly ping?

docs/practices/labels.md
```diff
@@ -0,0 +1,41 @@
...
+
+The label conventions presented in this document are not required
+for using Prometheus, but can serve as both a style-guide and a collection of
+best practices. Individual organizations may want to approach some of these
+practices, e.g. naming conventions, differently.
+
+## Labels
+
+Prometheus labels can come from both the target and from
+[relabeling in discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) as well as from the target itself.
```
krajorama:
```suggestion
Prometheus labels can come from both the target itself and from
[relabeling in discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
```

docs/practices/labels.md
```diff
@@ -0,0 +1,41 @@
...
+
+By default Prometheus configures two primary discovery target labels.
+
+- `job`
+  - The `job`  is a default target label set by the scrape configs and is used to identify metrics scraped from the same target/exporter.
```
krajorama:
```suggestion
  - The `job` is a default target label set by the scrape configs and is used to identify metrics scraped by the same scrape config.
```

docs/practices/labels.md
```diff
@@ -0,0 +1,41 @@
...
+  - If not specified in PromQL expressions, they will match unrelated metrics with the same name. This is especially true in a multi system or multi tenant installation
```
krajorama: It might be the intention of the user to not distinguish between unrelated metrics, for example when aggregating across jobs. So I'd turn this into some positive statement instead (use "specify" instead of "not specified"). Multi system seems vague and I don't think multi tenancy has anything to do with this.

```suggestion
  - If specified in PromQL expressions, they will match metrics scraped by the same job.
```

docs/practices/labels.md
```diff
@@ -0,0 +1,41 @@
...
+
+WARNING: When using `without`, be careful not to strip out the `job` label accidentally.
+
+- `instance`
+  - The `instance` label will include the `ip:port` what was scraped
```
krajorama:
```suggestion
  - The `instance` label will include the `ip:port` what was scraped, identifying the target instance.
```

docs/practices/labels.md
```diff
@@ -0,0 +1,41 @@
...
+Prometheus labels can come from both the target and from
+[relabeling in discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) as well as from the target itself.
+
+By default Prometheus configures two primary discovery target labels.
+
+- `job`
+  - The `job`  is a default target label set by the scrape configs and is used to identify metrics scraped from the same target/exporter.
+  - If not specified in PromQL expressions, they will match unrelated metrics with the same name. This is especially true in a multi system or multi tenant installation
+
+WARNING: When using `without`, be careful not to strip out the `job` label accidentally.
```
krajorama: It might be intentional , so I think this needs to be conditional.

docs/practices/labels.md
```diff
@@ -0,0 +1,41 @@
...
+
+- `instance`
+  - The `instance` label will include the `ip:port` what was scraped
+
```
krajorama, line 35: Don't you need a similar warning for "instance" , depending on usage?

docs/practices/rules.md
```diff
@@ -27,6 +30,18 @@ If there is no obvious operation to use, use `sum`.  When taking a ratio by
 doing division, separate the metrics using `_per_` and call the operation
 `ratio`.
 
+## Labels
+
+NOTE: Omitting a label in a PromQL expression is the functional equivalent of specifying `label=*`
```
krajorama: Please use proper sentences with punctuation. And that's not valid PromQL label selector.
 
```suggestion
NOTE: Omitting a label in a PromQL expression is the functional equivalent of specifying `label=~".*"`.
```

docs/practices/rules.md
```diff
@@ -27,6 +30,18 @@ If there is no obvious operation to use, use `sum`.  When taking a ratio by
...
+
+* In both recorded rules and alerting expressions, always specify a `job` label to prevent expression mismatches from occuring.
```
krajorama, line 37: I think the need to specify `job` is very circumstantial , so again I think it needs to be conditional on what you want to achieve. Also specify `job` is very vague in itself.

docs/practices/rules.md
```diff
@@ -27,6 +30,18 @@ If there is no obvious operation to use, use `sum`.  When taking a ratio by
...
+  This is especially important in multi-tenant systems where the same metric names may be exported by different jobs or the
```
krajorama, line 38: I don't think multi-tenant has anything to do with job and instance labels.

docs/practices/rules.md
```diff
@@ -40,10 +55,6 @@ Instead keep the metric name without the `_count` or `_sum` suffix and replace
 the `rate` in the operation with `mean`. This represents the average
 observation size over that time period.
 
-* Always specify a `without` clause with the labels you are aggregating away.
```
krajorama, line 43: This seems relevant here, why remove it?

krajorama (changes requested):
Thanks for working on this. I feel these additions are currently too specific for your use cases to be included in the generic documentation.

docs/practices/rules.md
conallob: As above, "Multi-tenent systems" may not be the best term, but I'm referring to a Prometheus, run as a platform for multiple teams (e.g by a DevEx or Platform Engineering team), to prevent every team running their own siloed Prometheus stack. In such a setup, all PromQL expressions should be scoped with a `job` label, to ensure the metrics are from the the expected exporters.

Or framed another way, in such a centralised stack, always write `up{job=bla}`, never `up{}`

docs/practices/labels.md
conallob: While `instance` is one of the standard scrape time labels, like `job`, stripping it doesn't have the same blast radius. Stripping `instance` will make metrics hard to debug, but should will still work.


For certain use cases that require using multiple layers of rules (e.g in a multi region, multi layered tree of Prometheus), you may want to strip out `instance` at the higher aggregation layers to manage label cardinality (e.g instance labels make sense to the per region aggregation, but can be problematic if aggregated at the global level)


I've added a warning that stripping `instance` can make it harder to debug scrape time issues with a metric though.

docs/practices/labels.md
conallob: > It might be the intention of the user to not distinguish between unrelated metrics, for example when aggregating across jobs. So I'd turn this into some positive statement instead (use "specify" instead of "not specified").

I agree that the user may want to aggregate across all, or a subset of jobs. But to do so, they should be explicit with the `job` label values. But I expect unintended label stripping to vastly outnumber cross-job aggregation use cases.

> Multi system seems vague and I don't think multi tenancy has anything to do with this.

"Multi-tenent systems" may not be the best term, but I'm referring to a Prometheus, run as a platform for multiple teams (e.g by a DevEx or Platform Engineering team), to prevent every team running their own siloed Prometheus stack. In such a setup, all PromQL expressions should be scoped with a `job` label, to ensure the metrics are from the the expected exporters.

Or framed another way, in such a centralised stack, always write `up{job=bla}`, never `up{}`

docs/practices/rules.md
conallob: Because it's guidance wrt labels, not naming, so it's moving to the new labels.md. But I forgot to add it to labels.md

Added now.

docs/practices/rules.md
conallob: Can you elaborate on why `job` is circumstantial?

Afaik, `job` will always be set on metrics unless it is explicitly stripped away.

conallob:
@krajorama , @SuperQ : friendly ping? I appreciate you're both busy with your day jobs and other prometheus responsibilities, but ideally I'd like to wrap this PR up before it is open for 12+ months (T-29 days)

PR #2692, opened 2025-07-15
chore(security.md): add more precisions regarding sec bugs reporting

machine424:

### DIFF
--- a/docs/operating/security.md
+++ b/docs/operating/security.md
@@ -23,12 +23,21 @@ This page describes the general security assumptions of Prometheus and the
 attack vectors that some configurations may enable.
 
 As with any complex system, it is near certain that bugs will be found, some of
-them security-relevant. If you find a _security bug_ please report it
-privately to the maintainers listed in the MAINTAINERS of the relevant
-repository and CC prometheus-team@googlegroups.com. We will fix the issue as soon
-as possible and coordinate a release date with you. You will be able to choose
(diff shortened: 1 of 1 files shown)


machine424:
cc @SuperQ

docs/operating/security.md
```diff
@@ -23,12 +23,21 @@ This page describes the general security assumptions of Prometheus and the
...
+publicly disclosed, please report it privately to the maintainers listed in the
+MAINTAINERS of the relevant repository and CC prometheus-team@googlegroups.com.
+We will fix the issue as soon as possible and coordinate a release date with
+you. You will be able to choose if you want public acknowledgement of your
+effort and if you want to be mentioned by name.
+
+For _security bugs_ that have already been publicly disclosed and require more than
+just a dependency version bump to fix, please open a
+[Bug report](https://github.com/prometheus/prometheus/issues) including all relevant
+details, unless one has already been filed.
```
SuperQ: Reading order. Let's put this first.

Let's also explicitly mention "publicly disclosed, for example CVEs, ...".

machine424: A CVE can be embargoed and we want the private channel to be used for that. I'll use `public CVEs`

machine424: good idea about the order, done.
we don't want it to be missed, I didn't put it at the end.

PR #2698, opened 2025-07-23
utf8: update doc for deprecated bool

ywwg:

### DIFF
--- a/docs/guides/utf8.md
+++ b/docs/guides/utf8.md
@@ -22,7 +22,7 @@ library](https://github.com/prometheus/client_golang) accept UTF-8 names by
 default.
 
 Previously, documentation recommended that users override the value of
-`model.NameValidationScheme` to select lrgacy validation by default. This
+`model.NameValidationScheme` to select legacy validation by default. This
 boolean is now deprecated and should always be set to UTF8Validation. Legacy
 validation enforcement, if desired, should be done by individual implementations
 calling the appropriate validation APIs and is no longer a library feature.


docs/guides/utf8.md
```diff
@@ -17,29 +17,20 @@ This document guides you through the UTF-8 transition details.
...
-}
-```
-
-If users want to enforce the legacy character set, they can set the validation
-scheme to `LegacyValidation`.
-
-Setting the validation scheme must be done before the instantiation of metrics
-and can be set on the fly if desired.
+Previously, documentation recommended that users override the value of
+`model.NameValidationScheme` to select lrgacy validation by default. This
```
sysadmind:
```suggestion
`model.NameValidationScheme` to select legacy validation by default. This
```

ywwg: applied manually, thanks!

docs/guides/utf8.md
```diff
@@ -98,7 +89,9 @@ is no way to enforce the legacy character set validation with Remote Write 2.0.
 
 ## OTLP Metrics
 
-OTLP receiver in Prometheus 3.0 still normalizes all names to Prometheus format by default. You can change this in `otlp` section of the Prometheus configuration as follows:
+OTLP receiver in Prometheus 3.0 still normalizes all names to Prometheus format
+by default. You can change this in `otlp` section of the Prometheus
+configuration as follows:
```
ArthurSens, line 94: Should we update the configuration below to `NoTranslation`?

ArthurSens (approved):
just one non-blocking question, LGTM

PR #2712, opened 2025-09-01
feat(downloads): mark first stable release as Latest (even if LTS)

roidelapluie:
- Ensure the very first non-prerelease version is always tagged as Latest
- Allow coexistence of Latest + LTS badges on the same release

I recommend using "Hide whitespace changes" option for a clearer diff.


roidelapluie:
https://deploy-preview-2712--prometheus-docs.netlify.app/download/

image

PR #2714, opened 2025-09-03
feat(om2): relax the acceptance rules around unit and total suffixes

krajorama:
Instead of relaxing the rules in general as in PR #2677. This proposal only relaxes the rules on what is acceptable by ingestors. This allows non Prometheus systems to expose names without suffixes and still get processed.


krajorama:
we'll follow #2750 approach

PR #2715, opened 2025-09-03
Rename blog-content-guide.md to README.md

nwanduka:
This PR renames the blog content guide file to README.md so that it automatically renders when viewing the blog folder in the GitHub UI.


bwplotka:
Thanks! 

We need to fix the website, but it's not related to this PR

PR #2713, opened 2025-09-01
docs: create blog-content-guide.md

nwanduka:
This PR adds a Blog Content Guide for contributors. The guide is meant to help community members understand what kind of content is a good fit for the Prometheus blog, how to structure posts, and the process for submitting and publishing them.

PR #2718, opened 2025-09-04
fix: fix build by allowing README file in blog dir

bwplotka:
Looks like we accidentally merged https://github.com/prometheus/docs/pull/2713 without seeing the CI failures.

Fixing the root cause, fixed assets dir bug - I think we have assets in public/assets

cc @nwanduka

### DIFF
--- a/blog-posts/README.md
+++ b/blog-posts/README.md
@@ -30,7 +30,7 @@ So, there are no rigid rules around style. But here a few formatting guidelines 
 `[Prometheus documentation](https://prometheus.io/docs/introduction/overview/)`.
 
 - **Images and diagrams:** Always include alt text so the content is accessible to everyone.
-  Store images in the `assets` folder (instead of linking from external sites) and reference them in your post.
+  Store images in the `<repo_root>/public/assets` folder (instead of linking from external sites) and reference them in your post.
 
 > [!NOTE]
 > See the general [Markdown Documentation Formatting Guide](../markdown-guide.md) for more formatting rules.
--- a/blog-posts/assets/.gitkeep
+++ b/blog-posts/assets/.gitkeep
@@ -1,1 +1,0 @@
-This folder stores images for blog posts.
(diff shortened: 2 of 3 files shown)


saswatamcode (approved):

bwplotka:
(debugging)

bwplotka:
Ok, this should work (:

nwanduka:
> I think we have assets in public/assets

I didn't notice. Thanks for fixing it 🙏

juliusv:
Ohhh, I didn't even notice that it wasn't a top-level file, but in the blog directory. How about moving it to the top level of the repo? I feel it would also be more visible there, and it seems cleaner to leave the content directory just for the actual content files.

nwanduka:
+1 on moving the assets folder to the top level so it’s more visible. 👍

I can raise a follow-up PR for this change, if it's okay..?

juliusv:
> 1 on moving the assets folder to the top level so it’s more visible.

Oh I meant moving the new guide Markdown file to the top level, out of the blog post content dir. Then we can remove that exception in the code (introduced by this PR) again.

> I can raise a follow-up PR for this change, if it's okay..?

Sure!

nwanduka:
Ohhhhkayyy... I think get what you mean. But just to be sure, you're suggesting we move the new blog content guide out of the blog post content dir where it currently sits and up one level, so it's on the same level as these other files (see image)? That's [here](https://github.com/prometheus/docs/tree/main)
image

juliusv:
@nwanduka Yep exactly, that was my initial idea on Slack :) Of course one could also argue that there some merit to having all blog-related files tucked away in a common sub-directory. But in that case, I think the actual blog post content files should still be separate from any other files by having another sub-dir in there. But just keeping the instructions in the root for now is probably good enough.

nwanduka:
Ohh I get it now, thanks for explaining 🙌
I’m in favor of having all blog-related files in a common folder and then putting the actual blog content in its own sub-directory within the blog-post folder.

juliusv:
> Ohh I get it now, thanks for explaining 🙌 I’m in favor of having all blog-related files in a common folder and then putting the actual blog content in its own sub-directory within the blog-post folder.

That's cool with me too :)

PR #2720, opened 2025-09-08
[blog] moved blog files to posts sub dir

nwanduka:
In this PR, I:
- Created a posts/ subdirectory inside blog-post/.
- Moved all existing blog post content files into posts/.


sysadmind:
This breaks the build. I think there are some other pieces missing if we move all these files. For example, we are looking at the blog-posts/ directory in this helpers file - https://github.com/prometheus/docs/blob/main/src/blog-helpers.ts#L5

nwanduka:
@sysadmind you're right. thanks for pointing it out.

sysadmind (approved):
LGTM

sysadmind:
@juliusv you had the feedback on the previous PR, do you want to weigh in on this one?

juliusv:
I guess it all goes back to the discussion I started at https://github.com/prometheus/docs/pull/2718#issuecomment-3253111299 - I don't mind that much whether the blog post instructions file is in the root or in a blog-specific subdir, I just found it cleaner to not have it in the same directory as all the blog post files themselves, since that requires an explicit code-level exception and also makes the new file harder to find by a human if it's in the same dir as all the blog post content files.

But yeah, maybe just calling the first dir `blog/...` and the content dir `blog/content` or `blog/posts` would be even clearer. Or just keep the new blog instructions in the repo root, I don't mind much either way :)

src/blog-helpers.ts
```diff
@@ -2,7 +2,7 @@ import fs from "fs";
 import path from "path";
 import matter from "gray-matter";
 
-const blogPostsDir = "blog-posts";
```
bwplotka, line 5: If we are doing this, let's also remove the filter from line 38 🙈 

Also maybe `blog/posts` might be clearer (as @juliusv mentioned)

bwplotka:
Ah got it, I didn't see the previous discussion on my fix.

Then let's remove the `filter` which seemed to be the main reason why we are moving files here (: 

Sounds like we have two options here then (happy with both)

A) Change dirs, remove coding filter.
* Move posts to `blog/posts` (instead of `blog-posts/posts`)
* Move README to `blog/README.md`
* Remove filtering introduced in https://github.com/prometheus/docs/pull/2718

B) Do nothing (: 

Thanks for helping @nwanduka!

juliusv:
Yeah, the previously introduced special-case filtering should be removed now. Then let's do option A I would say.

nwanduka:
Thanks a lot everyone (@sysadmind @bwplotka @juliusv) for reviewing and leaving your feedback. I appreciate it. I've made the necessary changes (that's option A) and I think this PR is ready for a final review.

PR #2721, opened 2025-09-09
[blog] Contribute to the Prometheus Blog

nwanduka:
This PR adds a new blog post announcing the revival of the Prometheus blog.

### DIFF
--- a/blog-posts/2025-09-09-contribute-to-the-prometheus-blog.md
+++ b/blog-posts/2025-09-09-contribute-to-the-prometheus-blog.md
@@ -1,0 +1,18 @@
+---
+title: Contribute to the Prometheus Blog
+created_at: 2025-09-09
+kind: article
+author_name: Victoria Nduka (@nwanduka)
+---
+
+We’re excited to share that we’re reviving the Prometheus blog 🎉! Going forward, you can expect more frequent updates on what’s happening across the Prometheus ecosystem including new releases, features, community events, tutorials, and behind-the-scenes stories.
+
+Prometheus maintainers, contributors, and community members alike are welcome to share:
+- How you use Prometheus in your projects and organizations
+- Lessons you’ve learned from running Prometheus in production
+- Experiences from contributing to or being part of the community
+- Tips, tutorials, and other content that can help fellow users
+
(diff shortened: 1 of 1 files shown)


nwanduka:
@bwplotka @juliusv could you review this when you get the chance? thank you

blog-posts/2025-09-09-contribute-to-the-prometheus-blog.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Contribute to the Prometheus Blog
+created_at: 2025-09-09
+kind: article
+author_name: Victoria Nduka (@nwanduka)
+---
+
+We’re excited to share that we’re reviving the Prometheus blog 🎉! Going forward, you can expect more frequent updates on what’s happening across the Prometheus ecosystem including new releases, features, community events, tutorials, and behind-the-scenes stories.
```
juliusv, line 8: Hmm, I'm a bit ambivalent about publishing "announcements about future announcements" and also the promise that there will be for sure more content - as long as we're not paying someone to write content on a regular basis, should we really promise that (rather than just doing it)? How about wording this a bit more as an aspiration and making it more about the call for action that you already have below, in the sense of "With the goal of reviving the Prometheus blog, ..." or similar?

Although then I'm still unsure if that deserves a blog post on its own or whether this post would be better published as a post to all our community channels (mailing lists, Slack channels, etc.) in order to hopefully produce more non-meta-level blog posts?

nwanduka: Thanks a lot for reviewing and for the feedback. I see what you mean. My initial thought was that having a post on the blog itself would act as a visible marker for the revival, but I’m open to either direction.

I’ve posted this in Slack (#prometheus-dev). Would someone with access to the mailing list be able to share it there as well so we reach more of the community?

juliusv: @nwanduka Great! So anyone should be able to post to the users and dev mailing lists by just sending an email to the list address (prometheus-dev@googlegroups.com and prometheus-users@googlegroups.com). Or were you thinking of using the prometheus-announce mailing list? I think we are only using that for component release announcements though.

nwanduka: Alrighty. I think the -dev and -users lists works just fine. I'll send the email and close this PR. Thanks a lot for always reviewing 🙂

PR #2723, opened 2025-09-18
remove matrix url as it is abandoned

Nexucis:


ll-SKY-ll:
thats weird, im moderating in that room and while its currently quiet there, its certainly not abandoned.

SuperQ:
Hey, it's not actually abandoned. Yes, there's a lot less activity there, but we still maintain the room.

Nexucis:
wait what ? I was pretty sure we have decided to migrate everything to slack and abandon the matrix channel.

TheArcaneBrony:
Did you remove the wrong link by chance? #prometheus-dev is still linked on the page, you just removed the community room.

ll-SKY-ll:
also, for a bit more context, I was/am currently in the process of getting that room listed in the matrix.org room directory again, so that it would actually be not abandoned anymore and can be discovered easier.

sysadmind:
I migrated from matrix to slack because it certainly seems dead. The current room has no messages from what I can tell. Do we want to send people to matrix if there is no activity?

ll-SKY-ll:
are you talking about #prometheus-dev:matrix.org ?  because as far as i remember, there went something wrong when they tried to upgrade the room but then never followed up with fixing it. The room #prometheus:matrix.org on the other hand is where im also at and which im trying to get more visibility for. Also that room actually has some messages here and there and i hope it will be more active once its in the room directory again.

sysadmind:
You are correct @ll-SKY-ll, I was thinking of prometheus-dev. To that end, is that a room that we still want to support for development purposes?

PR #2725, opened 2025-09-18
native histograms: clarify schema reduction and schemas 9-52

beorn7:
Call out that schemas can be reduced upon ingestion and that the "reserved" schemas 9 to 52 MAY already be accepted if they are reduced to the currently valid schemas -4 to 8.

Update the section about limiting resolution accordingly.

Related to https://github.com/prometheus/prometheus/issues/14168


krajorama (approved):

beorn7:
I'll only merge this once we have merged the other related PRs.

PR #2728, opened 2025-09-21
docs: remove removed visualization by consoles

dancer1325:
## how to review?
* https://github.com/prometheus/prometheus/pull/14807
* check by yourself in the source code, it does NOT exist anymore -- https://github.com/prometheus/prometheus


jan--f:
#2812 addressed this I assume. Please re-open if there is still something to fix.

PR #2733, opened 2025-09-23
Fix: Respect user's browser theme preference on initial load

milinddethe15:
### What
Set the default color scheme to "auto" to respect the user's browser/system theme preference when first loading the site.

### Changes
- Updated `MantineProvider` to use `defaultColorScheme="auto"`
- Updated `ColorSchemeScript` to use `defaultColorScheme="auto"`

### Before
Site always loaded in light mode regardless of user's browser preference.

### After
Site now loads with the user's preferred theme (dark/light) based on their browser/system settings, while maintaining the ability to manually override via the theme selector.


milinddethe15:
@jan--f / @bwplotka Can you PTAL?

PR #2734, opened 2025-09-23
Fix typo in metric name

MohammadAlavi1986:
I believe the referenced recording rule should be `instance_path:request_latency_seconds_count:rate5m`.


krajorama (approved):
thank you

PR #2745, opened 2025-10-02
feat(om1.1): relax the acceptance rules around unit and total suffixes

krajorama:
This proposal only relaxes the rules on what is acceptable by ingestors. This allows non Prometheus systems to expose names without suffixes and still get processed.

The reason is that Prometheus strongly recommends using the `_total` and unit suffixes for usability and to avoid conflicting metric names.

Generally, to be transparent, this is a future compatibility ingestor breaking change guarded by the content negotiation that we decided (OM 2.0 WG) to proceed with this, because many implementations already do this (and some cases require for OM to be useful).

### DIFF
--- a/docs/specs/om/open_metrics_spec_1_1.md
+++ b/docs/specs/om/open_metrics_spec_1_1.md
@@ -146,6 +146,8 @@ Type specifies the MetricFamily type. Valid values are "unknown", "gauge", "coun
 
 Unit specifies MetricFamily units. If non-empty, it MUST be a suffix of the MetricFamily name separated by an underscore. Be aware that further generation rules might make it an infix in the text format.
 
+Ingestors MAY accept MetricFamily name without unit suffix. Be aware that exposing such metrics directly to end-users reduces usability due to confusion about what the metric's unit is.
+
 ##### Help
 
 Help is a string and SHOULD be non-empty. It is used to give a brief description of the MetricFamily for human consumption and SHOULD be short enough to be used as a tooltip.
@@ -488,7 +490,7 @@ A valid example for a foo_seconds metric with a unit of "seconds":
 # UNIT foo_seconds seconds
 ```
 
-An invalid example, where the unit is not a suffix on the name:
+An invalid example (which MAY still be accepted by ingestors), where the unit is not a suffix on the name:
 
 ```
 # TYPE foo counter
@@ -634,6 +636,8 @@ foo 18.0 456
 ##### Counter
 
 The MetricPoint's Total Value Sample MetricName MUST have the suffix `_total`. If present the MetricPoint's Created Value Sample MetricName MUST have the suffix `_created`.
+
+Ingestors MAY accept MetricPoint Total Value name without the `_total` suffix. Be aware that exposing such metrics directly to end-users reduces usability due to confusion about what the metric's type is.
 
 An example with a Metric with no labels, and a MetricPoint with no timestamp and no created:
 


docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -146,6 +150,8 @@ Type specifies the MetricFamily type. Valid values are "unknown", "gauge", "coun
 
 Unit specifies MetricFamily units. If non-empty, it MUST be a suffix of the MetricFamily name separated by an underscore. Be aware that further generation rules might make it an infix in the text format.
 
+Ingestors MAY accept MetricFamily name without unit suffix.
```
krajorama: Question to reviewers: a slightly more strict version would be:
```suggestion
Ingestors MAY accept MetricFamily name without unit suffix, provided that there is a mechanism to retrieve the unit after ingestion.
```

docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -635,6 +641,8 @@ foo 18.0 456
 
 The MetricPoint's Total Value Sample MetricName MUST have the suffix `_total`. If present the MetricPoint's Created Value Sample MetricName MUST have the suffix `_created`.
 
+Ingestors MAY accept MetricPoint Total Value name without the `_total` suffix.
```
krajorama: Question to reviewers, a slightly more strict version:
```suggestion
Ingestors MAY accept MetricPoint Total Value name without the `_total` suffix, provided that there is a mechanism to retrieve the counter type information after ingestion.
```

docs/specs/om/open_metrics_spec_1_1.md
bwplotka: Is is needed?

There will be cases where e.g. Prometheus will ingest no unit series with NO unit label etc, and it's still acceptable, no?

docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -118,6 +118,8 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
 
 MetricFamily names are a string and MUST be unique within a MetricSet. Names SHOULD be in snake_case. Metric names MUST follow the restrictions in the ABNF section.
 
+Ingestors MAY accept duplicate MetricFamily names, provided that there is a mechanism to distinguish the resulting MetricPoints.
```
bwplotka: Can we have a bit more strict definition? Like we know it's literally type and unit being different, maybe help. IMO we should block duplicate MetricFamilies that has the same type and unit (and maybe help), it makes no sense.

Help is a good question, push based systems like Otel have different help too, but this can cause some overhead and Prometheus will have to drop the diff (we will likely pick first help). Perhaps for clean approach help should be within a "key".

Anyway, this needs a bit more explantation (e.g. on valid cases).

Some yolo attempt:

```suggestion
Ingestors MAY accept duplicate MetricFamily names, provided that either of type, unit or help is unique within duplicated MetricFamily names. This is discouraged due to overhead and user experience, notably Exposers MAY duplicate MetricFamilies when aggregating external metrics from different sources. 
```

docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -126,6 +128,8 @@ MetricFamily names beginning with underscores are RESERVED and MUST NOT be used
 
 The name of a MetricFamily MUST NOT result in a potential clash for sample metric names as per the ABNF with another MetricFamily in the Text Format within a MetricSet. An example would be a gauge called "foo_created" as a counter called "foo" could create a "foo_created" in the text format.
 
+Ingestors MAY accept conflicting MetricFamily names, provided that there is a mechanism to distinguish the resulting MetricPoints. It follows that ingestors MAY accept MetricFamily name without the type specific suffix.
```
bwplotka:
```suggestion
Ingestors MAY accept conflicting MetricFamily names, especially resulting from the MetricFamily names without the type or unit specific suffix (see [MetricFamily metadata](#MetricFamily-metadata)).
```

docs/specs/om/open_metrics_spec_1_1.md
bwplotka: Again, I don't understand what are those mechanisms. However I would mention something like 
```suggestion
Ingestors MAY accept MetricPoint Total Value name without the `_total` suffix, but only if the TYPE is explicitly set to Counter.
```

bwplotka (approved):
Look good, added some suggestions.

Generally, to be transparent, this is a future compatibility ingestor breaking change guarded by the content negotiation that we decided (OM 2.0 WG) to proceed with this, because many implementations already do this (and some cases require for OM to be useful).

docs/specs/om/open_metrics_spec_1_1.md
krajorama: That's my question in a different form? What's the answer :) ?

docs/specs/om/open_metrics_spec_1_1.md
krajorama: That doesn't make sense. We're already only talking about counters, where the TYPE was Counter.

What I'm trying to get at is to protect the user from themselves.

If we allow no suffixes willy-nilly  and don't force/suggest some way to get to the information afterwards, we loose the niceness  of the Prometheus model completely where you could tell from the name that something was a counter. I'd like to at least say "ingestors SHOULD provide the way to retrieve the type information later". For example by `__type__` label or some other means, I don't care.

docs/specs/om/open_metrics_spec_1_1.md
krajorama: Yes, I think being more specific in a standard is better. 

Again I want to go back to what I'd like to solve, which is to not loose the ability to find and distinguish the series. So instead of saying that "This is discouraged due to ... user experience" we should encourage Ingestors to solve this. Maybe use the wording "SHOULD" in this version, to not stop adoption right away though :)

docs/specs/om/open_metrics_spec_1_1.md
krajorama: suggesting something different now.

docs/specs/om/open_metrics_spec_1_1.md
krajorama: suggesting something different now

docs/specs/om/open_metrics_spec_1_1.md
krajorama: Updated the test to be more specific.

docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -126,6 +128,8 @@ MetricFamily names beginning with underscores are RESERVED and MUST NOT be used
 
 The name of a MetricFamily MUST NOT result in a potential clash for sample metric names as per the ABNF with another MetricFamily in the Text Format within a MetricSet. An example would be a gauge called "foo_created" as a counter called "foo" could create a "foo_created" in the text format.
 
+Ingestors MAY accept duplicate MetricFamily names, provided that either of HELP, TYPE, or UNIT is unique within the duplicated MetricFamily names. Notably Exposers MAY duplicate MetricFamilies when aggregating external metrics that have the same MetricFamily name, but come from different sources and have either different HELP, or TYPE, or UNIT. Be aware that exposing such metrics directly to end-users reduces usability due to confusion about which Metric belongs to which MetricFamily.
```
ywwg: I was aware of type and unit being identified, but this is the first I'm hearing of HELP being identifying -- is that a case we need to support?

ywwg: when I say "this is the first I'm hearing" I mean that literally -- I often miss / forget conversations 😅

docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -118,6 +118,8 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
 
 MetricFamily names are a string and MUST be unique within a MetricSet. Names SHOULD be in snake_case. Metric names MUST follow the restrictions in the ABNF section.
 
+Ingestors MAY accept duplicate MetricFamily names, provided that either of HELP, TYPE, or UNIT is unique within the duplicated MetricFamily names. Notably Exposers MAY duplicate MetricFamilies when aggregating external metrics that have the same MetricFamily name, but come from different sources and have either different HELP, or TYPE, or UNIT. Be aware that exposing such metrics directly to end-users reduces usability due to confusion about which Metric belongs to which MetricFamily.
```
dashpole: IMO we can still require metric family names to be unique within the text exposition format while being more tolerant of other naming. It properly places the burden of ensuring unique metric family names on the exposers that choose to deviate from naming recommendations (e.g. OTel SDKs), and doesn't impose an additional burden on Ingestors.

krajorama: Would you say that's the OpenTelemetry use case? To put the burden on exposer? Would that be enough for OM 1.1?

Tolerating non-unique metric family names comes from the federation, push-gateway and similar use cases. We could discard these use cases in OM 1.1 and only discuss for OM 2.0. I'm not opposed to that.

docs/specs/om/open_metrics_spec_1_1.md
krajorama: let's decide https://github.com/prometheus/docs/pull/2745#discussion_r2399091513 first

docs/specs/om/open_metrics_spec_1_1.md
dashpole: Yes, that would meet the OpenTelemetry use-case. IMO non-unique metric family names seems more disruptive than other changes we've been discussing, so I would prefer punting that to OM 2.0.  But I don't have a good sense of how important the federation, push-gateway use-cases are, so i'm open to other opinions.

krajorama: Not allowing federation/push-gateway isn't a regression, it has always been the case. I'll remove these additions from the PR.

krajorama:
On further reflection, making the relaxed rule only apply to the ingestor is a bad idea as that would mean that different ingestors would treat the same text differently . Meaning that for example if a developer exposed a metric without the _total suffix and tested it with a lenient ingestor , they would not notice that it will fail for others.

Superseded by #2750

PR #2746, opened 2025-10-02
Update OM2.0 for UTF-8 metric names and labels

ywwg:
part of https://github.com/prometheus/prometheus/issues/16093

### DIFF
--- a/docs/specs/om/open_metrics_spec_2_0.md
+++ b/docs/specs/om/open_metrics_spec_2_0.md
@@ -92,7 +92,7 @@ Strings MUST only consist of valid UTF-8 characters and MAY be zero length. NULL
 
 Labels are key-value pairs consisting of strings.
 
-Label names beginning with underscores are RESERVED and MUST NOT be used unless specified by this standard. Label names MUST follow the restrictions in the ABNF section.
+Label names beginning with two underscores are RESERVED and MUST NOT be used unless specified by this standard. Label names SHOULD follow the restrictions in the ABNF section under the `label-name` section. Label names MAY be any quoted escaped UTF-8 string as described in the ABNF section. Be aware that exposing UTF-8 metrics is still experimental and may reduce usability.
 
 Empty label values SHOULD be treated as if the label was not present.
 
@@ -128,7 +128,7 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
 
 ##### Name
 
-MetricFamily names are a string and MUST be unique within a MetricSet. Names SHOULD be in snake_case. Metric names MUST follow the restrictions in the ABNF section.
+MetricFamily names are a string and MUST be unique within a MetricSet. Names SHOULD be in snake_case. Names SHOULD follow the restrictions in the ABNF section under `metricname`. Metric names MAY be any quoted and escaped UTF-8 string as described in the ABNF section. Be aware that exposing UTF-8 metrics is still experimental and may reduce usability, especially when suffixes are not included.
 
 Colons in MetricFamily names are RESERVED to signal that the MetricFamily is the result of a calculation or aggregation of a general purpose monitoring system.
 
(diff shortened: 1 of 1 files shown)


docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -366,16 +372,16 @@ HASH = "#"
...
+metricname-utf8 = DQUOTE escaped-string DQUOTE
 
+label-key = label-name / DQUOTE escaped-string DQUOTE
 label-name = label-name-initial-char *label-name-char
-
 label-name-char = label-name-initial-char / DIGIT
 label-name-initial-char = ALPHA / "_"
 
-escaped-string = *escaped-char
+escaped-string = 1*escaped-char
```
ywwg: I am led to believe that this will prevent empty strings from being valid

krajorama: Yes, you can write `escaped-char *escaped-char` if you want to be more specific, but 1*escaped-char is totally fine, see https://en.wikipedia.org/wiki/Augmented_Backus–Naur_form "variable repetition".

docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -312,22 +316,24 @@ metricset = *metricfamily
...
 metric = *sample
 
 metric-type = counter / gauge / histogram / gaugehistogram / stateset
 metric-type =/ info / summary / unknown
 
-sample = metricname [labels] SP number [SP timestamp] [exemplar] LF
+sample = metricname-and-labels SP number [SP timestamp] [exemplar] LF
 
-exemplar = SP HASH SP labels SP number [SP timestamp]
+metricname-and-labels = metricname [labels-in-braces] / name-in-braces
```
krajorama: maybe?
```suggestion
metricname-and-labels = metricname [labels-in-braces] / name-and-labels-in-braces
```

krajorama (commented):
looks good, one nit comment

docs/specs/om/open_metrics_spec_1_1.md
ywwg: applied manually

docs/specs/om/open_metrics_spec_1_1.md
krajorama: Isn't this a breaking change for labels? https://prometheus.io/docs/specs/om/open_metrics_spec/#label
The format allows empty string as far as I understand.

ywwg: ah, correct -- ok we need to allow empty strings for label values but not label or metric names

krajorama:
@ywwg I'd suggest to scan the "Design considerations" section, in particular https://prometheus.io/docs/specs/om/open_metrics_spec/#metric-naming-and-namespaces .

I've done it as a follow-up to relaxed suffix rules, see https://github.com/prometheus/docs/pull/2751

docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -942,7 +951,7 @@ For high availability and ad-hoc access a common approach is to have multiple in
 # EDITOR’S NOTE:  This section might be good for a BCP paper.
 -->
 
-We aim for a balance between understandability, avoiding clashes, and succinctness in the naming of metrics and label names. Names are separated through underscores, so metric names end up being in “snake_case”.
+We aim for a balance between understandability, avoiding clashes, and succinctness in the naming of metrics and label names. Names are separated through underscores, so metric names end up being in “snake_case”. While we strongly recommend the practices recommended in this document, other metric systems have different philosophies regarding naming conventions. OpenMetrics allows metrics to be exposed, but without the conventions and suffixes recommended here there is an increased risk of collisions and incompatibilities along the chain of services in a metrics system. Users wishing to use alternative conventions will need to take special care and expend additional effort to ensure that the entire ecosystem is consistent.
```
dashpole:
```suggestion
We aim for a balance between understandability, avoiding clashes, and succinctness in the naming of metrics and label names. Names are separated through underscores, so metric names end up being in “snake_case”. While we strongly recommend the practices recommended in this document, other metric systems have different philosophies regarding naming conventions. OpenMetrics allows these metrics to be exposed, but without the conventions and suffixes recommended here there is an increased risk of collisions and incompatibilities along the chain of services in a metrics system. Users wishing to use alternative conventions will need to take special care and expend additional effort to ensure that the entire system is consistent.
```

ywwg:
we are going to go directly to 2.0 and skip 1.1 after all, so I'll adapt this change to 2.0 instead

ywwg:
(I am not deleting 1.1 because we need to port some more changes from it into 2.0 and it's useful to keep it around for now)

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -384,13 +386,16 @@ metricname = metricname-initial-char 0*metricname-char
...
+metricname-utf8 = DQUOTE escaped-string-non-empty DQUOTE
 
+label-key = label-name / DQUOTE escaped-string-non-empty DQUOTE
 label-name = label-name-initial-char *label-name-char
 
 label-name-char = label-name-initial-char / DIGIT
 label-name-initial-char = ALPHA / "_"
 
-escaped-string = *escaped-char
+escaped-string = escaped-char
```
krajorama:
```suggestion
escaped-string = *escaped-char
```

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -327,22 +327,24 @@ metricset = *metricfamily
...
+metric-descriptor =/ HASH SP help SP (metricname / metricname-utf8) SP escaped-string LF
+metric-descriptor =/ HASH SP unit SP (metricname / metricname-utf8) SP *metricname-char LF
 
 metric = *sample
 
 metric-type = counter / gauge / histogram / gaugehistogram / stateset
 metric-type =/ info / summary / unknown
 
-sample = metricname [labels] SP number [SP timestamp] [SP created] [exemplar] LF
+sample = metricname-and-labels SP number [SP timestamp] [exemplar] LF
```
krajorama:
```suggestion
sample = metricname-and-labels SP number [SP timestamp] [SP created] [exemplar] LF
```

ywwg:
removed diffs with 1.1 and cleaned things up a bit

krajorama (changes requested):
Looks good. I really think we should add some examples that are checked against the ABNF. It's ok if you want to do it in a next PR as well.

ywwg:
Yeah can we do that in a followup? I'll file an issue

ywwg:
https://github.com/prometheus/docs/issues/2766

Issue #2747, opened 2025-10-03
escaping=allow-utf-8 or escaping=allow-utf8


martynd:
There appears to be some contradictions in the docs regarding the correct form of `utf8` to use with `escaping=` in the `Accept` header

Based on what works functionally I believe `allow-utf-8` to be the correct option

The content negotiation page at https://prometheus.io/docs/instrumenting/content_negotiation/ uses `allow-utf8` exclusively. The value appears in the list of valid values, in the 2 instances in the default `Accept` header value, and the 2 instances in the default protobuf `Accept` header value

Edit: Preemptively added a PR https://github.com/prometheus/docs/pull/2748

beorn7:
Yes, `allow-utf-8` is correct. Has the PR fixed all occurrences?

I assume so, so closing this, but please follow up if you have seen more.

PR #2748, opened 2025-10-03
docs: Fix utf8 escaping parameter in content negotiation docs

martynd:
Corrects what appears to be a mistake in the docs regarding the `allow-utf-8` `escaping` option as described in https://github.com/prometheus/docs/issues/2747


krajorama (approved):
LGTM, based on code and https://prometheus.io/docs/instrumenting/escaping_schemes/

PR #2750, opened 2025-10-03
feat(om1.1): relax the requirements around unit and total suffixes

krajorama:
Allow exposers to expose metrics that do not put `_total` or unit as suffix to metric names.

Alternatives considered:

### DIFF
--- a/docs/specs/om/open_metrics_spec_1_1.md
+++ b/docs/specs/om/open_metrics_spec_1_1.md
@@ -144,7 +144,9 @@ Type specifies the MetricFamily type. Valid values are "unknown", "gauge", "coun
 
 ##### Unit
 
-Unit specifies MetricFamily units. If non-empty, it MUST be a suffix of the MetricFamily name separated by an underscore. Be aware that further generation rules might make it an infix in the text format.
+Unit specifies MetricFamily units. If non-empty, it SHOULD be a suffix of the MetricFamily name separated by an underscore. Be aware that further generation rules might make it an infix in the text format.
+
+Be aware that exposing metrics without the unit being a suffix of the MetricFamily name directly to end-users may reduce the usability due to confusion about what the metric's unit is.
 
 ##### Help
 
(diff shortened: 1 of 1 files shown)


docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -488,7 +490,7 @@ A valid example for a foo_seconds metric with a unit of "seconds":
 # UNIT foo_seconds seconds
 ```
 
-An invalid example, where the unit is not a suffix on the name:
+A valid example, where the unit is not a suffix on the name:
```
dashpole:
```suggestion
A valid, but discouraged example, where the unit is not a suffix on the name:
```

docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -665,6 +669,13 @@ foo_total 17.0 1520879607.789
 foo_created 1520430000.123 1520879607.789
 ```
 
+An example with a Metric with no labels, and a MetricPoint without the `_total` suffix and with a timestamp and a created:
+
+```openmetrics-add-eof
+# TYPE foo counter
+foo 17.0 1520879607.789
+foo_created 1520430000.123 1520879607.789
```
dashpole, line 677: needs three ticks to fix formatting

ywwg:
PR description needs to be updated

docs/specs/om/open_metrics_spec_1_1.md
```diff
@@ -633,7 +635,9 @@ foo 18.0 456
 
 ##### Counter
 
-The MetricPoint's Total Value Sample MetricName MUST have the suffix `_total`. If present the MetricPoint's Created Value Sample MetricName MUST have the suffix `_created`.
+The MetricPoint's Total Value Sample MetricName SHOULD have the suffix `_total`. If present the MetricPoint's Created Value Sample MetricName MUST have the suffix `_created`.
```
ywwg, line 638: hmmm, I'm not sure how Otel deals with created metrics.  I am tempted not to have any MUSTs around names so that we can truly allow anything and not be against the spec (no matter how much we discourage it)

dashpole: I think we need a MUST for the _created series, otherwise it wouldn't be the `_created` series :).

ywwg (approved):

PR #2751, opened 2025-10-08
feat(om1.1): follow up relaxed suffix rules

krajorama:
I've scanned the "Design considerations" section of the doc and there's one place that talks about suffixes (unit) specifically. This commit updates the statement about unit suffix to be a recommendation not a rule.


krajorama:
We're moving this over to 2.0 I'll open a new PR.

PR #2752, opened 2025-10-08
Various native histogram spec updates

beorn7:
See commit descriptions for details.

Let's only merge this once we are positive about how and when to tackle the related issues:

- https://github.com/prometheus/prometheus/issues/17308 (I plan to create a PR today or tomorrow).
- https://github.com/prometheus/prometheus/issues/17255 (which already has a PR under review)

FYI:
- @linasm (Please check if I caught the implications of NHCB reconciliations correctly.)
- @crush-on-anechka (WRT counter reset hint implications – but as said, I plan to create a PR and will cc you on it.)

### DIFF
--- a/docs/specs/native_histograms.md
+++ b/docs/specs/native_histograms.md
@@ -504,15 +504,15 @@ their upper boundary within the custom values list.
 
 The lower exclusive boundary is defined by the custom value preceding the upper
 boundary. For the first custom value (at position zero in the list), there is
-no preceding value, in which case the lower boundary is considered to be
-`-Inf`. Therefore, the custom bucket with index zero counts all observations
-between `-Inf` and the first custom value. In the common case that only
-positive observations are expected, the custom bucket with index zero SHOULD
-have an upper boundary of zero to clearly mark if there have been any
-observations at zero or below. (If there are indeed only positive observations,
-the custom bucket with index zero will stay unpopulated and therefore will
-never be represented explicitly. The only cost is the additional zero element
-at the beginning of the custom values list.)
+no preceding value, in which case the lower boundary is considered to be `-Inf`
+inclusively. Therefore, the custom bucket with index zero counts all
+observations between (and including) `-Inf` and the first custom value. In the
+common case that only positive observations are expected, the custom bucket
+with index zero SHOULD have an upper boundary of zero to clearly mark if there
+have been any observations at zero or below. (If there are indeed only positive
+observations, the custom bucket with index zero will stay unpopulated and
+therefore will never be represented explicitly. The only cost is the additional
+zero element at the beginning of the custom values list.)
 
 Custom values MUST NOT be `+Inf`. Observations greater than the last custom
 value go into an overflow bucket with an upper boundary of `+Inf`. This
@@ -1687,7 +1687,7 @@ can be converted on the fly to meet these compatibility criteria:
 
 - NHCBs (schema -53) are only compatible with each other. Different custom
   values need to be reconciled by conversion in the following way:
-  - Identify the custom values that are present in each of the original NHCBs.
+  - Identify the custom values that are shared by all of the original NHCBs.
     These are the new reconciled custom values.
   - Convert each original NHCB to the new custom values by merging its buckets
     into the unified bucket set described by the new custom values.


linasm:
Related question: it is not very clear from the spec, whether `-Inf` is a valid (first, obviously) custom bound value. Currently, `Validate()` would reject that, likely accidentally (no tests are failing if I change that check).

Also the spec should state that `-Inf` is **inclusive** first lower bound. Otherwise it is not clear where observations with `-Inf` value would be counted. For the sake of symmetry - observations with `+Inf` value are counted in the last bucket.

beorn7:
Right, we should make the handling of -Inf explicit (as we do for the standard exponential schemas).

beorn7:
I added a commit for that.

WRT -Inf as the 1st custom value: I think it should be allowed. That's implicit from the spec. I don't think we have to make it explicit in the spec. Just fix the Validate method. (Will create a PR momentarily.)

beorn7:
See https://github.com/prometheus/prometheus/pull/17320

beorn7:
I'll give https://github.com/prometheus/prometheus/pull/17278 a final review ASAP. But since we already know that we want to merge it, this can be merged now, too.

@krajorama maybe give it a final pass?

docs/specs/native_histograms.md
```diff
@@ -1159,8 +1166,16 @@ histogram. Via the boolean scrape config option
...
+While NHCBs support [automatic reconciliation between different bucket
+layouts](#compatibility-between-histograms), their mergeability is still
+fundamentally limited. The reconciliation only retains exact matches of bucket
+boundaries between the involved NHCBs. This yields useful results, if most
+bucket boundaries match. However, abitrary changes in the bucket layout can
+easily create a situation where none of the boundaries match, resulting in a
+histogram with only one bucket (the overflow bucket).
+
+A key advantage of NHCBs is that they are generally much less expensive to
+store.
```
krajorama: Maybe add that this enables the use case of adding more buckets to existing NHCBs in case classic histograms had cardinality issues.

docs/specs/native_histograms.md
```diff
@@ -1665,14 +1679,22 @@ retrieved from the TSDB.
...
-  have the exact same custom values. (In principle, there are possible
-  differences in custom values that could be reconciled, but PromQL doesn't yet
-  consider those.)
+histograms involved need to have the same schema, the same zero bucket width,
+and (if applicable) the same custom values. Within certain limits, histograms
+can be converted on the fly to meet these compatibility criteria:
+
+- NHCBs (schema -53) are only compatible with each other. Different custom
+  values need to be reconciled by conversion in the following way:
+  - Identify the set of buckets with the highest possible number of buckets
```
krajorama: This is a bit too complicated, maybe we can say:

"Take the  custom values from the original NHCBs that occur in both NHCBs (i.e. intersect). These are the new custom values. For example in case of [1, 2, 4, 8, 16] and [1, 4, 16], the new custom values would be [1, 4, 16], thus creating the (-Inf, 1], (1, 4], (4, 16], (16, +Inf] buckets that overlap exactly with the original buckets."

docs/specs/native_histograms.md
```diff
@@ -1883,13 +1905,28 @@ histogram samples (for the reason stated in parentheses):
 
 The `sum` aggregation operator work with native histograms by summing up the
```
krajorama:
```suggestion
The `sum` aggregation operator works with native histograms by summing up the
```

docs/specs/native_histograms.md
```diff
@@ -1883,13 +1905,28 @@ histogram samples (for the reason stated in parentheses):
...
 histogram to be aggregated in the same way as described for the `+` operator
-above, including the implications regarding counter vs. gauge histograms.
-(Generally, only gauge histograms should be aggregated in this way.) The `avg`
-aggregation operator works in the same way, but divides the sum by the number
-of aggregated histogram (in the same way as described for the `/` operator
-above). Both aggregation operators remove elements from the output vector that
-would require the aggregation of float samples with histogram samples. Such a
-removal is flagged by a warn-level annotation.
+above. The `avg` aggregation operator works in the same way, but divides the
+sum by the number of aggregated histogram (in the same way as described for the
```
krajorama:
```suggestion
sum by the number of aggregated histograms (in the same way as described for the
```

docs/specs/native_histograms.md
beorn7: I extended this paragraph along this thought.

docs/specs/native_histograms.md
beorn7: Yes, simpler indeed.

I adjusted the explanation, but I did not add the example. I think that is not necessary in a reference documentation that is already very long.

docs/specs/native_histograms.md
beorn7: I extended this paragraph along this thought.

beorn7:
I manoeuvred GH into a state where it marks all my comments as "pending" but I cannot submit them.

beorn7:
image

beorn7:
image

beorn7:
PHAL.

docs/specs/native_histograms.md
```diff
@@ -1665,14 +1681,21 @@ retrieved from the TSDB.
...
-  have the exact same custom values. (In principle, there are possible
-  differences in custom values that could be reconciled, but PromQL doesn't yet
-  consider those.)
+histograms involved need to have the same schema, the same zero bucket width,
+and (if applicable) the same custom values. Within certain limits, histograms
+can be converted on the fly to meet these compatibility criteria:
+
+- NHCBs (schema -53) are only compatible with each other. Different custom
+  values need to be reconciled by conversion in the following way:
+  - Identify the custom values that are present in each of the original NHCBs.
```
krajorama:
```suggestion
  - Identify the custom values that are present in both of the original NHCBs.
```

docs/specs/native_histograms.md
```diff
@@ -1881,15 +1904,30 @@ histogram samples (for the reason stated in parentheses):
...
-removal is flagged by a warn-level annotation.
+above. The `avg` aggregation operator works in the same way, but divides the
+sum by the number of aggregated histograms (in the same way as described for
+the `/` operator above).
+
+Both `sum` and `avg` remove elements from the output vector that would require
+the aggregation of float samples with histogram samples. Such a removal is
+flagged by a warn-level annotation.
+
+Both `sum` and `avg` should only be applied to gauge histograms. PromQL allows
```
krajorama, line 1917: Optional:

```suggestion
Both `sum` and `avg` should only be applied to gauge histograms, the recommended way to turn counter histograms into gauge histograms is to calculate their rate first as in `sum(rate(histogram[1m]))`. PromQL allows
```

krajorama (changes requested):
Small fix and an optional nit suggestion

docs/specs/native_histograms.md
beorn7: "First the rate, then aggregate." This is literally a Prometheus proverb. I don't think we should teach users Prometheus fundamentals in this spec. (It's long enough already.) Furthermore, a gauge histogram can also come directly from a target, so it's not even generally true (though very common) that a gauge histogram is always created from a `rate`.

docs/specs/native_histograms.md
beorn7: An aggregation operator could aggregate more than two NHCBs. We have even said "two or more" in the previous paragraph. I think "each" makes more sense than "both" here.

beorn7 (commented):
Thanks for the review.

However, I would like to keep both commented sections as they are. 😁

docs/specs/native_histograms.md
krajorama: then maybe "every one" ? I feel like each sounds like it could mean you do a union, and if I misunderstand it, others will

docs/specs/native_histograms.md
krajorama: fair

docs/specs/native_histograms.md
beorn7: Settled on a new wording.

PR #2757, opened 2025-10-22
Add Oodle as a remote write receiver

mgaurav:
A working example is in Oodle playground: https://play.oodle.ai/settings?integration=PROMETHEUS

cc: @juliusv


juliusv:
:+1: Thanks!

PR #2762, opened 2025-10-28
RW 2.0-rc.4 (breaking): Move CT from TimeSeries to Sample; Rename to ST (Start Timestamp)

bwplotka:
Given the recent movement for Prometheus native support of ST ([PROM-60](https://github.com/prometheus/proposals/pull/60)) and plans for delta temporality
([PROM-48](https://github.com/prometheus/proposals/pull/48)) it might be beneficial to make (hopefully) last change to Remote Write 2.0 before stabilizing, so:

* Rename Created Timestamp to Start Timestamp.
* Move CT/ST from TimeSeries to Sample and Histogram messages.
* Clarified optionality (0 value meaning unset).

### Rationales (for moving CT/ST to Sample)

The biggest change is the move from CT/ST per TimeSeries to be per Sample. This makes sample a triplet `(st, t, v)` not a pair `(t, v)`. 

Let's go through pros & cons of this change:

**Pro**
* It logically cleaner (ST relates to sample semantics around timestamps) for both cumulatives and delta one day.
* Future consistency with 1P (first party): [PROM-60](https://github.com/prometheus/proposals/pull/60) for native support of ST, all APIs will logically allow ST per sample. It will be awkward to pass this to per series for RW (and back).
* Consistency with 3P (third party system: Known systems that supports CT/ST store it per sample (Otel, Google).
* Has no practical impact on current performance because Prometheus always sends one sample per timeseries message.
* If, one day, use cases emerge to send more than one sample per TimeSeries with different ST values (e.g. counter reset or delta), it's will much easier to read and write (vs duplicating timeseries message, so essentially creating a new series) and potentially less efficient (duplicate TimeSeries message).

**Cons**
*  If, one day, use cases emerge to send more than one sample per TimeSeries with the same ST values, there will be some overhead of extra, duplicated int64 (there's no diff compression like in chunk).
   * However, this is exactly the same for t,v and we didn't optimize it because RW never had use cases beyond one-sample streams. If we start seeing more use cases around bulk imports using RW with dozens-hundrends of samples per timeSeries, we can do 2.1 that adds chunks or tailored semantics for samples.
* We make a breaking change in an experimental RW (however, the impact is low, see below)

### Low Breaking Change Impact

This change is in theory **breaking**. However, practically we impact should be minimal. See [implementation change](https://github.com/prometheus/prometheus/pull/17411) that will follow.  Notice that only receiver part was implemented for CT/ST. Given no sending part was done we expect this feature (ST/CT) not being used, thus breakage impact is minimal. This has been confirmed with early adopters like Mimir (Grafana), Chronosphere, Thanos, Cortex and Google.

See previous discussions and 3 explicit approvals: https://github.com/prometheus/prometheus/pull/17036

#### Alternatives Considered

A. Don't do move ST. Rejected, as we miss lot's of consistency, cleanup and arguably efficiency factors.
B. Keep ST per series for cumulatives; Add ST per sample. Rejected, as it makes protocol much more complex to explain and implement without benefits. We send only one sample so there's no efficiency different and if we send more eventually cumulatives will change ST causing duplicated timeseries which is not great either. Let's optimize for multiple-sample case when we will have use cases.
C. Add special semantics for ST per sample to only add it when needed (e.g. when it's different). Similar issues as with (B), maybe something to optimize if more multi-sample per series cases emerge.

### Additional Spec Improvements

* I updated link to proto.
* Updated links to new compliance tests.
* Update native histogram spec link.

cc @alexgreenbank @cstyan @ArthurSens @aknuds1 @krajorama @bboreham @kgoudeaux

### DIFF
--- a/docs/specs/prw/remote_write_spec_2_0.md
+++ b/docs/specs/prw/remote_write_spec_2_0.md
@@ -305,7 +305,7 @@ message Sample {
   // or delta type metrics.
   //
   // For cumulative metrics, the start timestamp represents the time when the
-  // counter started counting (sometimes referred to as start timestamp), which
+  // counter started counting (sometimes referred to as created timestamp), which
   // can increase the accuracy of certain processing and query semantics (e.g. rates).
   //
   // Note:


docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -312,7 +296,30 @@ message Sample {
...
   // timestamp represents timestamp of the sample in ms.
+  //
+  // For Go, see github.com/prometheus/prometheus/model/timestamp/timestamp.go
+  // for conversion from/to time.Time to Prometheus timestamp.
   int64 timestamp = 2;
+  // start_timestamp represents an optional start timestamp for the sample,
+  // in ms format. This information is typically used for counter, histogram (cumulative)
+  // or delta type metrics.
+  //
+  // For cumulative metrics, the start timestamp represents the time when the
```
krajorama, line 307: nit: if you're going to explain cumulative metrics, I'd also say something about delta metrics. E.g. "For delta type metrics, the start timestamp represents the time to which the delta value is relative to ? Probably something better.

docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -271,27 +270,11 @@ message TimeSeries {
...
-  //
-  // Note that the "optional" keyword is omitted due to
-  // https://cloud.google.com/apis/design/design_patterns.md#optional_primitive_fields
-  // Zero value means value not set. If you need to use exactly zero value for
-  // the timestamp, use 1 millisecond before or after.
-  int64 created_timestamp = 6;
 }
 
-// Exemplar represents additional information attached to some series' samples.
+// Exemplar is an additional information attached to some series' samples.
```
krajorama, line 275: nit: not related to start time

docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -249,13 +246,15 @@ message Request {
 
 // TimeSeries represents a single series.
 message TimeSeries {
+  reserved 6;
```
krajorama, line 249: nit: why not put it in order ? I'd move to the end

docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -42,10 +45,10 @@ In this document, the following definitions are followed:
 * a `Sender` is something that sends Remote-Write data.
 * a `Receiver` is something that receives (writes) Remote-Write data. The meaning of `Written` is up to the Receiver e.g. usually it means storing received data in a database, but also just validating, splitting or enhancing it.
 * `Written` refers to data the `Receiver` has received and is accepting. Whether or not it has ingested this data to persistent storage, written it to a WAL, etc. is up to the `Receiver`. The only distinction is that the `Receiver` has accepted this data rather than explicitly rejecting it with an error response.
-* a `Sample` is a pair of (timestamp, value).
-* a `Histogram` is a pair of (timestamp, [histogram value](https://github.com/prometheus/docs/blob/b9657b5f5b264b81add39f6db2f1df36faf03efe/content/docs/concepts/native_histograms.md)).
+* a `Sample` is a triplet of (start timestamp, timestamp, value).
```
krajorama, line 48: But start time is optional, so maybe 
```suggestion
* a `Sample` is a pair (timestamp, value) or a triplet of (start timestamp, timestamp, value).
```
and similar for Histograms.

docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -338,10 +345,11 @@ message Metadata {
   uint32 unit_ref = 4;
 }
 
-// A native histogram, also known as a sparse histogram.
-// See https://github.com/prometheus/prometheus/blob/remote-write-2.0/prompb/io/prometheus/write/v2/types.proto#L142
-// for a full message that follows the native histogram spec for both sparse
-// and exponential, as well as, custom bucketing.
+// A native histogram message, supporting
```
krajorama, line 348: nit: not related to start time

docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -405,6 +412,8 @@ Rationales: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-p
 -->
 Senders MUST send `samples` (or `histograms`) for any given `TimeSeries` in timestamp order. Senders MAY send multiple requests for different series in parallel.
 
+Sample's or Histogram's `start_timestamp` SHOULD be provided for types that follow counter semantics (e.g. counters and histograms). Receivers MAY reject those series without `start_timestamp` being set. Given optionality, the 0 value MUST be treated by receivers as unset value. To represent the unlikely 0 unix timestamp in milliseconds, "1" or "-1" value MUST be used.
```
krajorama: Histogram might be gauge.

```suggestion
Sample's or Histogram's `start_timestamp` SHOULD be provided for types that follow counter semantics (e.g. counters and counter histograms). Receivers MAY reject those series without `start_timestamp` being set. Given optionality, the 0 value MUST be treated by receivers as unset value. To represent the unlikely 0 unix timestamp in milliseconds, "1" or "-1" value MUST be used.
```

docs/specs/prw/remote_write_spec_2_0.md
bwplotka: This is for brevity. This is just shorter snippet of proto.

docs/specs/prw/remote_write_spec_2_0.md
bwplotka: They don't exist now and we can do it later. I'd skip for now.

docs/specs/prw/remote_write_spec_2_0.md
bwplotka: Yes. This is intended, see description. It's better to make changes in atomic version bump especially cosmetic like this.

docs/specs/prw/remote_write_spec_2_0.md
bwplotka: Nice I like it, although one could argue `(0, timestamp, value)` ==  `(timestamp, value)`, no?

docs/specs/prw/remote_write_spec_2_0.md
bwplotka: Yes. This is intended, see description. It's better to make changes in atomic version bump especially cosmetic like this.

bwplotka:
I addressed all comments @krajorama 

Kind ping @ArthurSens @cstyan

dashpole (approved):

docs/specs/prw/remote_write_spec_2_0.md
krajorama: Sure, but it makes it look like we always allocate 3 items, even though apparently there's no overhead if there's only two.

docs/specs/prw/remote_write_spec_2_0.md
krajorama: i still think it looks ugly, but not that important

docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -312,7 +296,30 @@ message Sample {
...
+  // start_timestamp represents an optional start timestamp for the sample,
+  // in ms format. This information is typically used for counter, histogram (cumulative)
+  // or delta type metrics.
+  //
+  // For cumulative metrics, the start timestamp represents the time when the
+  // counter started counting (sometimes referred to as start timestamp), which
+  // can increase the accuracy of certain processing and query semantics (e.g. rates).
+  //
+  // Note:
+  // * That some receivers might require start timestamps for certain metric
```
krajorama, line 312: I wonder if we should state what's expected in special cases:
start timestamp == timestamp  (ignore?)
start timestamp > timestamp (require reject? / ignore? )
Or if we don't want to be / can be specific, maybe say that receivers may reject / ignore start timestamp at will?

docs/specs/prw/remote_write_spec_2_0.md
bwplotka: We do allocate 3 items (e.g. on generated Go structure from proto) .

> even though apparently there's no overhead if there's only two.

What "no overhead" you mention? On proto encoding itself, there's no overhead if you skip AFAIK, practically there's no overhead because we used to have this field per timeseries and Prometheus (and ecosystem) only sends one sample anyway.

But generally there is one field in the tuple to allocate (: 

So it is worth mentioning "no overhead" semantics?

docs/specs/prw/remote_write_spec_2_0.md
bwplotka: Great point!

My first instinct is to follow OpenMetrics model of not specifying those details, given we just start to think about consuming this information in Prometheus. 

There are some Chronosphere (cc @matthiasr and @https://github.com/prometheus/prometheus/issues/17143, do you have opinion; should protocol guarantee some semantics?) and [Google semantics  ](https://docs.cloud.google.com/monitoring/api/ref_v3/rpc/google.monitoring.v3#timeinterval) that will be applied on receiving (rejecting samples).

Similarly for timestamps, we don't tell much either e.g. ts is required and there's certain float unit, but that's it. We never said what happens if you send samples with ts[1] == ts[2] or OOO. Potentially because those things change over time and you don't want to change protocol to change?

Would be nice to find a good framework when to specify vs when to underspecify. I'd vote to underspecify for now given we are just shaping the exact semantics on append in [PROM-60](https://github.com/prometheus/proposals/pull/60)

bwplotka: Otel [mentions "interval" aspect](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#sums:~:text=A%20time%20window%20), but we do here too:

```
For cumulative metrics, the start timestamp represents the time when the counter started counting .
```

docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -312,7 +296,30 @@ message Sample {
...
+  //
+  // For Go, see github.com/prometheus/prometheus/model/timestamp/timestamp.go
+  // for conversion from/to time.Time to Prometheus timestamp.
   int64 timestamp = 2;
+  // start_timestamp represents an optional start timestamp for the sample,
+  // in ms format. This information is typically used for counter, histogram (cumulative)
+  // or delta type metrics.
+  //
+  // For cumulative metrics, the start timestamp represents the time when the
+  // counter started counting (sometimes referred to as start timestamp), which
```
bwplotka:
```suggestion
  // counter started counting (sometimes referred to as created timestamp), which
```

docs/specs/prw/remote_write_spec_2_0.md
krajorama: ok, we'll see home much overhead this creates for us

docs/specs/prw/remote_write_spec_2_0.md
ArthurSens: I'm happy with "underspecifying" for now, too. I believe implementors will naturally reject when `start time > timestamp`. We can be more strict once we notice confusion in the community

docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -4,7 +4,7 @@ nav_title: "2.0"
 sort_rank: 2
 ---
 
-* Version: 2.0-rc.3
+* Version: 2.0-rc.4
 * Status: **Experimental**
 * Date: May 2024
```
bboreham, line 9: Drive-by comment: I expected to see somewhere that the most recent change was November 2025.

PR #2771, opened 2025-11-06
Update 2025-10-30-non-code-contribution.md

nwanduka:
I noticed that the hyperlink to the blog content guide was leading to a 404 error on the live site, even though it worked correctly on GitHub. I think the issue was caused by using a relative link. I’ve updated the link to use an absolute URL so it resolves properly in the published blog.

@juliusv could you review please. thank you


juliusv:
Ah yes, thanks! :+1:

PR #2772, opened 2025-11-07
docs(exposition): clarify required Content-type header for scrapes

LukoJy3D:
In https://github.com/prometheus/prometheus/pull/16155, it was suggested to clarify it in exposition formats. 

While writing it, I feel that some aspects overlap with those mentioned in https://prometheus.io/docs/instrumenting/content_negotiation/. Therefore, it may be sufficient to mention it in the migration guide, but I will give this a try.

### DIFF
--- a/docs/instrumenting/exposition_formats.md
+++ b/docs/instrumenting/exposition_formats.md
@@ -167,6 +167,18 @@ When such features are enabled either by feature flag
 appropriate configuration option (`scrape_native_histograms: true`) then
 Protobuf will be favored over other exposition formats.
 
+## HTTP Content-Type requirements
+
+Starting with Prometheus 3.0, scrape targets **must** return a valid `Content-Type` header for the metrics endpoint. If the `Content-Type` is missing, unparsable, or not a supported media type, **the scrape will fail**. See changes in [scrape protocols](https://prometheus.io/docs/prometheus/latest/migration/#scrape-protocols) in the migration guide for details.
+
+See each of the exposition format sections for the accurate HTTP content types.
+
+### ScrapeProtocols vs Content-Type
+
+Prometheus scrape config offers scrape protocol negotiation based on the content-type using the [`scrape_protocols`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) config. For the Prometheus user convenience the scrape protocols are referenced by a unique name that maps to the concrete content-type. See [Protocol Headers](./content_negotiation.md#protocol-headers) for details.
+
+However, the targets should expose metrics in the exposition format with the absolute, response content-type (e.g. `application/openmetrics-text;version=1.0.0`) and only one.
+
 ## Historical versions
 
 For details on historical format versions, see the legacy


docs/instrumenting/exposition_formats.md
```diff
@@ -167,6 +167,22 @@ When such features are enabled either by feature flag
...
+- `application/openmetrics-text`
+- `text/plain`
+- `application/vnd.google.protobuf`
+
+Adding parameters such as `version=` (e.g. `text/plain;version=1.0.0`) is **optional**. Prometheus will correctly negotiate the protocol based solely on the media type. For the canonical list of protocols, parameters, and the negotiation rules, see [Scrape protocol content negotiation](./content_negotiation.md).
+
+### Common pitfall
+
+The value of the `Content-Type` header must be a **valid MIME type** — plain tokens like `OpenMetricsText1.0.0` are **not valid** and the scrape will fail. Use the full media type instead, `application/openmetrics-text;version=1.0.0'`. See [Protocol Headers](./content_negotiation.md#protocol-headers) for details.
+
```
bwplotka: Hm, this is bit misleading. Content type depends on what you actually expose, but it reads here like it's a random string to put. (: 

What if we make it clearer what the content type is for each section above?

Prometheus text format already mentions content type.


* In Proto section let's mention `application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited`
* In OpenMetrics section `application/openmetrics-text;version=<version>`

Version is NOT always optional. It matters. Especially with OpenMetrics, because there will be breaking changes between `application/openmetrics-text;version=1.0.0` and (upcoming) `application/openmetrics-text;version=2.0.0`

Then we could say:

```suggestion

See each of the exposition format sections for the accurate HTTP content types.

### ScrapeProtocols vs Content-Type

Prometheus scrape config offers to configure negotiation based on the content-type using the [`scrape_protocols`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) config. For the Prometheus user convenience the scrape protocols are referenced by a unique name that maps to the concrete content-type. See the current (v3.7.3) mapping [here](https://github.com/prometheus/prometheus/blob/v3.7.3/config/config.go#L556).

However, the targets should expose metrics in the exposition format with the absolute, response content-type (e.g. `application/openmetrics-text;version=1.0.0`) and only one.
```

bwplotka (commented):
Thanks!

This is great addition, but I propose to word things a bit differently, see the comment. WDYT? Is that helpful? (:

docs/instrumenting/exposition_formats.md
bwplotka: Actually I just noticed https://prometheus.io/docs/instrumenting/content_negotiation lol, so maybe a bit simpler:

```suggestion
See each of the exposition format sections for the accurate HTTP content types.

### ScrapeProtocols vs Content-Type

Prometheus scrape config offers scrape protocol negotiation based on the content-type using the [`scrape_protocols`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) config. For the Prometheus user convenience the scrape protocols are referenced by a unique name that maps to the concrete content-type. See [Protocol Headers](./content_negotiation.md#protocol-headers) for details.

However, the targets should expose metrics in the exposition format with the absolute, response content-type (e.g. `application/openmetrics-text;version=1.0.0`) and only one.
```

LukoJy3D: Yes, I've noticed the content negotiation page mid PR :smile: I like your suggestion, and I applied it. Yet honestly, I'm not sure if there is much to add without overlapping with the mentioned page.

Maybe a slightly different structure would help make this more noticeable.

PR #2773, opened 2025-11-12
Redirect top level spec/<proj> dir to the latest version

bwplotka:
This is to allow redirect of openmetrics.io domain to https://prometheus.io/docs/specs/om/ on CNCF side and then be able to dynamically switch latest on our side, also convenience.

Perhaps the domain hosting CNCF uses allows HTTP redirects something like that. If not we could play with some netlify redirection 🤔


juliusv:
Seems ok to me :) 👍

PR #2777, opened 2025-11-14
Add PromQL `info` function blog post

aknuds1:
I'm proposing to add a blog post on the PromQL `info` function! During the last PromCon, a revelation was that the `info` function is too poorly marketed. One potential remediation mentioned was writing a blog post :)


blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+kind: article
+author_name: Arve Knudsen
+---
+
+Enriching metrics with metadata labels can be surprisingly tricky in Prometheus, even if you're a PromQL wiz!
+Traditionally, complex PromQL join syntax is required in Prometheus to add even basic information like Kubernetes cluster names or cloud provider regions to queries.
+The new, still experimental `info()` function, promises a simpler way, making label enrichment as simple as wrapping your query in a single function call.
+
+In Prometheus 3.0, we introduced the [`info()`](https://prometheus.io/docs/prometheus/latest/querying/functions/#info) function, a powerful new way to enrich your time series with labels from info metrics.
+`info` doesn't only suffer simpler syntax however.
```
ywwg:
```suggestion
`info` doesn't only offer a simpler syntax however.
```

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+It also solves a subtle yet critical problem that has plagued join queries for years: The "churn problem" that causes queries to fail when "non-identifying" info metric labels change.
+Identifying labels here in practice means those that are joined on.
```
ywwg: This sentence is a bit awkward, is this better (and still  accurate?)

```suggestion
In practice, "identifying labels" refers to those labels that the join is performed on.
```

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+    rate(http_server_request_duration_seconds_count[2m]),
+    {k8s_cluster_name=~".+"}
+  )
+)
+```
+
+Much more comprehensible, no?
+The real magic happens under the hood though: **`info()` automatically selects the time series with the latest sample**, eliminating churn-related join failures entirely.
+
+### Basic Syntax
```
ywwg: I know Patrick ran into issues where `info()` had to be put in very specific places, especially when a rate function was involved.  I think his intuition was to "wrap the metric name in info()" but that doesn't work when there's a rate function?  So I think we should have an aside with some advice on how to know where the info function will need to go.

blog/posts/2025-12-16-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+
+**How it works:** When non-identifying info metric labels change (e.g., a pod is re-created), there's a brief period where both old and new series might exist.
+The `info()` function simply selects whichever has the most recent sample, ensuring your queries keep working.
+
+### 2. Better Performance
+
+The `info()` function is more efficient than traditional joins:
+- Only selects matching info series
+- Avoids unnecessary label matching operations
+- Optimized query execution path
```
ywwg, line 248: I know some people were suspicious about the performance of the target_info concept ... Do we have numbers comparing the performance of label promotion vs info() joining?  If it's small enough, that may help discourage people from promoting everything by default.

blog/posts/2025-12-16-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+
+# With aggregation
+sum by (k8s_cluster_name) (
+  info(up, {k8s_cluster_name=~".+"})
+)
+```
+
+## Current Limitations and Future Plans
+
+The current implementation is an **MVP (Minimum Viable Product)** designed to validate the approach and gather user feedback.
```
ywwg, line 262: Can we add a link to directly file an issue?  "If you experience any issues with info() please report them here" etc

blog/posts/2025-12-16-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+- Dynamically determine identifying labels based on the info metric's structure
+
+**Important:** Because this is an experimental feature, the behavior may change in future Prometheus versions, or the function could potentially be removed from PromQL entirely based on user feedback.
+
+## Conclusion
+
+The experimental `info()` function represents a significant step forward in making PromQL more accessible and reliable.
+By simplifying metadata label enrichment and automatically handling the churn problem, it removes two major pain points for Prometheus users, especially those adopting OpenTelemetry.
+
+We encourage you to try the `info()` function and share your feedback:
```
ywwg, line 292: link again

ywwg (changes requested):
This is great! I made a few comments

blog/posts/2025-12-16-introducing-info-function.md
aknuds1: I don't know about perf comparisons versus label promotion. I just know there's an obvious performance benefit over join queries, due to being able to filter what to select on the RHS :)

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+---
+
+Enriching metrics with metadata labels can be surprisingly tricky in Prometheus, even if you're a PromQL wiz!
+Traditionally, complex PromQL join syntax is required in Prometheus to add even basic information like Kubernetes cluster names or cloud provider regions to queries.
+The new, still experimental `info()` function, promises a simpler way, making label enrichment as simple as wrapping your query in a single function call.
+
+In Prometheus 3.0, we introduced the [`info()`](https://prometheus.io/docs/prometheus/latest/querying/functions/#info) function, a powerful new way to enrich your time series with labels from info metrics.
+`info` doesn't only offer a simpler syntax however.
+It also solves a subtle yet critical problem that has plagued join queries for years: The "churn problem" that causes queries to fail when "non-identifying" info metric labels change.
+In practice, "identifying labels" refers to those labels that the join is performed on.
```
ywwg: I think we should expand a section laying out what "identifying" really means, compared to "data labels" or "nonidentifying labels", referencing the otel data model (which also uses the word "descriptive"!)

aknuds1: Do you want to write a section and make a PR to my branch or something? Maybe the easiest.

ywwg: yeah, can do

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
+---
+title: Introducing the Experimental info() Function
+created_at: 2025-11-14
+kind: article
+author_name: Arve Knudsen
+---
+
+Enriching metrics with metadata labels can be surprisingly tricky in Prometheus, even if you're a PromQL wiz!
+Traditionally, complex PromQL join syntax is required in Prometheus to add even basic information like Kubernetes cluster names or cloud provider regions to queries.
```
beorn7: This sounds like it's just a problem of syntax. This blog post might be an opportunity to debunk this naive assumption that is in everyone's head.
It could also be a good segue into the parts of the info function that still need to be finished before it can unleash its true power.

To reiterate the reasons for the complexity of the join query:
- You have to specify on which labels to join (segue into what "identifying labels" means, see @ywwg's comment, and that we want to store this information in the future).
- You have to specify the info metric to join with (segue into storing the type information in the future to know what info metrics exist).
- You have to specify which "data labels" to copy over from the info metric.

Of course, you cannot explain all of this in the opening paragraph, but maybe a bit of it, something like:

```suggestion
The PromQL join query traditionally used for this is inherently quite complex because it has to specify the labels to join on, the info metric to join with, and the labels to enrich with. 
```

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+The new, still experimental `info()` function, promises a simpler way, making label enrichment as simple as wrapping your query in a single function call.
+
+In Prometheus 3.0, we introduced the [`info()`](https://prometheus.io/docs/prometheus/latest/querying/functions/#info) function, a powerful new way to enrich your time series with labels from info metrics.
+`info` doesn't only offer a simpler syntax however.
```
beorn7: Again, I don't think the fundamental nature of the problem is syntactical.

`info` is simpler because you do not have to specify the identifying labels, you do not have to specify the info metric to join with, and if you just want to enrich with all data labels, you do not even have to specify the data labels you want to enrich with.

I think each of the three features is more important (both in practical terms and as a key insight how info works and why it is useful) than the next one (the "churn problem"), among others because the churn problem only occurs with broken staleness handling (which could be fixed in OTLP ingestion), and it should actually never occur if OTel folks just used resource attributes as they are meant to be used. (Of course, we don't need to explain those subtleties here, I just want to put things into perspective.)

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+- Which labels are the "identifying" labels to join on (`job`, `instance`)
+- Which data labels you want to add (`k8s_cluster_name`)
+- The proper PromQL join syntax (`on`, `group_left`)
+
+This requires expert-level PromQL knowledge and makes queries harder to read and maintain.
+
+**2. The Churn Problem (The Critical Issue):**
+
+Here's the subtle but serious problem: What happens when a Kubernetes pod gets recreated?
+The `k8s_pod_name` label in `target_info` changes, and Prometheus sees this as a completely new time series.
```
beorn7: I think that in this situation, also the `instance` label will change (and has to change according to OTel conventions), so this problem shouldn't occur.

Maybe we can still keep using this example, but we should then also mention that this is a broken setup (where `instance` doesn't change upon pod recreation).

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+
+If the old `target_info` series isn't properly marked as stale immediately, both the old and new series can exist simultaneously for up to 5 minutes (the default lookback delta).
```
beorn7: Maybe don't phrase this as an "if". It will never be marked properly as stale because of the way OTLP ingestion works. So on the one hand, we can clearly say that the staleness handling doesn't work right now, but we could also mention that this is an issue that could be fixed in the OTLP ingestion layer.

blog/posts/2025-11-14-introducing-info-function.md
beorn7: I agree in general, but the rationale is not that `info` is special or needlessly complicated here, but the opposite: `info` is just a normal function. Nothing new here. You place it where you place any other function that receives an instant vector. (You could call that "very specific places". I would call it "just the normal expected places".) For example, `label_replace` is placed at exactly the same places where you would use `info`. (Buth functions not only take an instant vector as an argument, they also share the property that they just manipulate labels, not values. All perfectly normal and expected – I might even use the infamous word "intuitive" for that.)

Patrick's issue has to do with PromQL in general, not `info` in particular. Obviously, the intuition of somebody not familiar with PromQL is very different from the intuition of somebody who knows PromQL. Catering for the one might very well be counterproductive for the other.

So we should argue here that `info` is just a normal function acting on any instant vector, not any new concept like a "decorator" or something.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+## The Solution: Simple, Reliable Label Enrichment
+
+The `info()` function solves both problems at once.
+Here's the same query using `info()`:
+
+```promql
+sum by (k8s_cluster_name, http_status_code) (
+  info(
+    rate(http_server_request_duration_seconds_count[2m]),
+    {k8s_cluster_name=~".+"}
```
beorn7: I'm wondering if we should better start with the easiest use case first, as this is about showcasing the simplicity. So let's leave out this line (it is also the one "weird" part of the info function, a new type of function parameter, using selector syntax without being a real selector).

It will promote all data labels, but that's not a problem because we aggregate them away in the `sum` anyway. (You can explain that later.)

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+  )
+)
```
beorn7:
```suggestion
sum by (k8s_cluster_name, http_status_code) (
  info(rate(http_server_request_duration_seconds_count[2m]))
)
```

As per my other comment.

Maybe then follow up with the detail that this uses all data labels, then explain that it doesn't matter because we are aggregating them away in the `sum`, and then introduce the selector to select just one label, in cases where that is needed.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+```
+
+- **`v`**: The instant vector to enrich with metadata labels
+- **`data-label-selector`** (optional): Label matchers in curly braces to filter which labels to include
+
+If you omit the second parameter, `info()` adds **all** data labels from `target_info`:
+
+```promql
+info(rate(http_server_request_duration_seconds_count[2m]))
+```
```
beorn7: As said, I would frame this is the primary mode of using `info`. See above.

blog/posts/2025-12-16-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+info(up, {__name__="build_info"})
+
+# Use multiple info metrics (combines labels from both)
+info(up, {__name__=~"(target|build)_info"})
+
+# Select build_info and only include the version label
+info(up, {__name__="build_info", version=~".+"})
+```
+
+**Note:** The current implementation always uses `job` and `instance` as the identifying labels for joining, regardless of which info metric you select.
```
beorn7, line 127: Maybe also note that the future intention is that PromQL "knows" what metrics in the TSDB are info metrics and automatically uses _all_ of them, _unless_ the selection is explicitly restricted by a name matcher like the above.

General question (not part of the review): Why didn't we make it possible to just write `info(up, build_info)` or `info(up, build_info{version=~".+"})`?
Maybe `info(up, {"build_info"})` or `info(up, {"build_info", version=~".+"})` works already?

blog/posts/2025-12-16-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+This works well for most standard info metrics but may have limitations with custom info metrics that use different identifying labels.
```
beorn7, line 128: Very infamously this would be `kube_pod_labels` from kube-state-metrics. It requires joining on `(namespace, pod)`. This is a pretty common use case. We should not pretend it's fringe and rarely happening, but call out that this doesn't work now but will be fixed in the future (by letting PromQL "know" what labels are identifying).

blog/posts/2025-12-16-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+The real magic happens under the hood though: **`info()` automatically selects the time series with the latest sample**, eliminating churn-related join failures entirely.
+
+### Basic Syntax
+
+```promql
+info(v instant-vector, [data-label-selector instant-vector])
+```
+
+- **`v`**: The instant vector to enrich with metadata labels
+- **`data-label-selector`** (optional): Label matchers in curly braces to filter which labels to include
```
beorn7, line 82: This could also be used to filter which metrics to include.

For example, the following is modified from the first example above, now only including clusters whose name begins with `eu-`:

```promql
sum by (k8s_cluster_name, http_status_code) (
  info(
    rate(http_server_request_duration_seconds_count[2m]),
    {k8s_cluster_name=~"eu-.*"}
  )
)
```

I don't think you have an example like this in the blog post so far, but I think it's a very neat feature.

blog/posts/2025-12-16-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+```promql
+# Add all OTel resource attributes
+info(rate(http_server_request_duration_seconds_sum[5m]))
+
+# Add only specific attributes
+info(
+  rate(http_server_request_duration_seconds_sum[5m]),
+  {k8s_cluster_name=~".+", k8s_namespace_name=~".+", k8s_pod_name=~".+"}
+)
+```
```
beorn7, line 155: And then add an example that filters for specific values of resource attributes.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+Enrich your metrics with Kubernetes-specific information:
+
+```promql
+# Add cluster and namespace information to request rates
+info(
+  sum by (job, http_status_code) (
+    rate(http_server_request_duration_seconds_count[2m])
+  ),
+  {k8s_cluster_name=~".+", k8s_namespace_name=~".+"}
+)
```
beorn7: How is this different from the above?

Maybe we can add an example here that is not tailored to OTel/`target_info`? Maybe something using `build_info`.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+If you omit the second parameter, `info()` adds **all** data labels from `target_info`:
+
+```promql
+info(rate(http_server_request_duration_seconds_count[2m]))
+```
+
+### Selecting Different Info Metrics
+
+By default, `info()` uses the `target_info` metric.
+However, you can select different info metrics (like `build_info`, `node_uname_info`, or `kube_pod_labels`) by including a `__name__` matcher in the data-label-selector:
```
beorn7: `kube_pod_labels` doesn't work because it will have the `job` and `instance` labels from KSM (which we would need to ignore in the join) and requires joining on `namespace` and `pod`, see comment below.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+  rate(cloud_storage_request_count[5m]),
+  {cloud_provider=~".+", cloud_region=~".+", cloud_availability_zone=~".+"}
+)
+```
+
+## Before and After: Side-by-Side Comparison
+
+Let's see how the `info()` function simplifies real queries:
+
+### Example 1: Basic Label Enrichment
```
beorn7: This should include an example where you want to use a lot of resource attributes (because that's very long in the join query, but you just use the default behavior of `info`). Just promoting a single RA is kind-of the worst case for `info` and the best case for the traditional join query, so we shouldn't list it is the 1st example.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+    target_info
+)
+```
+
+**With info():**
+```promql
+sum by (k8s_cluster_name, k8s_namespace_name, http_status_code) (
+  info(
+    rate(http_server_request_duration_seconds_count[2m]),
+    {k8s_cluster_name=~".+", k8s_namespace_name=~".+"}
```
beorn7: Because of the sum, we don't really need this line.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+  )
+)
+```
+
+The intent is much clearer with `info`: We're enriching `http_server_request_duration_seconds_count` with cluster and namespace information, then aggregating by those labels and `http_status_code`.
```
beorn7: I would also add an Example 3 with filtering for label _values_. That's also possible with the join query, but makes things even more verbose, while `info` does it in a very natural way. Another strength of `info` we can and should showcase here.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Technical Benefits
+
+Beyond cleaner syntax, the `info()` function provides several technical advantages:
```
beorn7: It's not just syntax, as said. `info` is utilizing information that has otherwise to be provided by the user. Sorry for hammering this in so much, but I think this is very crucial to understand why `info` is so cool (and also what is still missing in its implementation).

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+info(
+  rate(http_server_request_duration_seconds_count[5m]),
+  {k8s_cluster_name=~".+"}
+)
+
+# With aggregation
+sum by (k8s_cluster_name) (
+  info(up, {k8s_cluster_name=~".+"})
+)
+```
```
beorn7: Not sure if there is value to repeat examples already given above.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+The current implementation is an **MVP (Minimum Viable Product)** designed to validate the approach and gather user feedback.
+It has some intentional limitations:
+
+### Current Constraints
+
+1. **Default info metric:** Only considers `target_info` by default
+   - Workaround: You can use `__name__` matchers like `{__name__=~"(target|build)_info"}` in the data-label-selector, though this still assumes `job` and `instance` as identifying labels
+
+2. **Fixed identifying labels:** Always assumes `job` and `instance` are the identifying labels for joining
+   - This works for most use cases but may not be suitable for all scenarios
```
beorn7: I think this over-promises. There are common use cases already where this doesn't work. You even dropped `kube_pod_labels` above. We should clearly say that this _is_ a real problem but also that we want to solve this in the future (by storing the information which labels are identifying).

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+### Future Development
+
+These limitations are meant to be temporary.
+The experimental status allows us to:
+- Gather real-world usage feedback
+- Understand which use cases matter the most
+- Iterate on the design before committing to a final API
+
+A future version of the `info()` function should:
+- Support all info metrics (not just `target_info`)
```
beorn7:
```suggestion
- Consider all info metrics by default (not just `target_info`)
```

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,298 @@
...
+- Dynamically determine identifying labels based on the info metric's structure
```
beorn7: It's not so much dynamic, but it's about (statically) storing the information what labels are identifying.

Note for future implementations (not related to the review, just happened to come to my mind while doing this review): For something like `kube_pod_labels`, we'll have the actual identifying labels (`namespace`, `pod`), then we'll have the data labels (in this case the actual pod labels), but when these metrics are ingested from KSM, we will also get a `job` and `instance` label attached (and whatever other target labels are configured). The future perfect version of `info` probably should just join on `namespace` and `pod` and only add the pod labels, but completely ignore `job` and `instance` and possibly other target labels (i.e. don't join on them, but also don't add them to the result either). Devil is in the detail here…

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: Revised, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: Revised, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: Revised, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: Revised, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: Revised, PTAL.

blog/posts/2025-12-16-introducing-info-function.md
aknuds1: Revised, PTAL.

aknuds1: > Why didn't we make it possible to just write info(up, build_info) or info(up, build_info{version=~".+"})?

I don't remember it being considered :)

blog/posts/2025-12-16-introducing-info-function.md
aknuds1: Revised, PTAL.

blog/posts/2025-12-16-introducing-info-function.md
aknuds1: I added an example, PTAL.

blog/posts/2025-12-16-introducing-info-function.md
beorn7: I vaguely remember that there was a reason, but I don't remember the reason itself. :grimacing:

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: PTAL.

blog/posts/2025-12-16-introducing-info-function.md
aknuds1: Done, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: Revised, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: Removed it, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: Revised, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: Revised, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: I removed them, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: Revised, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: I added this as example 2 (dropped the original one), PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: I turned example 1 into such an example, PTAL.

blog/posts/2025-12-16-introducing-info-function.md
aknuds1: Done, PTAL.

blog/posts/2025-12-16-introducing-info-function.md
aknuds1: Done, PTAL.

blog/posts/2025-12-16-introducing-info-function.md
aknuds1: It could be that it's difficult to support the syntax technically. Maybe with the current state of the implementation, these ideas could be revisited?

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: PTAL.

beorn7:
Will have a look ASAP. (I'm saying that to many these days… it might not be very soon…)

aknuds1:
Thanks @beorn7!

blog/posts/2025-12-16-introducing-info-function.md
```diff
@@ -0,0 +1,301 @@
...
+
+Here's the subtle but serious problem: What happens when an OTel resource attribute changes in a Kubernetes container, while the identifying resource attributes stay the same?
+An example could be the resource attribute `k8s.pod.labels.app.kubernetes.io/version`.
+Then the corresponding `target_info` label `k8s.pod.labels.app.kubernetes.io/version` changes, and Prometheus sees a completely new `target_info` time series.
+
+As the OTLP endpoint doesn't mark the old `target_info` series as stale, both the old and new series can exist simultaneously for up to 5 minutes (the default lookback delta).
+During this overlap period, your join query finds **two distinct matching `target_info` time series** and fails with a "many-to-many matching" error.
+
+This could in practice mean your dashboards break and your alerts stop firing when infrastructure changes are happening, perhaps precisely when you would need visibility the most.
+
```
beorn7, line 60: Something is missing here. A headline like "The Solution in its simplest form" or something?

Currently, this appears as just another paragraph in the section describing the 2nd part of the problem…

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,301 @@
...
+
+This could in practice mean your dashboards break and your alerts stop firing when infrastructure changes are happening, perhaps precisely when you would need visibility the most.
+
+```promql
+sum by (k8s_cluster_name, http_status_code) (
+  info(rate(http_server_request_duration_seconds_count[2m]))
+)
+```
+
+Much more comprehensible, no?
```
beorn7: Language nit:

```suggestion
Much more comprehensible, isn't it?
```

Unless you intentionally want to use a specific sociolect that is not standard English…

blog/posts/2025-12-16-introducing-info-function.md
beorn7: Yes, I would love that.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,301 @@
...
+
+## Before and After: Side-by-Side Comparison
+
+Let's see how the `info()` function simplifies real queries:
+
+### Example 1: OpenTelemetry Resource Attribute Enrichment
+
+**Traditional approach:**
+```promql
+sum by (http_status_code, k8s_cluster_name, k8s_container_name, k8s_cronjob_name, k8s_daemonset_name, k8s_deployment_name, k8s_job_name, k8s_namespace_name, k8s_pod_name, k8s_replicaset_name, k8s_statefulset_name) (
```
beorn7: Is this a realistic set of labels to sum by? I don't know the semantic conversions by heart, but isn't `k8s_pod_name` essentially unique per pod? There wouldn't be much to sum up in this way… Maybe sum by fewer labels, something that is more likely to happen in a real query.

(Mostly I want to prevent that readers just dismiss the query as contrived.)

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,301 @@
...
+
+### Example 2: Filtering by Label Value
+
+**Traditional approach:**
+```promql
+sum by (http_status_code) (
+    rate(http_server_request_duration_seconds_count[2m])
+  * on (job, instance) group_left ()
+    target_info{k8s_cluster_name="us-east-1"}
+)
```
beorn7: To better demonstrate how `info` reduces redundancies, maybe let's filter for a multitude of clusters, like this:

```suggestion
sum by (http_status_code, k8s_cluster_name) (
    rate(http_server_request_duration_seconds_count[2m])
  * on (job, instance) group_left (k8s_cluster_name)
    target_info{k8s_cluster_name=~"us-.*"}
)
```

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+**With info():**
+```promql
+sum by (http_status_code) (
+  info(
+    rate(http_server_request_duration_seconds_count[2m]),
+    {k8s_cluster_name="us-east-1"}
+  )
+)
```
beorn7: With above change, this would look like:

```suggestion
sum by (http_status_code, k8s_cluster_name) (
  info(
    rate(http_server_request_duration_seconds_count[2m]),
    {k8s_cluster_name=~"us-.*"}
  )
)
```

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+Here we filter to only include metrics from the `us-east-1` cluster. The `info()` version integrates the filter naturally into the data-label-selector.
```
beorn7:
```suggestion
Here we filter to only include metrics from clusters in the US (whose name starts with `us-`). The `info()` version integrates the filter naturally into the data-label-selector.
```

beorn7 (dismissed):
A few more things, but looking very good already.

Note that the date in the filename needs to be updated to the actual date of publication.

blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -0,0 +1,301 @@
+---
+title: Introducing the Experimental info() Function
+created_at: 2025-11-14
```
beorn7: This also needs to be updated with the actual date of publication.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: I revised to what might be a more realistic set of labels. PTAL.

blog/posts/2025-12-16-introducing-info-function.md
aknuds1: You're right. On reviewing the section, I realized that a sub-section was missing. I added it, PTAL.

blog/posts/2025-11-14-introducing-info-function.md
aknuds1: We don't know yet when the date will be, right :) I thought to leave it until the PR has approval?

blog/posts/2025-12-16-introducing-info-function.md
ywwg, line 127: "current implementation" implies that this may change, do we have any plans to? Maybe we can call out that we are looking for feedback on this point

blog/posts/2025-11-14-introducing-info-function.md
beorn7: Just wanted to make sure you are aware.

blog/posts/2025-12-16-introducing-info-function.md
aknuds1: > "current implementation" implies that this may change, do we have any plans to?

Yes the plan is relatively clear, the last sentence of the same paragraph already explains how it's supposed to work in the future. I don't think we need to solicit feedback, since we know how it's supposed to work on top of persisted metadata.

aknuds1:
> Can we make "Giving Feedback" a heading so it's more eyecatching?

@ywwg @beorn7 I moved requests for feedback to a single section "Giving Feedback", just before the "Conclusion" section. WDYT?

beorn7:
I'm traveling until Dec 16. My remaining items were not very significant, and I trust you to address them appropriately, so as soon as the other reviewers are happy, assume I am, too, and don't wait for my explicit approval.

PR #2778, opened 2025-11-17
Update exporters.md add DPtech exporter

dptech-dev:
add dptech_exporter link to Other monitoring systems, exporter info in  github.


jan--f:
Thanks for the contribution. I'm afraid we need to decline this for now. We are happy to feature exporters that are actively developed and used. Unfortuntately this on has so far only 6 commits and no stars over its life time.

PR #2780, opened 2025-11-18
Add info function blog post

ywwg:
is this accurate?

Signed-off-by: Arve Knudsen <arve.knudsen@gmail.com>


blog/posts/2025-11-14-introducing-info-function.md
```diff
@@ -12,7 +12,10 @@ The new, still experimental `info()` function, promises a simpler way, making la
...
+
+The `info()` function helps make this separation explicit by separating labels into two buckets: those on the metric itself (identifying) and those on the target_info metric (non-identifying). When a query is run, the value of labels provided by `info()` can change over time, but the response will appear as a single, non-interrupted series, without the need for further `sum()` aggregations. The question of whether a label is "identifying" -- whether it indicates a distinct series or not -- can vary between installations and is controlled by the user.
```
aknuds1, line 18: I don't really think this makes sense, sorry.

It's true that metric labels are always identifying in Prometheus, but in the _specific context_ of joining with info metrics and calling `info`, a _subset_ is considered identifying. With regards to the `target_info` metric, the `instance` and `job` labels are the identifying subset.

Identifying labels in this particular context are exactly like foreign keys in SQL, you need to know that `instance` and `job` are the labels to join on (unless you use `info`, because it hides that detail).

PR #2781, opened 2025-11-21
Improve HA FAQ entry

SuperQ:
Update the high availaibliy FAQ entry with information about data deduplication.

Remove wording about memberlist as it's confusing to readers that it's part of the Prometheus HA design.


jan--f:
Not sure how I can retrigger the failing checks...

Issue #2732, opened 2025-09-23
[Formal SPEC Discussion clientlibs] Same metric name, different set of label names


incertum:
(Follow for existing community discussion here:  https://github.com/prometheus/client_java/issues/696).

Hi everyone,

To better support dynamic use cases, could we consider relaxing the label names rule from MUST NOT to SHOULD NOT?

And to maintain clarity, what if we added a specific clause that a metric name MUST NOT be used as both labeled and unlabeled at the same time? The idea is to let users manage aggregation of the resulting time series in PromQL (see https://github.com/swift-server/swift-prometheus/pull/131#issuecomment-3304222408).

Does this seem like a reasonable trade-off for more flexibility? 
Curious to hear your perspectives. Thank you very much in advance!

---

The ask:
https://github.com/prometheus/client_java/issues/696#issue-997248606 (example has been slightly adjusted)
```
# HELP app_responses_total responses
# TYPE app_responses_total counter
app_responses_total{uri="/", outcome="SUCCESS"} 10
app_responses_total{uri="/", outcome="FAILURE", error="INVALID_REQUEST"} 2
app_responses_total{uri="/", outcome="FAILURE", error="DATA_NOT_FOUND"} 3
app_responses_total{uri="/", outcome="FAILURE", error="ERROR2"} 20
app_responses_total{uri="/", outcome="FAILURE", error="ERROR3"} 30
```

Conforming to current guidance:
https://github.com/prometheus/client_java/issues/696#issuecomment-920353692
```
# HELP app_responses_total responses
# TYPE app_responses_total counter
app_responses_total{uri="/", outcome="SUCCESS"} 10
app_responses_total{uri="/", outcome="FAILURE"} 5
# HELP app_responses_error_total responses error
# TYPE app_responses_error_total counter
app_responses_error_total{uri="/", outcome="FAILURE", error="ERROR2"} 20
app_responses_error_total{uri="/", outcome="FAILURE", error="ERROR3"} 30
```

For instance, if you use vector addition, you get incorrect aggregations:

```
sum(app_responses_total) by (outcome) + sum(app_responses_error_total) by (outcome)

{outcome="FAILURE"} 55
```

Users must be cautious in selecting a correct query, such as the one provided below:

```
sum(app_responses_total or app_responses_error_total) by (outcome)

{outcome="SUCCESS"} 10
{outcome="FAILURE"} 55
```

---

Retrieved on September 22, 2025:

[...]

PR #2782, opened 2025-11-21
relax stance on metrics with same name but different labels

zeitlinger:
Fixes https://github.com/prometheus/docs/issues/2732


juliusv:
Thanks, seems ok to me 👍

Issue #2766, opened 2025-10-31
OM2.0: Add examples for new ABNF


ywwg:
We should include new examples that exercise the expanded ABNF, particularly UTF-8 support

PR #2784, opened 2025-12-01
OM2.0: Add info and examples for UTF-8

ywwg:
Fixes https://github.com/prometheus/docs/issues/2766

### DIFF
--- a/docs/specs/om/open_metrics_spec_2_0.md
+++ b/docs/specs/om/open_metrics_spec_2_0.md
@@ -439,6 +439,39 @@ go_goroutines 69
 # UNIT process_cpu_seconds seconds
 # HELP process_cpu_seconds Total user and system CPU time spent in seconds.
 process_cpu_seconds_total 4.20072246e+06
+# TYPE "foodb.read.errors" counter
+# HELP "foodb.read.errors" The number of errors in the read path for fooDb.
+{"foodb.read.errors","service.name"="my_service"} 3482
+# EOF
+```
+
+##### UTF-8 Quoting
+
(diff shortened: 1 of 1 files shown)


docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -439,6 +439,29 @@ go_goroutines 69
...
+# HELP "foodb.read.errors" The number of errors in the read path for fooDb.
+{"foodb.read.errors","service.name"="my_service"} 3482
+# EOF
+```
+
+##### UTF-8 Quoting
+
+Metric and label names not conforming to the ABNF definition of `metricname` MUST be enclosed in double quotes and the alternative UTF-8 syntax MUST be used.
+Quoting and the alternative metric name syntax MAY be used for any metric name, regardless of whether the name conforms to `metricname` or not.
+The metric names MUST be enclosed in double quotes in TYPE, UNIT, and HELP lines. The metric name MUST be moved inside the brackets without an operator, in accordance with the ABNF.
```
krajorama:
```suggestion
The metric names MUST be enclosed in double quotes in TYPE, UNIT, and HELP lines. In MetricPoints the metric name MUST be moved inside the brackets without an operator, in accordance with the ABNF.
```

krajorama: "operator" is not defined anywhere, maybe say "without any label name, right after the opening bracket" ?

ywwg:
TODO: add a paragraph about what kinds of names don't need to be quoted.  Regex is probably fine for this vs ABNF

PR #2791, opened 2025-12-03
om2: MetricFamilyName is MetricName

bwplotka:
Fixes https://github.com/prometheus/OpenMetrics/issues/305

~We can either merge this with TODOs before complex values/types or wait for complex type/values.~

### DIFF
--- a/docs/specs/om/open_metrics_spec_2_0.md
+++ b/docs/specs/om/open_metrics_spec_2_0.md
@@ -165,7 +165,7 @@ MetricFamily names beginning with underscores are RESERVED and MUST NOT be used 
 
 ###### Reserved Suffixes
 
-MetricFamily name SHOULD NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically, a name SHOULD NOT create a MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec). Ingestors MAY reject such MetricFamily or MetricSet.
+MetricFamily name SHOULD NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically, a name SHOULD NOT create a MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec). Ingestors MAY reject such MetricFamily.
 
 A non-compliant example would be a gauge called `foo_bucket` and a histogram called `foo`. Exposers negotiating the older OpenMetrics or Text formats, or ingestors which support only the older data model could end up storing the `foo` histogram in the classic representation (`foo_bucket`, `foo_count`, `foo_sum`), which would clash with the gauge and cause a scrape rejection or dropped data.
 


bwplotka:
I am bit surprised but I couldn't find any special rule for those suffixes in the ABNF to change... also tests passes despite me adding suffixes to MF names. Something smells like ABNF is too relaxed.

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -128,27 +130,21 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
-Colons in MetricFamily names are RESERVED to signal that the MetricFamily is the result of a calculation or aggregation of a general purpose monitoring system.
+MetricFamily name:
 
-MetricFamily names beginning with underscores are RESERVED and MUST NOT be used unless specified by this standard.
+* MUST be string.
+* MUST be unique within a MetricSet.
+* MUST be the same as every MetricPoint's MetricName in the family. TODO(bwplotka): This assumes complex values will be introduced. Otherwise, we need to keep suffixes section for histograms and summaries.
 
-###### Suffixes
+> NOTE: [OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec/#suffixes) was allowing different MetricFamily Name vs MetricName in the same family. For example MetricFamily Name `foo` for a counter with a MetricName `foo_total`. This is no longer accepted thanks to ComplexValue and for parser reliability and future compatibility.
```
ArthurSens:
```suggestion
> NOTE: [OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec/#suffixes) allowed different MetricFamily Name vs MetricName in the same family. For example, MetricFamily Name `foo` for a counter with a MetricName `foo_total`. This is no longer accepted for parser reliability and future compatibility.
```

I'm not sure it's needed to mention what thing enabled this to happen 🤔

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -128,27 +130,21 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
 
-The name of a MetricFamily MUST NOT result in a potential clash for sample metric names as per the ABNF with another MetricFamily in the Text Format within a MetricSet. An example would be a gauge called "foo_total" as a counter called "foo" could create a "foo_total" in the text format.
+Names SHOULD be in snake_case. Names SHOULD follow the restrictions in the ABNF section under `metricname`. Metric names MAY be any quoted and escaped UTF-8 string as described in the ABNF section. Be aware that exposing UTF-8 metrics is still experimental and may reduce usability, especially when `_total` or unit suffixes are not included.
```
ArthurSens:
```suggestion
Names SHOULD be in snake_case. Names SHOULD follow the restrictions in the ABNF section under `metricname`. Metric names MAY be any quoted and escaped UTF-8 string as described in the ABNF section.
```

Debatable if this should be removed at all, and definitely not in this PR... but isn't it weird that OpenMetrics mentions experimental features in Prometheus?

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -128,27 +130,21 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
+ 
+Colons in MetricFamily names are RESERVED to signal that the MetricFamily is the result of a calculation or aggregation of a general purpose monitoring system.
```
ArthurSens, line 162: I don't know if this holds any value in the spec. When we mention 'RESERVED', are we instructing SDKs not to allow colons?

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -190,13 +186,13 @@ A gauge MAY be used to encode an enum where the enum has many states and changes
...
-A MetricPoint in a Metric with the type Counter MUST have one value called Total. A Total is a non-NaN and MUST be monotonically non-decreasing over time, starting from 0.
-
-A MetricPoint in a Metric with the type Counter SHOULD have a Timestamp value called Start Timestamp. This can help ingestors discern between new metrics and long-running ones it did not see before.
+A Counter MetricPoint's value MUST be a Number. The value MUST NOT be NaN and MUST be monotonically non-decreasing over time, starting from 0.
 
-A MetricPoint in a Metric's Counter's Total MAY reset to 0. If present, the corresponding Start Timestamp MUST also be set to the timestamp of the reset.
+A Counter MetricPoint SHOULD have a Timestamp value called Start Timestamp. This can help ingestors discern between new metrics and long-running ones it did not see before.
 
-A MetricPoint in a Metric's Counter's Total MAY have an exemplar.
+A Counter MetricPoint MAY reset to 0. If present, the corresponding Start Timestamp MUST also be set to the timestamp of the reset.
```
ArthurSens:
```suggestion
A Counter MetricPoint MAY reset to 0. If so, the corresponding Start Timestamp MUST also be set to the timestamp of the reset.
```

ArthurSens (commented):
I could review only half of the PR 😬, I have some small comments

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -128,27 +130,21 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
-The name of a MetricFamily MUST NOT result in a potential clash for sample metric names as per the ABNF with another MetricFamily in the Text Format within a MetricSet. An example would be a gauge called "foo_total" as a counter called "foo" could create a "foo_total" in the text format.
+Names SHOULD be in snake_case. Names SHOULD follow the restrictions in the ABNF section under `metricname`. Metric names MAY be any quoted and escaped UTF-8 string as described in the ABNF section. Be aware that exposing UTF-8 metrics is still experimental and may reduce usability, especially when `_total` or unit suffixes are not included.
+ 
+Colons in MetricFamily names are RESERVED to signal that the MetricFamily is the result of a calculation or aggregation of a general purpose monitoring system.
 
-Exposers SHOULD avoid names that could be confused with the suffixes that text format sample metric names use.
+MetricFamily names beginning with underscores are RESERVED and MUST NOT be used unless specified by this standard.
 
-* Suffixes for the respective types are:
-* Counter: `_total`
```
bwplotka, line 162: Feedback: on storage clash can happen (STORAGE classic clash)

```
# TYPE foo_count gauge
foo_count ...

# TYPE foo summary
foo { .....}
```

In storage you have
* foo_count
* foo_count classic representation


Unrelated: This is not possible in OM 2.0 (TODO double check if it's clear). (MetricFamily Type and unit uniqueness) 

```
# TYPE foo counter
foo ...

# TYPE foo gauge
foo{ .....}
```

bwplotka: Consensus: We add section around suffix clashes. 

Q: SHOULD avoid collisions vs MUST avoid collision 

Proposal: Keep MUST NOT for now?

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -489,53 +486,60 @@ Timestamps SHOULD NOT use exponential float rendering for timestamps if nanoseco
...
 
 ```openmetrics-add-eof
-# TYPE foo counter
+# TYPE foo_total counter
 ```
 
 If no TYPE is exposed, the MetricFamily MUST be of type Unknown.
 
-If a unit is specified it MUST be provided in a UNIT metadata line. In addition, an underscore and the unit SHOULD be the suffix of the MetricFamily name.
+If a unit is specified it MUST be provided in a UNIT metadata line. In addition, an underscore and the unit SHOULD be part of the suffix chain of the MetricFamily name.
```
bwplotka: Just mentioning it's not a clear suffix anymore (it's not the last trailing part).

bwplotka: TODO: Link to prometheus convention??

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -660,35 +667,37 @@ foo 18.0 456
 
 ##### Counter
 
-The MetricPoint's Total Value Sample MetricName SHOULD have the suffix `_total`. If present, the MetricPoint's Start Timestamp MUST be inlined with the Metric point with a `st@` prefix. If the value's timestamp is present, the Start Timestamp MUST be added right after it. If exemplar is present, the Start Timestamp MUST be added before it.
+The Counter MetricPoint's MetricName SHOULD have the suffix `_total`. Be aware that exposing metrics
+without `_total` suffix may reduce the usability due to confusion about what the metric's type is.
 
-Be aware that exposing metrics without `_total` being a suffix of the MetricFamily name directly to end-users may reduce the usability due to confusion about what the metric's type is.
+// TODO Shouldn't it be before t@?
```
bwplotka: Was queries

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -737,19 +744,21 @@ foo{entity="replica",foo="ccc"} 1.0
 
 ##### Info
 
-The Sample MetricName for the value of a MetricPoint for a MetricFamily of type Info MUST have the suffix `_info`. The Sample value MUST always be 1.
+The Info MetricPoint's MetricName MUST have the suffix `_info`. // TODO(bwplotka): Shouldn't this "SHOULD"?
```
bwplotka: Consensus: Info MUST is fine for Otel. (no conflicting rule, no native info type). If we encode we can follow Prometheus convention.
```suggestion
The Info MetricPoint's MetricName MUST have the suffix `_info`.
```

docs/specs/om/open_metrics_spec_2_0.md
dashpole: It many cases it _required_ different MetricFamily Name vs MetricName.

docs/specs/om/open_metrics_spec_2_0.md
dashpole: I know this is just moved, but we should clarify that `exposing UTF-8 metrics is still experimental` doesn't apply to its support in this protocol, just to its support elsewhere in the Prometheus ecosystem.

docs/specs/om/open_metrics_spec_2_0.md
dashpole: I also don't know what RESERVED means? I don't think it is one of the keywords (like MUST, SHOULD, MUST NOT).

dashpole: I think we can say "MUST NOT be used unless ..."

ywwg:
do we still intend to merge this?

docs/specs/om/open_metrics_spec_2_0.md
bwplotka: I will keep RESERVED as it was there before my change. We use RESERVED word 6 times already in this document, so let's have a dedicated PR to change it.

I think it's fine, but to be strict with https://datatracker.ietf.org/doc/html/rfc2119 we could check MUST NOT.

Let's focus on suffixes in this PR, WDYT?

docs/specs/om/open_metrics_spec_2_0.md
bwplotka: yes, I updated it

docs/specs/om/open_metrics_spec_2_0.md
bwplotka: Good point, not experimental, killing

docs/specs/om/open_metrics_spec_2_0.md
bwplotka: Added entry.

bwplotka:
I think so. Much simpler now with Composite Types merged in.

bwplotka:
PTAL @krajorama @dashpole @ywwg

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -87,7 +87,7 @@ CompositeValue MUST contain all information necessary to recreate a sample value
 The following Metric Types MUST use CompositeValue for Metric Values:
 
 * [Histogram](#histogram) MetricFamily Type.
-* [GaugeHistogram](#gauge-histogram) MetricFamily Type.
+* [GaugeHistogram](#gaugehistogram) MetricFamily Type.
```
bwplotka, line 90: FYI: Broken link

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -146,27 +146,40 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
+when expanded to [an OpenMetrics 1.0 Text format](https://prometheus.io/docs/specs/om/open_metrics_spec).
+
+An example would be a gauge called `foo_bucket` and a histogram called `foo`. Ingestors which implements older data model
+could end up storing `foo` histogram in a classic representation (`foo_bucket`, `foo_count`, `foo_sum`) causing scrape rejection
+or dropped data.
 
-The name of a MetricFamily MUST NOT result in a potential clash for sample metric names as per the ABNF with another MetricFamily in the Text Format within a MetricSet. An example would be a gauge called "foo_total" as a counter called "foo" could create a "foo_total" in the text format.
+Exposers MUST avoid names that could conflict with the Text OpenMetrics 1.0 [histogram](https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec.md#histogram-1), [gaugehistogram](https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec.md#gaugehistogram-1) and [summary](https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec.md#summary-1) suffixes.
```
bwplotka: Another way to put it is that gauges, counters, stateset names MUST NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket` - kind of simple rule. Would that be more clear?

krajorama: I'd just simplify further:

MetricFamily name MUST NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`.

It's true for anything except Histograms and Gauge Histograms , including Summary since that has series without magic suffixes. But I don't think it's worth making an exception for Histograms, Gauge Histograms. Especially because I don't think people will actually follow this unless SDKs prohibit it and I think we need to add a line about ingestors:

Ingestors MAY accept non compliant MetricFamily names and only error out when a naming clash occurs.

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -146,27 +146,40 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
+Names SHOULD be in snake_case. Names SHOULD follow the restrictions in the ABNF section under `metricname`. Metric names MAY be any quoted and escaped UTF-8 string as described in the ABNF section. Be aware that exposing UTF-8 metrics may reduce usability, especially when `_total` or unit suffixes are not included in the names.
 
 Colons in MetricFamily names are RESERVED to signal that the MetricFamily is the result of a calculation or aggregation of a general purpose monitoring system.
 
 MetricFamily names beginning with underscores are RESERVED and MUST NOT be used unless specified by this standard.
 
-###### Suffixes
+###### Forbidden Suffixes 
+
+This specification is following a shift in Prometheus ecosystem towards [composite values](#compositevalues) instead of [the "classic" representation](https://prometheus.io/docs/specs/om/open_metrics_spec/#histogram-1). However, this transformation will take time. To support ingestors
```
krajorama: I think the rule must come first, explanation later.

docs/specs/om/open_metrics_spec_2_0.md
bwplotka: Generally good point, but especially _count might be common as a standalone gauge or even counter. This line would prohibit for compatibility reason (so tmp time) and actually is not needed if the same target does NOT expose `<metric without _count_but_histogram>`.

> Ingestors MAY accept non compliant MetricFamily names and only error out when a naming clash occurs.

Again, if this needs to be stronger that we need to move `MetricFamily name MUST NOT end with _count, _sum, _gcount, _gsum, _bucket.` to SHOULD NOT.

Every "must" MAY be accepted by some implementations, no? 

I proposed SHOULD NOT for wide rule, and MUST NOT for collision explanation - framed as "OpenMetrics 2 converted to OpenMetrics 1 or Prom TEXT should not cause collisions)

docs/specs/om/open_metrics_spec_2_0.md
bwplotka: done

docs/specs/om/open_metrics_spec_2_0.md
dashpole: thats also fine by me

docs/specs/om/open_metrics_spec_2_0.md
dashpole: Yeah, I agree with SHOULD NOT.  MUST NOT presumably means that we place more of a burden on clients exposing the metrics to ensure it doesn't happen (like suffixes in OM 1.0). We either would fail scrapes with bad suffixes, or we would have to "fix" names by modifying them to make them valid (which we sometimes did for OM 1.0).

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -146,27 +146,34 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
 Colons in MetricFamily names are RESERVED to signal that the MetricFamily is the result of a calculation or aggregation of a general purpose monitoring system.
 
 MetricFamily names beginning with underscores are RESERVED and MUST NOT be used unless specified by this standard.
 
-###### Suffixes
+###### Reserved Suffixes
 
-The name of a MetricFamily MUST NOT result in a potential clash for sample metric names as per the ABNF with another MetricFamily in the Text Format within a MetricSet. An example would be a gauge called "foo_total" as a counter called "foo" could create a "foo_total" in the text format.
+MetricFamily name SHOULD NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically name MUST NOT create 
+MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec).
```
ywwg: nit - are we manually word-wrapping or no? Other paragraphs seem not to have newlines mid-paragraph. The markdown format doesn't care but we might as well be consistent.

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -146,27 +146,34 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
 
 Colons in MetricFamily names are RESERVED to signal that the MetricFamily is the result of a calculation or aggregation of a general purpose monitoring system.
 
 MetricFamily names beginning with underscores are RESERVED and MUST NOT be used unless specified by this standard.
 
-###### Suffixes
+###### Reserved Suffixes
 
-The name of a MetricFamily MUST NOT result in a potential clash for sample metric names as per the ABNF with another MetricFamily in the Text Format within a MetricSet. An example would be a gauge called "foo_total" as a counter called "foo" could create a "foo_total" in the text format.
+MetricFamily name SHOULD NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically name MUST NOT create
```
ywwg:
```suggestion
MetricFamily name SHOULD NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically a name MUST NOT create a MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec).
```

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -146,27 +146,34 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
-* Suffixes for the respective types are:
-* Counter: `_total`
-* Summary: `_count`, `_sum`, `` (empty)
-* Histogram: `_count`, `_sum`, `_bucket`
-* GaugeHistogram: `_gcount`, `_gsum`, `_bucket`
-* Info: `_info`
-* Gauge: `` (empty)
-* StateSet: `` (empty)
-* Unknown: `` (empty)
+> This rule is because this specification is following a shift in Prometheus ecosystem towards [composite values](#compositevalues) instead of [the "classic" representation](https://prometheus.io/docs/specs/om/open_metrics_spec/#histogram-1). However, this transformation will take time.
```
ywwg:
```suggestion
> This rule exists because this specification is following a shift in Prometheus ecosystem towards [composite values](#compositevalues) instead of [the "classic" representation](https://prometheus.io/docs/specs/om/open_metrics_spec/#histogram-1). However, this transformation will take time.
```

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -146,27 +146,34 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
+###### Reserved Suffixes
 
-The name of a MetricFamily MUST NOT result in a potential clash for sample metric names as per the ABNF with another MetricFamily in the Text Format within a MetricSet. An example would be a gauge called "foo_total" as a counter called "foo" could create a "foo_total" in the text format.
+MetricFamily name SHOULD NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically name MUST NOT create 
+MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec).
 
-Exposers SHOULD avoid names that could be confused with the suffixes that text format sample metric names use.
+A non-compliant example would be a gauge called `foo_bucket` and a histogram called `foo`. Exposers negotiating the older
+OpenMetrics or Text formats, or ingestors which support only the older data model could end up storing `foo` histogram 
+in a classic representation (`foo_bucket`, `foo_count`, `foo_sum`) causing clash with the gauge and scrape rejection or dropped data.
```
ywwg:
```suggestion
A non-compliant example would be a gauge called `foo_bucket` and a histogram called `foo`. Exposers negotiating the older OpenMetrics or Text formats, or ingestors which support only the older data model could end up storing the `foo` histogram in the classic representation (`foo_bucket`, `foo_count`, `foo_sum`), which would clash with the gauge and cause a scrape rejection or dropped data.
```

docs/specs/om/open_metrics_spec_2_0.md
bwplotka: We don't need to, will revert. We could use autoformatter too to remove those if we want

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -146,27 +146,30 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
 
 Colons in MetricFamily names are RESERVED to signal that the MetricFamily is the result of a calculation or aggregation of a general purpose monitoring system.
 
 MetricFamily names beginning with underscores are RESERVED and MUST NOT be used unless specified by this standard.
 
-###### Suffixes
+###### Reserved Suffixes
 
-The name of a MetricFamily MUST NOT result in a potential clash for sample metric names as per the ABNF with another MetricFamily in the Text Format within a MetricSet. An example would be a gauge called "foo_total" as a counter called "foo" could create a "foo_total" in the text format.
+MetricFamily name SHOULD NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically a name MUST NOT create a MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec).
```
bwplotka: From WG:

* Let's have other guidance for naming conventions (snake_case, unit) (TODO)

```suggestion
MetricFamily a name MUST NOT create a MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec). This can happen for MetricFamily names with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`.
```

* What if exposer has this metric, what should be done? We need consistent logic for violating MUST NOT. Implementations are used to be relaxing. TODO research failure modes for violating various MUST NOT rules? e.g. server side will reject this.  We don't want to put burden on clients. SDK blocking.
* General usability generic protocol vs e2e Prometheus experience?
* Collision move to should not and be around naming?
* Stronger case than just naming suffix.
* leaning into "SHOULD NOT and explain ingestors/Prometheus will likely reject such scrape for foreseeable future (for this case)"

krajorama: Assuming SHOULD NOT for the suffixes

## MUST NOT conflict

To ensure that the user doesn't run into this, the SDK must validate. We don't do this now, so introducing the check is technically breaking (you can register foo_total gauge and foo counter).

SDKs already check for some collisions ([go](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/registry.go#L298), [java](https://github.com/prometheus/client_java/blob/8c229df26e7779246e88f16694a14c379abf2d26/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/registry/PrometheusRegistry.java#L154). 

But not specifically for this. So I think it's not that hard to implement on top of existing checks !? 

It might force a major version, as this can be technically breaking, but on the other hand, I think we'll need a major version anyway to be able to introduce some strict mode that checks that Prometheus  naming conventions are followed. SDK strict mode should be the default and I think it should be possible to turn it off generally or per metric.

_Even with this validation, the user might run into problems later due to federation or some other feature._

## SHOULD NOT conflict

As you say, we can explain why it's bad, but no burden on anyone, the ingestor will just reject the sample  if it runs into it. This is simpler. It would be super annoying if we rejected the whole scrape due to one metric.

## Conclusion

I think to avoid breaking change due to OM2.0 I'd also say SHOULD NOT.

But I'll continue to advocate for SDKs with default strict mode that checks that the name specified follows Prometheus naming conventions.

dashpole: As long as strict mode is opt-in, I don't think you'd need a major version bump.

krajorama: > As long as strict mode is opt-in, I don't think you'd need a major version bump.

Ok, we don't have to decide here, I think it's a question related to the dev summit item [Fabian: How to determine the exposition format](https://docs.google.com/document/d/1uurQCi5iVufhYHGlBZ8mJMK_freDFKPG0iYBQqJ9fvA/edit?tab=t.0#heading=h.hcuvlxdktinx)

bwplotka: I assume SHOULD NOT you refer to is for both "nots"?

> MetricFamily name THIS end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically a name THIS create a MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec).

bwplotka: If yes then changing to:

> MetricFamily name SHOULD NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically a name SHOULD NOT create a MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec). Ingestor MAY reject such samples.

Although I think it's not ideal for protocol to specify: You can do this, but ingestor can ignore it for wide cases (it's wide because Prometheus will do it now). I think I can accept this as it might be lifted later on (if Prometheus implements native storage for composite types).

docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -146,27 +146,30 @@ A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, H
...
 
 Colons in MetricFamily names are RESERVED to signal that the MetricFamily is the result of a calculation or aggregation of a general purpose monitoring system.
 
 MetricFamily names beginning with underscores are RESERVED and MUST NOT be used unless specified by this standard.
 
-###### Suffixes
+###### Reserved Suffixes
 
-The name of a MetricFamily MUST NOT result in a potential clash for sample metric names as per the ABNF with another MetricFamily in the Text Format within a MetricSet. An example would be a gauge called "foo_total" as a counter called "foo" could create a "foo_total" in the text format.
+MetricFamily name SHOULD NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically, a name SHOULD NOT create a MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec). Ingestors MAY reject such MetricFamily or MetricSet.
```
bwplotka:
```suggestion
MetricFamily name SHOULD NOT end with `_count`, `_sum`, `_gcount`, `_gsum`, `_bucket`. Specifically, a name SHOULD NOT create a MetricName collision when converted to [the Text OpenMetrics 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec). Ingestors MAY reject such MetricFamily.
```

PR #2799, opened 2025-12-15
Fix broken prom.lib link in console docs

### LINKED ISSUE #2649: Broken documentation link to prom.lib example file
> Custom formats can be defined. See [prom.lib](https://github.com/prometheus/prometheus/blob/main/console_libraries/prom.lib) for examples.

This line of documentation in https://prometheus.io/docs/visualization/consoles/ results in a 404.

It seems to me that this console library mentioned is necessary for the example in the Consoles documentation.

cnaples79:
## Summary
- replace the dead prom.lib link with a reference to the console_libraries examples shipped in Prometheus release artifacts

## Rationale
- prom.lib was removed upstream, so the existing link 404s; this keeps the guidance without pointing to a missing file.

## Testing
- not run (docs-only change)

Fixes #2649

### DIFF
--- a/docs/visualization/consoles.md
+++ b/docs/visualization/consoles.md
@@ -120,8 +120,9 @@ Valid output formats for the third argument to `prom_query_drilldown`:
   This is usually used with `B` as the second argument to produce units such as `KiB` and `MiB`.
 * `printf.3g`: Display 3 significant digits.
 
-Custom formats can be defined. See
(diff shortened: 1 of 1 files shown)


jan--f:
Hey @cnaples79, #2812 recently added a warning to the docs referring back the the 2.x code stream. Given that we no longer include the examples in the release payload, perhaps this change here is no longer needed?

jan--f:
Please feel free to re-open if something still needs to be addressed.

PR #2802, opened 2025-12-18
update cadvisor doc to support inlini config

devarsh10:
This PR addresses [#17613](https://github.com/prometheus/prometheus/issues/17613)

Added documentation for using inline Docker configs as an alternative to bind mounts when deploying Prometheus with Docker Compose.

This approach is particularly useful when deploying to remote Docker hosts, as it eliminates the need to manually manage configuration files on the host filesystem. All configuration is embedded in the `docker-compose.yml` file.

**Key additions:**
- Complete example using Docker's `configs` feature
- Required `uid`, `gid`, and `mode` parameters (with explanation of the error users encounter if omitted)
- Guidance on finding correct UID/GID values for different images
- Use cases for when this approach is beneficial

### DIFF
--- a/docs/guides/cadvisor.md
+++ b/docs/guides/cadvisor.md
@@ -103,15 +103,7 @@ version: '3.8'
 configs:
   prometheus_config:
     content: |
-      global:
-        scrape_interval: 15s
-        evaluation_interval: 15s
-
       scrape_configs:
-        - job_name: 'prometheus'
-          static_configs:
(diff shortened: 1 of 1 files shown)


divStar (approved):
Especially the config block configuration looks exactly like what my experience setting it up was. Thank you!

devarsh10:
> Especially the config block configuration looks exactly like what my experience setting it up was. Thank you!

Thank you.

PR #2811, opened 2026-01-09
feat(om2.0): allow float histograms

krajorama:
Fixes: https://github.com/prometheus/OpenMetrics/issues/309

OM1.0 said that count and bucket values are integers, which disallowed NaN and +-Inf and float values implicitly. Allowing floats means we have to explicitly say if we want to disallow NaN or +-Inf.

After this PR:

The semantics still require counter like (monotonic until reset) behavior for Histograms.

There are some issues with float precision: we require Count be equal to bucket values added together , but that might not be possible due to precision issues. I'm glossing over this. Same for requirement that difference between the two is equal to number of NaN when using Native Buckets.

The semantics still prohibit negative values for Count and bucket values for Histograms, but not anymore for GaugeHistograms. The ABNF doesn't care.

The semantics now allow NaN in the Count and bucket values for both Histograms and GaugeHistograms. The ABNF also allows NaN.

The semantics now allow +-Inf in the Count and bucket values. The ABNF also allows +-Inf value.

Notes:

We did consider separate ABNF definitions for integer and float histograms and special rules for making sure we can tell from the `count` field if the native histogram is going to be all integers or not. However the fact that Prometheus stores integer and float valued histograms separately is just an optimization. In fact during PromQL evaluation, all integers are converted to floats anyway.

From https://github.com/prometheus/OpenMetrics/issues/309#issuecomment-3718621952 :
> Since the most expensive operation is always the memory allocations (and histograms have a main body + 2x spans, 2x buckets potentially), we could simply do two passes:

> -   first (opt: find the boundaries of the values and) check if the counter/zero counter/bucket values are all integer or not, while also count spans and buckets
> -   allocate the correct type and correctly sized spans and bucket arrays and fill in the values

> This would avoid complicated switchover from integer to float type and also resizing buffers. My guess is that the histogram sample complex value (excluding exemplars) will probably fit into CPU L1 cache, so these two passes will be very fast.

### DIFF
--- a/docs/specs/om/open_metrics_spec_2_0.md
+++ b/docs/specs/om/open_metrics_spec_2_0.md
@@ -250,7 +250,7 @@ Histograms measure distributions of discrete events. Common examples are the lat
 
 A Histogram MetricPoint MUST contain Count and Sum values.
 
-The Count value MUST be equal to the number of measurements taken by the Histogram. The Count is a counter semantically. The Count SHOULD be an integer. The Count MUST NOT be +Inf, NaN, or negative.
+The Count value MUST be equal to the number of measurements taken by the Histogram. The Count is a counter semantically. The Count SHOULD be an integer. The Count MUST NOT be negative. The Count SHOULD NOT be +Inf, NaN.
 
 Float Count is allowed to make it possible to expose results of arithmetic operations on histograms, such as addition that may result in values beyond the range of integers.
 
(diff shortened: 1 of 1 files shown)


docs/specs/om/open_metrics_spec_2_0.md
```diff
@@ -250,13 +250,17 @@ Histograms measure distributions of discrete events. Common examples are the lat
 
 A Histogram MetricPoint MUST contain Count and Sum values.
 
-The Count value MUST be equal to the number of measurements taken by the Histogram. The Count is a counter semantically. The Count MUST be an integer and MUST NOT be NaN or negative.
+The Count value MUST be equal to the number of measurements taken by the Histogram. The Count is a counter semantically. The Count SHOULD be an integer. The Count MUST NOT be +Inf, NaN, or negative.
```
bwplotka: From WG sync: Sounds like we want to allow this given it rare cases of math ops it's possible.

PR #2814, opened 2026-01-18
Update .NET / C# link to maintained alternative

GGG-KILLER:
Replace the link pointing to [prometheus-net](https://github.com/prometheus-net/prometheus-net) (which seems to be abandoned at this point in time, the last release was 3 years ago) with an official (language creator/maintainer-wise) documentation link on how to do it.

The tutorial itself shows screenshots of a working sample using both Prometheus and Grafana.

@RichiH


GGG-KILLER:
There also seems to be an [alternative using OTLP](https://github.com/open-telemetry/opentelemetry-dotnet/blob/core-1.14.0/docs/metrics/getting-started-prometheus-grafana/README.md), however I'm not sure what's the standard on this page.

Perhaps both could be listed for people who don't want to enable OTLP collection in their setup?

Issue #2794, opened 2025-12-09
Add link validation to CI


ArthurSens:
We've been seeing expiring links on the website quite often. Currently, we're not very effective at identifying and removing expired links, so we may want to implement automation to address this issue.

that-avg-dev:
would like to work on this issue , can you please assign it to me and guide a bit on what level of validation is needed or just the non expired is gtg?

ArthurSens:
Hey @kairveeehh, you can take a look at bwplotka/mdox or other similar tools. I'm pretty these kind of tools are being used in different Prometheus repositories.

I'd like to avoid assigning issues to non-members because we often assign issues, and contributors disappear afterwards. Feel free to open a PR even without the assignment!

that-avg-dev:
> Hey @kairveeehh, you can take a look at bwplotka/mdox or other similar tools. I'm pretty these kind of tools are being used in different Prometheus repositories.
> 
> I'd like to avoid assigning issues to non-member because we often getting assigned to issues and disapearing afterwards. Feel free to open a PR even without the assignment!

Sure ! Will do this , thanks.

that-avg-dev:
> Hey [@kairveeehh](https://github.com/kairveeehh), you can take a look at bwplotka/mdox or other similar tools. I'm pretty these kind of tools are being used in different Prometheus repositories.
> 
> I'd like to avoid assigning issues to non-members because we often assign issues, and contributors disappear afterwards. Feel free to open a PR even without the assignment!

should I make the internal or local links to be ignored or they should be notified about as well ? Asking because during local testing they are to be found more prominently.

Another imp thing - that should it block PR's ( if yes it will block mine as well ) , I can thou add up another issue and work over cleaning the invalid links for once and then it can be this way  
Another is to automate the removal but then I am not sure if you would want it to be that way because there are a lot of broken links( internal and external , image's link etc)  
My take would be this - 
Runs weekly and finds broken HTTP(S) links
Automatically removes them from markdown files `(converts [text](broken-url) to just text)`
Commits and pushes the changes back to the repo with message `"chore: remove broken links from documentation [automated]"`
let me know what approach works for you! 
thanks.

ArthurSens:
> should I make the internal or local links to be ignored or they should be notified about as well ? Asking because during local testing they are to be found more prominently.

Do those broken internal links end up as broken links in the website, too? If they are broken on the website, then it's a good idea to notify us as well!

[...]

PR #2820, opened 2026-01-29
CI: add link validation

that-avg-dev:
solves #2794 
@ArthurSens  , it checks and lets you have an artifact after it runs , scheduled weekly and I have tested manually as well.
There are certain cases where you might want to ignore the fact that the link is broken but those are very specific to be added inside the rules for the CI , kindly review and let me know if anything is to be modified.


that-avg-dev:
@ArthurSens kindly review it.

that-avg-dev:
hii @ArthurSens , @jan--f  I have made the changes as requested and tested the CI manually on my forked repo too , it fails as expected.
kindly review.

.github/workflows/link-validation.yml
```diff
@@ -0,0 +1,29 @@
+name: Link Validation (Weekly Check)
+
+on:
+  schedule:
+    - cron: '0 6 * * 1'  # Mondays at 06:00 UTC
```
ArthurSens:
```suggestion
```

As Jan mentioned, maybe let's remove the schedule since we don't know how often we can keep up with notifications. If we use manual triggers, at least we can control when we want to see the reports.

Makefile
```diff
@@ -0,0 +1,4 @@
+.PHONY: mdox-check
+
+mdox-check:
+	@mdox fmt --links.validate --check $$(find . -type d -name node_modules -prune -o -type d -name .next -prune -o -type d -name .git -prune -o -name '*.md' -print)
```
ArthurSens, line 3:
```suggestion
.PHONY: mdox-check
mdox-check:
	@mdox fmt --links.validate --check $$(find . -type d -name node_modules -prune -o -type d -name .next -prune -o -type d -name .git -prune -o -name '*.md' -print)
```

This empty line isn't needed

scripts/remove-broken-links.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
ArthurSens: Could we remove this script? I'm not sure removing broken links is the way we wanna fix everything. I suspect we'll need to manually review everything and decide on a case-by-case approach, whether we want to remove or replace the link

that-avg-dev:
> Apologies for the delayed review here! I was traveling during the last week and couldn't keep up with GitHub notifications.
> 
> Just two comments, and it's good to merge in my opinion!
> 
> I can see some improvement opportunities, but let's tackle them in follow-up PRs and issues:
> 
> * The Makefile assumes whoever runs this has mdox installed. We can use some cool tricks to download mdox for the user automatically if it's not there
> * Some links don't work locally, but work on the [website](https://prometheus.io/docs/visualization/browser/)... these warnings could be ignored.

No problem , it is completely ok :) 
Sure, will work further over these improvements.

did the suggested changes as well, can be reviewed for the same.

ArthurSens:
I've created https://github.com/prometheus/docs/issues/2828 to discuss next steps

PR #2823, opened 2026-02-04
Histogram spec: Rework and amend operators section

beorn7:
- Divide operators section into sub-sections and reformat accordingly.

- Add that negative results of multiplication count as gauge histogram.

- Correct and complete the sub-section about trim operators.

### DIFF
--- a/docs/specs/native_histograms.md
+++ b/docs/specs/native_histograms.md
@@ -1743,7 +1743,8 @@ in that case, treating it as if it were a counter histogram.
 
 ### Interpolation within a bucket
 
-When estimating quantiles or fractions, PromQL has to apply interpolation
+When estimating quantiles or fractions of observations in a histogram, or when
+removing observations from a histogram, PromQL has to apply interpolation
 within a bucket. In classic histograms, this interpolation happens in a linear
 fashion. It is based on the assumption that observations are equally
 distributed within the bucket. In reality, this assumption might be far off.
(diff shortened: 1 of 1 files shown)


beorn7:
Now that we have nailed down what to do in https://github.com/prometheus/prometheus/pull/17904 , we can finalize this. @krajorama & @linasm what do you think about this? I'm wondering how detailed we should describe the edge cases in the spec (or pseudo-spec, as I prefer to call it) here. One could argue that they are all implied with the current wording, but another line of argument would say we should spell out the edge cases here explicitly. (I currently tend towards the former, as it is more in line with the other parts of this document.)

beorn7:
As per discussion over on https://github.com/prometheus/prometheus/pull/17904, I'll add a detailed edge case description here (hopefully tomorrow).

beorn7:
I think the current version should cover all edge cases.

@krajorama @linasm please verify.

docs/specs/native_histograms.md
```diff
@@ -1853,14 +1872,20 @@ circumstances involving the `HistogramStatsIterator`, which includes additional
...
 scalar). This will lead to “scaled” and sometimes even negative histograms,
 which is usually only useful as intermediate results inside other expressions
-(see also [notes above](#unary-minus-and-negative-histograms)). Multiplication
-works for both counter histograms and gauge histograms, and their flavor is left
-unchanged by the operation.
+(see also [notes above](#unary-minus-and-negative-histograms)).
+
+Multiplication works for both counter histograms and gauge histograms, and
+their flavor is left unchanged by the operation, with the exception that a
+negative histogram is always considered to be a gauge histogram.
```
krajorama: I don't think we check for negative value, so maybe

```suggestion
their flavor is left unchanged by the operation, with the exception that
multiplying by a negative value always results in a gauge histogram.
```

docs/specs/native_histograms.md
```diff
@@ -1870,30 +1895,86 @@ and sum of observations all set to `+Inf`, `-Inf`, or `NaN`, depending on their
...
+
+Note that this estimation of the sum of observations is inaccurate (even in the
+case where the cutoff threshold conicides with a bucket boundary), up to a
+point where it could yield results that are obviously wrong. For example, after
+removing some positive observation, the estimated sum of observations could be
+larger than the sum of observations in the original histogram. The estimation
+algorithm could be refined to take such cases into account, but it is kept
+deliberately simple to make it easier to reason with. In general, the
+histograms resulting from trim operations are meant to be used for quantile
+estimation. Their sum of observations is considered to be of limited use.
```
krajorama: I'd like to get my use case in here, because I think it's particularly useful.

I'm suggesting the following, but this buries "Their sum of observations is considered to be of limited use." a little bit, so needs some work:

```suggestion
deliberately simple to make it easier to reason with. 
Their sum of observations is considered to be of limited use.
In general, the histograms resulting from trim operations are
meant to be used for quantile estimation or filtering. The trim
operator can be a good replacement of  `histogram_fraction` in
cases when one boundary is infinite, such as calculating number
of observations below a certain threshold for an alert.
For example using `histogram_count(h </ x)` instead of
 `histogram_fraction(-Inf, x, h) * histogram_count(h)` is simpler
 and also yields `0` instead of `NaN` in case `h` has no observations,
 because the fraction would divide 0 by 0 in the later case. 
```

Or we can flip it around and just mention this case here and elaborate in `histogram_fraction` ?

krajorama (changes requested):
looks good over all, made two suggestions

docs/specs/native_histograms.md
beorn7: I think your suggestion is fine, it just needs a bit more wordsmithing. See what I made of it after the next push…

beorn7: I also think trim operators work just fine even if both boundaries are not Inf.

beorn7:
Done. PHAL.

docs/specs/native_histograms.md
```diff
@@ -1870,30 +1895,93 @@ and sum of observations all set to `+Inf`, `-Inf`, or `NaN`, depending on their
...
+
+The removal is only precise if the threshold coincides with a bucket boundary.
+Otherwise, interpolation within the affected buckets has to be used, as
+described [above](#interpolation-within-a-bucket) with the following exceptions
+for NHCBs:
+
+- If the lowest bucket has a lower boundary of -Inf, all observation in that
+  bucket are considered to be of value -Inf (rather than the upper boundary).
+- If the highest bucket has an upper boundary of +Inf, all observation in that
+  bucket are considered to be of value -Inf (rather than the lower boundary).
```
krajorama: Typo?

```suggestion
  bucket are considered to be of value +Inf (rather than the lower boundary).
```

krajorama (approved):

PR #2829, opened 2026-02-08
Add webhook integrations

SoloJacobs:
I was going to add these integrations to the Alertmanager documentation, but I then stumbled on this page.


SoloJacobs:
That confused me too, but I think it actually is just a confusing disclaimer. It says right after:
>  Integrations that were created before version 1.3.21 (1 August 2023) were marked as (Legacy) and migrated.

And I think they actually mean this part of the docs (which you can reach by scrolling as well).
https://grafana.com/docs/grafana-cloud/alerting-and-irm/irm/configure/integrations/integration-reference/oncall/alertmanager/#note-about-legacy-integration

I spent at least 30 min trying to find a better link, and am 90% sure there isn't a better one.

SoloJacobs:
To make this a bit clearer: I don't think the header about the legacy integrations applies to the documentation, which is linked, it is just shown at the top.

jan--f:
Thanks for checking :)

PR #2269, opened 2023-01-31
Document extra scrape metrics.

vaxvms:
As theses metrics are disabled to avoid increasing cardinality and not because they are breaking changes or are considered experimental, it make sense to include them is the documentation.

PR #2891, opened 2026-03-12
Document extra scrape metrics.

vaxvms:
As theses metrics are disabled to avoid increasing cardinality and not because they are breaking changes or are considered experimental, it make sense to include them is the documentation.

Follow up of #2269 with requested changes

### DIFF
--- a/docs/concepts/jobs_instances.md
+++ b/docs/concepts/jobs_instances.md
@@ -41,6 +41,15 @@ the following time series:
 * `scrape_series_added{job="<job-name>", instance="<instance-id>"}`:
    the approximate number of new series in this scrape. *New in v2.10*
 
+if you enable [`extra-scrape-metrics` feature](/docs/prometheus/latest/feature_flags/#extra-scrape-metrics), you will also get:
+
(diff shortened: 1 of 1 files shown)


docs/concepts/jobs_instances.md
```diff
@@ -41,6 +41,15 @@ the following time series:
 * `scrape_series_added{job="<job-name>", instance="<instance-id>"}`:
    the approximate number of new series in this scrape. *New in v2.10*
 
+if you enable [`extra-scrape-metrics` feature](/docs/prometheus/latest/feature_flags/#extra-scrape-metrics), you will also get:
```
krajorama, line 44: These are already documented below.
I think a more important change would be to document that they should be enabled by configuration and that the feature flag is being phased out . See https://github.com/prometheus/prometheus/pull/17606

jan--f: Hello from the bug scrub!
I agree with the duplication here.
@vaxvms do you want to address the request to document the configuration?

jan--f:
Hello from the bug scrub!

PR #2932, opened 2026-03-26
feat: add blog post on django prom/otel instrumentation

Wineshuga:
### Summary
This PR adds a blog post that documents my experience instrumenting a Django application with both Prometheus and OpenTelemetry using only their official documentation.

### DIFF
--- a/blog/posts/2026-03-26-instrument-django-with-prometheus-and-opentelemetry.md
+++ b/blog/posts/2026-03-26-instrument-django-with-prometheus-and-opentelemetry.md
@@ -1,0 +1,55 @@
+---
+title: "What Instrumenting a Django App Taught Me About Prometheus and OpenTelemetry Docs"
+created_at: 2026-03-26
+kind: article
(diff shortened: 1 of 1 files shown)


ArthurSens (commented):
Hey @Wineshuga, thanks for the blog proposal and feedback on the docs!

I believe the content looks great, honestly, but we need to be careful about how this will be perceived by end users. This is the official Prometheus blog after all.

A blog post advertising how our docs could be improved adds no value to the Prometheus end user; it actually has the opposite effect 😅, I can see users deciding not to use Prometheus after reading this blog.

To proceed here, I'd encourage you to publish this blog post on your personal account, and we'd appreciate any feedback or help with our own docs! I think your experience is valid, and we can make improvements based on your learnings!

Wineshuga:
Hello @ArthurSens. 

Thanks for the feedback and suggestion. I’ve published the blog on my personal page as recommended.

I’ll go ahead and close this PR.

PR #2973, opened 2026-04-16
docs: clarify basic auth wording

ANISHSAJIKUMAR:
Small docs-only cleanup in the basic auth guide. The sentence now talks about configuration instead of setup, which reads more naturally for the reader.


jan--f:
Hello from the bug scrub!

Given that we have PRs opened for every commit in this PR here and none have signed the DCO, I assume that these are bot generated. I also disagree with some of these changes here. Closing.

PR #2974, opened 2026-04-17
fix: license notice

adielcristo:
Update the footer license notice with text copied from the `page.tsx` file.

https://github.com/prometheus/docs/blob/main/src/app/page.tsx#L68


juliusv (approved):
Thanks!

PR #2975, opened 2026-04-17
feat: serve raw markdown via content negotiation

roidelapluie:
Add a Netlify edge function that serves raw .md files when the client sends Accept: text/markdown or text/plain. A build-time script copies markdown sources into out/_md/docs/ for the edge function to proxy.

Apparently useful for agents.

### DIFF
--- a/netlify/edge-functions/md-negotiate.ts
+++ b/netlify/edge-functions/md-negotiate.ts
@@ -1,0 +1,32 @@
+import type { Config, Context } from "@netlify/edge-functions";
+
+export default async function handler(req: Request, ctx: Context) {
+  const accept = req.headers.get("Accept") ?? "";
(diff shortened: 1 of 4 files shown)


jan--f (approved):
Nice! This lgtm.

juliusv (approved):
Generally looks ok to me too. I want to call out two considerations that are probably ok, but at least we should be explicitly aware of them:

* We are making the site deployment way more Netlify-dependent than before (it's no longer just serving static files).
* I wonder what the performance hit will be to serve every request through an edge function rather than just serving a static file. Probably fine? Does Netlify's caching along with the `Vary: Accept` take care of it still?

roidelapluie:
We might be better of with a llms.txt.

PR #2976, opened 2026-04-21
feat: add sitemap generation for static, docs, and blog routes

roidelapluie:

### DIFF
--- a/src/app/sitemap.ts
+++ b/src/app/sitemap.ts
@@ -1,0 +1,58 @@
+import fs from "fs";
+import path from "path";
+import { MetadataRoute } from "next";
+import { docsCollection } from "@/docs-collection";
+import { getAllPostFileNames, postFileNameToPath } from "@/blog-helpers";
+import docsConfig from "../../docs-config";
+
+export const dynamic = "force-static";
+
+function getStaticAppRoutes(): string[] {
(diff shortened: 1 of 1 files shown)


src/app/sitemap.ts
```diff
@@ -0,0 +1,54 @@
...
+export default function sitemap(): MetadataRoute.Sitemap {
+  const base = docsConfig.siteUrl;
+
+  const staticEntries: MetadataRoute.Sitemap = getStaticAppRoutes().map(
+    (route) => ({ url: `${base}${route}` })
+  );
+
+  const docsEntries: MetadataRoute.Sitemap = Object.values(docsCollection)
+    .filter((doc) => {
+      if (doc.type === "local-doc") return true;
```
juliusv: Nit: I'm not a fan of Claude's (I assume) usual omission of braces around single-line if-condition bodies, it makes the `return` control flow harder to spot for me. Could we do this here and below:

```typescript
if (doc.type === "local-doc") {
  return true;
}
```

juliusv (dismissed):

src/app/sitemap.ts
roidelapluie: Done

juliusv (approved):

PR #2982, opened 2026-04-30
Update exporters.md

twekkel:
Add a Lite version of the official node_exporter


jan--f:
Hello from the bug scrub!
Thank you for your contribution, I really like the project. I'm afraid we need to decline this for now. We are happy to feature exporters that are actively developed and used. Unfortunately this one has so far only a limited community behind it and seemingly low adoption. Please try again at a later stage we are also thinking about how we can be less selective while keeping our maintenance load manageable.

PR #2978, opened 2026-04-22
Add banner for Prom exporter vs Collector receiver survey

ArthurSens:
Adds a banner promoting the survey we're doing in collaboration with OTel's End User SIG.

Here's the link for the doc we're using to organize the survey: https://docs.google.com/document/d/1gfE8UmeHaAmFG6ZjRiVDO2Phi-5Vk3mLpMU3YJQ58PU/edit?tab=t.0#heading=h.zbk80121fq6r

Leaving this as a draft until we have the sibling PR opened in OTel's website :)

PR #2983, opened 2026-05-06
Fix banner build

ArthurSens:
Follow up to #2978 

The logs made me think that the problem was while rendering old release pages from the Prometheus repository, but after further investigation, the problem was indeed on the banner!

The banner was already broken before the changes in #2978, but the change to the active date finally caused the build to render it, whereas it had not been rendered before.

I'm definitely not a front-end developer, so I rely a lot on LLM tools to make the fixes here. Please let me know if you have better ways to fix this!

Logs for the build failure on main can be seen here: https://app.netlify.com/projects/prometheus-docs/deploys/69fb29ec6ee9e60008fe05f2

Root cause analysis from the LLM:
```
`src/app/docs/layout.tsx` was a Server Component that rendered Mantine's
`Popover`, `PopoverTarget`, and `PopoverDropdown` directly for the mobile docs
navigation. `PopoverTarget` clones its child and reads `child.props.className`.
That is safe in a Client Component, but brittle when this interactive Mantine
popover is rendered through the Server Component tree during static export.

The announcement commit exposed the issue because it activated a client-rendered
header branch for the first time in this date window. The fetched repository docs
were only where Next happened to be prerendering when the shared docs layout
crashed; they were not the cause.
```

### DIFF
--- a/src/app/docs/DocsMobileNav.tsx
+++ b/src/app/docs/DocsMobileNav.tsx
@@ -1,0 +1,39 @@
+"use client";
+
+import {
+  Button,
+  Popover,
+  PopoverDropdown,
+  PopoverTarget,
+  ScrollAreaAutosize,
+} from "@mantine/core";
+import { IconMenu2 } from "@tabler/icons-react";
+import LeftNav from "./LeftNav";
+
+export default function DocsMobileNav() {
+  return (
+    <Popover position="bottom" withArrow shadow="md">
+      <PopoverTarget>
+        <Button
+          hiddenFrom="sm"
+          variant="outline"
+          mb="lg"
+          leftSection={<IconMenu2 stroke={1.5} />}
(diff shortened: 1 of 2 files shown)


sysadmind:
I have not had to deal with server vs client components in Mantine, but according to their docs, all components are already client components. ~~https://help.mantine.dev/q/server-components#should-i-add-use-client-directive-to-all-files.~~

Actually turns out that sometimes `use client` is the right answer if you look further down at those errors on that link.

ArthurSens:
Ah gotcha, so the only fix here is that we're adding `use client;` at the top?

sysadmind:
I'm not sure @ArthurSens. Javascript has a lot of "magic" that I find hard to reason about. In this case, it's what decides if a component is a client component or a server component. I think these changes are probably fine.

sysadmind (approved):
LGTM

ArthurSens:
Thank you for the help @sysadmind !

PR #2989, opened 2026-05-13
ci: pin contents: read on md-validation and openmetrics

arpitjain099:
Pins the default `GITHUB_TOKEN` to read-only on the two PR-time validation workflows. Both only checkout and run local scripts (`make check` for markdown link checking, ABNF validator for the OpenMetrics 2.0 spec); no GitHub API writes.


jan--f (changes requested):
This makes sense. @arpitjain099 once you sign the commit to pass the DCO test we can merge this.

arpitjain099:
> This makes sense. @arpitjain099 once you sign the commit to pass the DCO test we can merge this.

@jan--f thanks! I signed the DCO. Please check again

arpitjain099:
Re-signed the commit with SSH (now shows Verified on GitHub) while I was at it. DCO was already passing from the earlier `-s` signoff, so no functional change here, just thought I'd attach the verification badge since I had signing set up. Ready when you are @jan--f.

PR #3004, opened 2026-05-31
Add F5 BIG-IP exporter to the list of exporters

Haameed:
This PR adds the F5 BIG-IP exporter to the list of third-party exporters.

- **Repository**: https://github.com/Haameed/f5_bigip_exporter
- **Port**: 11000 (registered in the [port allocation wiki](https://github.com/prometheus/prometheus/wiki/Default-port-allocations))
- **License**: MIT

It's a Prometheus exporter for F5 BIG-IP devices using the iControl REST API.
It follows the multi-target `/probe` pattern (like blackbox_exporter) and
supports monitoring multiple devices concurrently. Metrics cover virtual
servers, pools, system/CPU/memory, disk, SSL certificates, HA/sync status,
and global traffic.


jan--f:
Hello from the bug scrub!

PR #3008, opened 2026-06-09
Update node-exporter.md

hayakauwu:
add in example of simple systemd service to make node-exporter running 24/7

Reasoning being: i have used it along with Prometheus over the years and would have appreciated it being there when i was a newbie. If you feel the addition isn't relevant anymore, feel free to close.

### DIFF
--- a/docs/guides/node-exporter.md
+++ b/docs/guides/node-exporter.md
@@ -70,6 +70,33 @@ Success! The Node Exporter is now exposing metrics that Prometheus can scrape, i
 curl http://localhost:9100/metrics | grep "node_"
 ```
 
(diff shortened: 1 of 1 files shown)


jan--f (changes requested):
Hi,
I don't think we should add this particular change to the documentation. Systemd services should be written carefully and the one added here will likely not work well for most users.
Typically most users would probably look to their distributions package manager to install node-exporter, which will usually add a systemd service (or equivalent for the distros chosen init system). We should point user to their distros first and foremost.
Additionally we have example units in the [node_exporter examples](https://github.com/prometheus/node_exporter/tree/master/examples/systemd) plus more for other init systems.

What we cloud do here is call out that the instructions run node_exporter in the foreground and users should look towards their distro packages for service units or use the one from the node_exporter repo.

jan--f:
Hello from the bug scrub!

Closing due to lack of response.

PR #3014, opened 2026-06-16
Add Prometheus 3.13 as an LTS release

roidelapluie:
The new website engine renders Markdown tables through Mantine, which
ignores the Bootstrap row classes (danger/success/info) that previously
color-coded support status. Convert the LTS table to a Markdown table
with an explicit Status column and bold the currently supported release
instead of relying on row colors.


krajorama (approved):
lgtm

PR #3015, opened 2026-06-16
docs: add cluster-wide inhibition example to alerting tutorial

### LINKED ISSUE #1353: Needs example of cluster wide inhibition
The documentation here states that inhibition can be used to suppress the same alert coming from an entire cluster:

- https://prometheus.io/docs/alerting/alertmanager/#inhibition

However, none of the examples I can find online show how this can be done easily in either of these places:

- https://github.com/prometheus/alertmanager/blob/master/doc/examples/simple.yml#L94
- https://github.com/prometheus/alertmanager/blob/master/examples/ha/alertmanager.yml#L14

yugstar:
Closes #1353

The Alertmanager docs mention that inhibition can suppress alerts from an entire cluster, but as noted in #1353 there is no concrete example showing how. As suggested on the issue, this adds an **"Inhibiting alerts from an entire cluster"** section to the *Alerting based on metrics* tutorial.

It shows an `inhibit_rules` configuration that mutes every alert sharing the same `cluster` label as a firing `ClusterUnreachable` alert, and explains `source_matchers`, `target_matchers`, and the `equal` label list (using the current matcher syntax).

### DIFF
--- a/docs/tutorials/alerting_based_on_metrics.md
+++ b/docs/tutorials/alerting_based_on_metrics.md
@@ -82,3 +82,27 @@ Open [http://localhost:9090/rules](http://localhost:9090/rules) in your browser 
 <iframe width="560" height="315" src="https://www.youtube.com/embed/xaMXVrle98M" frameborder="0" allowfullscreen></iframe>
 
 Similarly Alertmanager can be configured with other receivers to notify when an alert is firing.
+
+## Inhibiting alerts from an entire cluster
+
+When a whole cluster (or instance) becomes unreachable, you usually don't want a separate notification for every alert that fires as a consequence. Alertmanager's [inhibition](/docs/alerting/latest/alertmanager/#inhibition) feature lets a single "cluster is down" alert mute all the dependent alerts coming from that same cluster, so you receive one meaningful notification instead of a flood.
+
+Inhibition is configured with `inhibit_rules` in `alertmanager.yml`. The following rule mutes every alert that shares the same `cluster` label value as a firing `ClusterUnreachable` alert:
+
+> alertmanager.yml
+
+```yaml
+inhibit_rules:
+  - source_matchers:
+      - 'alertname = "ClusterUnreachable"'
+    target_matchers:
+      - 'alertname != "ClusterUnreachable"'
+    equal:
+      - 'cluster'
+```
+
+- `source_matchers` selects the alert that suppresses others when it is firing (here, `ClusterUnreachable`).
+- `target_matchers` selects the alerts to mute. `ClusterUnreachable` is excluded so the source alert itself is still delivered.
+- `equal` lists the labels whose values must match between the source and target alerts for the inhibition to apply. Alerts are muted only when they share the **same** `cluster` value, so an outage in one cluster never hides alerts from another.
+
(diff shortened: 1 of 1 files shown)


docs/tutorials/alerting_based_on_metrics.md
```diff
@@ -81,4 +81,28 @@ Open [http://localhost:9090/rules](http://localhost:9090/rules) in your browser
 
 <iframe width="560" height="315" src="https://www.youtube.com/embed/xaMXVrle98M" frameborder="0" allowfullscreen></iframe>
 
-Similarly Alertmanager can be configured with other receivers to notify when an alert is firing.
\ No newline at end of file
+Similarly Alertmanager can be configured with other receivers to notify when an alert is firing.
+
+## Inhibiting alerts from an entire cluster
+
+When a whole cluster (or instance) becomes unreachable, you usually don't want a separate notification for every alert that fires as a consequence. Alertmanager's [inhibition](/docs/alerting/latest/alertmanager/#inhibition) feature lets a single "cluster is down" alert mute all the dependent alerts coming from that same cluster, so you receive one meaningful notification instead of a flood.
```
jan--f, line 88:
```suggestion
When an entire cluster becomes unreachable, you usually don't want separate notifications for every alert that fires as a consequence. Alertmanager's [inhibition](/docs/alerting/latest/alertmanager/#inhibition) feature lets a single, firing "cluster is down" alert mute all alerts coming from that same cluster, so you receive one meaningful notification instead of a flood.
```

docs/tutorials/alerting_based_on_metrics.md
```diff
@@ -81,4 +81,28 @@ Open [http://localhost:9090/rules](http://localhost:9090/rules) in your browser
...
+
+```yaml
+inhibit_rules:
+  - source_matchers:
+      - 'alertname = "ClusterUnreachable"'
+    target_matchers:
+      - 'alertname != "ClusterUnreachable"'
+    equal:
+      - 'cluster'
+```
```
jan--f, line 102: Let's show users the full `alertmanager.yml` here, similarly to how this is done previously in this file. You can re-use the previous listing and just add the inihibit rule.

docs/tutorials/alerting_based_on_metrics.md
```diff
@@ -81,4 +81,28 @@ Open [http://localhost:9090/rules](http://localhost:9090/rules) in your browser
...
+
+- `source_matchers` selects the alert that suppresses others when it is firing (here, `ClusterUnreachable`).
+- `target_matchers` selects the alerts to mute. `ClusterUnreachable` is excluded so the source alert itself is still delivered.
```
jan--f, line 105: Below it says the triggering alert is always delivered. Maybe its worth quoting the docs here: 

> To prevent an alert from inhibiting itself, an alert that matches both the target and the source side of a rule cannot be inhibited by alerts for which the same is true (including itself). However, we recommend to choose target and source matchers in a way that alerts never match both sides. It is much easier to reason about and does not trigger this special case.

jan--f (changes requested):
Thanks, this looks pretty good already.
I have a few inline suggestions.

Additionally it might be good to add a listing for a `ClusterUnreachable` alert to complete the tutorial.

Maybe also link the doc reference https://prometheus.io/docs/alerting/latest/configuration/#inhibit_rule to further reading at the end?

jan--f:
Hello from the bug scrub!

PR #3017, opened 2026-06-17
Serve machine-readable download.json at site root

roidelapluie:
Add a download.json, served at https://prometheus.io/download.json, listing the releases of every repository shown on the downloads page. The per-release shape mirrors Go's version.json (version, stable, files with os/arch/sha256/ size), with Prometheus-specific "latest" and "lts" flags added.

The file is generated by fetch-downloads-info.ts into generated/ and exposed at the site root via a committed symlink in public/, matching the existing repo-docs-assets mechanism.

### DIFF
--- a/public/download.json
+++ b/public/download.json
@@ -1,0 +1,1 @@
+../generated/download.json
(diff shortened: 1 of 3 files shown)


bwplotka (approved):
Thanks, makes sense!

Image

PR #3022, opened 2026-06-25
Add alternative Haskell client library

brandonchinn178:
`prometheus` is another popular client library for Prometheus; this documentation should probably not take a stance on which one is recommended

### DIFF
--- a/docs/instrumenting/clientlibs.md
+++ b/docs/instrumenting/clientlibs.md
@@ -27,8 +27,8 @@ Unofficial third-party client libraries:
 * [Delphi](https://github.com/marcobreveglieri/prometheus-client-delphi)
 * [Elixir](https://github.com/deadtrickster/prometheus.ex)
 * [Erlang](https://github.com/deadtrickster/prometheus.erl)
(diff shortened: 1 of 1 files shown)


jan--f:
Hello from the bug scrub! Thank you for your contribution.

Iiuc the two hackage packages are generated from https://github.com/fimad/prometheus-haskell. So this doesn't add anything unless I'm mistaken. By convention I would prefer to keep links to the source code over language packages.

Regarding the multilevel list: I'm not against this but it departs from a currently consistent style. If you would like to propose introducing this, please open a PR for the purpose and also consider other similar list (like https://prometheus.io/docs/instrumenting/exporters/ and possible others) and add a string argument why  multilevel lists are an improvement.

In the mean time I'll close this PR here. Please feel free to reopen if you disagree with any of the above.

brandonchinn178:
@jan--f No, these are two completely separate libraries. prometheus is not generated from prometheus-client

brandonchinn178:
I'm unable to reopen the PR. Can you please reopen it?

jan--f:
Ah I see now, the other haskell lib lives at https://github.com/bitnomial/prometheus.

jan--f (approved):

PR #3034, opened 2026-07-08
fix: link governance to GitHub repository

### LINKED ISSUE #3028: New governance is not linked on the website
The website currently still shows https://prometheus.io/governance/. This should be updated to link to https://github.com/prometheus/governance/blob/main/GOVERNANCE.md

jan--f:
Update governance links on the community page and FAQ to point to the canonical governance document in the prometheus/governance repository. Remove the local /governance/ page that hosted an outdated copy.

Fixes: https://github.com/prometheus/docs/issues/3028


SuperQ:
I'm thinking we should add the governance repo to the `githubMarkdownSources` in `/docs-config.ts` so we can inline the contents rather than link to it.

jan--f:
Added inlining only for the governance doc, which requires rewriting some links in the document. 

Happy to pull in  `prometheus/governance` repo, if there is a preference for that. Perhaps we should even include the election documents in the webpage?

SuperQ (approved):

PR #3035, opened 2026-07-09
docs: improve 'first steps' with data verification and metric discovery

### LINKED ISSUE #3032: docs: Issue with "First steps with Prometheus"
**Page:** https://prometheus.io/docs/introduction/first_steps/

**Describe the issue:**
My overall goal is to capture Opentelemetry data and display it. I'm an experienced mainframe person, but have only just started using prometheus.

I've been through the getting started, and I'm stuck on what next.   I'm looking for the step after getting started.
From  status -> health I believe I have connected opentelemetry  to Prometheus  both running in docker, but I do not know if any data is flowing

It would be good if you could provide instructions on the second step
For example
- Provide some sample data in a file.
- Explain how I know what data I can query.   For example the documentation mentions promhttp_metric_handler_requests_total.   How do I list the available fields, so I can use them in a query.  In the OTEL records, there are many fields.   
-  I issued a curl query and got a list of fields, but it does not have any records from OTEL.   How am I meant to know what fields are available?
- Provide a metric _records_processed_ and it shows all the  scrapers - their status and how many records they have processed ( and perhaps last read time)
- Then as part of the second step say issue the query.... to show how many records you have read from which source.

Third step.
Im not sure if I can provide saved queries to save me having to type long queries every time.
The third step should provide an example of creating/editing/deleting a saved rule on the sampled data provided

This information may already exist - I just can't find it.

Thank you

Colin Paice

Atharva-Ramawat:
Fixes #3032

Hi team! This PR addresses the friction points mentioned in issue #3032 for new users going through the "First Steps" guide.

I have added:
* Instructions on verifying data flow using the **Status > Targets** page.
* Guidance on discovering available metrics using the **Metrics Explorer** and autocomplete.
* A brief section introducing **Recording Rules** for saving complex queries, linking out to the full documentation.

### DIFF
--- a/docs/introduction/first_steps.md
+++ b/docs/introduction/first_steps.md
@@ -73,8 +73,9 @@ To start Prometheus with our newly created configuration file, change to the dir
 
 Prometheus should start up. You should also be able to browse to a status page about itself at http://localhost:9090. Give it about 30 seconds to collect data about itself from its own HTTP metrics endpoint.
 
-You can also verify that Prometheus is serving metrics about itself by
-navigating to its own metrics endpoint: http://localhost:9090/metrics.
+You can verify that Prometheus is serving metrics about itself by navigating to its own metrics endpoint: http://localhost:9090/metrics. 
+
+To confirm that Prometheus is actively scraping this data, navigate to **Status > Targets** in the top menu of the Prometheus UI. You should see the `prometheus` endpoint listed with a state of `UP`. This is the easiest way to verify that data is successfully flowing from your targets into Prometheus.
 
 ## Using the expression browser
 
@@ -82,9 +83,11 @@ Let us try looking at some data that Prometheus has collected about itself. To
 use Prometheus's built-in expression browser, navigate to
 http://localhost:9090/query and choose the "Table" tab.
 
-As you can gather from http://localhost:9090/metrics, one metric that
-Prometheus exports about itself is called
-`promhttp_metric_handler_requests_total` (the total number of `/metrics` requests the Prometheus server has served). Go ahead and enter this into the expression console:
+To see what metrics are available to query, you can use the **Metrics Explorer**. Click the globe icon next to the "Execute" button, which reveals a list of all metrics currently being scraped. Alternatively, you can start typing in the expression bar to trigger autocomplete suggestions. 
+
+A great metric to start with is `up`, which shows the current health status of all configured scrapers (a value of `1` means healthy). To see how many data points your scrapers are actively processing, you can query `scrape_samples_scraped`.
+
+For this tutorial, we will use a metric that Prometheus exports about itself called `promhttp_metric_handler_requests_total` (the total number of `/metrics` requests the Prometheus server has served). Go ahead and enter this into the expression console:
 
 ```
 promhttp_metric_handler_requests_total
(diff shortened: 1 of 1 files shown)


docs/introduction/first_steps.md
```diff
@@ -73,18 +73,21 @@ To start Prometheus with our newly created configuration file, change to the dir
 
 Prometheus should start up. You should also be able to browse to a status page about itself at http://localhost:9090. Give it about 30 seconds to collect data about itself from its own HTTP metrics endpoint.
 
-You can also verify that Prometheus is serving metrics about itself by
-navigating to its own metrics endpoint: http://localhost:9090/metrics.
+You can verify that Prometheus is serving metrics about itself by navigating to its own metrics endpoint: http://localhost:9090/metrics. 
+
+To confirm that Prometheus is actively scraping this data, navigate to **Status > Targets** in the top menu of the Prometheus UI. You should see the `prometheus` endpoint listed with a state of `UP`. This is the easiest way to verify that data is successfully flowing from your targets into Prometheus.
```
jan--f, line 78: Let's use a localhost link here as well instead of giving navigation instructions?

docs/introduction/first_steps.md
```diff
@@ -73,18 +73,21 @@ To start Prometheus with our newly created configuration file, change to the dir
...
 ## Using the expression browser
 
 Let us try looking at some data that Prometheus has collected about itself. To
 use Prometheus's built-in expression browser, navigate to
 http://localhost:9090/query and choose the "Table" tab.
 
-As you can gather from http://localhost:9090/metrics, one metric that
-Prometheus exports about itself is called
-`promhttp_metric_handler_requests_total` (the total number of `/metrics` requests the Prometheus server has served). Go ahead and enter this into the expression console:
+To see what metrics are available to query, you can use the **Metrics Explorer**. Click the globe icon next to the "Execute" button, which reveals a list of all metrics currently being scraped. Alternatively, you can start typing in the expression bar to trigger autocomplete suggestions.
```
jan--f: The Metrics Explorer doesn't really tell you about available metrics. Checking the `/metrics` endpoint does that, so I'd prefer to keep the old wording here. Perhaps mention that the targets page also has a link to the metric endpoints being scraped.

Linking the query interface on localhost here as well is a good idea though.

There is a way to query prometheus for all metrics, but we should _not_ add that to a first steps tutorial (or at all really) as this is a potentially DoS causing query. Hitting a particular metrics endpoint is much better.

docs/introduction/first_steps.md
```diff
@@ -119,6 +122,14 @@ rate(promhttp_metric_handler_requests_total{code="200"}[1m])
...
 You can experiment with the graph range parameters and other settings.
 
+## Saving queries (Recording Rules)
+
+If you find yourself typing the same long queries repeatedly, Prometheus allows you to save them as **Recording Rules**. 
+
+Instead of re-evaluating a complex query every time you refresh a dashboard, a recording rule evaluates the query in the background on a schedule and saves the result as a brand new, pre-computed metric. This is essential for both performance and convenience as your monitoring infrastructure grows.
+
+To learn how to configure your first saved query, read the [Recording rules documentation](/docs/prometheus/latest/configuration/recording_rules/).
+
```
jan--f: Not sure we should add this in a first steps tutorial.

The convenience aspect of recording rules is perhaps also a bit overstated here.

Perhaps we should rather mention recording rules in `/docs/introduction/overview`, which is linked at the end of the document anyway.

jan--f (changes requested):
Thanks, I think this improves things even though I have some change requests.

jan--f:
Also please read https://prometheus.io/docs/guides/contributing/

Atharva-Ramawat:
Thanks for the feedback @jan--f! I've applied the requested changes: I reverted the metrics discovery to focus on the /metrics endpoint and completely removed the Recording Rules section. I added these fixes in a new commit and made sure to include my DCO sign-off there. Let me know if it looks good!

jan--f:
Hello from the bug scrub!

This still doesn't pass the DCO check, seems like the author is not paying attention to our contribution guide.

PR #3042, opened 2026-07-21
Add OpenObserve remote storage integration

hengfeiyang:
## What changed

Add OpenObserve to the Remote Endpoints and Storage integrations as a Prometheus remote-write destination.

## Why

OpenObserve natively accepts Prometheus remote-write samples, but it is not currently listed among the compatible remote storage endpoints.

The entry links directly to the OpenObserve Prometheus ingestion documentation, which includes a representative `remote_write` configuration and endpoint.

## Validation

- Ran `make check`
- Ran `git diff --check`
- Verified the linked documentation returns HTTP 200

@RichiH @juliusv


juliusv (approved):
Thanks 👍

PR #3050, opened 2026-07-30
Fix subtitle notation in open_metrics_spec.md

natenjoy:


jan--f (changes requested):
Thanks, good catch.
Would you mind adding the same fix to https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec_2_0.md please?

natenjoy:
> Thanks, good catch. Would you mind adding the same fix to https://github.com/prometheus/docs/blob/main/docs/specs/om/open_metrics_spec_2_0.md please?

Added.

jan--f (approved):
Thanks

PR #3053, opened 2026-08-04
docs: update TSDB storage comparison

### LINKED ISSUE #1031: Fix Prometheus TSDB file layout explanation in comparison doc
https://prometheus.io/docs/introduction/comparison/ has "Prometheus also creates one local file per time series". This is no longer true for Prometheus 2.x and needs to be updated.

JasonColapietro:
## Summary

- replace both obsolete per-series-file descriptions of Prometheus storage
- describe the in-memory head and WAL separately from persisted block directories
- link to the maintained on-disk layout documentation

## Testing

- focused assertions for removal of the stale claims and presence of the current layout terms
- `git diff --check`

Site lint/build could not start locally because dependencies were absent and npm registry resolution was blocked; this draft relies on hosted CI for those gates.

Closes #1031.

### DIFF
--- a/docs/introduction/comparison.md
+++ b/docs/introduction/comparison.md
@@ -56,10 +56,12 @@ RRD-style database that expects samples to arrive at regular intervals. Every
 time series is stored in a separate file, and new samples overwrite old ones
 after a certain amount of time.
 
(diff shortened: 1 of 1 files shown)


jan--f (approved):
Thanks!

PR #3060, opened 2026-08-09
docs: guide for running Prometheus behind a reverse proxy

locker95:
Short guide on proxying Prometheus: `--web.external-url`, Host / X-Forwarded-* headers, nginx and Caddy snippets, and a note about lifecycle/admin paths. Linked from the security model page.

Fixes #2950

### DIFF
--- a/docs/guides/reverse-proxy.md
+++ b/docs/guides/reverse-proxy.md
@@ -1,0 +1,88 @@
+---
+title: Running Prometheus behind a reverse proxy
+nav_title: Reverse proxy
+sort_rank: 12
+---
+
+Prometheus is often placed behind a reverse proxy (nginx, Caddy, Traefik,
+Apache, etc.) for TLS termination, path prefixes, or access control. This guide
+covers common settings that keep the UI and HTTP API working correctly.
+
+NOTE: Reverse proxies do **not** replace network isolation. Treat Prometheus
+HTTP endpoints as sensitive (see the [security model](/docs/operating/security/)).
+Do not expose them on the public internet without authentication and careful
+rate limiting.
+
+## External URL
+
+If the proxy serves Prometheus under a public hostname or URL prefix, set
+[`--web.external-url`](https://prometheus.io/docs/prometheus/latest/command-line/prometheus/)
+to that public base URL (including path prefix if any). This is used for
+generated links, redirects, and the expression browser.
+
+Example when the UI is at `https://monitoring.example.com/prometheus/`:
+
+```bash
+prometheus \
+  --config.file=/etc/prometheus/prometheus.yml \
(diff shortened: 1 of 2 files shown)


docs/operating/security.md
```diff
@@ -15,6 +15,9 @@ components, and the `/pprof` endpoint of server components implemented in Go.
 Furthermore, it is easily possible to overload and ultimately DoS servers with
 requests to these endpoints.
 
+If you terminate TLS or add authentication at a reverse proxy, see
+[Running Prometheus behind a reverse proxy](/docs/guides/reverse-proxy/).
+
```
jan--f, line 20:
```suggestion
```
Not sure this is right place to cross link this. Maybe better places are `/docs/guides/tls-encryption.md' and `/docs/guides/basic-auth.md`?

docs/guides/reverse-proxy.md
```diff
@@ -0,0 +1,88 @@
...
+for the combination that matches your setup).
+
+## Headers to forward
+
+Proxies should pass through at least:
+
+| Header | Purpose |
+|--------|---------|
+| `Host` | Original host as seen by clients (or the public hostname you intend Prometheus to see) |
+| `X-Forwarded-Proto` | `https` when TLS is terminated at the proxy so redirects stay on HTTPS |
```
jan--f, line 43: Iirc Prometheus uses relative redirects internally. While its good practice to suggest bot headers, the purpose is not quite accurate, as Promteheus doesn't use the information.
Maybe just

```suggestion
| `Host` | Original host the client used, not used by Prometheus but good practice |
| `X-Forwarded-Proto` | `https` when TLS is terminated at the proxy, not used by Prometheus but good practice|
```

docs/guides/reverse-proxy.md
```diff
@@ -0,0 +1,88 @@
...
+HTTP endpoints as sensitive (see the [security model](/docs/operating/security/)).
+Do not expose them on the public internet without authentication and careful
+rate limiting.
+
+## External URL
+
+If the proxy serves Prometheus under a public hostname or URL prefix, set
+[`--web.external-url`](https://prometheus.io/docs/prometheus/latest/command-line/prometheus/)
+to that public base URL (including path prefix if any). This is used for
+generated links, redirects, and the expression browser.
```
jan--f, line 21:
```suggestion
generated links.
```
I'm not sure about the last two statements. If you're sure about the let me know, otherwise let's err on the side of caution and drop them?

jan--f (changes requested):
Thanks, nice addition! Just a few comments.

jan--f:
Hello from the bug scrub!
I will close this for now due to lack of response. Please feel free to reopen at any time.

PR #3065, opened 2026-08-12
build: skip prometheus v0 releases during doc import

jan--f:
Some releases have version strings that are not semver compatible, e.g. v0.18.0rc1.


bwplotka (approved):
Thanks!

PR #3076, opened 2026-08-19
Add postgres exporter docs to website

ArthurSens:
100% AI generated for now, adds https://github.com/prometheus-community/postgres_exporter/pull/1361 to the website :)


jan--f (changes requested):
Not sure why 7b2a05efecd051103822329b3c7117855c0ac8a4 is needed. I'm not against moving exporter related content under a new exporter focused section but including this here makes thing unnecessarily complex.

Including the postgres docs also seems broken. In the preview the link gives 404.

ArthurSens:
Addressed the feedback during the docs call :) 

> Including the postgres docs also seems broken. In the preview the link gives 404.

That is because there's no release tag of the Postgres exporter that includes the `/docs` folder. I've added the commit https://github.com/prometheus/docs/pull/3076/commits/752d5d066a3a4be08d6b31e2bd598bb5a9d20bd5 that changes the logic from release tag to SHA just so things are visible, but I'd like to remove that commit before merging here

ArthurSens:
https://github.com/prometheus-community/postgres_exporter/pull/1361 also needs to include FrontMatter headers

PR #3081, opened 2026-08-25
rw2: Async Receiver Support

bwplotka:
Context: https://docs.google.com/document/d/1AzEt6kOz4MJ855fBxtzvyyS-kDu8ynXiJy6VCSUQcSI/edit

PR #3080, opened 2026-08-25
rw2: rc.5 version

bwplotka:
Feature flag for stacked rc.5 changes.

Pending: 
* https://github.com/prometheus/docs/pull/3081
* https://github.com/prometheus/docs/pull/3085

PR #3085, opened 2026-08-26
RW 2.0-rc.5: allow exemplar-only time series

thc1006:
This changes the 2.0 specification so that a `TimeSeries` carrying only exemplars is described rather than forbidden.

## Why the current wording does not cover what a sender produces

Prometheus stores exemplars per series, not per sample. `populateV2TimeSeries` in `storage/remote/queue_manager.go` writes one output `TimeSeries` per queue item, and for an exemplar item it appends only an exemplar, having already reset that entry's samples and histograms. So an exemplar can leave the sender in a `TimeSeries` of its own.

The specification currently says:

> At least one element in `samples` or in `histograms` MUST be provided. A `TimeSeries` MUST NOT include both `samples` and `histograms`.

That shape therefore has no wording behind it, and a receiver has nothing to rely on when it arrives.

## What this changes

Two normative sentences, plus the proto copy this document carries:

- the "at least one element" rule accepts `exemplars` as well, and says what a `TimeSeries` carrying only exemplars means;
- the `#### Exemplars` section says the same thing from the exemplar's side;
- the copied proto comments say the same as the rules above them, so the two do not disagree inside one document.

No field, field number or wire layout changes, and nothing is asked of a sender that was not asked before.

## What the first version also had

It added a SHOULD that exemplars travel in the same request as the samples or histograms of their series. That is a rule about how a sender batches rather than a statement about what the wire format allows, and it is the part of this change review is stuck on, so it is out as of 81c879af.

Co-location is still worth writing down somewhere. `QueueManager.Append` and `AppendExemplars` both enqueue on the series ref, and `shards.enqueue` picks the shard with `ref % len(queues)`, so a series' samples and its exemplars go into the same shard queue in order. They are usually in one request and a full batch is what splits them. That belongs wherever the receiver profile question lands, not in this rule.

## What this leaves alone

The receiver tests in `prometheus/compliance` only cover exemplars sent alongside a sample. Covering a `TimeSeries` that carries only exemplars is a separate change. The canonical proto comments are in prometheus/prometheus#19530, which carries the source version bump to rc.5 as well.

## On versioning

Retitled with the version, the way rc.4 was done in #2762, so the title is the record of what went into the release. #3080 still carries the single bump and #3081 is the other change in this one.

Related to prometheus/prometheus#17857 and prometheus/prometheus#16944.

### DIFF
--- a/docs/specs/prw/remote_write_spec_2_0.md
+++ b/docs/specs/prw/remote_write_spec_2_0.md
@@ -362,7 +362,7 @@ For every `TimeSeries` message:
 <!---
 Rationales: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-proposal/proposals/2024-04-09_remote-write-20.md#partial-writes#samples-vs-native-histogram-samples
 -->
-* At least one element in `samples` or in `histograms` MUST be provided. A `TimeSeries` MUST NOT include both `samples` and `histograms`. For series which (rarely) would mix float and histogram samples, a separate `TimeSeries` message MUST be used.
+* At least one element in `samples`, in `histograms`, or in `exemplars` MUST be provided. A `TimeSeries` MUST NOT include both `samples` and `histograms`. For series which (rarely) would mix float and histogram samples, a separate `TimeSeries` message MUST be used. A `TimeSeries` carrying exemplars but neither samples nor histograms represents series-level exemplars for the series identified by `labels_refs`.
 
 <!---
 Rationales: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-proposal/proposals/2024-04-09_remote-write-20.md#always-on-metadata
@@ -452,6 +452,8 @@ Rationales: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-p
 -->
 * MAY contain labels e.g. referencing trace or request ID. If the exemplar references a trace it SHOULD use the `trace_id` label name, as a best practice.
 * MUST contain a timestamp. While exemplar timestamps are optional in Prometheus/Open Metrics exposition formats, the assumption is that a timestamp is assigned at scrape time in the same way a timestamp is assigned to the scrape sample. Receivers require exemplar timestamps to reliably handle (e.g. deduplicate) incoming exemplars.
+* MAY be sent in a `TimeSeries` that carries neither samples nor histograms. In that form the exemplars are associated with the series identified by `labels_refs`, rather than with a particular sample or histogram.
+* SHOULD be sent in the same request as the samples or histograms of the series they belong to, including when they are sent in a `TimeSeries` of their own.
 
 ## Out of Scope
 
--- a/docs/specs/prw/remote_write_spec_2_0.md
+++ b/docs/specs/prw/remote_write_spec_2_0.md
@@ -257,22 +257,23 @@ message TimeSeries {
   // or start timestamp.
   repeated uint32 labels_refs = 1;
 
-  // Timeseries messages can either specify samples or (native) histogram samples
-  // (histogram field), but not both. For a typical sender (real-time metric
-  // streaming), in healthy cases, there will be only one sample or histogram.
+  // TimeSeries messages must specify at least one float sample, native histogram sample
+  // or exemplar. Samples and histograms must not be specified at the same time.
+  // Exemplars may be specified without samples or histograms. For a typical sender
+  // (real-time metric streaming), in healthy cases, there will be only one sample or histogram.
   //
   // Samples and histograms are sorted by timestamp (older first).
   repeated Sample samples = 2;
   repeated Histogram histograms = 3;
 
-  // exemplars represents an optional set of exemplars attached to this series' samples.
+  // exemplars represents an optional set of exemplars for this series.
   repeated Exemplar exemplars = 4;
 
   // metadata represents the metadata associated with the given series' samples.
   Metadata metadata = 5;
 }
 
-// Exemplar is an additional information attached to some series' samples.
+// Exemplar contains additional information associated with a series.
 // It is typically used to attach an example trace or request ID associated with
 // the metric changes.
 message Exemplar {


bwplotka:
This looks what wanted as per https://docs.google.com/document/d/1ZcyKVF87Xf6FhaGkb33LgVLb27FH5NIlRXWEGO2Cc4Q/edit?tab=t.0#bookmark=id.nllt8bbnl0i decision. However we need strict versioning. 

I initially planned to pack this and [async change](https://github.com/prometheus/docs/pull/3080) under rc.5 for an easy reference -- we don't use git tags so it's a bit challenging to pack two separate changes under same version (:

thc1006:
Retitled with the version, the way rc.4 was done in #2762, so the title carries the record that git tags would otherwise hold. #3080 keeps the single bump and #3081 is the other change in rc.5, so nothing has to be packed together beyond the two titles saying the same version.

Also added the second half of the decision, which I had left out: exemplars SHOULD be sent in the same request as the samples or histograms of their series, including when they travel in a `TimeSeries` of their own. Without it the wording read as a licence to detach them freely, which is not what the sender does today.

Out of draft now.

bwplotka:
I changed base to https://github.com/prometheus/docs/pull/3085 so we can squash it all as a rc.5, hope that's ok

bwplotka:
Ups I think you need to manually rebase, sorry @thc1006  - do you mind doing this?

docs/specs/prw/remote_write_spec_2_0.md
krajorama: This would be updated once the proto is updated then?

krajorama (commented):
LGTM, except the question about the copied protobuf

docs/specs/prw/remote_write_spec_2_0.md
bwplotka: Let's update it at one go and prep a PR.

thc1006:
Rebased onto `rw2-spec-rc5`, so this is one commit against that branch again. No trouble at all, and moving the base was the right call for keeping rc.5 in one place.

@krajorama good catch on the copied protobuf. The comment you pointed at turned out to be one of three saying the same thing: that one says a `TimeSeries` specifies samples or histograms, the `exemplars` field says they belong to the series' samples, and the `Exemplar` message says it too. All three are in this PR now.

The originals live in `prompb/io/prometheus/write/v2/types.proto`, so those go in prometheus/prometheus#19530. That one does not depend on this PR: `populateV2TimeSeries` clears `Samples` and `Histograms` and appends only an exemplar when the queue item is a `tExemplar`, so the comment already disagrees with the sender sitting next to it.

One thing I would rather ask than guess. In prometheus/prometheus#17857 back in January you proposed two halves, "Add SHOULD for per sample, but allow per series by spec". This PR has the second one. The `SHOULD` I added after your first comment here is about sending exemplars in the same request, not about preferring per sample association. Should the per sample `SHOULD` go in as well, or does the August decision replace it? Your 26 August comment on that issue reads as moving away from the per sample model for now, which is why I left it out.

Last thing, unrelated to the diff: prometheus/prometheus#18014 has been open since February taking the other route, dropping exemplars that arrive without a matching sample. @AftAb-25 wrote it against the spec as it read then. If this lands, they deserve to hear that the direction moved. The bot has already marked it stale.

docs/specs/prw/remote_write_spec_2_0.md
thc1006: Both done: the copy here in this PR, the original in prometheus/prometheus#19530.

docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -257,22 +257,23 @@ message TimeSeries {
...
   // Samples and histograms are sorted by timestamp (older first).
   repeated Sample samples = 2;
   repeated Histogram histograms = 3;
 
-  // exemplars represents an optional set of exemplars attached to this series' samples.
+  // exemplars represents an optional set of exemplars for this series.
   repeated Exemplar exemplars = 4;
 
-  // metadata represents the metadata associated with the given series' samples.
+  // metadata represents the metadata associated with this series.
```
bwplotka: This is problematic, we likely don't want to allow sender to send metadata without samples, that's a step back. There's no benefit to existing ecosystem to allow that too.

To do so we not only revert the wording here, but also somehow adjust ` MAY be sent in a `TimeSeries` that carries neither samples nor histograms.` 🤔  it's getting cumbersome with some fields needing sample vs series though.


```suggestion
  // metadata represents the metadata associated with the given series' samples.
```

docs/specs/prw/remote_write_spec_2_0.md
```diff
@@ -452,6 +453,8 @@ Rationales: https://github.com/prometheus/proposals/blob/alexg/remote-write-20-p
 -->
 * MAY contain labels e.g. referencing trace or request ID. If the exemplar references a trace it SHOULD use the `trace_id` label name, as a best practice.
 * MUST contain a timestamp. While exemplar timestamps are optional in Prometheus/Open Metrics exposition formats, the assumption is that a timestamp is assigned at scrape time in the same way a timestamp is assigned to the scrape sample. Receivers require exemplar timestamps to reliably handle (e.g. deduplicate) incoming exemplars.
+* MAY be sent in a `TimeSeries` that carries neither samples nor histograms. In that form the exemplars are associated with the series identified by `labels_refs`, rather than with a particular sample or histogram.
```
bwplotka, line 456: We need to think about other elements like metadata. I don't think we want to allow them hanging without samples.

bwplotka (changes requested):
General thought when I look a the changes: It feels this change is a step back when it comes to compatibility with systems oriented around OpenTelemetry model. As-in, you can use PRW2 to have a reliable transport against those, but it requires being strict on various elements beyond what protocol requires from sender e.g.:

* Avoid unknown metadata.
* Use NHCB instead of classic histograms.

Now with this change it also technically needs:
* Exemplar per sample (or at least within same request).

In the same time this relaxation of rules for exemplars allows (slightly) simpler protocol for existing Prometheus systems, and it's unlikely the exemplar part would change for those given the current APIs.

I wonder how to tackle this redundancy of implementing PRW2 around those intentions to battle [ossification/proliferation and being too lean on sending](https://www.rfc-editor.org/rfc/rfc9413.html#name-protocol-robustness) so "I implement PRW2 sending" is not ambiguous. 

Perhaps what we need is to introduce some "mode" or consider OTLP export again given advances we have in the storage.

docs/specs/prw/remote_write_spec_2_0.md
thc1006: Applied here and in prometheus/prometheus#19530. You are right that it does not belong in this PR: it is about exemplars, and I changed a neighbouring line without asking what it was holding up.

docs/specs/prw/remote_write_spec_2_0.md
thc1006: The at-least-one rule above names samples, histograms and exemplars, so metadata alone does not satisfy it and a metadata-only `TimeSeries` stays invalid. If this line reads as an opening for the other fields, I can say so where the rule is rather than leaving it to be inferred.

thc1006:
One piece of data first, since it may change the shape of the question.

Prometheus already sends metadata in a message with no samples. `AppendExemplars` reads `seriesMetadata` when it enqueues, and `populateV2TimeSeries` fills `Metadata` before it looks at what kind of item it has, so an exemplar-only `TimeSeries` goes out carrying type, unit and help. That is today's traffic rather than something this change introduces.

The same holds for the OTel side. A receiver reading real Prometheus traffic already had to decide what to do with exemplar-only messages, and the specification told it they were invalid. This does not hand it a new case, it names one it already had.

What the change does add is the SHOULD about the same request, and that one is mine to defend or drop. It is there so a receiver has a bounded window to associate within. If the cost is another implicit requirement on senders, dropping it leaves the MAY describing the wire format and nothing more.

On modes and on OTLP export I do not have a view worth your time. What would help me is knowing whether you want this narrowed to the wire-format statement, held until that larger question settles, or something else.

bwplotka:
> Prometheus already sends metadata in a message with no samples. AppendExemplars reads seriesMetadata when it enqueues, and populateV2TimeSeries fills Metadata before it looks at what kind of item it has, so an exemplar-only TimeSeries goes out carrying type, unit and help. That is today's traffic rather than something this change introduces.

~That is true, and it's the tech debt due to abandoning the idea of using https://github.com/prometheus/prometheus/issues/15911 - if we want to bring this back, we need to fix it.~
~The path using  `__type__` and `__unit__` feature leverages metadata per samples correctly, so we do have path to use it.~

I understood it wrongly -- for exemplars sure, but why exemplars even have metadata then? 🙈

thc1006:
Thanks, that framing helps, and the `__type__`/`__unit__` pointer is one I had not followed.

Correcting something in what I wrote above: I leaned on `enableTypeAndUnitLabels` as though it described how things run, but it sits behind `--enable-feature=type-and-unit-labels`, so it is the path you are pointing at rather than the current state. The argument I built on it does not carry the weight I gave it.

What is left is narrower. Exemplars are enqueued separately from their samples whichever way metadata goes, so that part does not resolve with it.

Happy to drop the SHOULD and leave this describing only the wire format, or to park it until the metadata direction settles. Your call.

thc1006:
I took the same-request SHOULD out in 81c879af. What is left is only that an exemplar-only `TimeSeries` is valid and that it identifies its series by `labels_refs`.

@bwplotka that was the same-request half of the third item on your list, and the only one of the three this PR was writing into the spec. Whether a sender keeps a series' exemplars with its samples is about interoperability, not about what the wire format allows, so it belongs with the mode or profile question rather than as a SHOULD in the base spec.

@krajorama flagging this because of https://github.com/prometheus/prometheus/issues/17857#issuecomment-3753736954. I read it as asking to soften the per-sample MUST so the mode in that plan would not be a violation, rather than as asking for a new rule about batching. This PR makes the shape valid instead, so there is nothing left to violate. If I have read you wrong, say so and I will put the line back.

Co-location is still worth writing down somewhere. `Append` and `AppendExemplars` both enqueue on the series ref, and `shards.enqueue` picks the shard with `ref % len(queues)`, so a series' samples and its exemplars go into the same shard queue in order. They are usually in one request, and a full batch is what splits them. I can open an issue for that.

Does this settle the protocol format part of your review?

docs/specs/prw/remote_write_spec_2_0.md
thc1006: Pushed 6a3e7c92, which adjusts the bullet you asked about here.

The exemplar bullet now says `metadata` on an exemplar-only `TimeSeries` describes that series and that Receivers MAY ignore it, so a sender cannot rely on that shape to deliver metadata. I also added a line to the at-least-one rule saying `metadata` alone does not satisfy it. That second one is more than you asked for: a `TimeSeries` carrying only metadata was already invalid, the rule just never said so out loud. Say the word and I will drop it and leave the exemplar bullet on its own.

On your other question: exemplars carry metadata because `populateV2TimeSeries` sets the `Metadata` field in its first block, above the `switch d.sType` that decides whether an entry is a sample or an exemplar. Every batch entry passes through it. Nothing picked that for exemplars, and I would be glad to open a fix against prometheus/prometheus if you want one.

PR #3088, opened 2026-08-26
Add hint to regarding CORS headers of Alertmanager

SoloJacobs:
The current documentation specifies that only trusted users should have access to the http endpoints of Alertmanager. However, the browsers default Single Origin Policy pertains more to network access.

Thus, documenting the `Access-Control-Allow-Origin: *` headers is a good idea. The `API Security` section already hints at the configuration, but mentioning it in a Alertmanager specific location shouldn't hurt.

### DIFF
--- a/docs/operating/security.md
+++ b/docs/operating/security.md
@@ -111,6 +111,11 @@ Any user with access to the Alertmanager HTTP endpoint has access to its data.
 They can create and resolve alerts. They can create, modify and delete
 silences.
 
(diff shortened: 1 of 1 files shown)


juliusv (approved):
Thanks! I still think this is dangerous / unexpected default behavior for most users, but documenting it is a good start. The Prometheus server has the same CORS issue, except that it's read-only by default and doesn't allow people to write any data.

PR #3096, opened 2026-09-04
 Move Node.js client library to official list

Faiq1818:
## Summary

Prometheus now has an official Node.js client library maintained under the Prometheus organization. This moves the Node.js entry from the unofficial third-party list to the official client libraries list.

  - Remove the old `siimon/prom-client` entry
  - Add `prometheus/client_js`


ArthurSens (approved):
Thanks!

PR #3105, opened 2026-09-11
Add Oodle remote storage integration

mvijaykarthik:
## What changed

Add Oodle to the Remote Endpoints and Storage integrations as a Prometheus remote-write destination.

## Why

Oodle natively accepts Prometheus remote-write samples and serves them back over the Prometheus query API, but it is not listed among the compatible remote storage endpoints.

The entry links to Oodle's Prometheus integration page, which has the `remote_write` configuration and endpoint, plus the equivalent for `vmagent` and `kube-prometheus-stack`.

## Validation

- Ran `git diff --check`
- Verified the linked documentation returns HTTP 200
- Entry is in alphabetical order (between New Relic and OpenObserve)


mvijaykarthik:
@jan--f would you be able to review?

jan--f (approved):

PR #3108, opened 2026-09-13
docs: fix 'allow to set' grammar in native histograms spec

mrchatam:
Fix grammar in the native histograms spec: "do not allow to set" → "do not allow setting".


jan--f:
Not sure what's wrong with the grammar.

PR #3115, opened 2026-09-18
ci: require version bump when OpenMetrics 2.0 spec changes

### LINKED ISSUE #3087: ci: Validate if spec change is bumping version on main branch
This is easy to miss, but we want to ensure spec changes are atomic and versioned.

jarqvi:
Fixes #3087

Adds a CI step to the existing `check-abnf` job that fails when
`docs/specs/om/open_metrics_spec_2_0.md` changes without its `- Version:`
line changing. The new `check_openmetrics_spec_version.py` script compares
the version between the PR head and its merge base with the target branch.

Can be run locally:
`python3 openmetrics-spec-tests/check_openmetrics_spec_version.py origin/main HEAD docs/specs/om/open_metrics_spec_2_0.md`


roidelapluie:
What if changes are only typos etc?

jarqvi:
> What if changes are only typos etc?

Yeah, right now even a typo fix would need a bump. How about a label
(something like `skip-spec-version-check`) that skips the check for cases
like that? Would that work for you?

